PR #43, opened 2020-01-09
Feature/chill log messages

futurechimp:


clients/nym-client/src/clients/mod.rs
```diff
@@ -39,14 +42,17 @@ impl MixTrafficController {
...
                 "[MIX TRAFFIC CONTROL] - got a mix_message for {:?}",
                 mix_message.0
             );
             let send_res = mix_client.send(mix_message.1, mix_message.0).await;
             match send_res {
-                Ok(_) => println!("We successfully sent the message!"),
-                Err(e) => eprintln!("We failed to send the message :( - {:?}", e),
+                Ok(_) => {
+                    print!(".");
+                    io::stdout().flush().ok().expect("Could not flush stdout");
```
jstuczyn, line 53: Do we need to do that?

clients/nym-client/src/clients/mod.rs
```diff
@@ -150,7 +156,7 @@ impl NymClient {
         topology: Topology,
     ) {
         loop {
-            println!("[LOOP COVER TRAFFIC STREAM] - next cover message!");
+            info!("[LOOP COVER TRAFFIC STREAM] - next cover message!");
```
jstuczyn, line 159: I would actually change this (and similar messages) to lower level, debug or possibly even trace

clients/nym-client/src/clients/mod.rs
futurechimp: Yeah. My goal is to have ".", "..", "..." etc progressively print out as loop cover traffic is received, and then show "real" messages with actual content in them on their own line. 

As it turns out, if you are printing a small message like "." using `print!` in Rust, nothing shows up. This is because `print!` *does not auto-flush* (but `println!` does auto-flush). So that `flush` is needed to get the effect I'm looking for.

PR #53, opened 2020-01-14
print public key for nym client tools

ghost:


jstuczyn:
You should rather put it in https://github.com/nymtech/nym/blob/develop/nym-client/src/commands/websocket.rs#L32 and https://github.com/nymtech/nym/blob/develop/nym-client/src/commands/tcpsocket.rs#L36 then use `to_b64_string()` on the public_key type. 

Also, do keep in mind that the 1:1 mapping of public key to address is temporary

jstuczyn (changes requested):
Please move the print to locations specified in the comment

PR #60, opened 2020-01-19
added lib on cargo and lib.rs

ghost:
it would be nice to have the lib.rs back to the base code, things like the experimental nym-mobile still depends on the client.


futurechimp:
Hi @mileschet we've now fixed this in `develop` and are planning to bang out a 0.3.3 release this afternoon to give it back to you :). @jstuczyn wanted to restructure things a little bit while putting the lib back in place.

PR #87, opened 2020-01-20
Using println rather than log for startup banner, it's not an error

futurechimp:


jstuczyn:
But you're actually printing/logging an error not a banner?

Issue #15, opened 2020-01-06
If a thread blows at startup, panic the entire application


futurechimp:
At the moment, if a subsidiary thread explodes during startup, the main application thread just keeps on trucking. Ideally the whole node application startup process would fail, so that node administrators don't think startup succeeded.

jstuczyn:
Is partially being dealt with in https://github.com/nymtech/nym/pull/89

PR #89, opened 2020-01-21
Release builds should no longer silently fail - everything will be im…

jstuczyn:
…mediately stopped on any panic


jstuczyn:
Solves https://github.com/nymtech/nym/issues/15

PR #102, opened 2020-01-24
Feature/base58

### LINKED ISSUE #28: Convert to non-url-safe base64 (everywhere)

futurechimp:
Almost closes #28 - but don't merge yet, there is still a bit of base64 left in `nym-client` and I'm not clear right now on whether that's being used as an internal network transport format or something that's more publicly usable. So I've left it alone thus far (I don't want our JavaScript client to break).


futurechimp:
Ok ready to roll on this.

PR #117, opened 2020-02-05
Feature/config files

### LINKED ISSUE #21: Create config struct for mixnode (possibly also for client)
Currently only our provider does nice argument parsing to create a proper config (i.e. `fn new_config(matches: &ArgMatches) -> provider::Config`. It would be nice if mixnode did something similar and we just passed the config itself to create the mix instance.

### LINKED ISSUE #26: nym-mixnode mandatory host option
currently the doc states:

> nym-mixnode run --layer 1 will start the mixnode in layer 1 

but this errors with 

> error: The following required arguments were not provided:
>     --host <host>

so either the README should state 
> nym-mixnode run --layer 1 --host localhost will start the mixnode in layer 1 on localhost

or (maybe preferred option) the host option should default to localhost

### LINKED ISSUE #69: Replace args with proper config files
At the moment, mixnode keys change on every node restart. This is going to cause friction for clients (especially if they don't refresh periodically) because a restarted node will change its key, causing MacIntegrity errors as clients send un-decryptable packets. Ultimately this breaks the path, a big no-no.

So we might as well store some persistent information on disk.

* make an `init` command for mixnodes and providers
* create persistent mixnode keys that survive across restarts
* we need to consider whether we need node human names for this

jstuczyn:
Closes https://github.com/nymtech/nym/issues/69
Closes https://github.com/nymtech/nym/issues/26 (indirectly)
Closes https://github.com/nymtech/nym/issues/21


common/clients/mix-client/src/packet.rs
```diff
@@ -43,22 +43,23 @@ pub fn loop_cover_message<T: NymTopology>(
...
+        average_delay_duration,
     )
 }
 
 pub fn encapsulate_message<T: NymTopology>(
     recipient: Destination,
     message: Vec<u8>,
     topology: &T,
-    average_delay: f64,
+    average_delay_duration: time::Duration,
```
jstuczyn, line 62: TODO: rename to just `average_delay`

common/clients/mix-client/src/poisson.rs
```diff
@@ -1,9 +1,10 @@
 use rand_distr::{Distribution, Exp};
+use std::time;
 
-pub fn sample(average_delay: f64) -> f64 {
+pub fn sample_from_duration(average_duration: time::Duration) -> time::Duration {
```
jstuczyn, line 4: TODO: just call `sample`

PR #142, opened 2020-03-13
Feature/concurrent connection managers

jstuczyn:
Note that this is not the final iteration of this feature as there are two major issues remaining:
- client has no idea if his request actually succeeded + gets no response
- there is some unexplained latency to faraway servers


futurechimp:
Let me know when you'd like a review on this one :)

PR #149, opened 2020-03-16
Feature/add topology to validator

futurechimp:


validator/src/network/rest/routes/topology.rs
```diff
@@ -0,0 +1,12 @@
+use super::*;
+use crate::network::rest::models::presence;
+
+pub fn get(_req: &mut Request) -> IronResult<Response> {
+    let topology = presence::Topology {
+        mix_nodes: Vec::<presence::MixNode>::new(),
+        service_providers: Vec::<presence::ServiceProvider>::new(),
+        validators: Vec::<presence::Validator>::new(),
+    };
+    let response = serde_json::to_string_pretty(&topology).unwrap();
```
jstuczyn, line 10: Any reason we prettify the response rather than just use `to_string` ?

validator/Cargo.toml
```diff
@@ -18,7 +18,10 @@ futures = "0.3.1"
 iron = "0.6.1"
 log = "0.4"
 pretty_env_logger = "0.3"
+router = "0.6.0"
 serde = "1.0.104"
+serde_derive = "1.0.104"
```
jstuczyn: We definitely don't need to import `serde_derive` as seperate dependency as everything is already included in base `serde` (as long as we use correct features). I think the same might be  true for `serde_json` but I haven't tested that claim

jstuczyn (changes requested):

validator/Cargo.toml
futurechimp: Good catch on the `dervive`. It looks the `serde_json` is needed.

validator/src/network/rest/routes/topology.rs
futurechimp: I just like emitting pretty json wherever possible.

Issue #160, opened 2020-03-31
Improve the healthchecker


jstuczyn:
- incorporate new TCP client into healthchecker (which should hopefully fix "early eof" warnings received by mixes)
- fix crashing issue (no map entry) when score threashold was set to for example 50.0

PR #165, opened 2020-04-01
Feature/minor healthchecker improvements

jstuczyn:
Mixnodes should no longer see frequent "early eof"


jstuczyn:
Deals with first part of https://github.com/nymtech/nym/issues/160

PR #171, opened 2020-04-03
Feature/service persistence

futurechimp:


validator/.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "cSpell.words": [
+        "bodyparser"
+    ]
+}
```
jstuczyn: Could you add this file to gitignore?

validator/Cargo.toml
```diff
@@ -9,10 +9,12 @@ edition = "2018"
 
 [dependencies]
 abci = "0.6.4"
+bodyparser = "0.8.0"
 byteorder = "1.3.2"
 clap = "2.33.0"
+# chrono = { version = "0.4.11", features = ["serde"] } TODO: remove this if I don't start using it again soon
```
jstuczyn, line 15: We will be using this very soon for better time units in config files : )

validator/src/network/tendermint/mod.rs
```diff
@@ -57,12 +57,12 @@ impl abci::Application for Abci {
 
     fn commit(&mut self, _req: &RequestCommit) -> ResponseCommit {
         // Create the response
-        let mut resp = ResponseCommit::new();
+        let mut response = ResponseCommit::new();
         // Convert count to bits
         let mut buf = [0; 8];
         BigEndian::write_u64(&mut buf, self.count);
```
jstuczyn, line 77: For future reference, there are nicer helpers functions for that functionality: https://doc.rust-lang.org/std/primitive.u64.html#method.to_be_bytes

validator/src/network/rest/mod.rs
```diff
@@ -1,29 +1,59 @@
+use crate::services::mixmining;
 use iron::prelude::*;
+use presence::mixnode;
+use presence::topology;
 use router::Router;
+use std::sync::{Arc, Mutex};
```
jstuczyn, line 19: I might be entirely wrong about this, but this is based on some random comments on the internet I've read (so the most reliable source there is), but I think in an async setup, like we have here, i.e. tokio runtime, we want to be using this mutex: https://docs.rs/futures/0.3.4/futures/lock/struct.Mutex.html . As from what I understand, waiting for `std::sync::Mutex` blocks more stuff than necessary, while `futures::lock::Mutex` affects only this particular [tokio] task.

validator/src/network/rest/mod.rs
```diff
@@ -1,29 +1,59 @@
...
+        }
     }
 
+    /// Run the REST API.
     pub async fn run(self) {
-        let port = 3000;
-        println!("* starting REST API on localhost:{}", port);
+        let port = 3000; // TODO: make this configurable
+        let address = format!("localhost:{}", port);
+        println!("* starting REST API on http://{}", address);
```
jstuczyn, line 41: Can we be consistent and use `log` for those messages? : )

validator/src/network/tendermint/mod.rs
jstuczyn, line 77: There are nice helper functions to achieve that functionality: https://doc.rust-lang.org/std/primitive.u64.html#method.to_be_bytes

validator/src/network/rest/mod.rs
jstuczyn, line 19: I might be totally wrong about this, however, based on some comments I've read on the internet (so the most reliable source on the planet), you want to be using this mutex: https://docs.rs/futures/0.3.4/futures/lock/struct.Mutex.html . As from what I understand, `std::sync::Mutex` would block the whole thread what could lead to blocking async executor hence causing a deadlock. Meanwhile `futures::lock::Mutex` would simply return `Poll::Pending` and cause the Future (or tokio task) to be polled again at later time.

validator/src/network/rest/mod.rs
jstuczyn, line 41: Can we be consistent and use `log` for those messages? : )

validator/src/services/mixmining/db.rs
```diff
@@ -0,0 +1,105 @@
...
+            capacity: 6,
+            mixnodes,
+        }
+    }
+
+    pub fn add(&mut self, mixnode: Mixnode) {
+        self.mixnodes.push(mixnode);
+    }
+
+    pub fn get_mixnodes(&self) -> &Vec<Mixnode> {
```
jstuczyn, line 24: I'm not entirely sure whether returning `&Vec<_>` is the best idea here. I can't give a good reason against it, but it just feels weird for me. It might as well be totally fine thing to do that does everything we want, so also keep that in mind.

validator/src/network/rest/mod.rs
```diff
@@ -1,29 +1,59 @@
...
+    mixmining_service: Arc<Mutex<mixmining::Service>>,
+}
 
 impl Api {
-    pub fn new() -> Api {
-        Api {}
+    pub fn new(mixmining_service: mixmining::Service) -> Api {
+        let service = Arc::new(Mutex::new(mixmining_service));
+        Api {
+            mixmining_service: service,
```
jstuczyn, line 33: Looking at how many places are using `Arc<Mutex<mixmining_service>>` perhaps we could make it into a type alias?

validator/src/services/mixmining/mod.rs
```diff
@@ -1 +1,163 @@
+use db::MixminingDb;
+use models::*;
+
+pub mod db;
 pub mod health_check_runner;
+pub mod models;
+mod tests;
+
+pub struct Service {
+    db: MixminingDb,
```
jstuczyn, line 24: From what I see, everything using `Service`, has it behind `Arc<Mutex<Service>>` and have to deal with locks, etc. Now, note that the next suggestion has a lot of personal bias, because I've been using the following pattern a lot and I'm still not entirely sure whether it's a correct approach, but I've been seeing it in `mio`, `tokio` and `futures`, so it might not be as bad.

What if `Service` was responsible for dealing with the locks and synchronization itself, making all callers oblivious to them, similarly to what I've done with the client ledger?
https://github.com/nymtech/nym/blob/develop/sfw-provider/src/provider/client_handling/ledger.rs#L35

jstuczyn: So say `Service::add(...)` would deal with the inner lock rather than whoever is calling it

jstuczyn (changes requested):

validator/Cargo.toml
futurechimp: Cool. I had it in my validator implementation for a little bit, then it came out. I was planning to either delete or uncomment when it become totally clear if I'd really need it.

validator/.vscode/settings.json
futurechimp: Whoops. Good idea.

validator/src/network/rest/mod.rs
futurechimp: I am blocking way more than I need to anyway - I'm planning to move the mutex from the Service (where it is now) to the level of each individual datastore (when those exist), because right now every request is basically going to hit the mutex for every other request. I'll switch mutexes when I do that next bit of work (right now I only have 1 datastore so it doesn't really make much difference) - if you happen to have the source on the futures vs std mutex please shoot it to me in Keybase, I'm interested to see!

validator/src/network/rest/mod.rs
futurechimp: There's a difference between log messages (which I think are mostly for developers) and output messages (which should always exist and not be configurable by log level). There is no situation in which I ever want the startup messages about components to *not* print out (e.g. if somebody knocked log level down). 

I think maybe we need to argue this one out a bit more, but I'd say the reverse: some of the things that are currently configured as log messages actually are not - they're important messages from the running system to the user. Things like major subsystems starting, progress messages etc are not things that I think should be configurable as log messages.

jstuczyn:
As discussed - change Mutex used + fix merge conflicts : )

PR #183, opened 2020-04-16
Feature/nym sphinx wasm

futurechimp:


common/nymsphinx/src/addressing/nodes.rs
```diff
@@ -31,7 +31,7 @@ pub enum NymNodeRoutingAddressError {
 /// Current representation of Node routing information used in Nym system.
 /// At this point of time it is a simple `SocketAddr`.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
-pub struct NymNodeRoutingAddress(SocketAddr);
+pub struct NymNodeRoutingAddress(pub SocketAddr);
```
jstuczyn: Any reason for making the field public? Generally I try to avoid that and instead use getters/setters

common/nymsphinx/src/addressing/nodes.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use sphinx::constants::NODE_ADDRESS_LENGTH;
+pub use sphinx::constants::NODE_ADDRESS_LENGTH;
```
jstuczyn, line 15: Why are we re-exporting that constant?

common/nymsphinx/src/addressing/nodes.rs
```diff
@@ -158,6 +158,7 @@ mod tests {
 
     #[test]
     fn nym_node_routing_address_can_be_converted_to_and_from_bytes_for_v4_address() {
+        // HERE!!!!
```
jstuczyn: What's here? : )

gateway/src/listener.rs
```diff
@@ -0,0 +1,42 @@
...
+    client_ref: Arc<Mutex<MultiClient>>,
+) -> Result<()> {
+    let mut ws_stream = accept_async(stream).await.expect("Failed to accept");
+
+    info!("New WebSocket connection: {}", peer);
+
+    while let Some(msg) = ws_stream.next().await {
+        let msg = msg?;
+        if msg.is_binary() {
+            mixnet_client::forward_to_mixnode(msg.into_data(), client_ref.clone()).await;
```
jstuczyn: This is just some suggestion I've read - if you're working with raw Arcs (or Rcs) without any wrappers, be explicit you're cloning the reference counter with `Arc::clone(foo)`. It's not like you may accidentally clone something you don't want. No. It just makes it slightly more explicit about what you mean, especially if you come back to the code later on.

gateway/src/mixnet_client.rs
```diff
@@ -0,0 +1,43 @@
...
+    );
+    let client = multi_tcp_client::Client::new(config);
+    Arc::new(Mutex::new(client))
+}
+
+pub async fn forward_to_mixnode(mut payload: Vec<u8>, client_ref: Arc<Mutex<MultiClient>>) {
+    let mut address_buffer = [0; NODE_ADDRESS_LENGTH];
+    let packet = payload.split_off(NODE_ADDRESS_LENGTH);
+    address_buffer.copy_from_slice(payload.as_slice());
+    let address = NymNodeRoutingAddress::try_from_bytes(&address_buffer)
```
jstuczyn, line 37: You don't need to do any reallocations or copies here. You may do `NymNodeRoutingAddress::try_from_bytes(&payload[..NODE_ADDRESS_LENGTH])` instead

nym-client/js-examples/websocket/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+<script src="socket.js" async></script>
```
jstuczyn, line 3: Where's the jquery import? : (

nym-client/js-examples/websocket/socket.js
```diff
@@ -0,0 +1,56 @@
...
+        displayFetch(response.messages);
+    }
+};
+
+function display(message) {
+    document.getElementById("output").innerHTML = message;
+}
+
+function displayFetch(message) {
+    document.getElementById("fetch").innerHTML = message;
```
jstuczyn, line 31: That's just my personal preference, but rather than changing `innerHTML` on every new (even empty) message, I'd just append to it so that we'd have history of all messages received

nym-sphinx-wasm/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+/// capacity of a single Sphinx packet, the extra information will be discarded.
+#[wasm_bindgen]
+pub fn create_sphinx_packet(raw_route: &str, msg: &str, destination: &str) -> Vec<u8> {
+    utils::set_panic_hook(); // nicer js errors.
+
+    let route = sphinx_route_from(raw_route);
+
+    let average_delay = Duration::from_secs_f64(0.1);
+    let delays = delays::generate_from_average_duration(route.len(), average_delay);
+    let dest_bytes = DestinationAddressBytes::from_base58_string(destination.to_owned());
```
jstuczyn, line 67: Personal note: make all `from_X_string` methods everywhere accept everything with `Into<String>` trait

nym-sphinx-wasm/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+use sphinx::SphinxPacket;
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use std::net::SocketAddr;
+use std::time::Duration;
+use wasm_bindgen::prelude::*;
+
+mod utils;
+
+const IDENTIFIER_LENGTH: usize = 16;
```
jstuczyn, line 33: Any reason we're creating this constant ourselves rather than importing the same one defined in sphinx?

nym-sphinx-wasm/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+/// Concatenate the first mix address bytes with the Sphinx packet.
+///
+/// The Nym gateway node has no idea what is inside the Sphinx packet, or where
+/// it should send a packet it receives. So we prepend the packet with the
+/// address bytes of the first mix inside the packet, so that the gateway can
+/// forward the packet to it.
+fn payload(sphinx_packet: SphinxPacket, route: Vec<Node>) -> Vec<u8> {
+    let mut packet = sphinx_packet.to_bytes();
+    let mut first_mix_address = route.first().unwrap().clone().address.to_bytes().to_vec();
+    let mut bytes: Vec<u8> = vec![];
```
jstuczyn: You don't need to allocate and copy all the bytes here. You could either use iterators (`packet.into_iter().chain(first_mix_address.into_iter()).collect()`) or append/extend methods directly

nym-sphinx-wasm/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+///
+/// # Panics
+///
+/// This function panics if `JsonRoute.nodes` doesn't contain at least 1
+/// node.
+///
+fn sphinx_route_from(raw_route: &str) -> Vec<Node> {
+    let json_route: JsonRoute = serde_json::from_str(raw_route).unwrap();
+
+    assert!(
```
jstuczyn, line 109: Should this panic or rather return a Result?

jstuczyn: (this is just a question, I don't know what the correct answer in this case is)

nym-sphinx-wasm/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+        let x = Node::try_from(node_data.clone()).expect("Malformed NodeData");
+        sphinx_route.push(x);
+    }
+    sphinx_route
+}
+
+// Converts a base58 &str into a public key (MontgomeryPoint)
+//
+// A TryInto would be nice here but this crate doesn't own MontgomeryPoint.
+fn public_key_from(s: &str) -> MontgomeryPoint {
```
jstuczyn, line 125: I forget, but what was the exact reason we're operating directly on MontgomeryPoints again? 
Also we could/should have `into` or `try_into` on `MixIndentityPublicKey`

scripts/start_local_network.sh
```diff
@@ -38,16 +23,17 @@ killall nym-sfw-provider
 echo "Press CTRL-C to stop."
 echo "Make sure you have started nym-directory !"
 
-cargo build --release --manifest-path nym-client/Cargo.toml --features=local
-cargo build --release --manifest-path mixnode/Cargo.toml --features=local
-cargo build --release --manifest-path sfw-provider/Cargo.toml --features=local
+cargo build --manifest-path nym-client/Cargo.toml --features=local
```
jstuczyn, line 26: Why are we not building it in release anymore?

nym-sphinx-wasm/www/index.js
```diff
@@ -0,0 +1,116 @@
...
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as wasm from "nym-sphinx-wasm";
+
+console.log("Retrieving topology...");
```
jstuczyn, line 17: Just a minor thing, but shouldn't that be called in `getTopology()` function rather than on module import?

jstuczyn (changes requested):

common/nymsphinx/src/addressing/nodes.rs
futurechimp: Something needed it, let me see if there's a getter for this already...

common/nymsphinx/src/addressing/nodes.rs
futurechimp: I need it in order to determine what length an address should be so that I can accomplish the gateway routing we talked about yesterday. Is there a different constant that would be better to depend on?

common/nymsphinx/src/addressing/nodes.rs
futurechimp: Is there a symbol for embarrassment?

gateway/src/listener.rs
futurechimp: Cool, it's a good call, give me a moment and I'll switch it.

nym-sphinx-wasm/src/lib.rs
futurechimp: Your passion for iterators is coming in handy :)

nym-sphinx-wasm/src/lib.rs
futurechimp: Yeah I'd rather not, if I can avoid it (same as the Sphinx thing I just mentioned in KeyBase). Is there an easy way we can avoid it? I do need a public key here, can I use `MixIdentityKey`, or would a type alias somewhere in `nymsphinx` help?

scripts/start_local_network.sh
futurechimp: The only downside of switching to Rust has been the slow compilation times. We can build this in release, but you may lose a team member to suicide. 

I can't see any particular reason that most people would want this built in release mode. Platform developers want the fastest compilation times. External developers who want to run a testnet for their own local development purposes (which is what I was doing yesterday) probably also just want it to work as quickly as they can get their own code running. 

If we needed to benchmark a local network, that's probably something we could handle pretty easily, and it's a less frequent task.

nym-sphinx-wasm/www/index.js
futurechimp: Which, the console.log?

common/nymsphinx/src/addressing/nodes.rs
futurechimp: Fixed. I reverted the public API change and used the Try that was, embarrassingly, already provided by a more thoughtful person.

common/nymsphinx/src/addressing/nodes.rs
futurechimp: Deleted.

gateway/src/listener.rs
futurechimp: Done

nym-sphinx-wasm/www/index.js
jstuczyn: "console.log("Retrieving topology...");"

gateway/src/mixnet_client.rs
futurechimp: As discussed I think this one is alright.

nym-sphinx-wasm/src/lib.rs
futurechimp: Iterated.

common/nymsphinx/src/addressing/nodes.rs
futurechimp: This is a re-export to avoid having the gateway depend directly on Sphinx.

nym-sphinx-wasm/www/index.js
futurechimp: All this js is going to get banged around real soon, no fear there

nym-sphinx-wasm/src/lib.rs
futurechimp: I'm not sure either. I guess the panic will spit back a serialization error message from Serde saying which field blew up, which is maybe all the user needs, so it will do the job?

nym-sphinx-wasm/src/lib.rs
futurechimp: I've mentioned this one in https://github.com/nymtech/nym/issues/184 so we don't miss it. Save it for a rainy day etc

PR #190, opened 2020-04-20
Feature/improve js example

futurechimp:


nym-client/js-examples/websocket/src/index.js
```diff
@@ -0,0 +1,108 @@
+var ourAddress;
+
+async function main() {
+    var port = '9001' // client websocket listens on 9001 by default, change if yours is different
```
jstuczyn, line 4: If it's a default, how about let's change main to `async function main(port = '9001')`?

futurechimp: I can't see any circumstances under which the person looking at the demo wouldn't need to modify their js file.  Let's just keep it 1998.

PR #242, opened 2020-05-19
Simplified some names and used the published npm package

futurechimp:


clients/webassembly/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "nym-client-wasm"
 authors = ["Dave Hrycyszyn <futurechimp@users.noreply.github.com>", "Jedrzej Stuczynski <andrew@nymtech.net>"]
-version = "0.7.0-pre"
+version = "0.7.2"
```
jstuczyn, line 4: So are we going away from same version for all?

clients/webassembly/client.js
```diff
@@ -197,14 +195,14 @@ export class Client {
...
+            this.onText(reader.result)
         };
 
         reader.readAsText(event.data);
     }
 
-    // Alternatively you may use default implementation and get everything as a text
-    onParsedBlobResponse(data) {
+    // Alternatively you may use default implementation and treat everything as text
+    onText(data) {
```
jstuczyn, line 205: As I said on keybase, I really don't like that name. It implies we received text from the gateway which is untrue. We just parsed whatever we received as text, which might very likely be bunch on unreadable or maybe even unprintable gibberish

clients/webassembly/js-example/package.json
```diff
@@ -20,12 +20,12 @@
...
+  "license": "Apache-2.0",
   "bugs": {
-    "url": "https://github.com/rustwasm/create-wasm-app/issues"
+    "url": "https://github.com/nymtech/nym/issues"
   },
-  "homepage": "https://github.com/rustwasm/create-wasm-app#readme",
+  "homepage": "https://nymtech.net/docs",
   "devDependencies": {
     "copy-webpack-plugin": "^5.0.0",
     "hello-wasm-pack": "^0.1.0",
```
jstuczyn, line 31: I think we don't need this one either?

clients/webassembly/Cargo.toml
futurechimp: I would like to keep the exact same version but I made 2 changes as I renamed things, so I had to republish. It's not possible to change an NPM package after publishing. I'll note in the docs that people should use the highest patch version of the minor version that the network is at.

clients/webassembly/client.js
futurechimp: It's text at this point, yes? We can explain it in the docs. Nobody wants a parsedBlobResponse, you see things like that you run like hell

futurechimp: What I mean is, if you send text through, by default you'll get text back. That's good enough for 99% of cases. I don't think there's any reason to expose the hideous underside of our binary parsing to users here.

clients/webassembly/Cargo.toml
jstuczyn: Cool. But we must be careful with this. To be honest I kinda assumed this situation would come up at some point

clients/webassembly/client.js
jstuczyn: We've already had lengthy discussion about it on keybase

PR #243, opened 2020-05-20
Very initial coconut API draft

jstuczyn:
Do not merge it. It's mostly some initial point for discussion


futurechimp:
I think we'll stick with the Go code for the moment, so I'll close this one now.

PR #260, opened 2020-06-07
add disabling feature 'offline-test' for network-dependent tests

ghost:
Some tests need network access, but builds are not seldomly performed in sandboxes which don't allow that.
This adds a feature `offline-test` which can be used to disable all tests that depend on network access.

In 0.6 this was only `it_returns_resolved_ip_on_resolvable_hostname`, which was easy to work around by giving the tests an alternative hosts file, but 0.7 added some tests that seem to have more complex network interactions.


futurechimp:
Thanks @hyperfekt :)

Issue #256, opened 2020-06-04
Client should try to reconnect to gateway on broken pipe


jstuczyn:
Currently if gateway goes down (or crashes), the client will continuously fail on:
``` 2020-06-04T09:25:10.082 ERROR nym_client::client::mix_traffic          > Failed to send sphinx packet to the gateway! - NetworkError(Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }))```
Instead it should either just fail and terminate or even better, try to reconnect after some backoff interval

futurechimp:
@jstuczyn is this still a thing? Please close if it's doing what it needs to do, you'll have a better handle on this one than me.

Issue #84, opened 2020-01-20
Reliable chunk transmission


futurechimp:
Once we have chunking working for both [send](#82)  and [receive](#83) we will want to ensure that we can do it all reliably. We'll probably need to implement reliability checks and some form of re-transmission protocol to ensure that the entire series of chunks arrives, and can be re-ordered properly.

PR #263, opened 2020-06-12
Feature/packet retransmission

### LINKED ISSUE #164: Fragment retransmission for split messages
Hypothetically if some message sent was split into 10,000 packets and even a single one of them was dropped at some point in the network, the receiver would never get anything and just be stuck with his 9,999 packets worth of memory used for garbage. We should introduce some re-transmission mechanism to request missing packets. Presumably SURBs would be perfect for that!

jstuczyn:
Closes https://github.com/nymtech/nym/issues/164
Initial take on https://github.com/nymtech/nym/issues/84 (is not fully implemented as described below)
With the limited testing this seems to work as we wanted. **However**, there's one limitation of the current approach. If you happen to send a mix message of length 21 bytes, it will not go through - it will be interpreted as an ACK instead. 
Do not worry though, this will be resolved with further enhancements we have discussed one of which includes gateway being unable to read plaintext message and padding on them (meaning "real" messages will always have one constant size and it will not be possible to confuse them with acks regardless of their content)

In this branch I've also slightly attacked https://github.com/nymtech/nym/issues/256 by telling the client to panic if it detects the gateway went down.


common/nymsphinx/src/utils/mod.rs
```diff
@@ -1,2 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use rand::Rng;
+use rand_distr::{Distribution, Exp};
+use std::time;
+
+// TODO: ask @AP why we are actually using Distribution::Exp(1/L) rather than just
```
jstuczyn, line 19: @aniampio I'm almost certain I've already asked this when we were initially working on Sphinx, but could you explain it again?

gateway/src/node/mixnet_handling/receiver/packet_processing.rs
```diff
@@ -163,15 +186,15 @@ impl PacketProcessor {
...
             }
             Ok(ProcessedPacket::ProcessedPacketFinalHop(client_address, _surb_id, payload)) => {
                 // in our current design, we do not care about the 'surb_id' in the header
                 // as it will always be empty anyway
-                let (payload_destination, message) = payload
-                    .try_recover_destination_and_plaintext()
-                    .ok_or_else(|| MixProcessingError::InvalidPayload)?;
+                let (payload_destination, message) =
+                    payload.try_recover_destination_and_plaintext()?;
+                // TODO: @AP, does that check still make sense?
```
jstuczyn, line 197: @aniampio

PR #272, opened 2020-06-29
Feature/gateway shared key generation

jstuczyn:
I think this branch finally reached some nice natural checkpoint where it could be merged into develop. So the added features include removal of 'AuthToken' in favour of a shared symmetric key that is established between client and gateway in a 'registration handshake' that happens on the first interaction between the entities. I also did an initial clean-up of the keys used, so proper external types are used now.

Note: to work correctly, this requires merging https://github.com/nymtech/nym-directory/pull/30 first.
Note2: wasm client does *NOT* work with the current iteration of the gateway.


clients/native/src/client/topology_control.rs
```diff
@@ -117,16 +117,16 @@ impl<T: NymTopology> TopologyAccessor<T> {
...
+    // pub(crate) async fn get_gateway_socket_url(&self, id: &str) -> Option<String> {
+    //     match &self.inner.read().await.0 {
+    //         None => None,
+    //         Some(ref topology) => topology
+    //             .gateways()
+    //             .iter()
+    //             .find(|gateway| gateway.identity_key == id)
+    //             .map(|gateway| gateway.client_listener.clone()),
+    //     }
+    // }
```
futurechimp, line 129: Is there a reason not to whack this commented code? It's always in source control if we want to go back to it, ja?

common/client-libs/directory-client/models/src/presence/gateways.rs
```diff
@@ -51,7 +53,8 @@ impl From<topology::gateway::Node> for GatewayPresence {
             location: mpn.location,
             client_listener: mpn.client_listener.to_string(),
             mixnet_listener: mpn.mixnet_listener.to_string(),
-            pub_key: mpn.pub_key,
+            identity_key: mpn.identity_key,
```
futurechimp, line 56: What is an "mpn"? Is it a possibly unrenamed "mix provider node"? Should it just be "gateway"?

clients/native/src/client/topology_control.rs
jstuczyn: The only reason I didn't whack it is because I forgot about it and it seems I didn't notice it before making the PR. It's gone now in my local branch.

common/client-libs/directory-client/models/src/presence/gateways.rs
jstuczyn: I think this is one of the remnants of when we just had providers and I wouldn't be surprised there were more cases of similar things. This is one of the things I'd ideally look at when refactoring `Topology`

PR #275, opened 2020-07-08
Added simple react example

keviinfoes:
Added simple react example in addition to the basic js example.


futurechimp:
Thanks very much @keviinfoes :). It looks so much better!

PR #312, opened 2020-08-20
Snake casing the surbs.

futurechimp:


jstuczyn:
As I was saying before - you need to update all example code because they read from JSON using the old name

jstuczyn:
Superseded by https://github.com/nymtech/nym/pull/322

PR #320, opened 2020-09-05
the port in the announce address is not optional

timkuijsten:
Remove a misleading comment about the port in the announce address
being optional and defaulting to the listening port. Testing shows
this is not the case. The dashboard shows the announce address
without the port and no packets are received by nym-mixnode.

Of course a better fix would be to make that the listening port is used in case one is omitted in the announce address.

### DIFF
--- a/mixnode/src/config/template.rs
+++ b/mixnode/src/config/template.rs
@@ -50,9 +50,7 @@ public_sphinx_key_file = '{{ mixnode.public_sphinx_key_file }}'
 # Optional address announced to the directory server for the clients to connect to.
 # It is useful, say, in NAT scenarios or wanting to more easily update actual IP address
 # later on by using name resolvable with a DNS query, such as `nymtech.net:8080`.
-# Additionally a custom port can be provided, so both `nymtech.net:8080` and `nymtech.net`
-# are valid announce addresses, while the later will default to whatever port is used for
-# `listening_address`.
+# Additionally a custom port can be provided.
 announce_address = '{{ mixnode.announce_address }}'
 
 # Directory server to which the server will be reporting their presence data.


jstuczyn:
Can you show me how you tried to run it? With: `./nym-mixnode init --host 127.0.0.1 --id foo --announce-host custom-foo-domain.com --layer 1` followed by usual `./nym-mixnode run --id foo` the dashboard shows my host as `custom-foo-domain.com:1789`. 

Same is true for: `./nym-mixnode init --host 127.0.0.1 --port 42 --id foo --announce-host custom-foo-domain.com --layer 1` with the dashboard showing `custom-foo-domain.com:42`.

timkuijsten:
I edited the config file afterwards by hand. I didn't add the --announce-host while I was running `nym-mixnode init` but later, after reading the generated config file, decided to wrap a domain name around my ip.

On a related note, I've just updated to nym-mixnode v0.8.0 (with my [pledge(2)](https://man.openbsd.org/pledge.2) patches for OpenBSD) and re-tried with an announce host without a port to see if the problem still exists, so `announce_address = 'tosk.netsend.nl'`. To my great surprise, not only my node but all nodes listed on https://dashboard.nymtech.net go down to 0 packets sent and received within a couple of seconds. As soon as I change my config back to `announce_address = 'tosk.netsend.nl:1789'` and restart my node, the whole network starts receiving and sending packets again. 🐙

timkuijsten:
I edited the config file afterwards by hand. I didn't add the --announce-host while I was running `nym-mixnode init` but later, after reading the generated config file, decided to wrap a domain name around my ip. So at first I manually enabled announce_address without a port, which didn't work.

I see that if you create the config file with the --announce-host option it does write the port to the announce_address property in the config file and the problem does not arise. But the port seems required in the config file, as my patch suggests. The problem still exists in v0.8.0.

jstuczyn:
Since https://github.com/nymtech/nym/pull/654 ports and host are explicitly separate and this pull request does not apply anymore.

PR #339, opened 2020-09-11
Making layer optional and generating random in 1-3 range

jstuczyn:


jstuczyn:
This should be changed in favour of the solution we discussed today, i.e. mixnode looking up topology on startup and choosing layer with fewest nodes.

PR #355, opened 2020-09-28
Selecting the gateway randomly from 2 known good ones.

futurechimp:
There's a problem with the config file in this commit though - the written file comes out with a blank `gateway_id`. I don't completely understand how the config file overrides work, but my guess is that something is using `matches` instead of `gateway_id`


jstuczyn:
Should close https://github.com/nymtech/nym/issues/352

PR #381, opened 2020-10-13
Feature/upgrade command

### LINKED ISSUE #367: Introduce an explicit `upgrade` command
Each time we release a new version, people are running `init` again and blowing their existing config values. 
`upgrade` should instead re-init your configs but read old values first.

jstuczyn:
Closes https://github.com/nymtech/nym/issues/367


futurechimp:
Supporting `dev` or `rc` versions in here seems like a recipe for madness, no? I suspect it might be simpler to say: 

1. we never upgrade from a release version (like `0.8.1`) to a non-supported pre-release version (like `0.9.0-dev` or `-rc1`). 
2. attempting to upgrade from a `-dev` or `-rcX` version using the `upgrade` command will similarly fail.

I'd expect that we'd put a lot of weird experimental stuff into the config files during developement (and just as often take it out out, rename variables etc). And we probably don't want to accidentally tempt people into wrecking their happy and valuable configs by upgrading them to some momentarily messed up version of e.g. `0.10.0-dev`.

PR #386, opened 2020-10-14
Feature/double init prevention

### LINKED ISSUE #368: Change `init` to not blow away existing keys (if exist)
Instead introduce extra `init` flag like `--purge-existing-keys` that would force fresh keygen.

jstuczyn:
Closes https://github.com/nymtech/nym/issues/368


futurechimp:
This does the job, I think we should chat with Alexis and maybe Hans Bricks about whether more messaging is needed in the case of people first-time initing a node, wanting to change something about their config, etc. But we can handle that in a separate issue if need be.

PR #401, opened 2020-10-28
Feature/directory server transition

### LINKED ISSUE #370: Remove presence notifications
We should remove presence notifications from all Rust nodes, we can now track things through node announcements in the blockchain.

### LINKED ISSUE #371: Mixnode and gateway blockchain registration
Mixnodes and gateways should register themselves at the validators when they first start up. 

We should include incentivization info in the registration, to take some load off of Alexis - so we can automatically tie node uptime to nymph and Liquid accounts. We can have a chat to figure out the details here.

### LINKED ISSUE #383: Gateway announcement
Like mixnodes, gateways are going to need to register with the validators so that they show up in the network topology. 

This will require:

[ ] some work in a validator client (replacing the directory client) so that the gateway makes this announcement
[ ] a Gateway entity in the blockchain which can store the Gateway info

### LINKED ISSUE #390: Clients should use only "active" nodes
We're going to change things in the next release so that only nodes that are configured properly are included in the active topology returned by the validator directory. 

Once we know what the URL path of that route's going to be, we will need to change the Nym clients (and anything topology-related) to get info from the active set, rather than the whole set of registered (but possibly misconfigured) nodes.

jstuczyn:
Closes https://github.com/nymtech/nym/issues/383
Closes https://github.com/nymtech/nym/issues/390
Closes https://github.com/nymtech/nym/issues/371
Closes https://github.com/nymtech/nym/issues/370

On top of that mixnodes are now using a separate metrics server.


jstuczyn:
Ah, hold your horses. Forgot about `upgrade` compatibility for clients.

PR #412, opened 2020-11-04
Feature/network monitor file topology

jstuczyn:


jstuczyn:
Going to pass validator as attribute. one second.

PR #414, opened 2020-11-04
Updated default validator endpoint

jstuczyn:


jstuczyn:
This thing is constantly changing so closing PR for time being until we decide what the correct solution is going to be

PR #416, opened 2020-11-05
Increased default monitor interval from 1min to 5min

jstuczyn:


futurechimp:
I think 1 minute is probably better to stick with, otherwise it's a long time before node operators can get a sense of how they are doing.

PR #437, opened 2020-11-11
Changing default validator location to make it clear we're testnet

### LINKED ISSUE #432: Change default directory location
At the moment we've got default validator set to https://validator.nymtech.net (not in use). The actual validator address that we're really running is https://testtnet-validator1.nymtech.net

futurechimp:
Closes #432


jstuczyn (changes requested):
My comments from https://github.com/nymtech/nym/issues/432 still stand - what about SSL and port?

PR #493, opened 2021-01-04
Feature/GitHub actions and clippy cleanup

jstuczyn:


jstuczyn:
~I would say the nightly failure is acceptable as I don't think it makes complete sense. To fix it we would basically need to modify Sphinx crate.~ - yeah, that was not true at all.

futurechimp:
A few questions then:

a) is this due to the introduction of Clippy, or is it unrelated to that and just happened to appear? 
b) Isn't nightly automatically going to become a problem if we just leave it?
c) isn't this going to show as a breaking build forever?

PR #500, opened 2021-01-19
Fixed possible crash on invalid topology

jstuczyn:
Fixes crash described in https://github.com/nymtech/nym/issues/494


jstuczyn:
~The nightly clippy failure seems to refer to code in network monitor that is getting completely changed in the other pull request so I think it can be safely ignored this time.~ no longer applies.

PR #501, opened 2021-01-19
Feature/wasm client compilation fixes

### LINKED ISSUE #458: Gateway reconnection (wasm)
Implement simple gateway reconnections for wasm.

jstuczyn:
I've fixed the compilation warnings in the wasm client. Actually, when I think about it, it should also close https://github.com/nymtech/nym/issues/458


jstuczyn:
~This clippy failure is the same one as described in https://github.com/nymtech/nym/pull/500 - it refers to code in network monitor that is getting completely changed in the other pull request (https://github.com/nymtech/nym/pull/499) so I think it can be safely ignored this time.~ no longer applies

PR #506, opened 2021-01-27
Add Dockerfile

Legogris:
This is intended as a base image to extend on.

`nym-*` binaries are available in `/usr/local/bin`, built from local source.


wouterds:
Could be interesting FYI: https://github.com/wouterds/nym-mixnode-docker

durch:
Closing, we have some alternative Docker configs now, thank you.

PR #545, opened 2021-03-24
Feature/has node query validator client

jstuczyn:
Cannot be merged until the following are done:

1. ~~https://github.com/nymtech/nym/pull/544 is merged first~~ - merged by @futurechimp 
2. ~~I need to setup my local validator to see whether it actually works.~~ - all seems to work fine locally


futurechimp:
#544 is now merged.

PR #563, opened 2021-04-08
A pull request for discussion about contract state variables

futurechimp:


jstuczyn:
~~I've literally just noticed - the pull request is not fully complete. It's still using the constant values for bonding rather the one in the state. I'm putting it in "draft" mode until I fix this omission.~~
Edit: done.

PR #569, opened 2021-04-13
Adding the wallet url in startup instructions

futurechimp:


jstuczyn:
It seems the nightly clippy was updated and produced some warnings on old code - they should be dealt with in separate pull request independent of this tiny change here.

PR #571, opened 2021-04-14
Feature/simple payments

jstuczyn:
We probably don't want to include those changes in this release so I'm leaving this as a draft for time being, but it should be merge-able


clients/validator/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nymproject/nym-validator-client",
-  "version": "0.10.0-rc5",
+  "version": "0.10.0-rc6",
```
jstuczyn: Change that to 0.10.1

futurechimp:
This is good to merge as soon as conflicts are resolved.

PR #586, opened 2021-04-22
Checking for tx success when sending coins

jstuczyn:


jstuczyn:
I've converted it into a draft as before we merge it, we would need to increase the version number of the package. And it is going to depend on when we are going to perform the merge.

PR #605, opened 2021-05-15
Feature/add node description api

futurechimp:


common/mixnode-common/src/rtt_measurement/mod.rs
```diff
@@ -33,7 +33,7 @@ pub(crate) mod packet;
 pub(crate) mod sender;
 
 // TODO: MUST BE UPDATED BEFORE ACTUAL RELEASE!!
-pub const MINIMUM_NODE_VERSION: &str = "0.10.1";
+pub const MINIMUM_NODE_VERSION: &str = "0.10.1-dev";
```
jstuczyn: Why are we putting the dev suffix for the version in here? It's not like we would want to keep this later on and in semver 0.0.0-foo < 0.0.0 anyway so we're all fine

mixnode/src/commands/describe.rs
```diff
@@ -0,0 +1,55 @@
...
+        .arg(
+            Arg::with_name("id")
+                .long("id")
+                .help("The id of the mixnode you want to describe")
+                .takes_value(true)
+                .required(true),
+        )
+}
+
+pub fn execute(matches: &ArgMatches) {
```
jstuczyn, line 21: I think it should also first try to load the configuration file to make sure the node actually exists before attempting to describe it.
As a side node, perhaps the functionality of `describe` should also be available during `init`? Maybe as flags or something?

mixnode/src/commands/describe.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let node_description = NodeDescription {
+        name,
+        description,
+        link,
+    };
+
+    // save the struct
+    let config_path = Config::default_config_directory(id);
+    NodeDescription::save_to_file(&node_description, config_path).unwrap();
```
jstuczyn: I would argue that NodeDecription should be part of the Config itself rather than a separate file

mixnode/src/node/http/description.rs
```diff
@@ -0,0 +1,10 @@
+use crate::node::node_description::NodeDescription;
+use rocket::State;
+use rocket_contrib::json::Json;
+
+/// Provides verifiable location (verloc) measurements for this mixnode - a list of the
```
jstuczyn: This is not verloc : )

mixnode/src/node/node_description.rs
```diff
@@ -0,0 +1,40 @@
...
+        NodeDescription {
+            name: "This node has not yet set a name".to_string(),
+            description: "This node has not yet set a description".to_string(),
+            link: "https://nymtech.net".to_string(),
+        }
+    }
+}
+
+impl NodeDescription {
+    pub(crate) fn load_from_file(mut config_path: PathBuf) -> io::Result<NodeDescription> {
```
jstuczyn: should this (and `save_to_file`) be given `config_path` as arguments? Should it rather be `node_description_path`? Also, as I've mentioned before, I think it should be part of config itself.

mixnode/src/node/node_description.rs
```diff
@@ -0,0 +1,40 @@
...
+        config_path.push("descriptor.json");
+        let json = fs::read_to_string(config_path)?;
+        ::serde_json::from_str(&json)
```
jstuczyn: Any particular reason for the strange mixture of json and toml? I.e. we're saving node configuration in a `.toml` file but for some reason we save description as a `.json`?

jstuczyn (changes requested):

mfahampshire (approved):

mixnode/src/commands/describe.rs
```diff
@@ -20,7 +21,17 @@ pub fn command_args<'a, 'b>() -> App<'a, 'b> {
 
 pub fn execute(matches: &ArgMatches) {
     // figure out which node the user is describing
-    let id = matches.value_of("id").unwrap();
+    let id = matches
+        .value_of("id")
+        .expect("Please provide the id of your mixnode");
+
+    // ensure that the mixnode has in fact been initialized
```
jstuczyn, line 27: I think it would be better to attempt to load the config explicitly `Config::load_from_file(id)` so then you don't need to do any checks for directory existence etc. And also protects us from an edge case where user deleted the node content but preserved the directory

common/mixnode-common/src/rtt_measurement/mod.rs
futurechimp: Heheh I was under the impression that you'd explicitly asked for that - I'll just knock it out again.

mixnode/src/commands/describe.rs
futurechimp: I feel like there's already enough going on during `init`, adding three more fields to that, or somehow grabbing keyboard input during that process, starts to feel a bit heavyweight to me. But I'm really not sure - let's see how people deal with `describe` and then make some further decisions once we can see some real usage. 

I have added a check for the directory at start now

common/mixnode-common/src/rtt_measurement/mod.rs
futurechimp: Actually now I remember why - if you don't update it, trying to `run` the node fails with: 

```
 2021-05-19T12:56:39.024Z ERROR nym_mixnode::node           > You seem to have not updated your mixnode configuration file - please run `upgrade` before attempting again
```

PR #612, opened 2021-05-20
Bugfix/unique node ownership

jstuczyn:
@futurechimp Before merging it, let's have a discussion about what to do with the migration. I see two options here:
1. Don't do anything, the problem _should_ solve itself as people unbond and rebond to update. 
2. During migration iterate through all bonded mixnodes and (assuming they are in order which I'm not sure of), use the first entry as ownership and insert that data.


futurechimp:
Let's talk about it tomorrow.

PR #649, opened 2021-06-16
Bugfix/delegation type fixes

jstuczyn:
Fixes issue where delegation owner (i.e. person who delegated stake) would mistakenly be referred to as "node_identity"


common/mixnet-contract/src/delegation.rs
```diff
@@ -1,40 +1,38 @@
...
+        Delegation { owner, amount }
     }
 
     pub fn amount(&self) -> &Coin {
         &self.amount
     }
 
-    pub fn node_identity(&self) -> &String {
-        &self.node_identity
+    pub fn owner(&self) -> &HumanAddr {
```
durch: Do we want to return `HumanAddr` here, it usually easier if a function returns an owned type

durch (approved):
Looks good, just a question around `owner()`

common/mixnet-contract/src/delegation.rs
jstuczyn: I've just looked at the usages of the `owner()` call and you're absolutely right. It would definitely be way easier as in all of the places the call is followed by a `clone()` anyway.

PR #650, opened 2021-06-16
Introduced type aliases for mixnode and gateway keys

jstuczyn:
It's unfortunately not the full-blown type assertion, merely type aliases, but I think it makes contract code more readable to be able to figure out whether a `String` refers to particular key.


durch (approved):
Looks good. I'm not sure if we have any format requirements on either of these, in case we need any sort of special handling we could abstract this further with a regular struct: 

```rust
struct IdentityKey {
    value: String
}
```

In this case they'd need to be explicitly constructed, so it would add another layer of verification

jstuczyn:
Yeah... In my ideal world this would have been `identity_key: crypto::asymmetric::identity::PublicKey` (or a wrapper for this type). I even finished writing the code for it which compiles and runs* (with a big asterisk). However, the wasmVM does not like it as one of the dependencies used by the curve libraries involves some undeterminism (my bet would be on `rand`)...

So since we're not able to fully check whether the key is a valid curve point, I don't think there's a point in doing some half measures, because realistically what we could only check at this point is whether the string is a valid base58 string which upon decoding is 32 bytes long. Unless you think this would still be worth it. (cc @durch @futurechimp)

Oh, and as for the construction, I think it is never explicitly constructed, it's deserialized by the contract from user transaction content (which currently only comes from our typescript code)

PR #655, opened 2021-06-24
Feature/describe location

neacsu:
Give a mixnode operator the possibility to `describe` the location of its mixnode. Also removed the location from the `init` output.

### DIFF
--- a/mixnode/src/commands/describe.rs
+++ b/mixnode/src/commands/describe.rs
@@ -48,6 +48,7 @@ pub fn execute(matches: &ArgMatches) {
     let description = desc_buf.trim().to_string();
 
     let example_url = "https://mixnode.yourdomain.com".bright_cyan();
+    let example_location = "City: YourCity, Country: YourCountry";
 
     print!("link, e.g. {}: ", example_url);
     io::stdout().flush().unwrap();
@@ -55,10 +56,17 @@ pub fn execute(matches: &ArgMatches) {
     io::stdin().read_line(&mut link_buf).unwrap();
     let link = link_buf.trim().to_string();
 
+    print!("location, e.g. {}: ", example_location);
+    io::stdout().flush().unwrap();
+    let mut location_buf = String::new();
+    io::stdin().read_line(&mut location_buf).unwrap();
+    let location = location_buf.trim().to_string();
+
     let node_description = NodeDescription {
         name,
         description,
         link,
+        location,
     };
 
     // save the struct
--- a/mixnode/src/commands/describe.rs
+++ b/mixnode/src/commands/describe.rs
@@ -48,7 +48,7 @@ pub fn execute(matches: &ArgMatches) {
     let description = desc_buf.trim().to_string();
 
     let example_url = "https://mixnode.yourdomain.com".bright_cyan();
-    let example_location = "City: YourCity, Country: YourCountry";
+    let example_location = "City: London, Country: UK";
 
     print!("link, e.g. {}: ", example_url);
     io::stdout().flush().unwrap();
--- a/mixnode/src/commands/run.rs
+++ b/mixnode/src/commands/run.rs
@@ -163,7 +163,6 @@ pub fn execute(matches: &ArgMatches) {
     Address: {}
     Mix port: {}
     Layer: {}
-    Location: {}
     Version: {}
     ",
         identity_keypair.public_key().to_base58_string(),
@@ -171,7 +170,6 @@ pub fn execute(matches: &ArgMatches) {
         config.get_announce_address(),
         config.get_mix_port(),
         config.get_layer(),
-        description.location,
         config.get_version(),
     );
     MixNode::new(config, description, identity_keypair, sphinx_keypair).run();
--- a/mixnode/src/commands/run.rs
+++ b/mixnode/src/commands/run.rs
@@ -140,6 +140,9 @@ pub fn execute(matches: &ArgMatches) {
         show_binding_warning(config.get_listening_address().to_string());
     }
 
+    let description =
+        NodeDescription::load_from_file(Config::default_config_directory(id)).unwrap_or_default();
+
     println!(
         "Validator servers: {:?}",
         config.get_validator_rest_endpoints()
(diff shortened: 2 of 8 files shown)


mixnode/src/commands/describe.rs
```diff
@@ -47,17 +47,25 @@ pub fn execute(matches: &ArgMatches) {
     let description = desc_buf.trim().to_string();
 
     let example_url = "https://mixnode.yourdomain.com".bright_cyan();
+    let example_location = "City: YourCity, Country: YourCountry";
```
jstuczyn: I'd change it to a more concrete example like "London, UK", as people will enter the literal string of "YourCity, YourCountry"

mixnode/src/commands/run.rs
```diff
@@ -172,17 +175,15 @@ pub fn execute(matches: &ArgMatches) {
     Sphinx key: {}
     Host: {}
     Layer: {}
-    Location: [physical location of your node's server]
+    Location: {}
```
jstuczyn: Isn't location going away from bonding hence this shouldn't be shown anymore?

mixnode/src/commands/upgrade.rs
```diff
@@ -184,12 +186,12 @@ fn pre_090_upgrade(from: &str, config: Config) -> Config {
...
 }
 
-fn minor_010_upgrade(
+fn minor_0_10_upgrade(
     config: Config,
     _matches: &ArgMatches,
     config_version: &Version,
     package_version: &Version,
-) -> Config {
+) -> Result<Config, (Version, String)> {
```
jstuczyn: I like the `upgrade` command returning a `Result`, but I'm not a fan of this tuple error type. Perhaps some proper `UpgradeError` enum would be better?

mixnode/src/commands/upgrade.rs
```diff
@@ -244,15 +245,87 @@ fn patch_010_upgrade(
...
+    }
+    let to_version = package_version;
+    let id = config.get_id();
+    let config_path = Config::default_config_directory(&id);
+
+    print_start_upgrade(&config_version, &to_version);
+
+    let upgraded_config = config.with_custom_version(to_version.to_string().as_ref());
+
+    upgraded_config.save_to_file(None).map_err(|err| {
```
jstuczyn: I wouldn't save upgraded config to a file until you've verified you managed to migrate the description as if that fails, the user's node might end up in a broken state.

mixnode/src/commands/upgrade.rs
```diff
@@ -244,15 +245,87 @@ fn patch_010_upgrade(
...
+        (
+            to_version.clone(),
+            format!("failed to overwrite config file! - {:?}", err),
+        )
+    })?;
+
+    let description_file_path: PathBuf = [config_path.to_str().unwrap(), DESCRIPTION_FILE]
```
jstuczyn, line 337: You can just use a `join()` on Path/PathBuf rather than create an array and convert the iterator into a PathBuf

mixnode/src/commands/upgrade.rs
```diff
@@ -244,15 +245,87 @@ fn patch_010_upgrade(
...
+                    format!("failed to deserialize description content! - {:?}", err),
+                )
+            })?;
+        let new_description = NodeDescription {
+            name: old_description.name,
+            description: old_description.description,
+            link: old_description.link,
+            ..Default::default()
+        };
+        let description_toml = toml::to_string(&new_description).map_err(|err| {
```
jstuczyn: Here you can just use `NodeDescription::save_to_file` which will do toml serialization for you

mixnode/src/commands/upgrade.rs
```diff
@@ -301,3 +383,39 @@ pub fn execute(matches: &ArgMatches) {
...
+        let old_version = Version::new(0, 10, 1);
+        let new_version = Version::new(0, 10, 2);
+        let new_config =
+            patch_0_10_2_upgrade(config, &matches, &old_version, &new_version).unwrap();
+        assert_eq!(new_config.get_version(), "0.10.2");
+    }
+
+    #[test]
+    #[serial]
+    fn test_patch_0_10_2_upgrade_error() {
```
jstuczyn, line 469: This tests makes me potentially nervous as it saves data where you might already have an initialised node. But I'm also not sure what would be a good alternative for it.

mixnode/src/commands/describe.rs
neacsu: Done.

mixnode/src/commands/run.rs
neacsu: You're right, it doesn't make sense anymore to have it here. I'll remove it.

mixnode/src/commands/upgrade.rs
neacsu: Do you mean a type alias for the error type? If so, then I agree, it would look better, I can add it in a following commit.

mixnode/src/commands/upgrade.rs
neacsu: Moved it towards the end of the function.

mixnode/src/commands/upgrade.rs
neacsu: According to the [man page](https://doc.rust-lang.org/std/path/struct.PathBuf.html), `join` (as well as `push`) should be used for dynamic situations. They give this iterator to `PathBuf` as example, which fits as well in our situation.

mixnode/src/commands/upgrade.rs
neacsu: Will do.

mixnode/src/commands/upgrade.rs
neacsu: I'll replace the id from `0` to `-42`, in the hope that someone running these tests won't think to use that magic number anytime soon :D . A longer term solution would be to have a configurable config directory.

mixnode/src/commands/upgrade.rs
jstuczyn: Fair enough - I haven't realised that

jstuczyn (approved):

PR #658, opened 2021-06-28
Use thiserror for validator-client

durch:


jstuczyn:
You will probably either want to either remove the comment I made regarding `CodedError` or move it to the new definition. Or ideally find out what's the proper type for that. The comment in question is:
```
// this is the case of message like
/*
{
  "code": 12,
  "message": "Not Implemented",
  "details": [
  ]
}
 */
// I didn't manage to find where it exactly originates, nor what the correct types should be
// so all of those are some educated guesses
```

durch:
```
> {
>   "code": 12,
>   "message": "Not Implemented",
>   "details": [
>   ]
> }
 ```

I've actually seen this today, it seems to pop if you hit an unimplemented route, for example if you just hit `/smart` you end up with `501 - Not implemented`. I'll take a look at validator code

jstuczyn:
Yep, that's exactly how I encountered it. I tried to find where it originated in the validator codebase, but I could not find it anywhere

jstuczyn (approved):

neacsu (approved):

PR #659, opened 2021-06-28
contracts: Removed mixnode location from the bonding process

neacsu:
Signed-off-by: Bogdan-Ștefan Neacșu <bogdan@nymtech.net>


jstuczyn (approved):
Looks good to me

PR #660, opened 2021-06-28
Feature/total delegation field

### LINKED ISSUE #657: Rename `PagedResponse` into `PagedMixnodeResponse`
Originally the paged responses for mixnodes were named `PagedResponse` as it was the only thing using paging. However, currently we also get information about gateways and delegations in the same way and hence the naming might be a bit confusing.

This should probably be done alongside another ticket as it's a very straightforward thing.

jstuczyn:
Closes https://github.com/nymtech/nym/issues/657


contracts/mixnet/src/transactions.rs
```diff
@@ -3,15 +3,63 @@ use crate::error::ContractError;
...
+// 5. The node rebonds with the same identity
+fn find_old_delegations(delegations_bucket: ReadonlyBucket<Uint128>) -> StdResult<Coin> {
+    // I think it's incredibly unlikely to ever read more than that
+    // but in case we do, we should guard ourselves against possible
+    // out of memory errors (wasm contracts can only allocate at most 2MB
+    // of RAM, so we don't want to box the entire iterator)
+    let mut total_delegation = Coin::new(0, DENOM);
+    let mut start = None;
+    loop {
+        let iterator = delegations_bucket
```
durch, line 29: Superficially worried about looping in a loop, but not too much

contracts/mixnet/src/transactions.rs
```diff
@@ -1511,7 +1667,8 @@ pub mod tests {
 
         let initial_bond = 100_000000;
```
durch, line 1668: Maybe change this to `100_000_000` for readability

durch, line 1661: Maybe change this to `100_000_000` for readability

durch (approved):
Looks good, but I'd like another set of eyes I don't think I have enough context to be 100% sure

contracts/mixnet/src/transactions.rs
jstuczyn: Yeah, ideally I would have avoided it, but this particular iterator doesn't have `size_hint()` properly defined hence I do not know its size beforehand. Also, I couldn't iterate everything at once as when the iterator is created, everything is being allocated and in the worst case scenario we would have gone over the 2MB limit of the wasmVM. 

I hope that the test I've written for it would have caught any infinite loop issues

contracts/mixnet/src/transactions.rs
jstuczyn: I've actually left it this as `100_000000` for readability as in the contract we operate on the micro-currencies. Hence this would imply `100` of the currency "proper". But I don't mind changing it if you think it'd be better

contracts/mixnet/src/transactions.rs
```diff
@@ -3,15 +3,63 @@ use crate::error::ContractError;
...
-    gateways_owners_read, gateways_read, increase_gateway_bond, increase_gateway_delegated_stakes,
-    increase_mix_delegated_stakes, increase_mixnode_bond, increment_layer_count, mix_delegations,
-    mixnodes, mixnodes_owners, mixnodes_owners_read, mixnodes_read, read_gateway_epoch_reward_rate,
-    read_mixnode_epoch_reward_rate, read_state_params, Layer,
+    config, config_read, decrement_layer_count, gateway_delegations, gateway_delegations_read,
+    gateways, gateways_owners, gateways_owners_read, gateways_read,
+    increase_gateway_delegated_stakes, increase_mix_delegated_stakes, increment_layer_count,
+    mix_delegations, mix_delegations_read, mixnodes, mixnodes_owners, mixnodes_owners_read,
+    mixnodes_read, read_gateway_epoch_reward_rate, read_mixnode_epoch_reward_rate,
+    read_state_params, Layer,
```
neacsu: Why not change it to `storage::*`?

contracts/mixnet/src/transactions.rs
```diff
@@ -512,6 +646,22 @@ pub(crate) fn try_remove_delegation_from_gateway(
...
+            {
+                // we should NEVER underflow here, if we do, it means we have some serious error in our logic
+                existing_bond.total_delegation.amount = existing_bond
+                    .total_delegation
+                    .amount
+                    .checked_sub(delegation)
+                    .unwrap();
+                gateways_bucket.save(gateway_identity.as_bytes(), &existing_bond)?;
+            }
+
```
neacsu, line 657: This (and the rest of the function) is pretty much duplicated code from the `try_remove_delegation_from_mixnode` function above. Maybe we should consider putting the mixnodes and gateways more and more together under some sort of inheritance pattern in the future. But probably now's not the time for it.

neacsu (approved):

contracts/mixnet/src/transactions.rs
jstuczyn: Yeah, gateways and mixnodes have a lot of duplicate code, maybe we should do something about that in the future. Though, I'm slightly hesitant towards it as back in the past when we had centralised directory server, we've been doing exactly that and it's been a pain after a while

contracts/mixnet/src/transactions.rs
jstuczyn: Yeah, I guess that would clean it up a bit

PR #664, opened 2021-06-30
Layer selection

neacsu:


jstuczyn:
I see you interpreted the ticket slightly differently than what I initially had in mind. I'm not saying it's good or wrong. I guess we would need an external pair of eyes on it, @durch or @futurechimp . What I had in mind was to completely get rid of `layer` argument from the mixnode side (i.e. remove it from the data you sent to the contract) - in the long run mixnode should have no say in what layer it's gonna reside on because it might change dynamically.

So basically when mixnode bonds, contract says "hey, thanks for bonding, you're now on layer X"

neacsu:
> What I had in mind was to completely get rid of layer argument from the mixnode side (i.e. remove it from the data you sent to the contract)

Yep, I thought of that originally too. I kind of changed my mind on the way though, as it seemed to offer more flexibility for the mixnode operator. 

> in the long run mixnode should have no say in what layer it's gonna reside on because it might change dynamically.

If that is the case, then I can remove it completely now and be done with it

mixnode/src/commands/run.rs
```diff
@@ -167,21 +161,25 @@ pub fn execute(matches: &ArgMatches) {
         "Announcing the following address: {}",
         config.get_announce_address()
     );
+    // TODO: Currently layer is always None in the mixnode and should be
+    // extracted at some point after the bonding process, by querying a
+    // validator
+    if let Some(layer) = config.get_layer() {
```
jstuczyn: Do we even need to store that in the config?

mixnode/src/commands/upgrade.rs
```diff
@@ -381,6 +381,32 @@ fn patch_0_10_2_upgrade(
     Ok(upgraded_config)
 }
 
+fn patch_0_10_3_upgrade(
```
jstuczyn: I think this should be put together to the `undetermined_version_upgrade` as technically we haven't made any release since 0.10.2

mixnode/src/config/mod.rs
```diff
@@ -360,7 +360,8 @@ pub struct MixNode {
     id: String,
 
     /// Layer of this particular mixnode determining its position in the network.
-    layer: u64,
+    /// Before getting the actual layer from the validators, this is marked as None.
```
jstuczyn: As I asked before, do we even need it? (I'm not 100% sure myself, but I kinda don't see any reasons for it)

jstuczyn (changes requested):
One additional comment: we could now remove `LayerDistribution.invalid` field as it will be impossible to set it. Perhaps it might be worth to have a `choose_with_fewest` method on it to more easily find appropriate layer?

mixnode/src/config/mod.rs
neacsu: Removed it.

mixnode/src/commands/run.rs
neacsu: Probably not, so I'm going to remove it and we'll see later if we need it or not.

mixnode/src/commands/upgrade.rs
neacsu: Got it,  will put everything in that one until we make the next release.

common/mixnet-contract/src/lib.rs
```diff
@@ -18,6 +18,17 @@ pub struct LayerDistribution {
     pub invalid: u64,
```
jstuczyn: So are we not removing it? : )

common/mixnet-contract/src/mixnode.rs
```diff
@@ -13,27 +14,36 @@ pub struct MixNode {
...
     pub http_api_port: u16,
-    pub layer: u64,
     pub sphinx_key: SphinxKey,
     /// Base58 encoded ed25519 EdDSA public key.
     pub identity_key: IdentityKey,
     pub version: String,
 }
 
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Serialize, JsonSchema)]
+pub enum Layer {
```
jstuczyn, line 24: More type-safety - I like it

jstuczyn (approved):

common/mixnet-contract/src/lib.rs
neacsu: Yes we are :) Just one more (tiny) commit :)

PR #665, opened 2021-06-30
Validator API server

durch:
+ Adds a specific cache implementation
+ Establishes `rocket` instance and state management
+ Runs on `8080` by default, on `127.0.0.1` in debug and `0.0.0.0` in release mode, both configurable via `Rocket.toml`


validator-api/Rocket.toml
```diff
@@ -0,0 +1,20 @@
+[development]
```
jstuczyn: Should this file be even committed by default?

validator-api/src/cache.rs
```diff
@@ -0,0 +1,34 @@
+use anyhow::Result;
+use mixnet_contract::MixNodeBond;
+use std::time::Instant;
+use validator_client::Client;
+
+pub struct MixNodeCache {
+    value: Vec<MixNodeBond>,
+    as_at: Instant,
```
jstuczyn, line 9: Hm, just a suggestion, how about changing `as_at` from `Instant` to a `u64` value, where that value would be the height of the chain when we grabbed that information? We can get it from, for example in the case of our validator `https://testnet-finney-validator.nymtech.net/block` (if we don't specify arguments for block, we get the latest one, and then we can grab it from `block.header.height`)

jstuczyn: Also, temporarily for linter sake, just slap a `#[allow(dead_code)]` on top of it.

validator-api/src/cache.rs
```diff
@@ -0,0 +1,34 @@
...
+    validator_client: Client,
+}
+impl MixNodeCache {
+    pub fn init(value: Vec<MixNodeBond>, validator_client: Client) -> Self {
```
jstuczyn: Since presumably we will have identical cache for gateways (or maybe  they should be kept together?), should we make it own a `validator_client`?

validator-api/src/main.rs
```diff
@@ -35,6 +48,7 @@ const NODE_STATUS_API_ARG: &str = "node-status-api";
 const DETAILED_REPORT_ARG: &str = "detailed-report";
 const GATEWAY_SENDING_RATE_ARG: &str = "gateway-rate";
 const MIXNET_CONTRACT_ARG: &str = "mixnet-contract";
+const CACHE_INTERVAL: &str = "cache-interval";
```
jstuczyn: Add an `_ARG` suffix to be more explicit please : )

validator-api/src/monitor/mod.rs
```diff
@@ -137,9 +136,9 @@ impl Monitor {
     }
 
     async fn test_run(&mut self) {
-        info!(target: "Monitor", "Starting test run no. {}", self.nonce);
+        info!("Starting test run no. {}", self.nonce);
```
jstuczyn: This is just a question, what is the reason for removing the `target` part of the macro?

jstuczyn (changes requested):

validator-api/src/monitor/mod.rs
durch: Great that you raised this, `rocket` overrode the logging macros, so I made them compatible to make it compile, forgot to put them back

validator-api/src/cache.rs
durch: I've no strong opinions on this one. I had a more general Caching implementation but scrapped it as it was getting complicated. It is probably cleaner to have `MixNodeCache` make its own client tough

validator-api/src/cache.rs
durch: I'd prefer to leave `as_at` to be time, so that we can invalidate cache based on time if we need to, I do like adding the block height

validator-api/Rocket.toml
durch: I've left it in as a sane default, we can remove it

durch: Actually we should keep it in to enforce default configuration

PR #667, opened 2021-07-01
Rework validator client requests

durch:
Reworks validator-client requests to loop over all available validator urls. The intent is to avoid having to pass a mutable client around as it only allows us to permute validators. There might be a better implementation available, ie don't loop forever, although if no validators are available is there even a point in running anything as far as the operators are concerned

### DIFF
--- a/common/client-libs/validator-client/Cargo.toml
+++ b/common/client-libs/validator-client/Cargo.toml
@@ -14,3 +14,4 @@ serde_json = "1"
 rand = "0.8"
 reqwest = { version = "0.11", features = ["json"] }
 thiserror = "1"
+log = "0.4"
--- a/common/client-libs/validator-client/Cargo.toml
+++ b/common/client-libs/validator-client/Cargo.toml
@@ -9,10 +9,13 @@ edition = "2018"
 [dependencies]
 base64 = "0.13"
 mixnet-contract = { path="../../../common/mixnet-contract" }
-serde = { version="1.0", features=["derive"] }
+serde = { version="1", features=["derive"] }
 serde_json = "1"
 rand = "0.8"
 reqwest = { version="0.11", features=["json"] }
 thiserror = "1"
 log = "0.4"
-tokio = { version="1.7", features=["time"] }
+wasm-timer = "0.2"
+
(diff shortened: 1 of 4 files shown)


common/client-libs/validator-client/Cargo.toml
```diff
@@ -8,9 +8,11 @@ edition = "2018"
...
-serde = { version = "1.0", features = ["derive"] }
+mixnet-contract = { path="../../../common/mixnet-contract" }
+serde = { version="1.0", features=["derive"] }
 serde_json = "1"
 rand = "0.8"
-reqwest = { version = "0.11", features = ["json"] }
+reqwest = { version="0.11", features=["json"] }
 thiserror = "1"
+log = "0.4"
+tokio = { version="1.7", features=["time"] }
```
jstuczyn: Well, you can't have any tokio imports for package that compiles to wasm

common/client-libs/validator-client/src/lib.rs
```diff
@@ -109,37 +87,52 @@ impl Client {
...
+                }
+            }
+            error!(
+                "No validators available out of {} attempted! Will try again in {} seconds. Listing all attempted:",
+                validator_urls.len(), sleep_secs
+            );
+            for url in validator_urls.iter() {
+                error!("{}", url)
             }
+            tokio::time::sleep(Duration::from_secs(sleep_secs)).await;
```
jstuczyn: To make it more wasm-friendly you can use the approach I've taken in gateway-client: https://github.com/nymtech/nym/blob/343c55f98136cb3a9cffe761fc66bdd6aa39cbdc/common/client-libs/gateway-client/src/client.rs#L243-L251

jstuczyn (changes requested):
I definitely over-engineered the solution to this relatively trivial problem. I like those changes

common/client-libs/validator-client/src/lib.rs
durch: I keep forgetting that this targets `wasm` as well, I'll make a script to catch this before I actually push :)

jstuczyn: CI already does it for you : )

jstuczyn (approved):

PR #668, opened 2021-07-01
Remove mut where possible, parallel cache requests

durch:
+ Remove `mut` where applicable
+ `join!` on cache requests


jstuczyn (approved):
Looks good, but you've left out one `mut` to `validator_client` in `clients/webassembly/src/client/mod.rs` line 258 (for some reason I couldn't leave my comment directly at it)

durch:
> Looks good, but you've left out one `mut` to `validator_client` in `clients/webassembly/src/client/mod.rs` line 258 (for some reason I couldn't leave my comment directly at it)

Great catch!

neacsu (approved):

PR #672, opened 2021-07-05
Use rocket 0.5-rc1, remove rocket-contrib

durch:


mixnode/Cargo.toml
```diff
@@ -4,40 +4,42 @@
 [package]
 name = "nym-mixnode"
 version = "0.10.2"
-authors = ["Dave Hrycyszyn <futurechimp@users.noreply.github.com>", "Jędrzej Stuczyński <andrew@nymtech.net>"]
+authors = [
```
jstuczyn, line 7: You should add yourself here as well : )

mixnode/Cargo.toml
```diff
@@ -4,40 +4,42 @@
...
+]
 edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bs58 = "0.4.0"
-clap = "2.33.0"
+bs58 = "0.4"
+clap = "2"
```
jstuczyn: Shouldn't we be versioning with explicit patch versions in case things become incompatible at some point?

durch: Semver crates with versions greater then `1` should only break API at major versions changes, and those `< 1` should only break at minor versions, so I'm being optimistic :)

jstuczyn: You're assuming everyone is going to be following it properly : ) 
But I guess it should mostly be the case for the major crates

durch: I'm an optimist, makes it easier to stay up to date on crates. Maybe we should remain paranoid on `0.*` crates...

neacsu (approved):

mixnode/Cargo.toml
jstuczyn: I've  just re-read my comment and I think I confused you (and myself). Even though I said "patch" i meant minor version. oops. So x.y rather than entire x.y.z

jstuczyn (approved):

mixnode/Cargo.toml
durch: I think we can skip minors on the `1+`, there are only a few of them, and they are developed by mature teams, so my optimistic proposal is we use major only for `1>` and minor for `<1`. In the paranoid variant I suggest patch from `<1`. I expect stable crates would not break APIs on minor versions, but we can be extra careful here as well and do minor.

durch:
Nightlies broken on a new clippy lint

PR #673, opened 2021-07-05
Update hmac and blake3

durch:
Leave as draft until new version of blake3 is out, we need the `traits` feature support in [crates.io](https://crates.io/crates/blake3)


durch:
CI nightly fails on https://rust-lang.github.io/rust-clippy/master/index.html#nonstandard_macro_braces for `rocket` mostly it seems

jstuczyn:
Updated with the most recent crates of `aes`, `cipher` and `blake3`

PR #674, opened 2021-07-05
Use cached topology for clients

neacsu:


clients/validator/src/caches/gateways.ts
```diff
@@ -41,4 +42,12 @@ export default class GatewaysCache {
         this.gateways = newGateways
         return newGateways;
     }
+
+    /// Makes requests to assemble a full list of gateways from validator-api
+    async refreshValidatorAPIGateways(url: string): Promise<GatewayBond[]> {
+        const validator_api_url = url.split(":", 2);
+        validator_api_url.push("8080");
```
jstuczyn: Shouldn't port be part of the URL? What if it's not 8080?

clients/validator/src/caches/mixnodes.ts
```diff
@@ -42,4 +43,12 @@ export default class MixnodesCache {
         this.mixNodes = newMixnodes
         return this.mixNodes;
     }
+
+    /// Makes  requests to assemble a full list of mixnodes from validator-api
+    async refreshValidatorAPIMixNodes(url: string): Promise<MixNodeBond[]> {
+        const validator_api_url = url.split(":", 2);
+        validator_api_url.push("8080");
```
jstuczyn: Same thing as with the gateways

common/client-libs/validator-client/src/lib.rs
```diff
@@ -30,6 +31,15 @@ impl Config {
         rest_servers_available_base_urls: Vec<String>,
         mixnet_contract_address: S,
     ) -> Self {
+        // URLs should be in the form: http://127.0.0.1:1317
+        if rest_servers_available_base_urls
+            .iter()
+            .map(|url| url.split(':').count())
```
jstuczyn: If you're checking for the validity of the URL, it will fail in some cases, for example in ipv6 where the valid address contains multiple `:` characters

common/client-libs/validator-client/src/lib.rs
```diff
@@ -185,6 +210,11 @@ impl Client {
         Ok(mixnodes)
     }
 
+    pub async fn get_cached_mix_nodes(&mut self) -> Result<Vec<MixNodeBond>, ValidatorClientError> {
```
jstuczyn: Any reason this methods (as well as all other validator methods) is taking `&mut self`? `query_validators` does not require a mutable reference

jstuczyn (changes requested):

clients/validator/src/caches/gateways.ts
neacsu: It is part of the URL, but it's constructed here instead of passed directly by the user, as is the case with the validator url. Ideally, the user should only pass the host address, and not the port, but some refactoring needs to be done in order to split those. That way, we can attach the necessary port, when needed (1317 for contract calls, 8080 or whatever for validator-api).
As for the port changing, I'll have to look into linking the value from the validator-api crate here. I will however put a global variable in `index.ts`, as it would be easier to maintain.

common/client-libs/validator-client/src/lib.rs
neacsu: Yep, I forgot about the ipv6. Better leave the parsing to specialized crates and packages.

common/client-libs/validator-client/src/lib.rs
neacsu: Good catch. No really sure how I ended with a `mut` here, but it is definitely not needed. Thanks.

clients/validator/src/caches/gateways.ts
jstuczyn: Correct me if I'm wrong, isn't it the case that in theory each validator is going to be able to configure to use a different port hence the user would need to pass one themselves rather than use the hardcoded one?

clients/validator/src/caches/gateways.ts
```diff
@@ -41,4 +42,18 @@ export default class GatewaysCache {
         this.gateways = newGateways
         return newGateways;
     }
+
+    /// Makes requests to assemble a full list of gateways from validator-api
+    async refreshValidatorAPIGateways(urls: string[]): Promise<GatewayBond[]> {
+        for (const url of urls) {
+            const validator_api_url = new URL(url);
+            validator_api_url.port = VALIDATOR_API_PORT;
```
jstuczyn, line 50: I'm not entirely sure what to feel about ignoring user-provided port and using the hardcoded one, but if somebody else takes a look at it and is fine with that, I'll also feel good about it

clients/validator/src/index.ts
```diff
@@ -61,12 +65,12 @@ export default class ValidatorClient {
         // if we have more than a single validator, try to perform initial connection until we succeed or run out of options
         if (validatorUrls.length > 1) {
             for (let i = 0; i < validatorUrls.length; i++) {
-                console.log("Attempting initial connection to", validatorUrls[0])
-                const netClient = await NetClient.connect(wallet, validatorUrls[0], stakeDenom).catch((_) => ValidatorClient.moveArrayHeadToBack(validatorUrls))
+                console.log("Attempting initial connection to", validatorUrls[i])
```
jstuczyn: It should be reading from [0] all the time, not [i] as on failure the head of the validators array will be moved to the back.  This was a total overcomplication on my end and if somebody refactors it at some point, it will be great (but I don't think it should be done in this PR)

jstuczyn (changes requested):

clients/validator/src/caches/gateways.ts
neacsu: I don't have a problem with asking the user to provide the port, but I'm not sure how the user would come to (reliably) know what port to mention for the validator-api. Having it as a hardcoded value seems like the good old way of doing things -> having one hardcoded port per service e.g. 80 for HTTP, 22 for SSH etc. Of course we can make it configurable on both ends (which at this moment is not, even in the validator-api), but imho it should have a value to default to, in case the user doesn't know what to provide.
If we go with a configurable port here, that change should be made in a future PR, which also makes the validator-api configurable as well.

clients/validator/src/caches/gateways.ts
neacsu: The user-provided port is not ignored, but it just isn't usable for the validator-api. In the future, my thoughts are that the user shouldn't have to provide even the validator port or the `http` scheme. An IP or DNS address should be everything that's needed.

clients/validator/src/index.ts
neacsu: Ah alright, it caught my eye on that 0, but it make sense, I'll just leave it as is for now.

PR #678, opened 2021-07-06
Using specific commit of rocket cors

jstuczyn:
Using the most recent version of `rocket_cors` was kinda stressing me out. Now at least we point to a specific commit so stuff is not going to break randomly. However, do we even need that dependency? @durch what was the reason it was introduced in the first place?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2725,7 +2725,7 @@ dependencies = [
 [[package]]
 name = "rocket_cors"
 version = "0.5.2"
-source = "git+https://github.com/lawliet89/rocket_cors?branch=master#dfd3662c49e2f6fc37df35091cb94d82f7fb5915"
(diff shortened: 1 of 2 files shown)


durch:
> @durch what was the reason it was introduced in the first place?

I always include it so that future frontends don't hit any hard times. We don't really need it now tough

durch (approved):

jstuczyn:
So I'd argue that perhaps we don't want to include it until rocket goes to proper 0.5 and this dependency itself has another release that we could point to?

futurechimp:
I haven't tried CORS yet with Rocket, but from experience, the `attempt to build a front-end` -> `CORS says no` -> `what do I need to do here?` -> `somebody enables CORS` loop is a sad-making experience for client builders. @durch I think if the CORS module fixes this I think it would be good to pre-include it.

PR #679, opened 2021-07-13
Hang coconut issuance off the validator-api

durch:
### Changes: 
+ Adds `coconut` mod to validator API
+ Adds  `tauri-client` to `clients`
+ Adds `coconut-validator-interface` to `common`
+ `validator-api` needs a cocnut keypair to run, one can be generated via [keygen-cli](https://github.com/nymtech/coconut/tree/develop/keygen-cli)

+ see [README](https://github.com/nymtech/nym/blob/7c0a2a37fbc7139a9e062082b1cf923a4a4e3ed1/clients/tauri-client/README.md) to run 

I'll go about merging `develop`, see if I can resolve the conflicts


validator-api/src/main.rs
```diff
@@ -267,6 +276,45 @@ async fn get_gateways(cache: &State<Arc<ValidatorCache>>) -> Json<Cache<Vec<Gate
     Json(cache.gateways().await)
 }
 
+//  All strings are base58 encoded representations of structs
+#[derive(Deserialize, CopyGetters)]
+struct BlindSignRequestBody {
```
jstuczyn: Could you move those to the `coconut` specific module? I've cleaned up the `main.rs` significantly in my PR that will come soon ™️   (https://github.com/nymtech/nym/pull/680). You could also move the routes away and perhaps create a full Fairing (https://rocket.rs/v0.5-rc/guide/fairings/#fairings).
Take a look at what I've done in my PR. I've modified your validator cache to work that way (and if you disagree with my decisions, even better! We could discuss that)

clients/tauri-client/src/components/Nav.svelte
```diff
@@ -0,0 +1,53 @@
+<!-- <script>
```
jstuczyn: Any point of including commented code?

clients/tauri-client/src/routes/_layout.svelte
```diff
@@ -0,0 +1,22 @@
+<script>
+	import Nav from '../components/Nav.svelte';
+
+	// export let segment;
```
jstuczyn, line 4: Commented out code

clients/tauri-client/src/routes/index.svelte
```diff
@@ -0,0 +1,127 @@
...
+    });
+    alert(response);
+  }
+
+  async function deleteCredential(idx) {
+    const response = await invoke("delete_credential", {
+      idx: idx,
+    });
+    signatures = response;
+    // console.log(signature);
```
jstuczyn: ibid. : )

clients/tauri-client/src/service-worker.js
```diff
@@ -0,0 +1,86 @@
...
+			})
+	);
+});
+
+self.addEventListener('activate', event => {
+	event.waitUntil(
+		caches.keys().then(async keys => {
+			// delete old caches
+			for (const key of keys) {
+				if (key !== ASSETS) await caches.delete(key);
```
jstuczyn, line 26: This is a personal bias but I really dislike one-line if statements without explicit brackets. They are very error prone. (I'm having flashbacks of making same mistake in C once...)

jstuczyn: Unless this code is autogenerated and comes from tauri, then ignore it

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,170 @@
+pub use coconut_rs::*;
+use coconut_rs::{Attribute, Base58, BlindSignRequest, BlindedSignature, PublicKey};
+use getset::{CopyGetters, Getters};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Getters, CopyGetters, Clone)]
+pub struct Credential {
+    #[getset(get = "pub")]
```
jstuczyn, line 21: Is using separate library for getters/setters really worth it? At some point we will definitely want to limit our dependencies to bare minimum so I wonder if we couldn't just write those couple of lines ourselves.

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    n_params: u32,
+    #[getset(get = "pub")]
+    theta: Theta,
+    public_attributes: Vec<String>,
```
jstuczyn, line 25: Why are the attributes Strings?

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub key: VerificationKey,
+}
+
+impl VerificationKeyResponse {
+    pub fn new(key: VerificationKey) -> VerificationKeyResponse {
+        VerificationKeyResponse { key }
+    }
+}
+
+fn get_verification_key(url: &str) -> Result<VerificationKey, String> {
```
jstuczyn: Considering all of our other code is working nicely in async context, shouldn't we rather use some async library for http calls? For example `reqwest` that we already use elsewhere?

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn public_attributes(&self) -> Vec<Attribute> {
+        self.public_attributes
+            .iter()
+            .map(|x| Attribute::try_from_bs58(x).unwrap())
+            .collect()
+    }
+
+    pub fn verify(&self, validator_urls: Vec<String>) -> bool {
```
jstuczyn: I think it would make more sense if the public API took verification key as an argument rather than urls of validators

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub fn get_aggregated_verification_key(
+    validator_urls: Vec<String>,
+) -> Result<VerificationKey, String> {
+    let mut verification_keys = Vec::new();
+    let mut indices = Vec::new();
+
+    for (idx, url) in validator_urls.iter().enumerate() {
+        verification_keys.push(get_verification_key(url.as_ref())?);
+        indices.push((idx + 1) as u64);
```
jstuczyn, line 233: The problem with the sucker is that you must ensure your validator urls are in correct orders otherwise you will mess up your indices (I guess it should rather be called `points` as technically those are x-coordinates of a polynomial used during key generation). Perhaps we should slightly modify `coconut` library itself so that the coordinate/index would already be included in the serialized verification key?

gateway/Cargo.toml
```diff
@@ -36,3 +36,4 @@ nymsphinx = { path = "../common/nymsphinx" }
 pemstore = { path = "../common/pemstore" }
 validator-client = { path = "../common/client-libs/validator-client" }
 version-checker = { path = "../common/version-checker" }
+# coconut-rs = { git = "https://github.com/nymtech/coconut.git", branch = "serde" }
```
jstuczyn: I don't think it's needed then for now? : )

gateway/gateway-requests/src/registration/handshake/gateway.rs
```diff
@@ -22,11 +22,12 @@ impl<'a> GatewayHandshake<'a> {
         ws_stream: &'a mut S,
         identity: &'a crypto::asymmetric::identity::KeyPair,
         received_init_payload: Vec<u8>,
+        validator_urls: Vec<String>,
```
jstuczyn: I don't like putting validator urls here. I think gateway should obtain the entire verification key on startup. Think about it. If say 10 clients decide to start talking with the gateway, it will have to talk with the same validators all 10 times only to produce exactly the same verification key. So instead I'd say this argument should be just the verification key itself.

gateway/gateway-requests/src/registration/handshake/gateway.rs
```diff
@@ -49,11 +50,28 @@ impl<'a> GatewayHandshake<'a> {
                 }
 
                 // init: <- pub_key || g^x
-                let (remote_identity, remote_ephemeral_key) = check_processing_error(
+                let init_message = check_processing_error(
```
jstuczyn, line 54: Please update this comment because it seems you're also sending the coconut credential alongside

gateway/gateway-requests/src/registration/handshake/gateway.rs
```diff
@@ -49,11 +50,28 @@ impl<'a> GatewayHandshake<'a> {
...
                     State::<S>::parse_init_message(received_init_payload),
                     &mut state,
                 )
                 .await?;
+
+                let credential = init_message.credential();
+
+                check_processing_error(
+                    {
+                        if !credential.verify(validator_urls) {
```
jstuczyn: This is the issue I described before, we really don't want to be talking with all validators here every single time.... Just use existing verification key.
Also, should the coconut verification be happening immediately? Out of all the steps in the handshake this is **the** most expensive step, so perhaps it should be left at the very end? Though I guess we already have the credential on hand so maybe not... (this is just some thinking aloud, I'm not sure what's the correct answer to that)

gateway/gateway-requests/src/registration/handshake/mod.rs
```diff
@@ -28,23 +28,32 @@ pub async fn client_handshake<'a, S>(
...
-    ClientHandshake::new(rng, ws_stream, identity, gateway_pubkey).await
+    ClientHandshake::new(rng, ws_stream, identity, gateway_pubkey, coconut_credential).await
 }
 
 pub async fn gateway_handshake<'a, S>(
     rng: &mut (impl RngCore + CryptoRng),
     ws_stream: &'a mut S,
     identity: &'a identity::KeyPair,
     received_init_payload: Vec<u8>,
+    validator_urls: Vec<String>,
```
jstuczyn: .

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -15,9 +16,57 @@ use futures::{Sink, SinkExt, Stream, StreamExt};
...
+    pub fn ephemeral_key(&self) -> encryption::PublicKey {
+        encryption::PublicKey::from_bytes(&self.ephemeral_key).unwrap()
+    }
+
+    pub fn credential(&self) -> Credential {
+        self.credential.clone().unwrap()
+    }
+
+    fn to_bytes(&self) -> Vec<u8> {
+        bincode::serialize(self).unwrap()
```
jstuczyn, line 59: I'm not entirely sure what to think about the use of bincode here. Sure, it is way cleaner this way, but is the sucker 100% deterministic all the time? Are we not signing the result of the sucker?

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -15,9 +16,57 @@ use futures::{Sink, SinkExt, Stream, StreamExt};
...
+    ) -> Self {
+        InitMessage {
+            local_id_pubkey: local_id_pubkey.to_bytes().to_vec(),
+            ephemeral_key: ephemeral_key.to_bytes().to_vec(),
+            credential: Some(credential),
+        }
+    }
+
+    pub fn local_id_pubkey(&self) -> identity::PublicKey {
+        identity::PublicKey::from_bytes(&self.local_id_pubkey).unwrap()
```
jstuczyn, line 45: We definitely do not want to do unwraps here! We don't trust the data client is sending the gateway! They can send whatever bullshit they want and it could potentially crash the gateway if unwrap fails

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -15,9 +16,57 @@ use futures::{Sink, SinkExt, Stream, StreamExt};
...
+    }
+
+    pub fn ephemeral_key(&self) -> encryption::PublicKey {
+        encryption::PublicKey::from_bytes(&self.ephemeral_key).unwrap()
```
jstuczyn, line 50: Same unwrap issue as above

gateway/gateway-requests/src/registration/handshake/state.rs
jstuczyn: Though maybe it's fine because then deserialization would have failed... Hm... Could we perhaps just store the concrete identity::PublicKey on the struct and deal with the conversion during deserialization?

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -55,8 +55,8 @@ pub(crate) struct Handle<R, S> {
     clients_handler_sender: ClientsHandlerRequestSender,
     outbound_mix_sender: MixForwardingSender,
     socket_connection: SocketStream<S>,
-
     local_identity: Arc<identity::KeyPair>,
+    validator_urls: Vec<String>,
```
jstuczyn, line 61: Just inject here the verification key rather than validator urls

gateway/src/node/client_handling/websocket/listener.rs
```diff
@@ -15,13 +15,19 @@ use tokio::task::JoinHandle;
 pub(crate) struct Listener {
     address: SocketAddr,
     local_identity: Arc<identity::KeyPair>,
+    validator_urls: Vec<String>,
```
jstuczyn, line 18: ibid.

validator-api/Cargo.toml
```diff
@@ -22,24 +23,34 @@ log = "0.4"
...
+tokio = { version = "1.4", features = [
+    "rt-multi-thread",
+    "macros",
+    "signal",
+    "time",
+] }
+rocket_cors = { git = "https://github.com/lawliet89/rocket_cors", branch = "master" }
 anyhow = "1"
+getset = "0.1.1"
+hex = "0.4.3"
```
jstuczyn: I'm sorry, why are we using hex here? Unless I've missed something I didn't see any use of that anywhere?

validator-api/Cargo.toml
```diff
@@ -22,24 +23,34 @@ log = "0.4"
...
 serde_json = "1.0"
-tokio = { version="1.4", features=["rt-multi-thread", "macros", "signal", "time"] }
-rocket_cors = { git="https://github.com/lawliet89/rocket_cors", branch="master" }
+tokio = { version = "1.4", features = [
+    "rt-multi-thread",
+    "macros",
+    "signal",
+    "time",
+] }
+rocket_cors = { git = "https://github.com/lawliet89/rocket_cors", branch = "master" }
```
jstuczyn: I thought I already merged a PR that changed cors dependency to point to specific commit instead?

validator-api/src/coconut/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        InternalSignRequest {
+            total_params,
+            public_attributes,
+            public_key,
+            blind_sign_request,
+        }
+    }
+}
+
+pub fn blind_sign(request: InternalSignRequest, config: &Config) -> BlindedSignature {
```
jstuczyn: Does it really make sense for "sign" function to take entire config as an argument? I think it should either take secret key or there should be some method defined on the secret key itself.

validator-api/src/coconut/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        request.public_attributes(),
+    )
+    .unwrap()
+}
+
+#[post("/blind_sign", data = "<blind_sign_request_body>")]
+//  Until we have serialization and deserialization traits we'll be using a crutch
+pub async fn post_blind_sign(
+    blind_sign_request_body: Json<BlindSignRequestBody>,
+    config: &State<Config>,
```
jstuczyn: (the same applies to get_verification_key). I'm not really sure how to feel about entire Config being managed by rocket. It seems like we don't need 99% of the stuff there. Why not just put a keypair there?

validator-api/src/config/mod.rs
```diff
@@ -2,18 +2,17 @@
...
 
 mod template;
 
-const DEFAULT_VALIDATOR_REST_ENDPOINTS: &[&str] = &[
-    "http://testnet-finney-validator.nymtech.net:1317",
-    "http://testnet-finney-validator2.nymtech.net:1317",
-];
-const DEFAULT_MIXNET_CONTRACT: &str = "hal1k0jntykt7e4g3y88ltc60czgjuqdy4c9c6gv94";
+const DEFAULT_VALIDATOR_REST_ENDPOINTS: &[&str] =
+    &["http://testnet-milhon-validator1.nymtech.net:1317"];
```
jstuczyn: I also thought I changed that to localhost?

validator-api/src/config/mod.rs
```diff
@@ -2,18 +2,17 @@
...
+const DEFAULT_MIXNET_CONTRACT: &str = "punk18vd8fpwxzck93qlwghaj6arh4p7c5n89m7r3lj";
```
jstuczyn: That's definitely incorrect. Our current contract is `punk10pyejy66429refv3g35g2t7am0was7yalwrzen`

validator-api/src/config/mod.rs
```diff
@@ -2,18 +2,17 @@
...
 
 const DEFAULT_NODE_STATUS_API: &str = "http://localhost:8081";
```
jstuczyn: And that doesn't exist anymore... Did you merge with develop? : )

validator-api/src/config/mod.rs
```diff
@@ -73,6 +72,8 @@ pub struct Base {
 
     /// Address of the validator contract managing the network
     mixnet_contract_address: String,
+    // Avoid breaking derives for now
+    keypair_bs58: String,
```
jstuczyn, line 84: This is more of a question than a comment. Should config contain keys themselves or the paths to key files (like we're doing with mixnodes and gateways? Over there we load the .pem files based on config paths)

validator-api/src/config/mod.rs
```diff
@@ -178,11 +180,23 @@ impl Default for TopologyCacher {
     }
 }
 
+fn strip_trailing_newline(input: &str) -> String {
+    input
+        .strip_suffix("\r\n")
```
jstuczyn: `.trim()` will do exactly that - it will remove all leading and trailing whitespaces (so it includes newlines)

validator-api/src/config/mod.rs
```diff
@@ -178,11 +180,23 @@ impl Default for TopologyCacher {
...
+        .to_string()
+}
+
 impl Config {
     pub fn new() -> Self {
         Config::default()
     }
 
+    pub fn keypair(&self) -> KeyPair {
+        KeyPair::try_from_bs58(self.base.keypair_bs58.clone()).unwrap()
```
jstuczyn, line 217: As I said before, should it contain the actual keys or rather paths?

validator-api/src/main.rs
```diff
@@ -446,8 +455,10 @@ async fn main() -> Result<()> {
 
     let write_validator_cache = Arc::clone(&validator_cache);
 
+    let caching_interval = config.get_caching_interval();
+
     tokio::spawn(async move {
-        let mut interval = time::interval(config.get_caching_interval());
+        let mut interval = time::interval(caching_interval);
```
jstuczyn: Why are you no longer getting it from config?

validator-api/src/main.rs
```diff
@@ -471,8 +482,17 @@ async fn main() -> Result<()> {
 
     rocket::build()
         .attach(cors)
-        .mount("/v1", routes![get_mixnodes, get_gateways])
+        .mount(
```
jstuczyn: Please merge with develop : )

jstuczyn (changes requested):

gateway/gateway-requests/src/registration/handshake/gateway.rs
neacsu: Fixed by making it async

clients/tauri-client/src/components/Nav.svelte
neacsu: I don't think so, I've removed it.

validator-api/Cargo.toml
neacsu: Good catch, I've removed it.

validator-api/src/config/mod.rs
neacsu: I'll just leave it like this for the moment, as I don't see any reason to complicate things by reading from paths.

gateway/gateway-requests/src/registration/handshake/state.rs
neacsu: Why wouldn't it be deterministic?

common/coconut-interface/src/lib.rs
neacsu: I kind of like these getters/setters too :D . Regarding limiting the dependencies, I'm thinking we should apply that to the actual `coconut-rs` crate, while this interface that we're using is not so sensitive.

common/coconut-interface/src/lib.rs
neacsu: We can change that in the future, once we know more of how an attribute should look like.

gateway/gateway-requests/src/registration/handshake/gateway.rs
neacsu: I see what you mean, but in order to do that optimization, we'll need to have some caching mechanism implemented in the gateway, for refreshing that verification key, right? Until we do that, I'll just make the interface more natural by having the verification key as argument, but the problem of querying the verification key at each handshake will remain.

common/coconut-interface/src/lib.rs
neacsu: I agree we should unite the two, but that should be done separately, probably in the `coconut-rs` library.

clients/native/src/client/mod.rs
```diff
@@ -189,24 +190,30 @@ impl NymClient {
...
-            self.key_manager.identity_keypair(),
-            gateway_identity,
-            Some(self.key_manager.gateway_shared_key()),
-            mixnet_message_sender,
-            ack_sender,
-            self.config.get_base().get_gateway_response_timeout(),
-        );
-
         self.runtime.block_on(async {
+            let coconut_credential =
```
jstuczyn: I guess this would be for another PR since we want to just merge it - there should be an option to load credential from the disk or something. In real world client wouldn't be obtaining brand new credential every time

clients/native/src/commands/init.rs
```diff
@@ -74,12 +75,17 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
 async fn register_with_gateway(
     gateway: &gateway::Node,
     our_identity: Arc<identity::KeyPair>,
+    validator_urls: Vec<String>,
 ) -> SharedKeys {
     let timeout = Duration::from_millis(1500);
+    let coconut_credential = Credential::init(validator_urls)
```
jstuczyn: why do we need coconut credential for init? We're not sending any traffic through the mixnet?

clients/socks5/src/commands/init.rs
```diff
@@ -75,12 +76,17 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
 async fn register_with_gateway(
     gateway: &gateway::Node,
     our_identity: Arc<identity::KeyPair>,
+    validator_urls: Vec<String>,
 ) -> SharedKeys {
     let timeout = Duration::from_millis(1500);
+    let coconut_credential = Credential::init(validator_urls)
```
jstuczyn: Same question as with native client

clients/tauri-client/src/routes/index.svelte
```diff
@@ -0,0 +1,119 @@
+<script>
+  import {invoke} from "@tauri-apps/api/tauri";
+  import {onMount} from "svelte";
+  import QRious from "qrious";
+
+  const validator_urls = ["http://localhost:8080"];
```
jstuczyn, line 6: Should this be pointing to localhost?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -139,7 +146,10 @@ impl GatewayClient {
     #[cfg(target_arch = "wasm32")]
     async fn _close_connection(&mut self) -> Result<(), GatewayClientError> {
         match std::mem::replace(&mut self.connection, SocketState::NotConnected) {
-            SocketState::Available(mut socket) => Ok(socket.close(None).await),
+            SocketState::Available(mut socket) => {
```
jstuczyn, line 149: Any reason for expanding this to multiple lines?

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            theta: theta.clone(),
+            public_attributes: public_attributes
+                .iter()
+                .map(|attr| attr.to_bs58())
+                .collect(),
+            signature: *signature,
+        }
+    }
+
+    pub async fn init(validator_urls: Vec<String>) -> Result<Self, String> {
```
jstuczyn: This method feels... wrong. You give it validator urls are get back credential? But how do you know what attributes is it created on? Also (in the future, so right now it doesn't really matter), how do you know who is it bound to?

Also in multiple places you do have `-> Result<T, String>`. This kinda feels like if an Error enum should have been defined.

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            &theta,
+            &*state.public_attributes,
+            &signature,
+        ))
+    }
+
+    pub fn public_attributes(&self) -> Vec<Attribute> {
+        self.public_attributes
+            .iter()
+            .map(|x| Attribute::try_from_bs58(x).unwrap())
```
jstuczyn, line 71: Are we sure an unwrap is fine here?

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub params: Parameters,
+    pub public_attributes: Vec<Attribute>,
+    pub private_attributes: Vec<Attribute>,
+}
+
+impl State {
+    pub fn init() -> State {
+        let n_attributes: u32 = 3;
+        let params = Parameters::new(n_attributes).unwrap();
+        let public_attributes = params.n_random_scalars(2);
```
jstuczyn: Just why? Why would you want to initialise it with random attributes with some arbitrary number of them? They should be user-provided.

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+) -> Result<VerificationKey, String> {
+    let mut verification_keys = Vec::new();
+    let mut indices = Vec::new();
+
+    for (idx, url) in validator_urls.iter().enumerate() {
+        verification_keys.push(get_verification_key(url.as_ref(), client).await?);
+        indices.push((idx + 1) as u64);
+    }
+
+    match aggregate_verification_keys(&verification_keys, Some(&indices)) {
```
jstuczyn: you can just use `map_err`

gateway/gateway-requests/src/registration/handshake/state.rs
jstuczyn: I don't know the ins and outs of bincode, but you can't always guarantee determinism from serializers. For example protobuf serialization is not deterministic.

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -15,9 +16,60 @@ use futures::{Sink, SinkExt, Stream, StreamExt};
...
+        self.credential.clone().unwrap()
+    }
+
+    fn to_bytes(&self) -> Vec<u8> {
+        bincode::serialize(self).unwrap()
+    }
+}
+
+impl TryFrom<&[u8]> for InitMessage {
+    type Error = String;
```
jstuczyn: I don't think `String` should be used as errors everywhere. Especially if they can convey different messages.

gateway/src/node/client_handling/websocket/connection_handler.rs
jstuczyn: At least remember to create a github issue for that change after the merge.

gateway/gateway-requests/src/registration/handshake/state.rs
jstuczyn: And we do have a `HandshakeError` enum available here

validator-api/src/coconut/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+};
+use getset::{CopyGetters, Getters};
+use rocket::fairing::AdHoc;
+use rocket::serde::json::Json;
+use rocket::State;
+use validator_client::validator_api::VALIDATOR_API_CACHE_VERSION;
+
+#[derive(Getters, CopyGetters, Debug)]
+pub struct InternalSignRequest {
+    // Total number of paraeters to generate for
```
jstuczyn: typo

validator-api/src/coconut/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    BlindedSignatureResponse, KeyPair, Parameters, VerificationKeyResponse,
+};
+use getset::{CopyGetters, Getters};
+use rocket::fairing::AdHoc;
+use rocket::serde::json::Json;
+use rocket::State;
+use validator_client::validator_api::VALIDATOR_API_CACHE_VERSION;
+
+#[derive(Getters, CopyGetters, Debug)]
+pub struct InternalSignRequest {
```
jstuczyn: If it's internal, why is it public?

validator-api/src/config/mod.rs
```diff
@@ -22,7 +30,7 @@ const DEFAULT_GATEWAY_CONNECTION_TIMEOUT: Duration = Duration::from_millis(2_500
 
 const DEFAULT_CACHE_INTERVAL: Duration = Duration::from_secs(60);
 
-#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
```
jstuczyn: Why do we need clones on the config?

validator-api/src/network_monitor/monitor/sender.rs
```diff
@@ -110,6 +113,14 @@ impl PacketSender {
         // TODO: future optimization: if we're remaking client for a gateway to which we used to be connected in the past,
         // use old shared keys
         let (message_sender, message_receiver) = mpsc::unbounded();
+
+        let coconut_credential = Credential::init(vec![format!(
+            "http://{}:{}",
+            DEFAULT_VALIDATOR_HOST, VALIDATOR_API_PORT
```
jstuczyn: Why are we using defaults here? Shouldn't we use whatever was provided in config file?

validator-api/src/config/mod.rs
jstuczyn: Another thing to please remember about

validator-api/src/config/mod.rs
jstuczyn: (also applies to all sub structs)

validator-api/src/config/mod.rs
```diff
@@ -232,6 +247,12 @@ impl Config {
         self
     }
 
+    pub fn with_keypair(mut self, path: &str) -> Self {
```
jstuczyn: Given the signature (i.e. taking a `&str`) and the name, I wouldn't have expected it to take a path but rather raw string value. Perhaps it should be modified to more closely represent what it's doing?

common/coconut-interface/src/lib.rs
jstuczyn: Well, an attribute is a Scalar : ) There's nothing more to it really

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+                .map(|attr| attr.to_bs58())
+                .collect(),
+            signature: *signature,
+        }
+    }
+
+    pub async fn init(validator_urls: Vec<String>) -> Result<Self, String> {
+        let mut state = State::init();
+        let client = ValidatorAPIClient::new();
+        let signature = get_aggregated_signature(validator_urls.clone(), &state, &client)
```
jstuczyn: With the said Error type you would be able to just use `?` : )

common/coconut-interface/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    pub fn public_attributes(&self) -> Vec<Attribute> {
+        self.public_attributes
+            .iter()
+            .map(|x| Attribute::try_from_bs58(x).unwrap())
+            .collect()
+    }
+
+    pub async fn verify(&self, verification_key: &VerificationKey) -> bool {
+        let params = Parameters::new(self.n_params).unwrap();
```
jstuczyn, line 76: Perhaps the params should be kept on self since it's derived purely from the internal state?

common/client-libs/validator-client/src/validator_api/mod.rs
```diff
@@ -2,13 +2,15 @@ pub mod error;
...
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use url::Url;
 
 pub const VALIDATOR_API_PORT: u16 = 8080;
 pub const VALIDATOR_API_CACHE_VERSION: &str = "/v1";
 pub(crate) const VALIDATOR_API_MIXNODES: &str = "/mixnodes";
 pub(crate) const VALIDATOR_API_GATEWAYS: &str = "/gateways";
+pub const VALIDATOR_API_VERIFICATION_KEY: &str = "/verification_key";
+pub const VALIDATOR_API_BLIND_SIGN: &str = "/blind_sign";
```
jstuczyn, line 13: This comment is not specifically for this, but perhaps we should get rid of all slashes in constants and instead only introduce them in urls directly? That way we'd prevent future double slash issues

gateway/gateway-requests/src/registration/handshake/gateway.rs
```diff
@@ -48,12 +50,29 @@ impl<'a> GatewayHandshake<'a> {
                     }
                 }
 
-                // init: <- pub_key || g^x
-                let (remote_identity, remote_ephemeral_key) = check_processing_error(
+                // init: <- pub_key || g^x || credential
```
jstuczyn, line 53: I wonder whether there might be some attack surface here if we're sending the credential in plain that is not bound to anyone. Perhaps it should be sent towards the end of the handshake so that it'd be encrypted with the derived shared key? I honestly don't know. Perhaps @aniampio might have an opinion

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -36,6 +88,7 @@ pub(crate) struct State<'a, S> {
     /// The known or received public identity key of the remote.
     /// Ideally it would always be known before the handshake was initiated.
     remote_pubkey: Option<identity::PublicKey>,
+    coconut_credential: Option<Credential>,
```
jstuczyn, line 88: Thinking about it, I don't think coconut credential should be part of the state at all. It's only ever used once to get verified and is not required to be kept throughout the entire handshake (I guess another thing to change in the future, if you could leave a comment or something indicating that, it'd be great)

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -63,40 +119,21 @@ impl<'a, S> State<'a, S> {
...
+        .to_bytes()
     }
 
     // this will need to be adjusted when REMOTE_ID_PUBKEY is removed
-    pub(crate) fn parse_init_message(
-        mut init_message: Vec<u8>,
-    ) -> Result<(identity::PublicKey, encryption::PublicKey), HandshakeError> {
-        if init_message.len() != identity::PUBLIC_KEY_LENGTH + encryption::PUBLIC_KEY_SIZE {
-            return Err(HandshakeError::MalformedRequest);
+    #[cfg(not(target_arch = "wasm32"))]
```
jstuczyn, line 128: I don't think `#[cfg(not(target_arch = "wasm32"))]` is necessary here?

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -63,40 +119,21 @@ impl<'a, S> State<'a, S> {
...
+    pub(crate) fn parse_init_message(init_message: Vec<u8>) -> Result<InitMessage, HandshakeError> {
+        match InitMessage::try_from(init_message.as_slice()) {
```
jstuczyn: You can just use `map_err`

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -193,6 +230,7 @@ impl<'a, S> State<'a, S> {
             .map_err(|_| HandshakeError::InvalidSignature)
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
```
jstuczyn, line 227: .

jstuczyn (changes requested):
Considering we want to merge that puppy rather soon, I'd say the biggest things I'd really want changed before it are:
- `State::init()` in coconut-interface
- introduction of some error enum in the same file, it looks kinda messy with the String there
- creation of github issues/user stories for things I mentioned in other comments

clients/native/src/commands/init.rs
neacsu: It's the step where we're currently doing the handshake with the gateway. It will probably need to be moved at `run` step in the future though.

clients/tauri-client/src/routes/index.svelte
neacsu: Most definitely not, but I think this is just a placeholder to have a working tauri client. It should be taken from a configuration file at some point, like we do for all the other clients.

common/client-libs/gateway-client/src/client.rs
neacsu: I ran `clippy` with the wasm target at some point and it suggested this change. It wouldn't appear in the CI as we have disabled it for now afaik.

common/coconut-interface/src/lib.rs
neacsu: This is mostly a placeholder for getting a dummy `Credential` in each client. The next step would indeed be to figure out what attributes we should place in that `Credential` (and where to get them from) and remove this dummy method.

Regarding the error type, we can refactor it into an enum, I'll push a commit for that.

common/coconut-interface/src/lib.rs
neacsu: Yep, agreed, that will be the next step into the implementation. This is just an intermediary step :D

validator-api/src/config/mod.rs
neacsu: I forgot about this. It was before removing config from the rocket management, as I needed it to be passed to rocket and there were some read-only operations on it after that, in `validator-api/src/main.rs`. But now that only `KeyPair` is managed by rocket, there's no need for the clones, I'll remove that.

validator-api/src/network_monitor/monitor/sender.rs
neacsu: Considering that network monitor runs on the same machine as the validator-api, the default `localhost` doesn't need much configuration. `DEFAULT_VALIDATOR_HOST ` is not even so much a default as is a global constant for this crate.

common/coconut-interface/src/lib.rs
neacsu: Yeah, I just wanted a more verbose error message :)

common/client-libs/validator-client/src/validator_api/mod.rs
neacsu: I'm not sure, but I think that's how `rocket` mounts them, which is why I left them with the slash prefix, so that I wouldn't need for `format!` the mount arguments

gateway/gateway-requests/src/registration/handshake/state.rs
neacsu: Sure, I'll save all future work comments from this PR into github issues that we can (hopefully) take on in the sprints.

gateway/gateway-requests/src/registration/handshake/state.rs
neacsu: Well, it is currently only used by the gateway server code, so without this you'll get a "function not used" warning.

gateway/gateway-requests/src/registration/handshake/state.rs
neacsu: Same.

gateway/gateway-requests/src/registration/handshake/state.rs
neacsu: I would say `bincode` _should_ be deterministic, at least for the struct that we're using now (with `Vec<u8>` and `Option<Credential>`). The only way I can see it potentially become indeterministic like `protobuf` is to have some map-like structure in there. But even then, I don't fully see the problem, as long as the deserialization uses the appropriate algorithm and knows how to get the struct out of the bytes, the message exchange should be successful, right?

common/coconut-interface/src/lib.rs
neacsu: Right. The reason Drazen went with String is probably because it's an easier representation, using base 58, in the context of (de)serializing it for the `InitMessage`. Going deeper into `coconut-rs` and implementing (de)serializing for `Attributes` is I guess another possibility.

validator-api/src/coconut/mod.rs
neacsu: Well, I guess it's internal to validator-api, but it's just in the `coconut` module, so it needs to be `pub` (or `pub(crate)` at least, which I'm going to change in a patch, just for sake of it :D )

common/coconut-interface/src/lib.rs
neacsu: It would make more sense to do that, yes, but right now `Parameters` doesn't implement (De)Serialize , so that it can be kept inside `Credential`

common/coconut-interface/src/lib.rs
```diff
@@ -175,10 +182,13 @@ pub struct State {
 }
 
 impl State {
-    pub fn init() -> State {
-        let n_attributes: u32 = 3;
+    pub fn init(public_key: Option<crypto::asymmetric::identity::PublicKey>) -> State {
+        let n_attributes: u32 = 2;
+        let attribute = public_key.unwrap_or_else(|| {
```
jstuczyn: I might not understand something, but why do we use 0-bytes public key if key is not given? Why not just ensure key is always there?

common/coconut-interface/src/lib.rs
```diff
@@ -175,10 +182,13 @@ pub struct State {
 }
 
 impl State {
-    pub fn init() -> State {
-        let n_attributes: u32 = 3;
+    pub fn init(public_key: Option<crypto::asymmetric::identity::PublicKey>) -> State {
+        let n_attributes: u32 = 2;
```
jstuczyn: This one still stresses me out : ) 
Can we pass it as an argument or something instead?

common/coconut-interface/src/lib.rs
```diff
@@ -175,10 +182,13 @@ pub struct State {
...
+        let attribute = public_key.unwrap_or_else(|| {
+            crypto::asymmetric::identity::PublicKey::from_bytes(&[0; PUBLIC_KEY_LENGTH]).unwrap()
+        });
         let params = Parameters::new(n_attributes).unwrap();
-        let public_attributes = params.n_random_scalars(2);
+        let public_attributes = vec![hash_to_scalar(attribute.to_bytes())];
         let private_attributes = params.n_random_scalars(1);
```
jstuczyn: Just get rid of those random scalars here : )

gateway/gateway-requests/src/registration/handshake/gateway.rs
neacsu: As per the offline discussion, I'll do the credential send after the handshake, so that it's protected with the shared key.

PR #681, opened 2021-07-14
Feature/update gateway config milhon

mfahampshire:
Update the default config of the gateway to new testnet


clients/webassembly/js-example/index.js
```diff
@@ -24,10 +24,16 @@ async function main() {
...
-    const validator = "http://testnet-validator1.nymtech.net:8081"; //"http://localhost:8081";
+    // validator we will use to get topology from
+    // If you run into CORS errors make sure to set `enable-unsafe-cors` as true
+    // and add `'Access-Control-Allow-Origin'` to the accepted headers
+    // in your validator's `config.toml`,
+    const validator = "http://mixnet.club:1317"; //  1317 = standard API port.
 
-    client = new NymClient(validator);
+    // mixnet contract address
+    const mixnet_contract = "hal1k0jntykt7e4g3y88ltc60czgjuqdy4c9c6gv94"
```
jstuczyn, line 34: Well, is that still true? : )

gateway/src/config/mod.rs
```diff
@@ -18,10 +18,10 @@ pub(crate) const MISSING_VALUE: &str = "MISSING VALUE";
 const DEFAULT_MIX_LISTENING_PORT: u16 = 1789;
 const DEFAULT_CLIENT_LISTENING_PORT: u16 = 9000;
 pub(crate) const DEFAULT_VALIDATOR_REST_ENDPOINTS: &[&str] = &[
-    "http://testnet-finney-validator.nymtech.net:1317",
-    "http://testnet-finney-validator2.nymtech.net:1317",
+    "http://testnet-milhon-validator1.nymtech.net:1317",
```
jstuczyn, line 21: I don't remember the details, but isn't it the case gateways are talking with validator API rather than `nymd`? Meaning that the port should probably be `8080`? - I don't know the answer for it, just asking. But if you run it and it worked, it's good enough for me.

clients/webassembly/js-example/index.js
futurechimp: I've already updated this in another commit I think. I'll resolve the conflict here.

futurechimp (commented):
Hi @mfahampshire is this one kind of the same as #685? I've now resolved the conflict but the code looks pretty alike, just wondering which way to go here.

PR #692, opened 2021-07-20
Feature/other containers

neacsu:
This adds support to block explorer and network explorer containers as well, but with the caveat that you need the network explorer token for it to build

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,7 +51,29 @@ services:
                 depends_on:
                         - "genesis_validator"
 
+        mongo:
+                image: mongo:latest
+                command:
+                        - --storageEngine=wiredTiger
+                volumes:
+                        - mongo_data:/data/db
+
+        block_explorer:
+                build:
+                        context: https://github.com/forbole/big-dipper.git#v0.41.x-7
+                image: block_explorer:v0.41.x-7
+                ports:
+                        - "3080:3000"
+                depends_on:
+                        - "mongo"
+                environment:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -51,7 +51,40 @@ services:
                 depends_on:
                         - "genesis_validator"
 
+        mongo:
```
jstuczyn, line 54: Wait, what's the reason for mongo here?

neacsu: It's the mongo service from the block explorer docker setup. Maybe a more suitable naming is needed?

samip5:
Helm chart for this too?

https://phoenixnap.com/kb/what-is-helm

tl;dr Related to Kubernetes, Helm charts are Helm packages consisting of YAML files and templates which convert into Kubernetes manifest files.

docker-compose.yml
jstuczyn: Oh, I didn't know block explorer was using mongo deep down. I just have personal bias against mongo : )

jstuczyn (approved):

neacsu:
> Helm chart for this too?
> 
> https://phoenixnap.com/kb/what-is-helm
> 
> tl;dr Related to Kubernetes, Helm charts are Helm packages consisting of YAML files and templates which convert into Kubernetes manifest files.

Thank you for the suggestion. Kubernetes support is not currently on the roadmap, but if it will be at some point, we can consider Helm at that point.

PR #693, opened 2021-07-20
fixed gateway config issue: now points at only 1 validator

mfahampshire:


jstuczyn:
If the validator API is running on both machines, this change should get reverted : )

PR #698, opened 2021-07-22
Bond and delegation alerts

fmtabbara:
Display alert on bonding and delegation pages when the user is allocating all of their tokens.


wallet-web/common/helpers.ts
```diff
@@ -138,11 +137,11 @@ export const basicRawCoinValueValidation = (rawAmount: string): boolean => {
 }
 
 export const getDisplayExecGasFee = (): string => {
-    const table = buildFeeTable(nymGasPrice(UDENOM), nymGasLimits, nymGasLimits)
+    const table = buildFeeTable(nymGasPrice(DENOM), nymGasLimits, nymGasLimits)
```
jstuczyn: Are you sure `UDENOM` should be changed to `DENOM` here?

wallet-web/components/bond/BondNodeForm.tsx
```diff
@@ -1,380 +1,480 @@
...
+  amount: string
+  host: string
+  http_api_port: string
+  mixPort: string
+  verlocPort: string
+  sphinxKey: string
+  identityKey: string
+  version: string
+  location?: string
+  clientsPort?: string
```
jstuczyn, line 49: Oh. Optional fields. I totally forgot they existed when I first wrote it

wallet-web/components/bond/BondNodeForm.tsx
```diff
@@ -1,380 +1,480 @@
...
+    } catch {
+      return false
     }
-
-    const validateLocation = (location: string): boolean => {
-        // right now only perform the stupid check of whether the user copy-pasted the tooltip... (with or without brackets)
-        return !location.trim().includes("physical location of your node")
+  }
+
+  const validateHost = (host: string): boolean => {
```
jstuczyn, line 202: This is not an error on your end, but omission in what I've done - could you check for white spaces at beginning (and end) of the host string? I've noticed people were bonding with for example `<SPACE-HERE>1.1.1.1` value. Or maybe just trim it for them.

wallet-web/hooks/useGetBalance.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ValidatorClientContext } from "../contexts/ValidatorClient";
+import { basicRawCoinValueValidation } from "../common/helpers";
+
+export const useGetBalance = () => {
+  const { client } = useContext(ValidatorClientContext);
+  const [isLoading, setIsLoading] = useState(false);
+  const [balanceCheckError, setBalanceCheckError] = useState(null);
+  const [accountBalance, setAccountBalance] = useState<Coin>();
+
+  console.log(basicRawCoinValueValidation(accountBalance?.amount));
```
jstuczyn: Should we be logging this all the time?

wallet-web/components/delegate/DelegateForm.tsx
```diff
@@ -1,121 +1,146 @@
...
+
+  const handleAmountChange = (event: any) => {
+    // don't ask me about that. javascript works in mysterious ways
+    // and this is apparently a good way of checking if string
+    // is purely made of numeric characters
+    let parsed = +event.target.value
+    if (isNaN(parsed)) {
+      setValidAmount(false)
+    } else {
+      if (parsed > 0 && parseInt(accountBalance.amount) - parsed < 1) {
```
jstuczyn: Could we change that to instead show the warning if after delegating user is going to have less the gas fee? (so that he would be able to undelegate in the future)

wallet-web/components/bond/BondNodeForm.tsx
```diff
@@ -1,380 +1,480 @@
...
+            fullWidth
+            InputProps={{
+              endAdornment: (
+                <InputAdornment position='end'>{DENOM}</InputAdornment>
+              ),
+            }}
+            onChange={(e: ChangeEvent<HTMLInputElement>) => {
+              if (
+                typeof parseInt(e.target.value) === 'number' &&
+                parseInt(accountBalance.amount) - parseInt(e.target.value) < 1
```
jstuczyn: Could we change that to instead show the warning if after bonding user is going to have less than 2 times the gas fee? (two because 1 is required for bonding and the other one for potential unbonding)
Also, maybe this check should be moved to another function? It's getting slightly busy in this element

jstuczyn: I meant component*

wallet-web/components/bond/BondNodeForm.tsx
fmtabbara: :smiley: I thought a type for the form input to include these values would help us to see what the form is dealing with.

wallet-web/common/helpers.ts
fmtabbara: As discussed this updated the gas fee alert to display PUNK instead of UPUNK. It shouldn't break anything else but happy to revisit once you've given it a test.

wallet-web/hooks/useGetBalance.ts
fmtabbara: Good spot. It shouldn't be there. I've removed it :+1:

wallet-web/components/bond/BondNodeForm.tsx
fmtabbara: I moved some of the logic into a helper function. You're right, it makes sense to do it that way. It also means the function can be reused for the delegation form.

wallet-web/components/delegate/DelegateForm.tsx
fmtabbara: Yep. That's done. It's part of a helper function now.

wallet-web/components/bond/BondNodeForm.tsx
fmtabbara: I've updated this. Hopefully that's what you're after.

jstuczyn (approved):

PR #702, opened 2021-07-26
Feature/docker improvements

neacsu:
* DRY the prefix and wasmd variables for validator and contract uploader
* Validator containers don't re-init when restarted, only if re-created

### DIFF
--- a/docker/validator/init_and_start.sh
+++ b/docker/validator/init_and_start.sh
@@ -6,38 +6,48 @@ PASSPHRASE=passphrase
 cd /root
 
 if [ "$1" = "genesis" ]; then
-	./nymd init nymnet --chain-id nymnet 2> /dev/null
-	sed -i 's/minimum-gas-prices = ""/minimum-gas-prices = "0.025u'"${BECH32_PREFIX}"'"/' /root/.nymd/config/app.toml
-	sed -i '0,/enable = false/s//enable = true/g' /root/.nymd/config/app.toml
-	sed -i 's/cors_allowed_origins = \[\]/cors_allowed_origins = \["*"\]/' /root/.nymd/config/config.toml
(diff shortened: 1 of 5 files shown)


docker/validator/init_and_start.sh
```diff
@@ -1,44 +1,53 @@
...
-	sed -i 's/create_empty_blocks = true/create_empty_blocks = false/' /root/.nymd/config/config.toml
-	sed -i 's/laddr = "tcp:\/\/127.0.0.1:26657"/laddr = "tcp:\/\/0.0.0.0:26657"/' /root/.nymd/config/config.toml
-	yes "${PASSPHRASE}" | ./nymd keys add node_admin 2>&1 >/dev/null | tail -n 1 > /genesis_volume/genesis_mnemonic
-	ADDRESS=$(yes "${PASSPHRASE}" | ./nymd keys show node_admin -a)
-	yes "${PASSPHRASE}" | ./nymd add-genesis-account "${ADDRESS}" 1000000000000000u${BECH32_PREFIX},1000000000000000stake
-	yes "${PASSPHRASE}" | ./nymd gentx node_admin 1000000000stake --chain-id nymnet 2> /dev/null
-	./nymd collect-gentxs 2> /dev/null
-	./nymd validate-genesis > /dev/null
-	cp /root/.nymd/config/genesis.json /genesis_volume/genesis.json
+  if [ ! -d "/root/.nymd" ]; then
```
jstuczyn, line 9: I love it

jstuczyn (approved):

docker/validator/init_and_start.sh
jstuczyn: Actually, just a question, does it even matter if stuff runs under root if it's inside a docker container?

neacsu: root is the default user for the ubuntu image we're using here. I don't even think there are other usable users on this image actually

PR #703, opened 2021-07-27
Allow users to sign out

### LINKED ISSUE #706: As a wallet user, I would like to be able to log out of the wallet

fmtabbara:


jstuczyn:
Upon `yarn build` I'm getting the following error:
```
❯ yarn build
yarn run v1.22.10
$ next build && next export
info  - Using webpack 5. Reason: future.webpack5 option enabled https://nextjs.org/docs/messages/webpack5
Failed to compile.

./components/undelegate/NodeUndelegation.tsx:68:10
Type error: Cannot find name 'undelegationStarted'.

  66 | 
  67 |     // we haven't clicked undelegate button yet
> 68 |     if (!undelegationStarted) {
     |          ^
  69 |       return (
  70 |         <React.Fragment>
  71 |           <NodeTypeChooser nodeType={nodeType} setNodeType={setNodeType} />
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

jstuczyn:
However, it does seem to run under `yarn dev` and it looks good - fix that compilation error and you're good to merge : )

wallet-web/components/unbond/UnbondNode.tsx
```diff
@@ -104,7 +104,7 @@ const UnbondNode = () => {
     }
 
     // we haven't clicked unbond button yet
-    if (!isLoading) {
+    if (isLoading === undefined) {
```
fmtabbara, line 107: I changed this slightly as `!isLoading` was preventing the `<Confirmation />` component from being displayed.

I see why before there was `isStarted` and `isFinished` states, but I think it's good to reduce the number of state variables where possible. This way we can achieve the same thing using one state variable.

fmtabbara: Same for the other components with this sort of structure.

PR #704, opened 2021-07-27
Feature/receive coins page + UI tweaks

### LINKED ISSUE #705: As a wallet user, I would like to have a "receive" page where I can see my own wallet address

fmtabbara:


jstuczyn:
I think you might be trying to squeeze stuff too aggressively, that's how the bond page looks in my browser:
image

fmtabbara:
Good point! I updated the breakpoints so should hopefully be better now! :crossed_fingers:

jstuczyn:
At some width it suddenly shrinks:
image

goes into 

image

fmtabbara:
Ah OK. I think width for larger screens needs to be increased. I've updated it. See what you think.

jstuczyn:
There's still a weird sudden shrink at 1920px width, but it's not as bad as it was before. I think it's good enough now : )

jstuczyn (approved):

fmtabbara:
There'll be a bit of movement as I've added some responsiveness to the pages, which should allow the content to look OK across all screen sizes. I'll have another quick look before merging. Thanks for the feedback :relaxed:

PR #715, opened 2021-08-02
Bug/form validation

### LINKED ISSUE #717: Wallet - Fix Bond Form validation issue

fmtabbara:


wallet-web/components/bond/BondNodeForm.tsx
```diff
@@ -120,59 +134,49 @@ export default function BondNodeForm(props: TBondNodeFormProps) {
...
+      validVersion: validateVersion(version),
+      validMixPort: validateRawPort(mixPort),
+      validLocation:
+        props.type == NodeType.Gateway ? validateLocation(location) : true,
+      validVerlocPort:
+        props.type === NodeType.Mixnode ? validateRawPort(verlocPort) : true,
+      validHttpApiPort:
+        props.type === NodeType.Mixnode ? validateRawPort(httpApiPort) : true,
+      validClientsPort:
+        props.type === NodeType.Gateway ? validateRawPort(clientsPort) : true,
```
fmtabbara, line 169: I'm just validating everything here including the values in the advanced area. If the advanced area is displayed then we validate the values that are supplied by the user. If the advanced area is not displayed then we are using the defaults from above which will be valid.

wallet-web/components/bond/BondNodeForm.tsx
```diff
@@ -120,59 +134,49 @@ export default function BondNodeForm(props: TBondNodeFormProps) {
...
-    let validHost = validateHost(host)
-    let validVersion = validateVersion(version)
-
-    let validLocation =
-      props.type == NodeType.Gateway ? validateLocation(location) : true
-
+    verlocPort = DEFAULT_VERLOC_PORT.toString(),
+    mixPort = DEFAULT_MIX_PORT.toString(),
+    httpApiPort = DEFAULT_HTTP_API_PORT.toString(),
+    clientsPort = DEFAULT_CLIENTS_PORT.toString(),
```
fmtabbara, line 141: If these values are undefined i.e. the advanced area is not displayed, then we fallback to our default values.

fmtabbara:
I think the form could benefit from a little rewrite. What we have is perfectly fine and works well but we could reduce the overall code a little if we were to rewrite some of it. Maybe for discussion in sprint planning.

wallet-web/components/bond/BondNodeForm.tsx
jstuczyn: That sounds like a good approach

wallet-web/components/bond/BondNodeForm.tsx
jstuczyn: I guess this would simplify stuff

jstuczyn (approved):

jstuczyn:
Feel free to rewrite whatever you think requires a rewrite. When I was writing this piece I knew it would be replaced relatively soon afterwards so I didn't put much thinking into it : )

fmtabbara:
In general I think the form is fine. I'll just look at changing how the react state is being updated (in a separate PR). Everything else can pretty much stay the same.

neacsu (approved):
Bug is fixed after the patches

PR #721, opened 2021-08-03
update state management and validation

fmtabbara:
I appreciate these some of these files might be tricky to compare due to the changes that I've made. I'm happy to talk through the changes if that's preferred.

In a nutshell I have split the original bond form component out into three parts

- Utility functions (utils.ts)
- State management (useBondForm.ts)
- UI (BondNodeForm.tsx)

The biggest change is running validation checks in real-time instead of on submission. The 'Bond' button will be disabled until all fields are valid. Validation functions remain the same apart from the validateHost function which I've updated.

I'll keep this as a draft PR until I've finished testing.


wallet-web/components/bond/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+export const isValidHostname = (value) => {
+  if (typeof value !== 'string') return false
+
+  const validHostnameChars = /^[a-zA-Z0-9-.]{1,253}\.?$/g
+  if (!validHostnameChars.test(value)) {
+    return false
+  }
+
+  if (value.endsWith('.')) {
+    value = value.slice(0, value.length - 1)
```
jstuczyn: I'm not entirely sure I understand that, why are we removing the last character if it's a `.`?

wallet-web/components/bond/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+    const bytes = bs58.decode(key)
+    // of length 32
+    return bytes.length === 32
+  } catch {
+    return false
+  }
+}
+
+export const isValidHostname = (value) => {
+  if (typeof value !== 'string') return false
```
jstuczyn: We have the niceness of typescript, so change your function definition to `export const isValidHostname = (value: string)` and the check will be done at compile time : )

wallet-web/components/bond/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  if (value.endsWith('.')) {
+    value = value.slice(0, value.length - 1)
+  }
+
+  if (value.length > 253) {
+    return false
+  }
+
+  const labels = value.split('.')
```
jstuczyn: What if we are faced with an ipv6 address that is separated by `:` (and has bunch of edge cases)? 
Also remember that a hostname like `nymtech.net` should also be valid

wallet-web/components/bond/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+    // of length 32
+    return bytes.length === 32
+  } catch {
+    return false
+  }
+}
+
+export const isValidHostname = (value: string) => {
+  // regex for ipv4 and ipv6 - source https://www.regextester.com/104038
+  const hostnameRegex =
```
jstuczyn, line 35: This is a personal bias, but I love this quote about regexes:
```
Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.
```
If we're doing proper ip validation, surely there exists some package for that? Also in our case, `foo.com` is a valid hostname : )

fmtabbara: :laughing: I like that expression too. I have heard it before

I try to avoid regex as much as possible! But I think in this case it might be a good option. Saying that I will look into some packages that we might be able to use as an alternative.

I have  updated this part so that it will now validate ips and hostnames.

I'm adding `foo.com` to the tests :D

wallet-web/common/helpers.ts
```diff
@@ -110,15 +110,13 @@ export const validateIdentityKey = (key: string): boolean => {
...
-  // first of all it must be an integer
-  const port = parseInt(rawPort)
-  if (port == null) {
-    return false
-  }
-  // and it must be a non-zero 16 bit unsigned integer
-  return port >= 1 && port <= 65535
-}
+export const validateRawPort = (rawPort: number): boolean =>
+  typeof rawPort === 'number' &&
```
jstuczyn: the `typeof rawPort === 'number'` check is redundant, we know for sure it's a number since typescript wouldn't allow anything else here

jstuczyn (approved):

PR #724, opened 2021-08-05
Rust nymd/cosmwasm client

jstuczyn:
Following @futurechimp suggestion, I decided to go against creating a massive pull request with all the features. I've reached a good checkpoint location in my branch to create a pull request.

What it does is basically is introduces a cosmwasm client (as well as the signing variant) capable of connecting and talking with cosmwasm validators. Think of it as simplified rust equivalent of cosmjs (https://github.com/cosmos/cosmjs). It might be possibly of some interest to @ethanfrey .

Note: it's **not** wasm compatible at this point due to the tendermint rpc client being wasm-incompatible.

Possibly all of this might be at some point moved into a pull request to https://github.com/cosmos/cosmos-rust/ 

Edit: Since creating this PR more features have been added. The clients have been made into async traits (at the cost of making signing operation slightly more expensive due to having to re-derive keys at every request, but in the grand scheme of things it's a small price) and an example stub for nymd api has been established.


ethanfrey:
Very interesting work. I have yet to review the code, but very happy to see the progress here.

As to compiling to wasm, I don't think that is very interesting. There are good existing clients in CosmJS for the browser as well as Node.js needs.

What is more interesting to experiment with IMO is embedding via FFI with other languages. For example, a number of people have asked for a Python CosmWasm client, with signing and Stargate support. If you could compile this lib to a dll that could be connected to Python, that would be very cool and open the path for many more languages.

I did look into such an idea a while ago but wrote no code. You should be able to use cbindgen on the Rust side to generate standard c bindings. And then http://www.swig.org/ may be useful to generate a Python stub around such c bindings, so this would not be too much work. (One would probably want to wrap the auto-generated api to be more pythonic, but someone else could do that if you could manage the language gap)

In any case, great work and thanks for the ping. I will try to take a look after our upcoming cosmwasm/wasmd release

common/client-libs/validator-client/src/nymd/wallet/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+use cosmos_sdk::bip32::{DerivationPath, XPrv};
+use cosmos_sdk::crypto::secp256k1::SigningKey;
+use cosmos_sdk::crypto::PublicKey;
+use cosmos_sdk::tx::SignDoc;
+use cosmos_sdk::{tx, AccountId};
+
+pub const DEFAULT_COSMOS_DERIVATION_PATH: &str = "m/44'/118'/0'/0/0";
+// TODO: probably if this is to be made into a generic-cosmos library, the default
+// prefix should be "cosmos"
+pub const DEFAULT_BECH32_ADDRESS_PREFIX: &str = "punk";
```
futurechimp: I think we have a centralized config location for these now?

common/client-libs/validator-client/src/nymd/wallet/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+pub struct DirectSecp256k1HdWalletBuilder {
+    /// The password to use when deriving a BIP39 seed from a mnemonic.
+    bip39_password: String,
+
+    /// The BIP-32/SLIP-10 derivation paths. Defaults to the Cosmos Hub/ATOM path `m/44'/118'/0'/0/0`
+    hd_paths: Vec<DerivationPath>,
+
+    /// The bech32 address prefix (human readable part). Defaults to "punk".
```
futurechimp: If you change to use the centralized config, this comment might need to change

futurechimp (commented):
Looks good overall. I think there's a new chunk of code which Bogdan and Max worked on which DRYs up config values all over the codebase (bech32 prefix, contract address, etc).

common/client-libs/validator-client/src/nymd/cosmwasm_client/logs.rs
```diff
@@ -0,0 +1,92 @@
...
+    msg_index: usize,
+    // unless I'm missing something obvious, the "log" type in cosmjs is always an empty string
+    // and launchpad cosmos validator was setting it to what essentially is just the raw version of what
+    // we received (and we don't care about launchpad, we, as the time of writing this, work on the stargate)
+    // log: String,
+    events: Vec<cosmwasm_std::Event>,
+}
+
+/// Searches in logs for the first event of the given event type and in that event
+/// for the first first attribute with the given attribute key.
```
neacsu: Double `first`

common/client-libs/validator-client/src/nymd/wallet/mod.rs
jstuczyn: Fair point. Though I don't think `DEFAULT_COSMOS_DERIVATION_PATH` should be moved anywhere as from what I understand, it's cosmos-validator specific

ethanfrey: This derivation path is not tied to validators (like Tendermint) but rather any sdk account.

The purpose of using this is to maintain compatibility with ledger and golang CLI. If you do not have ledger support in this library, that is not so important (the cosmos ledger app only works for `m/44'/108'` prefix).

The other point would be that someone could import the same mnemonic in your nym rust client and in the wasmd cli tool and they would produce the same address. If this is important, it is good to stick with the default. If it is not, then there is no strong need to stick with it and you can let users configure it freely (every client may use a different one even). 

In CosmJS we allow library users to pick what they want. And it only has really been an issue for ledger compatibility.

PR #730, opened 2021-08-09
Feature/nymd client fee handling

jstuczyn:
Small pull request that introduces fee handling mechanism for the nymd client.

### DIFF
--- a/common/client-libs/validator-client/src/nymd/fee_helpers.rs
+++ b/common/client-libs/validator-client/src/nymd/fee_helpers.rs
@@ -47,7 +47,7 @@ impl Operation {
         // - gas limit
         // - fees
         let gas_limit = gas_limit.unwrap_or_else(|| self.default_gas_limit());
-        let fee = calculate_fee(&gas_price, gas_limit);
+        let fee = calculate_fee(gas_price, gas_limit);
         Fee::from_amount_and_gas(fee, gas_limit)
     }
 }
--- a/common/client-libs/validator-client/src/nymd/fee_helpers.rs
+++ b/common/client-libs/validator-client/src/nymd/fee_helpers.rs
@@ -2,14 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::nymd::GasPrice;
-use crate::ValidatorClientError;
-use config::defaults;
 use cosmos_sdk::tx::{Fee, Gas};
-use cosmos_sdk::{Coin, Denom};
-use cosmwasm_std::{Decimal, Uint128};
-use std::str::FromStr;
+use cosmos_sdk::Coin;
+use cosmwasm_std::Uint128;
 
-pub(crate) enum Operation {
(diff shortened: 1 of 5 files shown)


common/client-libs/validator-client/src/nymd/fee_helpers.rs
```diff
@@ -0,0 +1,88 @@
...
+impl Operation {
+    pub(crate) fn default_gas_limit(&self) -> Gas {
+        match self {
+            Operation::Upload => 2_500_000u64.into(),
+            Operation::Init => 500_000u64.into(),
+            Operation::Migrate => 200_000u64.into(),
+            Operation::ChangeAdmin => 80_000u64.into(),
+            Operation::Send => 80_000u64.into(),
+
+            Operation::BondMixnode => 175_000u64.into(),
```
neacsu, line 39: Would it make sense to have those values as `const`s in one place, so we can modify them more easily?

common/client-libs/validator-client/src/nymd/fee_helpers.rs
```diff
@@ -0,0 +1,88 @@
...
+    Upload,
+    Init,
+    Migrate,
+    ChangeAdmin,
+    Send,
+
+    BondMixnode,
+}
+
+pub(crate) fn calculate_fee(gas_price: &GasPrice, gas_limit: Gas) -> Coin {
```
neacsu, line 20: Maybe it's a naming convention that I'm not familiar with, in which case just ignore this comment, but isn't it more of a `gas_fee` rather than a `gas_limit`? Same goes for every appearance of this naming in this PR.

common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -97,7 +118,7 @@ impl<C> NymdClient<C> {
     }
 
     // now the question is as follows: will denom always be in the format of `u{prefix}`?
-    fn denom(&self) -> Denom {
+    pub fn denom(&self) -> Denom {
         format!("u{}", self.contract_address.prefix())
```
neacsu, line 122: Why didn't we get the `denom` from the defaults instead of deriving it from the contract address?

neacsu (approved):

common/client-libs/validator-client/src/nymd/fee_helpers.rs
jstuczyn: I'm 99% sure it's `limit` as if during contract execution we reach the amount specified here, the execution will come to a halt and tx fill fail to be included.

common/client-libs/validator-client/src/nymd/mod.rs
jstuczyn: We are already grabbing contract address from the defaults and everything has to match whatever is specified by the contract.

common/client-libs/validator-client/src/nymd/fee_helpers.rs
jstuczyn: Honestly, I don't know. I couldn't find any good place for it

PR #731, opened 2021-08-09
Explorer API - add port check and node description/stats proxy

mmsinclair:
This PR adds new features to the Explorer API:

- on-demand cached request to all open ports on a mix node (as defined in the mix node bond fields `mix_port`,`verloc_port`, `http_api_port`)
- on-demand cached proxy `/description` and `/stats` from mix node http API to allow the Network Explorer UI to make requests for these without mixed content errors (http fetches from an app served on https)

The default cache TTL is 30 minutes.


explorer-api/src/http/mod.rs
```diff
@@ -15,8 +16,9 @@ pub(crate) fn start(state: ExplorerApiStateContext) {
 
         rocket::build()
             .configure(config)
-            .mount("/countries", make_default_routes())
+            .mount("/countries", country_statistics_make_default_routes())
             .mount("/ping", ping_make_default_routes())
+            .mount("/mix-node", mix_node_make_default_routes())
             .mount("/swagger", make_swagger_ui(&get_docs()))
             // .register("/", catchers![not_found])
```
jstuczyn: Any reason for leaving this commented out?

explorer-api/src/mix_node/http.rs
```diff
@@ -0,0 +1,124 @@
...
+        .get_description(pubkey.to_string())
+        .await
+    {
+        Some(cache_value) => {
+            trace!("Returning cached value for {}", pubkey);
+            Some(Json(cache_value))
+        }
+        None => {
+            trace!("No valid cache value for {}", pubkey);
+            match state.inner.get_mix_node(pubkey).await {
```
jstuczyn, line 32: To collapse it a little bit, you can use `map()` instead of `match`

explorer-api/src/mix_node/cache.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    pub(crate) fn get(&self, identity_key: IdentityKey) -> Option<T>
+    where
+        T: Clone,
+    {
+        self.inner
+            .get(&identity_key)
+            .filter(|cache_item| cache_item.valid_until > SystemTime::now())
```
jstuczyn, line 22: Just a suggestion, perhaps there should be some way of removing stale entries from the cache? Right now everything will just stay there forever

explorer-api/src/mix_node/http.rs
```diff
@@ -0,0 +1,124 @@
...
+#[get("/<pubkey>/description")]
+pub(crate) async fn get_description(
+    pubkey: &str,
+    state: &State<ExplorerApiStateContext>,
+) -> Option<Json<NodeDescription>> {
+    match state
+        .inner
+        .mix_node_cache
+        .clone()
+        .get_description(pubkey.to_string())
```
jstuczyn: Looking at the cache implementation, you can get rid of the `to_string()` call here and just pass references all the way down (but you'd need to change the types of the functions down the stack)

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+                valid_until: SystemTime::now() - Duration::from_secs(60), // in the past
+            })),
+        }
+    }
+
+    pub(crate) async fn get(self) -> MixNodesResult {
+        // check ttl
+        let valid_until = self.inner.clone().read().await.valid_until;
+
+        if valid_until.lt(&SystemTime::now()) {
```
jstuczyn: You can use `<` operator directly

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub(crate) fn new() -> Self {
+        ThreadsafeMixNodesResult {
+            inner: Arc::new(RwLock::new(MixNodesResult {
+                value: vec![],
+                valid_until: SystemTime::now() - Duration::from_secs(60), // in the past
+            })),
+        }
+    }
+
+    pub(crate) async fn get(self) -> MixNodesResult {
```
jstuczyn: you probably want to use `&self here

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        if valid_until.lt(&SystemTime::now()) {
+            // force reload
+            self.refresh().await;
+        }
+
+        // return in-memory cache
+        self.inner.clone().read().await.clone()
+    }
+
+    pub(crate) async fn refresh_and_get(self) -> MixNodesResult {
```
jstuczyn: more &self action here

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        self.refresh().await;
+        self.inner.clone().read().await.clone()
```
jstuczyn: You don't need to clone inner here

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    async fn refresh(&self) {
+        // get mixnodes and keep
+        let value = retrieve_mixnodes().await;
+        self.inner
+            .clone()
```
jstuczyn: Another redundant clone

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+            .write()
+            .await
+            .clone_from(&MixNodesResult {
+                value,
+                valid_until: SystemTime::now() + Duration::from_secs(60 * 10), // valid for 10 minutes
+            });
```
jstuczyn: (this is not a complaint or anything, just a tip in case you weren't aware). You can also assign to value behind the lock like this: 
```
        *self.inner.write().await = MixNodesResult {
            value,
            valid_until: SystemTime::now() + Duration::from_secs(60 * 10), // valid for 10 minutes
        }
```

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        self.inner.clone().read().await.clone()
+    }
+
+    pub(crate) async fn refresh_and_get(self) -> MixNodesResult {
+        self.refresh().await;
+        self.inner.clone().read().await.clone()
+    }
+
+    async fn refresh(&self) {
+        // get mixnodes and keep
```
jstuczyn: and keep what? : )

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub(crate) async fn retrieve_mixnodes() -> Vec<MixNodeBond> {
+    let client = new_validator_client();
+
+    info!("About to retrieve mixnode bonds...");
+
+    let bonds: Vec<MixNodeBond> = match client.get_cached_mix_nodes().await {
+        Ok(result) => result,
+        Err(e) => panic!("Unable to retrieve mixnode bonds: {:?}", e),
```
jstuczyn: Should we panic here or rather just return an error?

explorer-api/src/ping/http.rs
```diff
@@ -1,19 +1,103 @@
...
 use rocket::serde::json::Json;
-use rocket::Route;
-use serde::Deserialize;
-use serde::Serialize;
+use rocket::{Route, State};
+
+use crate::ping::models::PingResponse;
+use crate::state::ExplorerApiStateContext;
+
+const CONNECTION_TIMEOUT_SECONDS: u64 = 10;
```
jstuczyn: You can be more explicit about it and define it directly as duration:
`const CONNECTION_TIMEOUT_SECONDS: Duration = Duration::from_secs(10);`

explorer-api/src/ping/http.rs
```diff
@@ -1,19 +1,103 @@
...
-#[derive(Deserialize, Serialize, JsonSchema)]
-pub(crate) struct PingResponse {
-    response_time: u32,
+#[openapi(tag = "ping")]
+#[get("/<pubkey>")]
+pub(crate) async fn index(
+    pubkey: &str,
+    state: &State<ExplorerApiStateContext>,
+) -> Option<Json<PingResponse>> {
+    match state.inner.ping_cache.clone().get(pubkey.to_string()).await {
```
jstuczyn: You don't need the clone here (and the `get` can get away with &str)

explorer-api/src/ping/http.rs
```diff
@@ -1,19 +1,103 @@
...
+            Some(Json(PingResponse {
+                ports: cache_value.ports,
+            }))
+        }
+        None => {
+            trace!("No cache value for {}", pubkey);
+
+            match state.inner.get_mix_node(pubkey).await {
+                Some(bond) => {
+                    let mut ports: HashMap<u16, bool> = HashMap::new();
```
jstuczyn: This test kinda looks like it wanted to live in a separate function

explorer-api/src/ping/http.rs
```diff
@@ -1,19 +1,103 @@
...
+                    trace!("Tested mix node {}: {:?}", pubkey, ports);
+
+                    let response = PingResponse { ports };
+
+                    // cache for 1 min
+                    trace!("Caching value for {}", pubkey);
+                    state
+                        .inner
+                        .ping_cache
+                        .clone()
```
jstuczyn: No need for clone

explorer-api/src/ping/http.rs
```diff
@@ -1,19 +1,103 @@
...
+                        .set(pubkey.to_string(), response.clone())
+                        .await;
+
+                    // return response
+                    Some(Json(response))
+                }
+                None => Option::None,
```
jstuczyn: You don't need to say `Option::None`, `None` by itself is sufficient

explorer-api/src/ping/models.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    pub(crate) async fn get(&self, identity_key: IdentityKey) -> Option<PingResponse> {
+        self.inner
+            .read()
+            .await
+            .get(&identity_key)
+            .and_then(|cache_item| {
+                if cache_item.valid_until.gt(&SystemTime::now()) {
```
jstuczyn: I think we can also use our `filter` and `map` approach here

explorer-api/src/ping/models.rs
```diff
@@ -0,0 +1,61 @@
...
+                    None
+                }
+            })
+    }
+
+    pub(crate) async fn set(&self, identity_key: IdentityKey, item: PingResponse) {
+        self.inner.write().await.insert(
+            identity_key,
+            PingCacheItem {
+                valid_until: SystemTime::now() + Duration::from_secs(5),
```
jstuczyn: Should it really be only valid for 5s?

explorer-api/src/ping/models.rs
```diff
@@ -0,0 +1,61 @@
...
+                ports: item.ports.clone(),
```
jstuczyn: You don't need to clone it here as you own the entire `item` object so you can just move the ports from it

explorer-api/src/state.rs
```diff
@@ -1,20 +1,40 @@
-use chrono::{DateTime, Utc};
-use serde::{Deserialize, Serialize};
 use std::fs::File;
 use std::path::Path;
 
+use chrono::{DateTime, Utc};
```
jstuczyn, line 4: Any reason for chrono here?

explorer-api/src/state.rs
```diff
@@ -1,20 +1,40 @@
...
     pub(crate) country_node_distribution: ConcurrentCountryNodesDistribution,
+    pub(crate) mix_nodes: ThreadsafeMixNodesResult,
+    pub(crate) mix_node_cache: ThreadsafeMixNodeCache,
+    pub(crate) ping_cache: ThreadsafePingCache,
+}
+
+impl ExplorerApiState {
+    pub(crate) async fn get_mix_node(&self, pubkey: &str) -> Option<MixNodeBond> {
+        self.mix_nodes
+            .clone()
```
jstuczyn: Redundant clone

explorer-api/src/state.rs
```diff
@@ -1,20 +1,40 @@
...
+            .get()
+            .await
+            .value
+            .iter()
+            .find(|node| node.mix_node.identity_key == pubkey)
```
jstuczyn, line 34: Since we're having to go through the whole vector each time we want to look up each mixnode, maybe vector is not the best data structure for it? Perhaps HashMap would have been better?

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use rocket::tokio::sync::RwLock;
+
+use mixnet_contract::MixNodeBond;
+use validator_client::Config;
+
+#[derive(Clone, Debug)]
+pub(crate) struct MixNodesResult {
+    pub(crate) valid_until: SystemTime,
+    pub(crate) value: Vec<MixNodeBond>,
```
jstuczyn, line 12: As discussed in one of my comments in `ExplorerApiState`, perhaps Vec is not the best data structure here

explorer-api/src/mix_node/http.rs
```diff
@@ -0,0 +1,124 @@
...
+#[openapi(tag = "mix_node")]
+#[get("/<pubkey>/stats")]
+pub(crate) async fn get_stats(
+    pubkey: &str,
+    state: &State<ExplorerApiStateContext>,
+) -> Option<Json<NodeStats>> {
+    match state
+        .inner
+        .mix_node_cache
+        .clone()
```
jstuczyn: Redundant clone

explorer-api/src/mix_node/http.rs
```diff
@@ -0,0 +1,124 @@
...
+                        }
+                    }
+                }
+                None => Option::None,
+            }
+        }
+    }
+}
+
+async fn get_mix_node_description(host: &str, port: &u16) -> Result<NodeDescription, ReqwestError> {
```
jstuczyn, line 106: I'm surprised clippy didn't yell at you for using `&u16`. You can just use `u16` directly. It's a 2byte primitive value that has `Copy` defined on it and that's cheaper than taking reference.

explorer-api/src/http/mod.rs
mmsinclair: Nope, just didn't get round to adding it back. Have added something similar to the mixnode http api.

explorer-api/src/mix_node/http.rs
mmsinclair: I tried this out, but the compiler was complaining about unstable references inside the anonymous async fn so I'm going to leave it as a match

explorer-api/src/mix_node/cache.rs
mmsinclair: Yeah, it is a bit wasteful generally - although I'm hoping the memory requirements for thousands of nodes will still be megabytes. I can add a cleanup background task sometime if it becomes a problem.

explorer-api/src/mix_node/http.rs
mmsinclair: 👍

explorer-api/src/mix_nodes/mod.rs
mmsinclair: 👍

explorer-api/src/mix_nodes/mod.rs
mmsinclair: the mix nodes of course 😉  - changed the comment

explorer-api/src/mix_nodes/mod.rs
mmsinclair: Dave was teaching me about panicking 😮  - have changed to a error log and returning an empty vec

explorer-api/src/ping/http.rs
mmsinclair: Much nicer 👍

explorer-api/src/ping/models.rs
mmsinclair: Done 😀

explorer-api/src/ping/models.rs
mmsinclair: Good spot - changed to 1 hour

explorer-api/src/state.rs
mmsinclair: It was so the response was formatted as an iso timestamp - I could change it to use `humantime-serde` if Chrono disgusts you?

explorer-api/src/ping/http.rs
mmsinclair: Yup, moved into separate function

explorer-api/src/state.rs
mmsinclair: Yes, it would be - mind if I add that in another PR where I add the geoip resolved location to the bond?

explorer-api/src/state.rs
jstuczyn: Oh, chrono is absolutely fine by me - I just couldn't see any of its usages in the code so I assumed it was a dead dependency : )

explorer-api/src/state.rs
jstuczyn: Sounds good

jstuczyn (approved):

jstuczyn (approved):

PR #732, opened 2021-08-10
Feature/completing nymd client api

jstuczyn:
Another nice checkpoint in the signing client story. 

This PR completes the nymd client's API. The only remaining thing is putting it in the proper `validator client` and making the split with validator API.

~~Since this PR builds on https://github.com/nymtech/nym/pull/730, it will remain as a draft until the former is merged.~~


jstuczyn:
~~Just realised I forgot about one thing - currently this API doesn't allow for easy contract upload/initialisation as you need to provide a valid contract address during construction.~~

common/client-libs/validator-client/src/nymd/fee_helpers.rs
```diff
@@ -37,6 +48,16 @@ impl Operation {
             Operation::Send => 80_000u64.into(),
 
             Operation::BondMixnode => 175_000u64.into(),
+            Operation::UnbondMixnode => 175_000u64.into(),
```
neacsu, line 51: In case those values are just an educated guess (which they appear to be), should we do some measurements at some point to see what is the actual cost of each operation, so that they are fair relative to each other?

common/client-libs/validator-client/src/nymd/gas_price.rs
```diff
@@ -51,6 +51,11 @@ impl Default for GasPrice {
 mod tests {
     use super::*;
 
+    #[test]
+    fn default_gas_price_is_valid() {
+        let _ = GasPrice::default();
+    }
+
```
neacsu, line 58: This doesn't look like a test :D

neacsu (approved):

common/client-libs/validator-client/src/nymd/gas_price.rs
jstuczyn: It actually is, if default is at some point broken, it will panic : )

common/client-libs/validator-client/src/nymd/fee_helpers.rs
jstuczyn: So I've actually only "measured" costs of mixnode bond and unbond which are slightly below the values here. But I definitely agree everything else has to be properly measured.

common/client-libs/validator-client/src/nymd/gas_price.rs
jstuczyn: For example if somebody sets an invalid `DENOM` or `GAS_PRICE_AMOUNT` in our defaults

PR #734, opened 2021-08-11
Explorer API: add new HTTP resource to decorate mix nodes with geoip locations

mmsinclair:
This PR adds a new HTTP resource that returns the list of mix nodes with an additional optional field for the location information from the geoip service.

It also:

1. Changes the cache of mix nodes from a `Vec` to a `HashMap`
2. Caches the resolved geoip locations for each mix node in memory and writes this to disk along with the country distributions so that there is location data for most of the mix nodes should the node restart. Any mix nodes that have joined since the restart will not have location information until the country distribution task runs.


explorer-api/src/state.rs
```diff
@@ -30,15 +32,15 @@ impl ExplorerApiState {
             .get()
             .await
             .value
-            .iter()
-            .find(|node| node.mix_node.identity_key == pubkey)
-            .cloned()
+            .get(pubkey)
+            .map(|bond| bond.bond.clone())
```
neacsu: Maybe the argument should be something like `bond_with_location`, to avoid the naming collision.

explorer-api/src/state.rs
```diff
@@ -105,3 +106,9 @@ impl ExplorerApiStateContext {
 fn get_state_file_path() -> String {
     std::env::var("API_STATE_FILE").unwrap_or_else(|_| STATE_FILE.to_string())
 }
+
+fn get_state_from_file(json_file_path: &Path) -> Result<ExplorerApiStateOnDisk, Box<dyn Error>> {
+    let file = File::open(json_file_path)?;
+    let state = serde_json::from_reader(file)?;
+    Ok(state)
```
neacsu: These two lines could simply be replaced by `serde_json::from_reader(file)`

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -21,12 +41,42 @@ impl ThreadsafeMixNodesResult {
...
             inner: Arc::new(RwLock::new(MixNodesResult {
-                value: vec![],
+                value: HashMap::new(),
                 valid_until: SystemTime::now() - Duration::from_secs(60), // in the past
+                location_cache: LocationCache::new(),
             })),
         }
     }
 
+    pub(crate) fn attach(location_cache: LocationCache) -> Self {
```
neacsu: I would rename this to `new_with_location_cache`, as attaching looks like doing work on something already existing.

neacsu (approved):
lgtm overall

explorer-api/src/country_statistics/mod.rs
```diff
@@ -51,17 +52,35 @@ impl CountryStatistics {
```
jstuczyn, line 41: Thid actually refers to line 46 - there's another unnecessary Arc clone there : )

explorer-api/src/country_statistics/mod.rs
```diff
@@ -51,17 +52,35 @@ impl CountryStatistics {
 
         info!("Locating mixnodes...");
         for (i, bond) in mixnode_bonds.iter().enumerate() {
```
jstuczyn: Since you only seem to care about values of the map, you might want to iterate via `mixnode_bonds.values()` to get rid of those `.1` uses

explorer-api/src/country_statistics/mod.rs
```diff
@@ -51,17 +52,35 @@ impl CountryStatistics {
...
+                        bond.1.bond.mix_node.host,
+                        three_letter_iso_country_code,
                     );
 
+                    self.state
+                        .inner
+                        .mix_nodes
+                        .set_location(
+                            &bond.1.bond.mix_node.identity_key,
+                            Location {
```
jstuczyn: To collapse it a little bit, maybe we could have a constructor on `Location` that takes `three_letter_iso_country_code` and `GeoLocation` as arguments? Then we'd just have `Location::new(three_letter_iso_country_code, location)`

explorer-api/src/mix_node/http.rs
```diff
@@ -1,13 +1,54 @@
...
+        state
+            .inner
+            .mix_nodes
+            .get()
+            .await
+            .value
+            .values()
+            .map(|i| {
+                let mix_node = i.bond.clone();
+                PrettyMixNodeBondWithLocation {
```
jstuczyn, line 41: Another case of maybe a constructor being useful?

explorer-api/src/mix_nodes/mod.rs
jstuczyn: Yep. I agree, `attach` seems like something that should have been called on `self`

jstuczyn: Perhaps you might change it into `pub(crate) fn attach(mut self, location_cache: LocationCache) -> Self` method that only modified the `location_cache` so that you could call `ThreadsafeMixNodesResult::new().attach(cache)`

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -21,12 +41,42 @@ impl ThreadsafeMixNodesResult {
...
+            inner: Arc::new(RwLock::new(MixNodesResult {
+                value: HashMap::new(),
+                valid_until: SystemTime::now() - Duration::from_secs(60), // in the past
+                location_cache,
+            })),
+        }
+    }
+
+    pub(crate) async fn get_location_cache(&self) -> LocationCache {
+        self.inner.read().await.location_cache.clone()
```
jstuczyn, line 93: Is there any particular reason you need access to the entire cache? Cloning few thousand values every time might not be the best solution : )

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -21,12 +41,42 @@ impl ThreadsafeMixNodesResult {
...
+        // add the location to the mix node
+        guard
+            .value
+            .entry(identity_key.to_string())
+            .and_modify(|item| item.location = Some(location));
+    }
+
     pub(crate) async fn get(&self) -> MixNodesResult {
         // check ttl
         let valid_until = self.inner.clone().read().await.valid_until;
```
jstuczyn: No need for Arc clone

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -48,9 +98,24 @@ impl ThreadsafeMixNodesResult {
     async fn refresh(&self) {
         // get mixnodes and cache the new value
         let value = retrieve_mixnodes().await;
+        let location_cache = self.inner.read().await.location_cache.clone();
         self.inner.write().await.clone_from(&MixNodesResult {
-            value,
+            value: value
+                .iter()
```
jstuczyn: Since you own the `value`, you can just call `.into_iter()` to prevent cloning later on

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -48,9 +98,24 @@ impl ThreadsafeMixNodesResult {
     async fn refresh(&self) {
         // get mixnodes and cache the new value
         let value = retrieve_mixnodes().await;
+        let location_cache = self.inner.read().await.location_cache.clone();
         self.inner.write().await.clone_from(&MixNodesResult {
```
jstuczyn: You don't need to clone here, you can just assign it directly with `*self.inner.write().await = &MixNodeResult {...}`

explorer-api/src/mix_nodes/mod.rs
```diff
@@ -48,9 +98,24 @@ impl ThreadsafeMixNodesResult {
...
-            value,
+            value: value
+                .iter()
+                .map(|bond| {
+                    (
+                        bond.mix_node.identity_key.to_string(),
+                        MixNodeBondWithLocation {
+                            bond: bond.clone(),
```
jstuczyn: `into_iter()` will get rid of this clone

explorer-api/src/state.rs
```diff
@@ -105,3 +106,9 @@ impl ExplorerApiStateContext {
 fn get_state_file_path() -> String {
     std::env::var("API_STATE_FILE").unwrap_or_else(|_| STATE_FILE.to_string())
 }
+
+fn get_state_from_file(json_file_path: &Path) -> Result<ExplorerApiStateOnDisk, Box<dyn Error>> {
```
jstuczyn: Somebody's being lazy with `Box<dyn Error>>` : )

explorer-api/src/mix_nodes/mod.rs
neacsu: Giving it a shot with `suggestion`
```suggestion
    pub(crate) fn new_with_location_cache(location_cache: LocationCache) -> Self {
```

explorer-api/src/state.rs
mmsinclair: Renamed to `cache_item`

explorer-api/src/country_statistics/mod.rs
mmsinclair: I just really want to increment the reference count - feels really wrong not to #cpp

explorer-api/src/country_statistics/mod.rs
mmsinclair: Refactored it a bit to simplify with the constructor

explorer-api/src/mix_nodes/mod.rs
mmsinclair: This was really just a function to hide the messy bits away when I need to get the cache to write it to file. Any suggestions on passing a reference back rather?

explorer-api/src/state.rs
mmsinclair: How hard can it be to open a file!! 😉 

Refactored this to something shorter and removed this function.

PR #736, opened 2021-08-12
Feature/nymd client integration

jstuczyn:
This pull request integrates the new validator client into the code base - mixnodes, gateways, clients and validator api are now using it.

The final remaining thing, which will be done in another pull request, is to make actual payments happen.

I'm sorry for whoever reviews it, but after the BIG MERGE ™️  this PR got rather bloated : (


clients/native/src/commands/init.rs
```diff
@@ -61,15 +60,29 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
         )
 }
 
+// this behaviour should definitely be changed, we shouldn't
+// need to get bandwidth credential for registration
```
neacsu, line 64: Why not? When do you think we should get it? What I had in mind was that we get it at registration/authentication and consume it during that session.

clients/native/src/commands/upgrade.rs
```diff
@@ -3,16 +3,21 @@
...
+#[allow(dead_code)]
+fn fail_upgrade<D1: Display, D2: Display>(from_version: D1, to_version: D2) -> ! {
+    print_failed_upgrade(from_version, to_version);
+    process::exit(1)
+}
 
 fn print_start_upgrade<D1: Display, D2: Display>(from: D1, to: D2) {
     println!(
-        "\n==================\nTrying to upgrade client from {} to {} ...",
+        "\n==================\nTrying to upgrade gateway from {} to {} ...",
```
neacsu: Why is this not the client that's being upgraded anymore?

common/coconut-interface/src/lib.rs
```diff
@@ -90,6 +63,15 @@ pub struct VerifyCredentialBody {
...
     public_attributes: Vec<String>,
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // TODO: CHECK THIS!!
+    // public_attributes: Vec<Attribute>,
```
neacsu: I'm not sure I get why this is repeated on multiple lines. Maybe this is something that could use the `SECURITY` tag Dave proposed (similar to `TODO`, but with higher priority) ?

common/coconut-interface/src/lib.rs
```diff
@@ -123,19 +105,28 @@ pub struct BlindSignRequestBody {
...
```
neacsu: Same as above

common/credentials/src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+use coconut_interface::{
+    aggregate_signature_shares, aggregate_verification_keys, prepare_blind_sign, prove_credential,
+    Attribute, BlindSignRequestBody, Credential, Parameters, Signature, SignatureShare,
+    VerificationKey,
+};
+use url::Url;
+
+// Note: list of validators must be correctly ordered by the polynomial coordinates used
+// during key generation and it is responsibility of the caller to ensure that correct
+// number of them is provided
```
neacsu: This could be doc comments

common/mixnode-common/src/verloc/mod.rs
```diff
@@ -88,54 +88,65 @@ impl ConfigBuilder {
...
+
+    pub fn validator_api_urls(mut self, validator_api_urls: Vec<Url>) -> Self {
+        self.0.validator_api_urls = validator_api_urls;
         self
     }
+
+    // pub fn mixnet_contract_address<S: Into<String>>(mut self, mixnet_contract_address: S) -> Self {
+    //     self.0.mixnet_contract_address = mixnet_contract_address.into();
+    //     self
+    // }
```
neacsu: Why not just remove it?

common/mixnode-common/src/verloc/mod.rs
```diff
@@ -88,54 +88,65 @@ impl ConfigBuilder {
...
-        if self.0.validator_urls.is_empty() {
+        if self.0.validator_api_urls.is_empty() {
             panic!("at least one validator endpoint must be provided")
         }
-        if self.0.mixnet_contract_address.is_empty() {
-            panic!("the mixnet contract address must be set")
-        }
+        // if self.0.mixnet_contract_address.is_empty() {
+        //     panic!("the mixnet contract address must be set")
+        // }
```
neacsu: Same as above

explorer-api/src/main.rs
```diff
@@ -12,8 +12,7 @@ mod mix_nodes;
 mod ping;
 mod state;
 
-const VALIDATOR_API: &str = "http://testnet-milhon-validator1.nymtech.net:8080";
-const CONTRACT: &str = "punk10pyejy66429refv3g35g2t7am0was7yalwrzen";
+const VALIDATOR_API: &str = "https://testnet-milhon-validator1.nymtech.net/api";
```
neacsu: Since this address was added above in `network-defaults`, wouldn't it be better to replace the const in this crate with the getter from there?

validator-api/Cargo.toml
```diff
@@ -27,35 +27,30 @@ reqwest = { version = "0.11", features = ["json"] }
...
-topology = { path = "../common/topology" }
-validator-client = { path = "../common/client-libs/validator-client" }
-version-checker = { path = "../common/version-checker" }
+crypto = { path="../common/crypto" }
+gateway-client = { path="../common/client-libs/gateway-client" }
+mixnet-contract = { path="../common/mixnet-contract" }
+nymsphinx = { path="../common/nymsphinx" }
+topology = { path="../common/topology" }
+validator-client = { path="../common/client-libs/validator-client", features = ["nymd-client"] }
+version-checker = { path="../common/version-checker" }
```
neacsu: Those changes seem unnecessary, especially since two lines bellow, there are still whitespaces around the `=`

validator-api/src/network_monitor/monitor/sender.rs
```diff
@@ -74,6 +81,8 @@ pub(crate) struct PacketSender {
     // behaviour is unlikely.
     active_gateway_clients: HashMap<[u8; PUBLIC_KEY_LENGTH], GatewayClient>,
 
+    // I guess that will be required later on if credentials are got per gateway
+    // aggregated_verification_key: Arc<VerificationKey>,
```
neacsu: I'd say we should add fields when we actually need them, I'm not really seeing the point in adding it now as commented code

clients/native/src/commands/init.rs
jstuczyn: My understanding is that bandwidth credential, as its name suggests, is used to access the mix network. However, during registration we are not sending any traffic through the network. Perhaps we should discuss this more with somebody else, maybe @aniampio ?

clients/native/src/commands/upgrade.rs
jstuczyn: 🤦🏻 because I've accidentally messed that up it seems

common/coconut-interface/src/lib.rs
jstuczyn: Ah. Another thing I forgot about. I usually make those multi-lines todo locally to check things out before making PR. Seems this one slipped out of my mind. Just for the context I wanted to see what's preventing us from using `Vec<Attribute>` directoy. Whether compiler is going to complain or something. I'll get back to you on that.

explorer-api/src/main.rs
jstuczyn: Yep. Absolutely. To be honest I forgot about `explorer-api` completely. I've only done any changes to it at the very end after CI started complaining about it.

validator-api/Cargo.toml
jstuczyn: Hm, seems my formatter dislikes whitespaces, but at the very least we do need that additional feature for `validator-client`

durch (approved):

clients/socks5/src/commands/upgrade.rs
```diff
@@ -3,16 +3,21 @@
...
+#[allow(dead_code)]
+fn fail_upgrade<D1: Display, D2: Display>(from_version: D1, to_version: D2) -> ! {
+    print_failed_upgrade(from_version, to_version);
+    process::exit(1)
+}
 
 fn print_start_upgrade<D1: Display, D2: Display>(from: D1, to: D2) {
     println!(
-        "\n==================\nTrying to upgrade client from {} to {} ...",
+        "\n==================\nTrying to client gateway from {} to {} ...",
```
neacsu: Still not good, but third time's a charm I guess :D

neacsu (approved):

neacsu:
Feel free to fix the `upgrade` message and merge the PR.

clients/socks5/src/commands/upgrade.rs
jstuczyn: How the hell did I manage to do that ... : D

PR #738, opened 2021-08-13
Add stake reward rates and bump version of client

neacsu:
This is needed as a separate PR so that we can publish the 0.17 version before changing the dependency version in the wallet


jstuczyn:
~~But the changes have not yet been made in the contract?~~ - this was referring to the early version of the PR that only modified the typescript code.

common/mixnet-contract/src/types.rs
```diff
@@ -34,6 +34,8 @@ pub struct StateParams {
     pub minimum_gateway_bond: Uint128, // minimum amount a gateway must bond to get into the system
     pub mixnode_bond_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
     pub gateway_bond_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
+    pub mixnode_stake_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
```
jstuczyn: I'd call it delegation

jstuczyn (commented):
Unless you want to do it in a separate PR, the delegation reward rates are actually not being used

common/mixnet-contract/src/types.rs
neacsu: Done

jstuczyn (approved):

PR #740, opened 2021-08-16
Feature/copy to clipboard

fmtabbara:


fmtabbara:
Config file removed.

Reference - https://nextjs.org/docs/messages/future-webpack5-moved-to-webpack5

jstuczyn (approved):

PR #741, opened 2021-08-16
Feature/use delegation rates

neacsu:


jstuczyn (commented):
Looks good to me, the only remaining thing would be writing some simple migration code to handle changes to `State` struct. For that you have two options:

a) via serde defaults
meaning you'd have to slap `#[serde(default = "your_default_fn")]` on `State` and `StateParams`, i.e.

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
pub struct StateParams {
    pub epoch_length: u32, // length of an epoch, expressed in hours

    pub minimum_mixnode_bond: Uint128, // minimum amount a mixnode must bond to get into the system
    pub minimum_gateway_bond: Uint128, // minimum amount a gateway must bond to get into the system
    pub mixnode_bond_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
    pub gateway_bond_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
    #[serde(default = "your_default_fn")]
    pub mixnode_delegation_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
    #[serde(default = "your_default_fn")]
    pub gateway_delegation_reward_rate: Decimal, // annual reward rate, expressed as a decimal like 1.25
    pub mixnode_active_set_size: u32,
}

//

#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
pub struct State {
    pub owner: Addr, // only the owner account can update state
    pub network_monitor_address: Addr,
    pub params: StateParams,

    // helper values to avoid having to recalculate them on every single payment operation
    pub mixnode_epoch_bond_reward: Decimal, // reward per epoch expressed as a decimal like 0.05
    pub gateway_epoch_bond_reward: Decimal, // reward per epoch expressed as a decimal like 0.05
    #[serde(default = "your_default_fn")]
    pub mixnode_epoch_delegation_reward: Decimal, // reward per epoch expressed as a decimal like 0.05
    #[serde(default = "your_default_fn")]
    pub gateway_epoch_delegation_reward: Decimal, // reward per epoch expressed as a decimal like 0.05
}

```

and then resaving it in the migration (so literally just load the state and save it again). Then after migration itself is finished, another PR should be made to remove those serde macros.

b) via intermediate structs

basically inside `migrate()` function define `struct OldState` and `struct OldStateParams`, deserialize to those, create the "proper" structs and save that data.

But whatever you do, please do the sanity check first, i.e. initialise contract with the old version on your localnet (before you made any changes), set some *non-default* parameters, migrate contract and try to do something to it.

neacsu:
Added the migration code and tested it locally. Thank you for the suggestions @jstuczyn . Will remove the last commit in a separate PR

jstuczyn (approved):
If you're saying the migration works locally for you, that's good enough for me

PR #742, opened 2021-08-17
explorer-api: add pending field to port check response

mmsinclair:
Adds a `pending` boolean flag to the response, that is true while the port check is being done to stop multiple port checks from being spawned.

This is the first step in improving this code - the next step will be to add a worker task with a queue that does the port checks, because this will allow rate limiting and controlling the number of active port check tasks.


jstuczyn (approved):
Looks good to me

PR #745, opened 2021-08-18
Feature/cred after handshake

neacsu:


common/client-libs/gateway-client/src/client.rs
```diff
@@ -379,21 +379,26 @@ impl GatewayClient {
...
             .map_err(GatewayClientError::RegistrationFailure),
             _ => unreachable!(),
         }?;
-
-        self.authenticated = true;
-        Ok(shared_key)
+        self.authenticated = match self.read_control_response().await? {
+            ServerResponse::Register { status } => Ok(status),
+            ServerResponse::Error { message } => Err(GatewayClientError::GatewayError(message)),
+            _ => unreachable!(),
```
durch: Maybe for clarity error out with what came out of the `read_control_response`, since we're returning an enum at some point in the future someone might extend the function to also return some other enum member, and at that time we blow up

common/client-libs/gateway-client/src/client.rs
```diff
@@ -438,8 +442,26 @@ impl GatewayClient {
...
+            .into();
+            self.authenticated = match self.send_websocket_message(msg).await? {
+                ServerResponse::Bandwidth { status } => Ok(status),
+                ServerResponse::Register { status: _ } => Err(GatewayClientError::NotAuthenticated),
+                ServerResponse::Error { message } => {
+                    self.authenticated = false;
+                    self.shared_key = None;
+                    Err(GatewayClientError::GatewayError(message))
+                }
+                _ => unreachable!(),
```
durch: Same as the above comment

durch (changes requested):
Looks good overall, I two small nits around `unreachable`.

common/client-libs/gateway-client/src/client.rs
neacsu: I used `unreachable!` as it is already part of the flow on the authentication path, but we can switch to error returns as well. The end result should be the same (the client blows up) because of some protocol mismatch.

common/client-libs/gateway-client/src/client.rs
```diff
@@ -438,8 +442,25 @@ impl GatewayClient {
...
+        }
+        // Authenticate with the bandwidth credential as well
+        if self.authenticated {
+            // Shared key should be set here, or else we wouldn't be authenticated
+            let msg = ClientControlRequest::new_enc_bandwidth_credential(
+                self.coconut_credential.clone(),
+                self.shared_key.as_ref().unwrap(),
+            )
+            .into();
+            self.authenticated = match self.send_websocket_message(msg).await? {
```
jstuczyn: Perhaps this should be extracted into separate function? On first glance it's not obvious why `self.authenticated` should be changed upon `self.send_websocket_message(msg)`

Also, should this really be part of authentication? Authentication is proving that you who you claim to be, i.e. you're the client who has established the shared key with the gateway. Perhaps there should be a separate flag/field for bandwidth credential? Perhaps something like "allowed to send traffic" or "traffic available" ?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -379,21 +379,26 @@ impl GatewayClient {
...
-                self.coconut_credential.clone(),
             )
             .await
             .map_err(GatewayClientError::RegistrationFailure),
             _ => unreachable!(),
         }?;
-
-        self.authenticated = true;
-        Ok(shared_key)
+        self.authenticated = match self.read_control_response().await? {
```
jstuczyn, line 391: I'm a bit confused.  Why should we get a random socket message after registration that should modify our auth state?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -379,21 +379,26 @@ impl GatewayClient {
...
+            ServerResponse::Register { status } => Ok(status),
+            ServerResponse::Error { message } => Err(GatewayClientError::GatewayError(message)),
+            _ => Err(GatewayClientError::NotAuthenticated),
+        }?;
+        if self.authenticated {
+            self.shared_key = Some(Arc::new(shared_key));
```
jstuczyn, line 397: Why are we assigning shared key conditionally? If we managed to derive it, we should be good to go, no?

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -50,11 +43,6 @@ impl InitMessage {
         encryption::PublicKey::from_bytes(&self.ephemeral_key).unwrap()
     }
 
-    #[cfg(not(target_arch = "wasm32"))]
-    pub fn credential(&self) -> Credential {
-        self.credential.clone().unwrap()
-    }
-
     fn to_bytes(&self) -> Vec<u8> {
         bincode::serialize(self).unwrap()
```
jstuczyn, line 47: I'm still not a fan of bincode here, but /shrug

gateway/gateway-requests/src/types.rs
```diff
@@ -126,6 +132,21 @@ impl ClientControlRequest {
             iv: iv.to_base58_string(),
         }
     }
+
+    pub fn new_enc_bandwidth_credential(credential: Credential, shared_key: &SharedKeys) -> Self {
+        ClientControlRequest::BandwidthCredential {
+            enc_credential: shared_key
+                .encrypt_and_tag(&bincode::serialize(&credential).unwrap(), None),
```
jstuczyn: I'm not 100% sure if a zero IV is fine here because if you send the same credential again, the ciphertext will be identical. We should ask @aniampio about her opinion. But my gut feeling is as follows: we should either 
a) always explicitly randomise the credential before sending it to ensure the bytes representation would be different each time even for the same underlying cryptographic claims
b) use non-zero IV

gateway/gateway-requests/src/registration/handshake/shared_key.rs
```diff
@@ -102,6 +103,36 @@ impl SharedKeys {
             .collect()
     }
 
+    pub fn decrypt_tagged(&self, mut enc_data: Vec<u8>) -> Result<Vec<u8>, GatewayRequestsError> {
```
jstuczyn: I don't think it should be taking ownership (and mutating) incoming data, I'd say `pub fn decrypt_tagged(&self, enc_data: &[u8]) -> Result<Vec<u8>, GatewayRequestsError>` type signature would be more appropriate

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -378,12 +379,29 @@ where
...
+                    *final_receiver = Some(mix_receiver);
                     self.handle_authenticate(address, enc_address, iv, mix_sender)
                         .await
                 }
                 ClientControlRequest::RegisterHandshakeInitRequest { data } => {
+                    *final_receiver = Some(mix_receiver);
                     self.handle_register(data, mix_sender).await
                 }
+                ClientControlRequest::BandwidthCredential { enc_credential } => {
+                    let credential = match ClientControlRequest::try_from_enc_bandwidth_credential(
```
jstuczyn: This looks like something that would like to live in `self.handle_bandwidth()`

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -378,12 +379,29 @@ where
...
+                        enc_credential,
+                        self.shared_key.as_ref().unwrap(),
+                    ) {
+                        Ok(c) => c,
```
jstuczyn: You can just use `map_err`

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -378,12 +379,29 @@ where
...
                         .await
                 }
                 ClientControlRequest::RegisterHandshakeInitRequest { data } => {
+                    *final_receiver = Some(mix_receiver);
                     self.handle_register(data, mix_sender).await
                 }
+                ClientControlRequest::BandwidthCredential { enc_credential } => {
+                    let credential = match ClientControlRequest::try_from_enc_bandwidth_credential(
+                        enc_credential,
+                        self.shared_key.as_ref().unwrap(),
```
jstuczyn: With this unwrap malicious client can trigger a gateway crash. He will just send `BandwidthCredential` request as the very first message exchanged before shared key is established,

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -366,6 +365,8 @@ where
     async fn handle_initial_authentication_request(
         &mut self,
         mix_sender: MixMessageSender,
+        mix_receiver: MixMessageReceiver,
```
jstuczyn: I'm not entirely sure what's the purpose of those arguments. It looks like you're assigning to `final_receiver` the value of `mix_receiver` you just passed in here?

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -378,12 +379,29 @@ where
...
+                        self.shared_key.as_ref().unwrap(),
+                    ) {
+                        Ok(c) => c,
+                        Err(e) => {
+                            return ServerResponse::Error {
+                                message: e.to_string(),
+                            }
+                        }
+                    };
+                    let status = credential.verify(&self.aggregated_verification_key).await;
```
jstuczyn: This function doesn't need to be async anymore

common/client-libs/gateway-client/src/client.rs
```diff
@@ -438,8 +442,25 @@ impl GatewayClient {
         if self.shared_key.is_some() {
```
jstuczyn, line 446: I don't quite follow, why is `perform_initial_authentication` returning an Arc now?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -379,21 +379,26 @@ impl GatewayClient {
...
+            _ => Err(GatewayClientError::NotAuthenticated),
+        }?;
+        if self.authenticated {
+            self.shared_key = Some(Arc::new(shared_key));
+        }
+        Ok(())
     }
 
-    pub async fn authenticate(
+    async fn authenticate(
```
jstuczyn, line 402: If authenticate is no longer returning authentication status then... what's the point of it?

gateway/gateway-requests/src/types.rs
```diff
@@ -111,7 +112,12 @@ pub enum ClientControlRequest {
         iv: String,
     },
     #[serde(alias = "handshakePayload")]
-    RegisterHandshakeInitRequest { data: Vec<u8> },
+    RegisterHandshakeInitRequest {
```
jstuczyn: Have you checked whether the wasm client still works after this renaming?

jstuczyn (changes requested):
It kinda looks like there should be another "step" in the client-gateway communication story. So before any coconut stuff we had 3 "states":
- unconnected - the client hasn't established any TCP connection to the gateway. Not much to add here
- un-authenticated - client is unable to send any traffic to the gateway, no shared key has been established or ownership of the key has not been proven
- authenticated - client has a proved shared key with the gateway and can send as much mix traffic as they want

It looks like 4th state relating to bandwidth credential should be introduced: 
- unconnected - as before
- un-authenticated - as before
- authenticated - client has a proved shared key with the gateway and secure channel is established. However, client is not yet allowed to send mix traffic
- "within-bandwidth-quota" (pending name) - the client has paid for some traffic and can send at most the specified amount of mix traffic
in the future once bandwidth credential have proper use I guess it will be possibly to transition from the "within-bandwidth-quota" state back to authenticated if the client runs out of bandwidth.

The bandwidth credential should not be part of authentication as it doesn't prove anything about you, it should only increase your traffic quota. So after authentication (assuming you haven't used that gateway before), you'd be at 0. So basically if you attempted to send any mix packets and gateway detected you're at 0 bandwidth left, it wouldn't forward anything for you.

Hm. Perhaps there could be another field on the `Handle` with `AtomicU64` indicating available bandwidth that's obtained from the ledger upon authentication to know how much traffic the client can send and is updated upon each received credential. But then when we introduce metering we'd need to somehow periodically and reliably flush those values back to the storage as the traffic is used... Or I guess it should be done later because the more I think about it, the more complexity it introduces.

common/client-libs/gateway-client/src/client.rs
neacsu: This is not so much random as is the protocol with all the other types of messages: you send a request and expect a response with the `status`. This happens for `register` as well, but up until now we haven't read that message from the connection and it just remained there until the connection eventually finished.

common/client-libs/gateway-client/src/client.rs
neacsu: Because the returned `status` of the `ServerResponse::Register` might be `false` for whatever reason, in which case the state is not authenticated and the shared key is not valid. This clearly doesn't happen right now, as we do all of the needed checks in the handshake communication, but in terms of the returned `ServerResponse::Register { status }`, it might be possible to fail _after_ all the handshake work.

common/client-libs/gateway-client/src/client.rs
neacsu: This is just a change to have `register` and `authenticate` functions more homogeneous. This makes them both do the changes on the mutable `self`, rather than one doing some changes in the function, and the other in the function caller, based on the returned value.

common/client-libs/gateway-client/src/client.rs
neacsu: The `register_with_gateway` functions of the native and socks5 clients called directly on the `register` function. With this change, they'll go through `perform_initial_authentication`, which in turn will call `register` for them + do the credential stuff too (which is anyway needed). The Arc was created anyway in the native and socks5 clients, only now it's all in one place.
So, to answer your question, the Arc was used anyway, but I did a bit of refactoring to have the creation of that Arc + the code for the coconut credential all in just one place.

common/client-libs/gateway-client/src/client.rs
neacsu: Yes, I've kind of merged the authenticated state with this new, credential enabled state. I'll have to figure out on how to separate them better, probably using a different field for that.

gateway/gateway-requests/src/registration/handshake/shared_key.rs
neacsu: Done

gateway/gateway-requests/src/types.rs
neacsu: No renaming here, just fmt I think, but I checked the wasm client anyway. It still works

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: This is sort of a hack, because the credential comes as part of the authenticate/register process, immediately after one of those messages, in the same loop that calls `handle_initial_authentication_request`. Because of that, two `mpsc` channels are created, instead of one, and only the first one is needed. Introducing a separated state for the credential message should remove this anyway.

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: I've added a check

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: `map_err` would work if the `handle_initial_authentication_request` function would return a `Result` type. Right now is just a plain enum, I don't really know if it's worth it to refactor that right now

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: Correct, I haven't seen that. I'll remove the async

gateway/gateway-requests/src/types.rs
neacsu: Switched to using a fresh IV each time a BandwidthCredential message is created.

gateway/gateway-requests/src/types.rs
```diff
@@ -111,7 +112,13 @@ pub enum ClientControlRequest {
...
+        data: Vec<u8>,
+    },
+    BandwidthCredential {
+        enc_credential: Vec<u8>,
+        iv: String,
```
jstuczyn: Why is the IV a String here while the credential is set properly as bytes? Why not send IV as `Vec<u8>`?
From what I recall, but I might be completely wrong because it's been literally over a year since I wrote it, those types only exist for wasm compatibility.

gateway/gateway-requests/src/types.rs
```diff
@@ -126,6 +133,29 @@ impl ClientControlRequest {
             iv: iv.to_base58_string(),
         }
     }
+
+    pub fn new_enc_bandwidth_credential(
+        credential: Credential,
+        shared_key: &SharedKeys,
+        iv: AuthenticationIV,
```
jstuczyn: We should probably rename some things, because this is not an `AuthenticationIV`

gateway/gateway-requests/src/types.rs
```diff
@@ -126,6 +133,29 @@ impl ClientControlRequest {
...
+    ) -> Self {
+        let enc_credential =
+            shared_key.encrypt_and_tag(&bincode::serialize(&credential).unwrap(), Some(iv.inner()));
```
jstuczyn: Are we sure unwrap is fine here?

gateway/gateway-requests/src/types.rs
```diff
@@ -158,6 +188,7 @@ impl TryInto<String> for ClientControlRequest {
 pub enum ServerResponse {
     Authenticate { status: bool },
     Register { status: bool },
+    Bandwidth { status: bool },
```
jstuczyn, line 195: Something for the future, as it's out of scope for this PR, maybe it could return the amount remaining?

gateway/gateway-requests/src/types.rs
```diff
@@ -210,35 +241,10 @@ pub enum BinaryRequest {
...
-        // together with a mutable one
-        let mut message_bytes_mut = &mut raw_req[mac_size..];
-
-        let zero_iv = stream_cipher::zero_iv::<GatewayEncryptionAlgorithm>();
-        stream_cipher::decrypt_in_place::<GatewayEncryptionAlgorithm>(
-            shared_keys.encryption_key(),
-            &zero_iv,
-            &mut message_bytes_mut,
-        );
+        let message_bytes_mut = &shared_keys.decrypt_tagged(&raw_req, None)?;
```
jstuczyn: Well, it doesn't look to be `mut`, does it?

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -342,12 +347,45 @@ where
         }
     }
 
-    // currently there are no valid control messages you can send after authentication
-    async fn handle_text(&mut self, _: String) -> Message {
-        trace!("Handling text message (presumably control message)");
+    async fn handle_bandwidth(&mut self, enc_credential: Vec<u8>, iv: String) -> ServerResponse {
```
jstuczyn: Just making a note for the future here.
While this is out of scope for this PR, I could only assume that in the long run some of the work would have to be offloaded to the `ClientsHandler` (I'm not a huge fan of that guy, but at the time of writing it, I couldn't come up with a better solution to solve the problem) that owns the `ClientLedger` so that it would check the amount on the credential remaining etc.

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -342,12 +347,45 @@ where
...
+            }
+        };
+        let credential = match ClientControlRequest::try_from_enc_bandwidth_credential(
+            enc_credential,
+            self.shared_key.as_ref().unwrap(),
+            iv,
+        ) {
+            Ok(c) => c,
+            Err(e) => {
+                return ServerResponse::new_error(e.to_string());
```
jstuczyn, line 368: It could potentially use some `map_err` action.

gateway/gateway-requests/src/types.rs
```diff
@@ -100,7 +101,7 @@ impl From<MixPacketFormattingError> for GatewayRequestsError {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize)]
```
jstuczyn: Just for curiosity sake, why did you remove Debug?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -449,13 +456,41 @@ impl GatewayClient {
         }
     }
 
+    pub async fn claim_bandwidth(&mut self) -> Result<(), GatewayClientError> {
```
jstuczyn, line 459: Just a super minor thing, but perhaps this method should be taking the bandwidth credential as an argument instead taking it from `self`? What do you think?

jstuczyn: I think it'd be more future proof so that you wouldn't need to create new clients when you run out bandwidth. Instead you'd just call this guy with fresh credential and keep on going.

jstuczyn (commented):
In general looks good to me, only a couple of small things I would have maybe changed. Let me know if you want to discuss that.

gateway/gateway-requests/src/types.rs
neacsu: I've set it to String just because that's how it is in the `Authenticate` request, but it is perfectly reasonable to pass it as a `Vec<u8>` as well.

gateway/gateway-requests/src/types.rs
neacsu: I've renamed it to simply `IV`

gateway/gateway-requests/src/types.rs
neacsu: I've handled the possible error now, just in case. Maybe it was possible to make it panic on the `public_attributes` content, I'm not sure.

gateway/gateway-requests/src/types.rs
neacsu: I'll add a comment on that

gateway/gateway-requests/src/types.rs
neacsu: Indeed it isn't :D . Renamed it

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: I agree, especially since this will need to be handled multiple times during a connection.

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: Again, that would be the case, should we refactor these into actual results rather than a `ServerResponse` enum that contains an `Error` variant.

gateway/gateway-requests/src/types.rs
neacsu: There was a reason behind it, but I don't remember the exact error I received and it doesn't reproduce anymore, so I'll put it back now.

common/client-libs/gateway-client/src/client.rs
neacsu: Not sure what you mean here. An existing client is needed anyway, for its underlaying connection to the gateway, among other things. The fact that the credential shouldn't necessarily be part of gateway client is indeed an improvement we'll need to do in the future, and that will force this function to take the credential as argument, as it won't be part of `self` anymore. But for now, I think it can live like this just as easily. (calling `self.claim_bandwidth(self.credential)` would look kind of strange tbh :D )

gateway/gateway-requests/src/authentication/iv.rs
```diff
@@ -2,15 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crypto::generic_array::{typenum::Unsigned, GenericArray};
-use crypto::symmetric::stream_cipher::{random_iv, NewStreamCipher, IV};
+use crypto::symmetric::stream_cipher::{random_iv, NewStreamCipher, IV as OtherIV};
```
jstuczyn: `OtherIV` is not a very descriptive name

gateway/gateway-requests/src/authentication/iv.rs
```diff
@@ -2,15 +2,15 @@
...
 use nymsphinx::params::GatewayEncryptionAlgorithm;
 use rand::{CryptoRng, RngCore};
 
 type NonceSize = <GatewayEncryptionAlgorithm as NewStreamCipher>::NonceSize;
 
 // I think 'IV' looks better than 'Iv', feel free to change that.
 #[allow(clippy::upper_case_acronyms)]
-pub struct AuthenticationIV(IV<GatewayEncryptionAlgorithm>);
+pub struct IV(OtherIV<GatewayEncryptionAlgorithm>);
```
jstuczyn: Hm, perhaps this entire file should be moved out of `authentication` module? Since we're using this `iv` for both, authentication and bandwidth. Maybe it should just live on the same level as `authentication`?

gateway/gateway-requests/src/authentication/iv.rs
neacsu: I've renamed it to `CryptoIV`, as is part of the `crypto` crate. I'm not feeling very inspired about this, so any suggestions are welcomed :D

gateway/gateway-requests/src/authentication/iv.rs
neacsu: I've thought of this as well, but figured the even bandwidth credential is still an authentication of some sort. But as we have a differentiation in our codebase for these names, I can just put it in the root of `gateway-requests`.

jstuczyn (approved):

PR #747, opened 2021-08-19
Feature/more reliable uptime calculation

jstuczyn:
This pull request changes how node uptimes are calculated. Up until this point the uptime was based on the percentage of "up" test runs during which the nodes was online. So for example if a mixnode was online during 100 network monitor test runs and it passed 90 of them it'd have 90% uptime. 

However, this approach has an issue. The calculation didn't care about number of actual runs that have happened. So say somebody was offline for 99% of given day and happened to be up for a single test run. They'd get assigned 100% uptime because they happened to pass that one test run. 

This pull request changes this and takes into consideration the total number of test runs. So in the previous case the node would have gotten score of 1% as there would be no information about them passing the runs during which they were offline.

### DIFF
--- a/validator-api/migrations/20210819120000_monitor_runs.sql
+++ b/validator-api/migrations/20210819120000_monitor_runs.sql
@@ -1,0 +1,7 @@
+-- keeping track of all monitor runs that have happened will help to
+-- solve an issue of mixnode being online only for a single check and yet being assigned 100% uptime
+CREATE TABLE monitor_run
+(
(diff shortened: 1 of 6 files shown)


jstuczyn:
I just had a sudden realisation that I didn't cover the case of run count being 0. Let me quickly fix it before restoring the PR.

neacsu (approved):

durch (approved):

PR #748, opened 2021-08-23
Feature/get own delegations

neacsu:
Create two new buckets with reverse delegation data: for each delegation owner address, store the list of delegated nodes (one bucket for mixnodes, one bucket for gateways)

The reason for storing only the node address and not other additional data (like the amount of stake, or other values that might be added in the future) is so that we don't store the same data in multiple places, as that would cause confusion on which data is the source of truth and also possible difficulties in keeping those two places in sync.

With that in ming, in order for one to obtain the full data (including the amount), one would have to query the list of nodes that a certain address delegated towards (query address1 -> [node1, node2, ...]) and then query independently for each node to obtain the amount that was delegated (and possibly other data in the future as well) (query (address1, node1) -> amount1; query (address1, node2) -> amount2 ...)

Although this incurs additional queries being performed, we can safely assume that an address is less likely to have a big number of nodes delegating towards (<10), so the additional number of queries should be small enough. The same cannot be said about the nodes, which could have delegations towards them in the order of hundreds or thousands.

<s> TODO list until finishing draft: </s>

- [x]  Modify existing contract tests and possibly add some new ones for these two buckets

- [x] Add contract migration code and perform a test: bind and delegate a node, check that after migration the node is present in the reversed bucket

### DIFF
--- a/contracts/mixnet/src/contract.rs
+++ b/contracts/mixnet/src/contract.rs
@@ -4,7 +4,7 @@
 use crate::helpers::calculate_epoch_reward_rate;
 use crate::state::State;
 use crate::storage::{
-    config, gateway_reverse_delegations, layer_distribution, mix_reverse_delegations,
+    config, reverse_gateway_delegations, layer_distribution, reverse_mix_delegations,
 };
 use crate::{error::ContractError, queries, transactions};
 use config::defaults::NETWORK_MONITOR_ADDRESS;
@@ -154,11 +154,11 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
             start_after,
             limit,
         )?),
-        QueryMsg::GetMixReverseDelegations {
+        QueryMsg::GetReverseMixDelegations {
             delegation_owner,
             start_after,
             limit,
-        } => to_binary(&queries::query_mixnode_reverse_delegations_paged(
+        } => to_binary(&queries::query_reverse_mixnode_delegations_paged(
             deps,
             delegation_owner,
             start_after,
@@ -182,11 +182,11 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
             start_after,
             limit,
         )?),
-        QueryMsg::GetGatewayReverseDelegations {
+        QueryMsg::GetReverseGatewayDelegations {
             delegation_owner,
             start_after,
             limit,
-        } => to_binary(&queries::query_gateway_reverse_delegations_paged(
+        } => to_binary(&queries::query_reverse_gateway_delegations_paged(
             deps,
             delegation_owner,
             start_after,
@@ -305,7 +305,7 @@ pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, C
     for mix_identity in mixnodes_identities {
         let delegations = get_all_mixnode_delegations(&deps, &mix_identity)?;
         for delegation in delegations {
-            mix_reverse_delegations(deps.storage, &delegation.owner())
+            reverse_mix_delegations(deps.storage, &delegation.owner())
                 .save(mix_identity.as_bytes(), &())?;
         }
     }
@@ -314,7 +314,7 @@ pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, C
     for gateway_identity in gateways_identities {
         let delegations = get_all_gateway_delegations(&deps, &gateway_identity)?;
         for delegation in delegations {
-            gateway_reverse_delegations(deps.storage, &delegation.owner())
+            reverse_gateway_delegations(deps.storage, &delegation.owner())
                 .save(gateway_identity.as_bytes(), &())?;
         }
     }
--- a/contracts/mixnet/src/contract.rs
+++ b/contracts/mixnet/src/contract.rs
@@ -205,100 +205,102 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
     Ok(query_res?)
 }
 
-fn get_all_mixnodes_identities(deps: &DepsMut) -> Result<Vec<IdentityKey>, ContractError> {
-    let mut mixnode_bonds = Vec::new();
-    let mut start_after = None;
-    loop {
-        let mut paged_response = queries::query_mixnodes_paged(deps.as_ref(), start_after, None)?;
-        mixnode_bonds.append(&mut paged_response.nodes);
-
-        if let Some(start_after_res) = paged_response.start_next_after {
-            start_after = Some(start_after_res)
-        } else {
-            break;
-        }
-    }
-    let mixnodes = mixnode_bonds
-        .into_iter()
-        .map(|bond| bond.mix_node.identity_key)
-        .collect();
-
-    Ok(mixnodes)
-}
-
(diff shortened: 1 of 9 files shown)


contracts/mixnet/src/queries.rs
```diff
@@ -139,6 +141,32 @@ pub(crate) fn query_mixnode_delegations_paged(
...
+) -> StdResult<PagedMixReverseDelegationsResponse> {
+    let limit = limit
+        .unwrap_or(DELEGATION_PAGE_DEFAULT_LIMIT)
+        .min(DELEGATION_PAGE_MAX_LIMIT) as usize;
+    let start = calculate_start_value(start_after);
+
+    let delegations = mix_reverse_delegations_read(deps.storage, &delegation_owner)
+        .range(start.as_deref(), None, Order::Ascending)
+        .take(limit)
+        .map(|res| res.map(|entry| String::from_utf8(entry.0).unwrap()))
```
jstuczyn: Sanity check: are we sure the unwrap is fine here? And if you're certain it's ok, could you write a comment explaining why is that? We really don't want our smart contract to randomly fail.

contracts/mixnet/src/queries.rs
```diff
@@ -190,6 +218,32 @@ pub(crate) fn query_gateway_delegations_paged(
...
+) -> StdResult<PagedGatewayReverseDelegationsResponse> {
+    let limit = limit
+        .unwrap_or(DELEGATION_PAGE_DEFAULT_LIMIT)
+        .min(DELEGATION_PAGE_MAX_LIMIT) as usize;
+    let start = calculate_start_value(start_after);
+
+    let delegations = gateway_reverse_delegations_read(deps.storage, &delegation_owner)
+        .range(start.as_deref(), None, Order::Ascending)
+        .take(limit)
+        .map(|res| res.map(|entry| String::from_utf8(entry.0).unwrap()))
```
jstuczyn: Same unwrap concern

contracts/mixnet/src/queries.rs
```diff
@@ -786,6 +840,248 @@ mod tests {
...
+        #[test]
+        fn mix_reverse_deletion_query_returns_error_if_delegation_doesnt_exist() {
+            let mut deps = helpers::init_contract();
+
+            let node_identity1: IdentityKey = "foo1".into();
+            let node_identity2: IdentityKey = "foo2".into();
+            let delegation_owner1 = Addr::unchecked("bar");
+            let delegation_owner2 = Addr::unchecked("bar2");
+
+            assert_eq!(
```
jstuczyn: This feels weird, you're asserting correct error type of a helper function you wrote for tests that only returns one error. Are we sure you're testing the correct thing here?

contracts/mixnet/src/queries.rs
```diff
@@ -786,6 +840,248 @@ mod tests {
...
+        use crate::storage::mix_reverse_delegations;
+
+        fn query_mixnode_reverse_delegation(
+            deps: Deps,
+            mix_identity: IdentityKey,
+            address: Addr,
+        ) -> Result<(), ContractError> {
+            mix_reverse_delegations_read(deps.storage, &address)
+                .may_load(mix_identity.as_bytes())?
+                .ok_or(ContractError::NoMixnodeDelegationFound {
```
jstuczyn: Is `NoMixnodeDelegationFound` the correct error type for what you're doing here?

contracts/mixnet/src/queries.rs
```diff
@@ -786,6 +840,248 @@ mod tests {
...
+                Option::from(per_page),
+            )
+            .unwrap();
+
+            // now we have 2 pages, with 2 results on the second page
+            assert_eq!(2, page2.delegated_nodes.len());
+        }
+
+        #[test]
+        fn mix_reverse_deletion_query_returns_current_delegation_value() {
```
jstuczyn: Does it? : ) 
You check for `Ok(())` also it seems you're checking output of your auxiliary function?

contracts/mixnet/src/queries.rs
```diff
@@ -786,6 +840,248 @@ mod tests {
...
+                .unwrap();
+
+            assert_eq!(
+                Ok(()),
+                query_mixnode_reverse_delegation(deps.as_ref(), node_identity, delegation_owner)
+            )
+        }
+
+        #[test]
+        fn mix_reverse_deletion_query_returns_error_if_delegation_doesnt_exist() {
```
jstuczyn: Again, you seem to be testing your auxiliary function rather than the actual query

contracts/mixnet/src/queries.rs
```diff
@@ -1010,4 +1306,246 @@ mod tests {
             query_gateway_delegation(deps.as_ref(), node_identity1, delegation_owner1)
         )
     }
+
+    #[cfg(test)]
```
jstuczyn, line 1243: I can only assume gateway tests follow the same pattern as mixnode tests

contracts/mixnet/src/contract.rs
```diff
@@ -180,8 +205,118 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
     Ok(query_res?)
 }
 
+fn get_all_mixnodes_identities(deps: &DepsMut) -> Result<Vec<IdentityKey>, ContractError> {
```
jstuczyn: I think those functions are only used in migration (?). Could we be more explicit about it? For example we could put it directly inside the migration, i.e.
```rust
#[entry_point]
pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
   fn get_all_mixnodes_identities(deps: &DepsMut) -> Result<Vec<IdentityKey>, ContractError> {
   ...
   }

...
}

jstuczyn (commented):
There's a couple of minor things that are unclear to me, but in general it looks good. Just one more additional comment:. It feels like it sounds better if you say "ReverseMixDelegation" rather than "MixReverseDelegation", but that's just a personal opinion. If you prefer the former, that's fine by me.

contracts/mixnet/src/queries.rs
neacsu: This is present in the current deployed contract as well, and it _should_ be fine, as these values are the ones we read from the storage, where we only write them from sender addresses. If the conversion fails, then we should definitely panic, because that means that the storage got corrupted. But I will add the panic on an `expect`, to have more verbosity.

contracts/mixnet/src/queries.rs
neacsu: You're right, it is pretty weird now that I think of it. The actual important thing here is not the error itself, but the fact that it is an error, and that delegation doesn't exist. Having said that, the correct thing is tested, but in an indirect way, I'll fix it in the next commits.

contracts/mixnet/src/queries.rs
neacsu: I've moved the code to `storage.rs` and removed the unnecessary auxiliary functions. This should also address a lot of the comments in this review, so I'll also resolve those.

contracts/mixnet/src/contract.rs
neacsu: Sure, I'll reduce their scope.

neacsu:
I don't have a strong preference for the naming tbh. I renamed every appearance of [node]reverse to reverse[node], where [node] is either mix, mixnode or gateway

durch (approved):
Nice big one :). Logic looks code, so does the code,we'll see if this bites us eventually

jstuczyn (approved):

PR #750, opened 2021-08-24
Feature/rust rewarding

jstuczyn:
Alright, the very basic version of rewards being distributed by validator API is there. There is, however, a couple of things I need to finalize before changing this PR from draft into a proper PR:

- [x] Make sure that the recent merge with develop didn't mess things up
- [x] Try it on a local network
- [x] Ideally try it on a QA network (@mfahampshire would that be feasible?)
- [ ] Discuss with @futurechimp whether the feature set is appropriate or more things are required. In particular how errors are being recorded.

Note: for this PR, apart from the usual, I'd really want @futurechimp to look at it, in particular because of certain decisions I've taken involving rewarding directly.


clients/native/src/commands/upgrade.rs
```diff
@@ -17,7 +17,7 @@ fn fail_upgrade<D1: Display, D2: Display>(from_version: D1, to_version: D2) -> !
 
 fn print_start_upgrade<D1: Display, D2: Display>(from: D1, to: D2) {
     println!(
-        "\n==================\nTrying to upgrade client from {} to {} ...",
+        "\n==================\nTrying to upgrade gateway from {} to {} ...",
```
neacsu: Again, those changes :) Not sure why they appear, but I don't think it's what you meant.

clients/socks5/src/commands/upgrade.rs
```diff
@@ -17,7 +17,7 @@ fn fail_upgrade<D1: Display, D2: Display>(from_version: D1, to_version: D2) -> !
 
 fn print_start_upgrade<D1: Display, D2: Display>(from: D1, to: D2) {
     println!(
-        "\n==================\nTrying to upgrade client from {} to {} ...",
+        "\n==================\nTrying to upgrade gateway from {} to {} ...",
```
neacsu: Same

validator-api/src/config/mod.rs
```diff
@@ -224,6 +256,11 @@ impl Config {
         self
     }
 
+    pub fn enabled_rewarding(mut self, enabled: bool) -> Self {
```
neacsu: Maybe
```suggestion
    pub fn with_rewarding_enabled(mut self, enabled: bool) -> Self {
```
to keep it in line with the other functions?

validator-api/src/main.rs
```diff
@@ -107,11 +124,26 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
+            Arg::with_name(KEYPAIR_ARG)
+                .help("Path to the secret key file")
+                .takes_value(true)
+                .long(KEYPAIR_ARG)
+        )
+        .arg(
+            Arg::with_name(FIRST_REWARDING_EPOCH_ARG)
+                .help("Datetime specifying beginning of the first rewarding epoch of this length. It must be a valid rfc3339 datetime.")
+                .takes_value(true)
+                .long(FIRST_REWARDING_EPOCH_ARG)
```
neacsu, line 146: We could also require that rewarding is enabled:
```suggestion
                .takes_value(true)
                .requires(REWARDING_ENABLED)
                .long(FIRST_REWARDING_EPOCH_ARG)
```

validator-api/src/main.rs
```diff
@@ -107,11 +124,26 @@ fn parse_args<'a>() -> ArgMatches<'a> {
...
+        )
+        .arg(
+            Arg::with_name(EPOCH_LENGTH_ARG)
+                .help("Length of the current rewarding epoch in hours")
+                .takes_value(true)
+                .long(EPOCH_LENGTH_ARG)
```
neacsu, line 153: Same as above

validator-api/src/network_monitor/monitor/sender.rs
```diff
@@ -81,6 +81,8 @@ pub(crate) struct PacketSender {
     // behaviour is unlikely.
     active_gateway_clients: HashMap<[u8; PUBLIC_KEY_LENGTH], GatewayClient>,
 
+    // I guess that will be required later on if credentials are got per gateway
+    // aggregated_verification_key: Arc<VerificationKey>,
```
neacsu, line 85: This seems to be part of some other PR

clients/native/src/commands/upgrade.rs
jstuczyn: Nope, I absolutely did not mean that. And my guess for those changes would be the nasty merge conflict I had to deal with.

validator-api/src/config/mod.rs
jstuczyn: Good idea. I will also change `enabled_network_monitor` to follow the same pattern.

validator-api/src/main.rs
jstuczyn: Oh. I didn't know we could do that. Nice!

validator-api/src/network_monitor/monitor/sender.rs
jstuczyn: To be honest, I'd prefer to leave it there as a note for the next person looking at this code as setting up multiple concurrent gateway clients is super tricky and any suggestion there might be useful (unless you really dislike that, then I will remove that commented line : P )

jstuczyn:
~~I found a possible infinite failure loop. Turning the PR into the draft until it's fixed.~~ should be fixed

validator-api/src/network_monitor/monitor/preparer.rs
```diff
@@ -167,10 +168,31 @@ impl PacketPreparer {
...
+        loop {
+            let gateways = self.validator_cache.gateways().await;
+            let mixnodes = self.validator_cache.mixnodes().await;
+            if self
+                .tested_network
+                .is_online(&mixnodes.into_inner(), &gateways.into_inner())
+            {
+                break;
+            } else {
+                info!(
```
neacsu, line 186: 30 seconds seems like quite a long time, maybe this should be a warning instead?

validator-api/src/main.rs
```diff
@@ -232,6 +241,15 @@ fn override_config(mut config: Config, matches: &ArgMatches) -> Config {
         config = config.with_epoch_length(Duration::from_secs(epoch_length * 60 * 60));
     }
 
+    if let Some(monitor_threshold) = matches
+        .value_of(REWARDING_MONITOR_THRESHOLD_ARG)
+        .map(|t| t.parse::<u8>())
+    {
+        let monitor_threshold =
+            monitor_threshold.expect("Provided monitor threshold is not a number!");
```
neacsu, line 257: We could also check the `0-100` value range here

validator-api/src/network_monitor/monitor/preparer.rs
jstuczyn: It's only happening at startup and I'd even argue that perhaps 30s is too short. This is basically to help to solve our chicken and the egg problem of network monitor requiring "good" topology to be online to run and mixnodes/gateways requiring the validator-api to be online in order to start themselves. With this delay we can start the basic validator-api functionality (without monitoring), while we wait for mixnodes/gateways to become online. After that happens the monitor will start

validator-api/src/main.rs
jstuczyn: ~~Why 0-100 range?~~ I looked at the wrong code 🤦🏻

neacsu: That's what I saw being mentioned in the template documentation

PR #755, opened 2021-08-27
Explorer API: port test now split out address resolution and add units tests

mmsinclair:
- splits of address resolution into a function that returns `Option<SocketAddr>` and handles the case of an empty iterator of addresses or an error result
- added unit tests

### DIFF
--- a/explorer-api/src/ping/http.rs
+++ b/explorer-api/src/ping/http.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::net::ToSocketAddrs;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::time::Duration;
 
 use rocket::serde::json::Json;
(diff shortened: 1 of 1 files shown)


explorer-api/Cargo.toml
```diff
@@ -20,6 +20,7 @@ okapi = { version = "0.6.0-alpha-1", features = ["derive_json_schema"] }
 rocket_okapi = "0.7.0-alpha-1"
 log = "0.4.0"
 pretty_env_logger = "0.4.0"
+dns-lookup = "1.0.8"
```
jstuczyn: Any particular reason for using new crate for that rather than using the resolution from the standard library?

explorer-api/src/ping/http.rs
```diff
@@ -81,12 +83,36 @@ async fn port_check(bond: &MixNodeBond) -> HashMap<u16, bool> {
...
 
 fn resolve_host(host: &str, port: u16) -> Option<SocketAddr> {
-    match format!("{}:{}", host, port).to_socket_addrs() {
-        Ok(mut addrs) => addrs.next(),
+    // do a DNS lookup on the host - this could be a hostname, ip address or garbage, so let
+    // the `dns_lookup` crate help us remove bad hostnames
+    match lookup_host(host) {
         Err(e) => {
-            error!("Failed to resolve {}:{} - {}", host, port, e);
+            error!("Failed to parse host {}:{} - {}", host, port, e);
```
jstuczyn: Should those really be reported as `error!`? It's not that we're going to crash (at least now :) ) if somebody provided us with bullshit data, I'd perhaps lower that to `warn!` or maybe even way down to `debug!`, but it's your call.

explorer-api/src/ping/http.rs
```diff
@@ -81,12 +83,36 @@ async fn port_check(bond: &MixNodeBond) -> HashMap<u16, bool> {
...
+            if ip_addrs.is_empty() {
+                error!(
+                    "DNS lookup for host {}:{} didn't yield any IP addresses",
+                    host, port
+                );
+                return None;
+            }
+
+            // the host string should hopefully parse and resolve into a valid socket address
+            let parsed_host = format!("{}:{}", host, port);
```
jstuczyn: You don't have to do intermediate string representation here. `SocketAddr` has a `new` method that takes `IpAddr` and `u16` port as arguments

explorer-api/src/ping/http.rs
```diff
@@ -141,6 +167,13 @@ mod tests {
         assert!(resolve_host("1", 1234).is_some()); // 0.0.0.1
```
jstuczyn: Windows is bad : (

jstuczyn (changes requested):

explorer-api/Cargo.toml
mmsinclair: I was hoping it would give more consistent behaviour cross platform. Sadly, it seems Windows has other ideas. I might revert this and just use https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html like before.

explorer-api/src/ping/http.rs
mmsinclair: Happy to change to a warning - I thought it might be best to get an idea of how much bad data is out there, hence the `error` because if the `host` is wrong, nothing is going to be able to connect to that mix node.

explorer-api/src/ping/http.rs
mmsinclair: I wanted it to use the same `host` as other parts of the system use, rather than using the first ip address that gets resolved by DNS. This is because it is unpredictable what that might be - e.g. it could return an ipv6 address one time and ipv4 the next time. Really I think I should probably test the ports of all resolved ip addresses.

explorer-api/src/ping/http.rs
mmsinclair: Yes it is. Not sure what to do about this 😞 🐼

durch (approved):

explorer-api/src/ping/http.rs
```diff
@@ -80,32 +82,112 @@ async fn port_check(bond: &MixNodeBond) -> HashMap<u16, bool> {
...
+    fn resolve_host_with_valid_hostname_returns_some() {
+        assert!(resolve_host("nymtech.net", 1234).is_some());
+    }
+
+    #[test]
+    fn resolve_host_with_malformed_ip_address_returns_some() {
+        // these have missing values filled in
+        assert!(resolve_host("1.2.3", 1234).is_some()); // 1.2.0.3
+        assert!(resolve_host("1.2", 1234).is_some()); // 1.0.0.2
+        assert!(resolve_host("1", 1234).is_some()); // 0.0.0.1
```
neacsu: Why not `assert_eq!` on the exact `Some` value?

mmsinclair: Because the return type is `Option<SocketAddr>` I'm only interested that it returns a value. These tests are for edge cases such as bad IP address formatting, invalid characters, etc. 

To test equailty on `SocketAddr`, I'd have to be reverse looking up something stable to ensure the test passed - for example `nymtech.net` could move around or resolve to multiple addresses in a random order. And as you can see from the Windows failures, testing the OS with real data is useful.

mmsinclair:
@jstuczyn I've added this as a simple check that trims and checks for empty strings:

https://github.com/nymtech/nym/pull/755/files#diff-f77edc076ae85e26dbf36dc0fc0030e4aaa42a3ebd67f00fb4e9e65d6ec60684R84-R91

explorer-api/Cargo.toml
mmsinclair: I've removed this dependency now

PR #756, opened 2021-08-30
Feature/add blockstamp

neacsu:


contracts/mixnet/src/contract.rs
```diff
@@ -301,21 +304,30 @@ pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, C
         Ok(delegations)
     }
 
+    let old_enough_blockstamp = env.block.height - TWENTY_FOUR_HOURS_BLOCKS;
```
jstuczyn: I'd just go with the safest  approach and set it to the current blockstamp of the migration

contracts/mixnet/src/queries.rs
```diff
@@ -632,14 +705,12 @@ mod tests {
     #[cfg(test)]
     mod querying_for_mixnode_delegations_paged {
```
jstuczyn, line 635: I think some tests for checking block height would be useful

contracts/mixnet/src/storage.rs
```diff
@@ -315,10 +315,18 @@ pub fn reverse_mix_delegations_read<'a>(
     ReadonlyBucket::multilevel(storage, &[PREFIX_REVERSE_MIX_DELEGATION, owner.as_bytes()])
 }
 
+// This function will need to be deleted, once the migration is completed
```
jstuczyn: To not pollute the storage with migration specific functions, you could move it to the "migrate" function scope

contracts/mixnet/src/storage.rs
```diff
@@ -355,6 +363,17 @@ pub fn reverse_gateway_delegations_read<'a>(
     )
 }
 
+// This function will need to be deleted, once the migration is completed
```
jstuczyn: ibid.

jstuczyn (approved):

contracts/mixnet/src/storage.rs
neacsu: Moved them inside that scope, it looks much "dirtier", but I guess it will be easier to remove.

contracts/mixnet/src/queries.rs
neacsu: Discussed offline that more testing will be possible once we start exposing more the block height

PR #757, opened 2021-08-31
Add block_height in the Delegation structure as well

neacsu:


contracts/mixnet/src/contract.rs
```diff
@@ -245,6 +245,8 @@ pub fn migrate(deps: DepsMut, env: Env, _msg: MigrateMsg) -> Result<Response, Co
                             "Non-UTF8 address used as key in bucket. The storage is corrupted!",
                         )),
                         coin(entry.1.u128(), DENOM),
+                        // dummy data, used as a placeholder in the context of migration
```
jstuczyn, line 248: Why are we using dummy data for migration? Let's just use the current block height because people will be very confused when they see height "42"

neacsu: We are using the current block height, it's just in the caller function, as this one's just for reading what's in there right now

neacsu: I.e. in `migration` we already put the `env.block.height` in the `RawDelegationData` that we write. This 42 value is just there so that we can pass the wrapper `Delegation` structure. `env` could also be passed down the function calls, but it's basically the same thing, and this code will be deleted anyways.

jstuczyn (approved):

PR #760, opened 2021-09-02
Feature/bond blockstamp

neacsu:
This will be made into a PR once the following are merged:

- [x] https://github.com/nymtech/nym/pull/757
- [x] https://github.com/nymtech/nym/pull/759

### DIFF
--- a/contracts/mixnet/src/storage.rs
+++ b/contracts/mixnet/src/storage.rs
@@ -567,6 +567,62 @@ mod tests {
         }
 
         #[test]
+        fn when_there_is_a_single_delegation_depending_on_blockstamp() {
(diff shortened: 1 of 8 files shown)


contracts/mixnet/src/storage.rs
```diff
@@ -193,11 +195,14 @@ pub(crate) fn increase_mix_delegated_stakes(
         );
 
         // and for each of them increase the stake proportionally to the reward
+        // if they have delegated for long enough
```
durch: I'd add a more verbose explanation for `long enough`, something along the lines, "At least N blocks have been created since they've delegated"

durch (approved):
Just a tiny docs nit.

PR #764, opened 2021-09-03
Feature/active sets

jstuczyn:
Most of the work done towards rewarding only active set of nodes.

This pull request introduces the concept of active sets - their sizes are set in the contract and you can query validator API in order to obtain nodes that are part of the said sets.

The only missing thing is actually only rewarding nodes that are present in the sets, but that requires merging https://github.com/nymtech/nym/pull/750 first. Then, I _think_ the only required change will be for the `Rewarder` to call `get_active_mixnodes()` (and gateways) as opposed to `get_mixnodes()`

Note: clients are **not** yet using active sets (they use full list of mixnodes). I see a potential problem with using active sets that we should probably discuss -> what if it happens that active set, for argument sake, has size of 100 and contains 98 nodes on layer 1, 1 on layer 2 and 1 on layer 3? Then we have no anonymity - we should solve that... somehow... . Perhaps active set would take n/3 of nodes from each layer instead?


validator-api/src/cache/mod.rs
```diff
@@ -117,12 +127,95 @@ impl ValidatorCache {
...
     }
 
-    async fn update_cache(&self, mixnodes: Vec<MixNodeBond>, gateways: Vec<GatewayBond>) {
+    // TODO: check if all nodes can be compared together,
+    // i.e. they all have the same denom for bonds and delegations
+    fn verify_mixnodes(&self, mixnodes: &[MixNodeBond]) -> bool {
+        if mixnodes.is_empty() {
+            return true;
+        }
+        let expected_denom = &mixnodes[0].bond_amount.denom;
```
durch, line 146: No action needed now, but in general I think we should formalize denom as an enum at least. Keeping it as string will bite us sooner or later. I've sort of formalized it internally in the tauri-wallet. Formalization would also allow us to trivially convert between denoms so checks like this could go away. We'd also be able to have a sane PartialOrd. Maybe something to discuss during a planning meeting.

common/client-libs/validator-client/src/validator_api/routes.rs
```diff
@@ -7,5 +7,7 @@ pub const API_VERSION: &str = VALIDATOR_API_VERSION;
 pub const MIXNODES: &str = "mixnodes";
 pub const GATEWAYS: &str = "gateways";
 
+pub const ACTIVE: &str = "active";
```
durch, line 10: Would it make sense to keep this info in an enum? Even if it only had one variant now, we might want to introduce further state variants in the future...

durch (approved):
Some discussion points, no blockers

PR #765, opened 2021-09-03
Update coconut-rs and use hash_to_scalar from there

neacsu:


jstuczyn:
Looks good to be, but was nothing already using `hash_to_scalar`, i.e. is it not going to break some existing code?

neacsu:
> Looks good to be, but was nothing already using `hash_to_scalar`, i.e. is it not going to break some existing code?

There's this `pub use coconut_rs::*;` that saves it

jstuczyn:
> > Looks good to be, but was nothing already using `hash_to_scalar`, i.e. is it not going to break some existing code?
> 
> There's this `pub use coconut_rs::*;` that saves it

Ah. makes sense

jstuczyn (approved):

PR #766, opened 2021-09-06
Feature/consumable bandwidth

neacsu:


durch (approved):
Looks good, something I'm wondering in general is how we will defend against double spend, and related to that what counts as spend, presenting the credential to the gateway or spending all the bandwidth. In the later case we need a way of keeping track of bandwidth available for a given credential globally. Keeping it in the gateway is simpler but it also limits consumers to only spending the credential with a single gateway. cc @futurechimp 

Took the discussion to kb

jstuczyn:
> Looks good, something I'm wondering in general is how we will defend against double spend, and related to that what counts as spend, presenting the credential to the gateway or spending all the bandwidth. In the later case we need a way of keeping track of bandwidth available for a given credential globally. Keeping it in the gateway is simpler but it also limits consumers to only spending the credential with a single gateway. cc @futurechimp
> 
> Took the discussion to kb

So basically the credential is going to have a pseudorandom serial number embedded inside of it (in the form of g^x, which is going to be proven in ZK, i.e. you never show x). Then once credential is spent, the serial number will be written onto the Blockchain. 

Essentially, whenever gateway attempts to redeem the received credential, the chain would check whether it was not spent before - meaning gateway itself doesn't have to store anything. I guess it could cache received values for some quick rejections

PR #767, opened 2021-09-06
remove alert

fmtabbara:
I've removed the alert and also made a reusable component called `AppAlert` that we can drop in to the app if we want to display any alerts in the future. 

API for the component - 

```
<AppAlert 
   title=<string>
   message=<string> 
   severity=<info | error | warning | success 
/>
```


Aid19801 (approved):
Approved on basis, presentational component, can see no issues.

PR #781, opened 2021-09-20
Routing, Navbar, AppBar, Dark/Light Theme and DataMap

Aid19801:
In This PR:

- Nav - contains MUI AppBar and Drawer components
- Light/Dark switch
- Datamap showing where nodes are situation on react-simple-maps implementation
- Nested Routing
- One 'main' context with all API reqs
- One 'Api' class for all data/fetching


explorer/src/api/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class Api {
+  static fetchMixnodes = async () => {
+    try {
+      const res = await fetch(MIXNODES_API);
+      const json = await res.json();
+      return json;
+    } catch (error: any) {
+      console.log('error ', error);
+      throw error.message;
```
mmsinclair: Rethrow the error and not just the message?

```suggestion
      throw error;
```

explorer/src/api/index.ts
```diff
@@ -0,0 +1,53 @@
+import {
+  GATEWAYS_API,
+  MIXNODES_API,
+  VALIDATORS_API,
+  BLOCK_API,
+} from './constants';
+
+export class Api {
```
mmsinclair, line 11: I'll show you how to add typings to the methods so that the response objects have nice typings. Sadly the OpenAPI crate the Explorer API is using (https://crates.io/crates/rocket_okapi) doesn't have enough features to autogenerate them from yet.

explorer/src/components/ContentCard.tsx
```diff
@@ -0,0 +1,28 @@
...
+      title={title}
+      avatar={Icon}
+      action={Action}
+      subheader={subtitle}
+    />
+    {children && <CardContent>{children}</CardContent>}
+    {errorMsg && (
+      <Typography
+        variant="body2"
+        sx={{ color: 'red', padding: (theme) => theme.spacing(2) }}
```
mmsinclair:
```suggestion
        sx={{ color: 'danger', padding: 2 }}
```

`padding: 2` is the same as `padding: (theme) => theme.spacing(2)` just shorter

And best to use the semantic names for colours i.e. show the user something bad happened, rather than `goldenrod === warning`

explorer/src/components/Nav.tsx
```diff
@@ -1,138 +1,276 @@
...
-              {sidebar ? (
-                <Close className={classes.hamburgerIcon} />
-              ) : (
-                <MenuIcon className={classes.hamburgerIcon} />
-              )}
-            </ListItemIcon>
+            <ListItemButton
+              sx={{
+                color: (theme) =>
+                  theme.palette.mode === 'light' ? '#000' : '#fff',
```
mmsinclair: The theme has a text colour in the palette that responds to the mode:

```suggestion
                color: 'text.primary',
```



image

image

explorer/src/components/Nav.tsx
```diff
@@ -1,138 +1,276 @@
...
+              }}
+            >
+              <ListItemIcon>{each.Icon}</ListItemIcon>
+              <ListItemText
+                primary={each.title}
+                sx={{
+                  color: (theme) =>
+                    theme.palette.mode === 'light' ? '#000' : '#fff',
```
mmsinclair, line 196: See above

explorer/src/components/Title.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const Title = ({ text }: TitleProps) => (
+  <Grid
+    item
+    xs={12}
+    sx={{
+      justifyContent: 'flex-start',
+      padding: (theme) => theme.spacing(2),
+      backgroundColor: (theme) => theme.palette.primary.dark,
```
mmsinclair:
```suggestion
      padding: 2
      backgroundColor: 'primary.dark',
```

explorer/src/components/Title.tsx
```diff
@@ -0,0 +1,33 @@
...
+    }}
+  >
+    <Box
+      sx={{
+        padding: (theme) => theme.spacing(3),
+        backgroundColor: (theme) => theme.palette.primary.light,
+      }}
+    >
```
mmsinclair:
```suggestion
    <Box
      padding="3",
      backgroundColor="primary.light"
    >
```

explorer/src/components/Title.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <Typography
```
mmsinclair, line 24:
```suggestion
      <Typography
        color="primary.main"
      >
```

explorer/src/components/WorldMap.tsx
```diff
@@ -1,111 +1,80 @@
...
-
-  const [scale, setScale] = useState<number>(200);
-  const [windowResizing, setWindowResizing] = useState<boolean>(false);
-
-  const projection = geoEqualEarth()
-    .scale(scale)
-    .translate([cx, cy])
-    .rotate([0, 0]);
-
-  const fetchGeographicData = (): void => {
```
mmsinclair, line 25: Change to `const fetchGeographicData = async () => {` and then something like:

```
const worldData = await (await fetch('/data/world-110m.json')).json();
```

You can add some `try {...} catch(e) { ... }` if you need to know the errors.

mmsinclair, line 25: You don't need to add the typing for `void`, that is implicit in `const someFunction = () => { ... }`

explorer/src/components/WorldMap.tsx
```diff
@@ -1,111 +1,80 @@
...
-      setWindowResizing(true);
-
-      timeout = setTimeout(() => {
-        setWindowResizing(false);
-      }, 200);
-    };
-    window.addEventListener('resize', handleResize);
-
-    return () => window.removeEventListener('resize', handleResize);
+  const [tooltipContent, setTooltipContent] = React.useState<string>('');
```
mmsinclair:
```suggestion
const [tooltipContent, setTooltipContent] = React.useState<string>(null);
```

or


```suggestion
const [tooltipContent, setTooltipContent] = React.useState<string>(undefined);
```

explorer/src/components/WorldMap.tsx
```diff
@@ -1,111 +1,80 @@
...
+  const [data, setData] = React.useState<Record<string, unknown>[]>([]);
```
mmsinclair: Define an `interface Data {...}` and say what the fields of each item will be will be, then do:

```suggestion
const [data, setData] = React.useState<Data[]>([]);
```

explorer/src/components/WorldMap.tsx
```diff
@@ -1,111 +1,80 @@
...
+  const { mode }: any = React.useContext(MainContext);
```
mmsinclair: `:any` isn't the best - does it have a type or a union type or enum you can use instead?

explorer/src/context/main.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
+import { Api } from '../api';
+
+export const MainContext = React.createContext({});
```
mmsinclair:
```suggestion
interface State {
  mode?: 'light' | 'dark';  // also consider `mode: light | dark` and set a default value of `light` or something from local storage
  // add other types for toggleMode, mixnodes, gateways, validators, block 
}

export const MainContext = React.createContext<State>({});
```

explorer/src/context/main.tsx
```diff
@@ -0,0 +1,67 @@
...
+    } catch (error) {
+      setBlock({ data: null, error });
+    }
+  };
+
+  React.useEffect(() => {
+    fetchMixnodes();
+    fetchGateways();
+    fetchValidators();
+    fetchBlock();
```
mmsinclair:
```
Promise.all([
    fetchMixnodes(),
    fetchGateways(),
    fetchValidators(),
    fetchBlock(),
]);
```
would be safer because if you forget to catch any exceptions, you stop the other fetch requests from starting. And you can also never tell when all of them have completed, e.g. global loading state

explorer/src/data/countriesData.ts
```diff
@@ -0,0 +1,39 @@
+export const countriesData = [
```
mmsinclair: You can fetch this from the explorer API - https://testnet-milhon-explorer.nymtech.net/api/countries

explorer/src/icons/NymLogoSVG.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const NymLogoSVG: React.FC = (props) => {
+  const theme = useTheme();
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 5389.9 5389.9"
+      width="40px"
+      height="40px"
+      style={{ margin: theme.spacing(2) }}
```
mmsinclair, line 12: Nice 😀

explorer/src/pages/Overview/index.tsx
```diff
@@ -1,5 +1,99 @@
...
+  const history = useHistory();
+  const { mixnodes, gateways, validators, block }: any =
+    React.useContext(MainContext);
+
+  return (
+    <>
+      <Box component="main" sx={{ flexGrow: 1 }}>
+        <Grid container spacing={0}>
+          <Grid item xs={12}>
+            <Typography sx={{ marginLeft: (theme) => theme.spacing(3) }}>
```
mmsinclair:
```suggestion
            <Typography sx={{ marginLeft: 3 }}>
```

explorer/src/theme.ts
```diff
@@ -1,62 +1,111 @@
...
-      // to be `contrastText`?
-    },
-    secondary: {
-      main: nymPalette.secondary.main,
-      light: nymPalette.secondary.light,
-      dark: nymPalette.secondary.dark,
-    },
-  },
-  shape: {
-    borderRadius: 24,
```
mmsinclair, line 46: Might be a good idea to make the `borderRadius` be a multiple of the spacing? (if it isn't already)

mmsinclair (changes requested):
Looking good so far - only a couple of things you could shorten and improve a bit. Then we can merge it and iterate.

explorer/src/components/Title.tsx
fmtabbara:
```
     p: 2
     bgColor: 'primary.dark',
```

explorer/src/components/WorldMap.tsx
fmtabbara: MUI has a tooltip component - Is it worth using that?

explorer/src/context/main.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }, []);
+
+  return (
+    <MainContext.Provider
+      value={{ mode, toggleMode, mixnodes, gateways, validators, block }}
+    >
+      {children}
+    </MainContext.Provider>
+  );
+};
```
fmtabbara, line 112: Nice work with the context - Looks good :+1:

explorer/src/components/WorldMap.tsx
fmtabbara: I think there is `PaletteType` that you can import from MUI and use here. It will give you "light" | "dark" as the type

fmtabbara: Add the type in the context though and TS will figure it out here

explorer/src/context/main.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
+import { Api } from '../api';
+
+export const MainContext = React.createContext({});
+
+export const MainContextProvider: React.FC = ({ children }: any) => {
```
fmtabbara: you can remove :any here

explorer/src/context/main.tsx
fmtabbara: I think it would be worth creating a type or interface for the global state.

```
type GlobalState = {
 mixnodes: <mixnodesType>
  gateways: <gatewaysType>
}
```
The types can all be taken from the Vue Explorer repo.

and then as per Mark's comment above
```
export const MainContext = React.createContext<State>({});
```

explorer/src/components/WorldMap.tsx
Aid19801: Hmmmm. i remember writing that function a while back assuming i'd need to fetch myself but react-simple-maps does the magic for me:

```
const geoUrl =
  'https://raw.githubusercontent.com/zcreativelabs/react-simple-maps/master/topojson-maps/world-110m.json';
  
  ...
...

      {data.length > 0 && (
            <Geographies geography={geoUrl}>
              {({ geographies }: any) =>
                geographies.map((geo: any) => {
```

Was this off an old commit or something?

explorer/src/components/WorldMap.tsx
Aid19801: I have played with the MUI tooltip but a) it doesnt pair out of the box the way these two play with each other nicely. b) may just be something simple like style override / position / zIndex - but can we use RTT for now and once PR is in i'm happy to explore the MUI one?

explorer/src/theme.ts
Aid19801: The theme as-is is commented out actually. Again, not sure if this is an old commit or something you're looking at?

Theme is currently defined in the AppWrapper. It will be broken out again though. This happened as part of the Dark mode refactor.

explorer/src/api/index.ts
```diff
@@ -5,47 +5,50 @@ import {
...
     try {
       const res = await fetch(GATEWAYS_API);
       const json = await res.json();
       return json;
-    } catch (error) {
+    } catch (error: any) {
       console.log('error ', error);
       return error;
     }
   };
```
mmsinclair:
```suggestion
static fetchGateways = async (): Promise<GatewayResponse> => {
  const res = await fetch(GATEWAYS_API);
  return await res.json();
};
```

mmsinclair: A couple of things:
- `error:any` should just be `error` because it is already typed as `Error`
- logging to the console should be done by the caller
- the error should be thrown and not returned because the signature return type is `Promise<GatewayResponse>` and not `Promise<GatewayResponse | Error>` (which is not a good idea because Promise already has a mechanism for throwing errors)

explorer/src/context/main.tsx
```diff
@@ -1,17 +1,33 @@
+import { PaletteMode } from '@mui/material';
 import * as React from 'react';
+import { GatewayResponse, MixNodeResponse, ValidatorsResponse } from 'src/typeDefs/node-status-api-client';
```
mmsinclair: Might be best to name that file `src/typeDefs/explorer-api`?

When I started this I was using another API before the Explorer API existed. Apologies for creating naming confusion in the vue-js project!

explorer/src/context/main.tsx
```diff
@@ -1,17 +1,33 @@
...
+// import { MixNodeResponse, GatewayResponse } from 'src/typeDefs/node-status-api-client';
 import { Api } from '../api';
 
-export const MainContext = React.createContext({});
+type NodeApiResponse = {
+  data: MixNodeResponse | GatewayResponse | ValidatorsResponse | number | null
+  error: string | null
+}
```
mmsinclair: This isn't a good idea - you're undoing the power of TypeScript to check you by basically defining `any`

explorer/src/context/main.tsx
```diff
@@ -1,17 +1,33 @@
...
+interface State {
+  mode: PaletteMode
+  toggleMode: () => void
+  mixnodes: NodeApiResponse | null
+  gateways: NodeApiResponse | null
+  validators: NodeApiResponse | null
+  block: NodeApiResponse | null
```
mmsinclair:
```suggestion
  mixnodes: MixNodeResponse | null
  gateways: GatewayResponse | null
  validators: ValidatorsResponse | null
  block: BlockResponse | null
```

mmsinclair: You may have been trying to do something like this:

```
interface ApiState<RESPONSE> {
  isLoading: boolean;
  response?: RESPONSE;
  error?: Error;
}

interface State {
  mode: PaletteMode
  toggleMode: () => void
  mixnodes?: ApiState<MixNodeResponse>,
  gateways?: ApiState<GatewayResponse>,
  validators?: ApiState<ValidatorsResponse>,
  block?: ApiState<BlockResponse>,
}
```

When `mixnode` is `undefined` it is "not preset" i.e. never loaded. When it starts loading, the value changes to `mixnode: { isLoading: true }`, when an error: `mixnode: { isLoading: false, error: { message: 'Oh no!' } }` and with data: `mixnode: { isLoading: false, response: {...}}`

explorer/src/context/main.tsx
```diff
@@ -1,17 +1,33 @@
...
+};
+
+export const MainContext = React.createContext({} as State);
```
mmsinclair: Create context has a generic parameter that defines the context's type.

```suggestion
export const MainContext = React.createContext<State>({});
```

mmsinclair (changes requested):
Almost there - only a couple of typings to tidy up on the `State`

explorer/src/context/main.tsx
Aid19801: Ah right. Yep, i'll rename then to something more appropriate. Cheers.

explorer/src/context/main.tsx
Aid19801: This yields a type error - `Argument of type '{}' is not assignable to parameter of type 'State'.
  Type '{}' is missing the following properties from type 'State': mode, toggleMode`
  
  So i've given it what i'm assuming is a default state of mode 'dark' and changed toggleMode to optional.

explorer/src/context/main.tsx
Aid19801: Thanks for this - i _think_ it's sorted now.

mmsinclair (approved):

PR #784, opened 2021-09-21
Feature/persistent gateway storage

### LINKED ISSUE #11: Database for gateway
Gateway should really be using some kinda of database for payload storage rather than raw storage on disk as that is kinda crude + thread unsafe.

jstuczyn:
This is something I started working on today while browsing for things to do that will definitely help in the long term (and also in the medium term when expanding on Coconut capabilities).

Basically the purpose of this pull request is going to be to unify the storage on the gateway with everything being stored in the single database. Prior to this change we had (well, technically still do as, at the time of writing this, it is still a draft):
- shared keys being stored in sled database
- messages for offline clients being stored in individual files directly in the file system (ughh)
- available client bandwidth stored in non-persistent in-memory HashMap

As a super nice byproduct I think we will also be able to significantly simplify internal gateway logic and completely get rid of `ClientsHandler` and all channel communication to it. Even though I wrote it myself over a year ago, I still need to think for a moment to figure out what is going on there. I can only imagine what everyone else must be going through when reading that code (reason for writing it that way was twofold: my own inexperience and the way we were storing messages for offline clients). So future changes should be easier for everyone.

The remaining work includes:
- [x] Start using new storage in the `ClientsHandler` 
- [x] Introduce managers + build API for bandwidth for the `GatewayStorage`
- [x] Use the new bandwidth manager
- [x] Figure out a way of staggering reading messages stored for offline clients (so that we would not be vulnerable to OOM attacks)
- [x] Remove the `ClientsHandler` and possibly introduce some centralised (or behind a lock) in-memory storage with open connections - I haven't yet investigated that part in depth
- [x] Update config file with the setup
- [ ] Possibly some upgrade magic - not really sure if this is going to be worth the hussle
- [x] Cleanup
- [x] Localnet testing 

Closes https://github.com/nymtech/nym/issues/11


gateway/src/node/storage/inboxes.rs
```diff
@@ -15,6 +16,123 @@ use tokio::fs::File;
 use tokio::io::AsyncWriteExt;
 use tokio_stream::wrappers::ReadDirStream;
 
+#[derive(Clone)]
+pub(crate) struct InboxManager {
+    connection_pool: sqlx::SqlitePool,
+    /// Maximum number of messages that can be obtained from the database per operation.
+    /// It is used to prevent memory overflows in the case of client receiving a lot of data while
```
neacsu: I think you meant preventing Out Of Memory errors rather than memory overflows.

gateway/src/node/client_handling/clients_handler.rs
```diff
@@ -93,51 +93,68 @@ impl ClientsHandler {
...
-        } else {
-            // but if all went well, we can now delete it
-            if let Err(e) = self.clients_inbox_storage.delete_files(paths).await {
-                error!(
-                    "Failed to remove client ({:?}) files - {:?}",
-                    client_address.as_base58_string(),
-                    e
-                );
+            if comm_channel.unbounded_send(messages).is_err() {
+                error!("Somehow we failed to stored messages to a fresh client channel - there seem to be a weird bug present!");
```
neacsu: nit: Failed to store

gateway/src/node/client_handling/websocket/connection_handler.rs
```diff
@@ -381,29 +422,42 @@ where
...
+
+        match Bandwidth::try_from(credential) {
+            Ok(bandwidth) => {
+                let mut bandwidth_value = bandwidth.value();
+                if bandwidth_value > i64::MAX as u64 {
+                    // note that this would have represented more than 1 exabyte,
+                    // which is like 125,000 worth of hard drives so I don't think we have
+                    // to worry about it for now...
+                    warn!("Somehow we received bandwidth value higher than 9223372036854775807. Going to cap it at that amount.");
+                    bandwidth_value = i64::MAX as u64;
```
neacsu: Even though this is a strange edge case, I don't think a warning and capping the value is the right approach here, as the client might have genuinely bought that much bandwidth, so we're just cutting down value here. We should return an error and not increase the GW bandwidth until the client sends an appropriate value.

gateway/src/node/storage/bandwidth.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// * `value`: amount of available bandwidth to be added to the client.
+    pub(crate) async fn increase_available_bandwidth(
+        &self,
+        client_address_bs58: &str,
+        value: i64,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query!(
+            r#"
+                UPDATE available_bandwidth
+                SET available = available + ?
```
neacsu, line 71: How does `sqlx` handle overflow situations?

gateway/src/node/storage/bandwidth.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// * `value`: amount of available bandwidth to be removed from the client.
+    pub(crate) async fn decrease_available_bandwidth(
+        &self,
+        client_address_bs58: &str,
+        value: i64,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query!(
+            r#"
+                UPDATE available_bandwidth
+                SET available = available - ?
```
neacsu, line 96: Same question

gateway/src/node/client_handling/websocket/connection_handler.rs
neacsu: I made the comment on the first appearance of this code, and I think it's still viable, but the code was moved somewhere else in another commit

gateway/src/node/storage/inboxes.rs
jstuczyn: Yeap - fixed.

gateway/src/node/client_handling/clients_handler.rs
jstuczyn: This code no longer exists anyway : )

gateway/src/node/client_handling/websocket/connection_handler.rs
jstuczyn: Personally I don't think client could have genuinely brought this much bandwidth - it's over 1 exabyte which is around 1% capacity of ALL datacenters in the world : ) But perhaps returning an error as you suggest might be a better approach than capping it - I'm going to change it.

gateway/src/node/storage/bandwidth.rs
jstuczyn: That is a very interesting question so I've run an experiment. Basically `sqlite` will happily store data bigger than maximum size of `int64` (so we never lose any "overflow" bandwidth). However, on query the results from `sqlx` are capped at the maximum value, meaning that in the unlikely scenario of client having more than few exabytes of traffic available, he is not going to see anything above it (but it would still exist in the database)

gateway/src/node/storage/bandwidth.rs
jstuczyn: And same answer : )

neacsu (approved):

PR #786, opened 2021-09-22
Feature/migrate hidden delegations

neacsu:
- [x] Bond, delegate, unbond, do the last three migrations that were done on milhon + this last one that should fix the hidden delegations problem and check that the old delegations on unbonded nodes are migrated
- [x] Write more unit tests for the new queries


common/client-libs/validator-client/src/client.rs
```diff
@@ -7,7 +7,7 @@ use crate::nymd::{
 };
 use crate::{validator_api, ValidatorClientError};
 use coconut_interface::{BlindSignRequestBody, BlindedSignatureResponse, VerificationKeyResponse};
-use mixnet_contract::{GatewayBond, MixNodeBond};
+use mixnet_contract::{GatewayBond, MixNodeBond, RawDelegationData};
```
jstuczyn: GitHub Actions says `RawDelegationData` import is unused : )

common/mixnet-contract/src/delegation.rs
```diff
@@ -7,7 +7,9 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 
-#[derive(Debug, Deserialize, PartialEq, Serialize)]
+pub type UnpackedDelegation<T> = (Addr, IdentityKey, T);
```
jstuczyn: Unless I missed something, it looks like the only variant ever constructed is `UnpackedDelegation<RawDelegationData>`, if that's the case, is the generic really required here?

jstuczyn: Also, perhaps for API/readability sake it should be a struct with named fields rather than a tuple? What do you think?

contracts/mixnet/src/helpers.rs
```diff
@@ -11,6 +15,9 @@ const HOURS_IN_YEAR: u128 = 8760;
 
 const DECIMAL_FRACTIONAL: Uint128 = Uint128(1_000_000_000_000_000_000u128);
 
+// cosmwasm bucket internal value
+const NAMESPACE_LENGTH: usize = 2;
```
jstuczyn, line 20: Is there no way to avoid using this? This looks like a ticking timebomb that will bite us in the future

contracts/mixnet/src/helpers.rs
```diff
@@ -65,9 +72,88 @@ pub(crate) fn scale_reward_by_uptime(
...
+pub(crate) fn identity_and_owner_to_bytes(identity: &IdentityKey, owner: &Addr) -> Vec<u8> {
+    let mut bytes = u16::to_be_bytes(identity.len() as u16).to_vec();
+    bytes.append(&mut identity.as_bytes().to_vec());
+    bytes.append(&mut owner.as_bytes().to_vec());
+
+    bytes
+}
+
+pub(crate) fn get_all_delegations_paged<T>(
+    bucket: &ReadonlyBucket<T>,
```
jstuczyn, line 120: perhaps we should take `Deps` rather than bucket as an argument for consistency sake with other `query_XXX_paged` methods?

contracts/mixnet/src/helpers.rs
```diff
@@ -65,9 +72,88 @@ pub(crate) fn scale_reward_by_uptime(
...
+    start_after: &mut Option<Vec<u8>>,
```
jstuczyn: seeing this as mutable is a big red flag...

contracts/mixnet/src/helpers.rs
```diff
@@ -65,9 +72,88 @@ pub(crate) fn scale_reward_by_uptime(
...
+where
+    T: Serialize + DeserializeOwned,
+{
+    let delegations = bucket
+        .range(start_after.as_deref(), None, Order::Ascending)
+        .filter(|res| res.is_ok())
+        .take(limit)
+        .map(|res| {
+            res.map(|entry| {
+                let (owner, identity) = extract_identity_and_owner(entry.0).expect("Invalid node identity or address used as key in bucket. The storage is corrupted!");
```
jstuczyn, line 133: I'm confused, why do we have to parse the value returned from the bucket? In previous cases it seemed to have always been returning deserialized data for us?

contracts/mixnet/src/helpers.rs
```diff
@@ -65,9 +72,88 @@ pub(crate) fn scale_reward_by_uptime(
...
+                (owner, identity, entry.1)
+            })
+        })
+        .collect::<StdResult<Vec<UnpackedDelegation<T>>>>()?;
+
+    *start_after = bucket
```
jstuczyn: I don't think that changing `start_after` should be responsibility of this function

contracts/mixnet/src/helpers.rs
```diff
@@ -65,9 +72,88 @@ pub(crate) fn scale_reward_by_uptime(
     Ok(uint128_to_decimal(scaled))
 }
 
+// Extracts the node identity and owner of a delegation from the bytes used as
+// key in the delegation buckets.
+fn extract_identity_and_owner(bytes: Vec<u8>) -> StdResult<(Addr, IdentityKey)> {
```
jstuczyn, line 78: This function seems like some unnecessary hack - is there really no way around this?

contracts/mixnet/src/queries.rs
```diff
@@ -144,6 +144,24 @@ pub(crate) fn query_mixnode_delegations_paged(
...
+    start_after: Option<Vec<u8>>,
+    limit: Option<u32>,
+) -> StdResult<PagedAllDelegationsResponse> {
+    let limit = limit
+        .unwrap_or(DELEGATION_PAGE_DEFAULT_LIMIT)
+        .min(DELEGATION_PAGE_MAX_LIMIT) as usize;
+
+    let bucket = all_mix_delegations_read::<RawDelegationData>(deps.storage);
+    let mut start = start_after.map(|mut v| {
+        v.push(0);
```
jstuczyn, line 158: why are you pushing the 0 byte here?

contracts/mixnet/src/queries.rs
```diff
@@ -230,6 +248,24 @@ pub(crate) fn query_gateway_delegations_paged(
...
+    start_after: Option<Vec<u8>>,
+    limit: Option<u32>,
+) -> StdResult<PagedAllDelegationsResponse> {
+    let limit = limit
+        .unwrap_or(DELEGATION_PAGE_DEFAULT_LIMIT)
+        .min(DELEGATION_PAGE_MAX_LIMIT) as usize;
+
+    let bucket = all_gateway_delegations_read::<RawDelegationData>(deps.storage);
+    let mut start = start_after.map(|mut v| {
+        v.push(0);
```
jstuczyn, line 261: same question

contracts/mixnet/src/storage.rs
```diff
@@ -304,6 +306,13 @@ pub fn gateways_owners_read(storage: &dyn Storage) -> ReadonlyBucket<IdentityKey
 }
 
 // delegation related
+pub fn all_mix_delegations_read<T>(storage: &dyn Storage) -> ReadonlyBucket<T>
```
jstuczyn, line 309: Can't we use a concrete type here?

contracts/mixnet/src/transactions.rs
```diff
@@ -29,42 +31,25 @@ pub(crate) const MINIMUM_BLOCK_AGE_FOR_REWARDING: u64 = 17280;
...
-) -> StdResult<Vec<(Vec<u8>, T)>> {
+) -> StdResult<Vec<UnpackedDelegation<T>>> {
     // I think it's incredibly unlikely to ever read more than that
     // but in case we do, we should guard ourselves against possible
     // out of memory errors (wasm contracts can only allocate at most 2MB
     // of RAM, so we don't want to box the entire iterator)
     let mut delegations = Vec::new();
-    // let mut total_delegation = Coin::new(0, DENOM);
-    let mut start = None;
+    let mut start_after = None;
```
jstuczyn: I haven't been closely watching changes to this file, but big red flag I can see right now is that you're putting everything into the `delegations` vector. We don't want to do this - if some node has a lot of delegations this will cause an OOM. we need to somehow stagger it or something...

jstuczyn (commented):
I've only glanced at the code - I will have another look at it later today/tomorrow

common/client-libs/validator-client/src/client.rs
neacsu: It's the `nymd-client` feature that causes this. I've fixed it now :)

contracts/mixnet/src/transactions.rs
durch: I’ve put that one in in order to be able to overwrite delegations of the old type. Given the OOM concern is there a better place to have something like this for migration purposes?

common/mixnet-contract/src/delegation.rs
neacsu: The generic links up with the generic from `get_all_delegations_paged`, which is called by `delegations`, which is also generic, so that it can be used (like in this PR, but also in future ones as well) for migrations regarding delegations that have different types. That way, it can be used for both the current form of `RawDelegationData` and any other variant we might need in future migrations.
Having it as a struct might be better, I'll do the changes

contracts/mixnet/src/helpers.rs
neacsu: Unfortunately I don't think there is much we can do about it. I think this namespace prefix appears because the node identity is part of the namespace of the bucket. One possible solution would be to move it into the key space, where the owner address also resides, by concatenation. I _think_ that might make this hardcoded value go away, but I'm not sure it's possible to iterate over partial keys the way I iterated here over partial namespaces. In any case, moving the node identity from namespace to key space would require yet another migration, so I definitely wouldn't do that now, that the delegations are in a somewhat weird state.
On the bright side, if anything changes in the internals of cosmwasm bucket regarding this, our unit tests should catch it.

contracts/mixnet/src/helpers.rs
neacsu: Answered above

contracts/mixnet/src/helpers.rs
neacsu: We wouldn't be able to have the generic type in that case.

contracts/mixnet/src/helpers.rs
neacsu: Maybe I went a little bit to C-like here :D
I don't really see it to be very dangerous, but I will change it to go into the returned value, wrapped around in a structure, if you'd like

contracts/mixnet/src/helpers.rs
neacsu: The reason is what I mentioned in previous comments: we were only iterating over the keys of the buckets. Now we need to iterate over the namespaces of the bucket, which are handled differently (with some prefix etc.)

contracts/mixnet/src/helpers.rs
neacsu: I'll put it in the return value

contracts/mixnet/src/storage.rs
neacsu: Generic type makes it usable for migrations as well, and eliminates the possible mistakes of re-writing a lot of code just for migration deserializations

contracts/mixnet/src/transactions.rs
neacsu: As per the offline discussion, I will look into having it as an iterator, so that it doesn't consume memory for storing all the delegations.

contracts/mixnet/src/queries.rs
neacsu: Appending a 0 byte basically gets you to the next possible key, in the key byte array. We do the same thing in other places to obtain the `start_after` places: (i.e. https://github.com/nymtech/nym/blob/020cad897d208cb21d3cd5c73294f02ea2881d05/contracts/mixnet/src/storage.rs#L187)

contracts/mixnet/src/helpers.rs
```diff
@@ -65,11 +73,181 @@ pub(crate) fn scale_reward_by_uptime(
...
+            })
+        })
+        .collect::<StdResult<Vec<UnpackedDelegation<T>>>>()?;
+
+    let start_next_after = bucket
+        .range(start_after.as_deref(), None, Order::Ascending)
+        .filter(|res| res.is_ok())
+        .take(limit)
+        .last()
+        .map(|res| res.unwrap().0);
```
durch: Is there a scenario in which `filter` returns an empty iterator? I'm a bit worried about that `unwrap`, could we do an `if let` instead of the `map`?

contracts/mixnet/src/transactions.rs
```diff
@@ -2,84 +2,27 @@
...
-        Ok(delegations) => Ok(Coin::new(
-            delegations.iter().fold(0, |acc, x| acc + x.1.amount.u128()),
-            "upunk",
-        )),
-        Err(e) => Err(e),
-    }
+    Ok(Coin::new(
+        Delegations::new(delegations_bucket)
+            .fold(0, |acc, x| acc + x.delegation_data.amount.u128()),
+        "upunk",
```
durch: Could we use the [`DENOM`](https://github.com/nymtech/nym/blob/develop/common/network-defaults/src/lib.rs#L69) const here?

durch:
Two comments, looks good otherwise.

contracts/mixnet/src/helpers.rs
neacsu: If `filter` returns an empty iterator, `take` will also return an empty iterator and `last` will return `None`, so `map`'s lambda won't even get invoked. The `unwrap` from the `map` is safe because of the above `filter` that checks all elements can safely be unwrapped.
That being said, I will change this approach to a more readable one, using `if let`.

contracts/mixnet/src/transactions.rs
neacsu: Looks like I've failed my own advice here :)) Will change it, thanks

durch: I think this one is not even yours, its adjacent :)

durch (approved):

PR #788, opened 2021-09-24
Feature/get all delegations

fmtabbara:


tauri-wallet/src-tauri/src/operations/delegate.rs
```diff
@@ -92,3 +95,33 @@ pub async fn undelegate_from_gateway(
...
   }
 }
+
+#[tauri::command]
+pub async fn get_reverse_mix_delegations_paged(
+  state: tauri::State<'_, Arc<RwLock<State>>>,
+) -> Result<PagedReverseMixDelegationsResponse, String> {
+  let r_state = state.read().await;
+  let client = r_state.client()?;
+  match client
```
jstuczyn: Rather than using `match`, you could use `map_err` here, i.e. 
```rust
client
  .get_reverse_mix_delegations_paged(Addr::unchecked(client.address().as_ref()), None, None)
  .await
  .map_err(|err| format_err!(err))
```

tauri-wallet/src-tauri/src/operations/delegate.rs
```diff
@@ -92,3 +95,33 @@ pub async fn undelegate_from_gateway(
...
+  state: tauri::State<'_, Arc<RwLock<State>>>,
+) -> Result<PagedReverseGatewayDelegationsResponse, String> {
+  let r_state = state.read().await;
+  let client = r_state.client()?;
+  match client
+    .get_reverse_gateway_delegations_paged(Addr::unchecked(client.address().as_ref()), None, None)
+    .await
+  {
+    Ok(r) => Ok(r),
+    Err(e) => Err(format_err!(e)),
```
jstuczyn: same `map_err` note here

jstuczyn (commented):
I only looked at the rust changes and I assume the typescript is fine here : ) But I presume it will be reviewed by somebody else

durch:
I'll add type generation, its gonna be a bit more involved then usual, as we'll have to wrap `cosmwasm-std::Address` to get it. I think its useful to do it anyway as we'll have more control over how we use it...

durch:
> I'll add type generation, its gonna be a bit more involved then usual, as we'll have to wrap `cosmwasm-std::Address` to get it. I think its useful to do it anyway as we'll have more control over how we use it...

This is a lot more involved then I anticipated, namely we have to migrate to our own Coin and Address types, and that pulls on a lot of threads, so I'd move that to a separate Roadmap item. I suggest we merge this as is, the work on the types. Ideally we should use our own types everywhere, and have them be exportable to TS.

durch (approved):

PR #792, opened 2021-09-27
Tauri wallet

durch:
Unified frontend and backend of the Tauri wallet


durch:
Merged through the CLI

PR #794, opened 2021-09-30
Total mix stake accounting

durch:
Adds total mix stake value to the contract, and threads it all the way to the `validator-api`. Total mix stake is increased on mixnode bonding and delegation and decreased on unbonding and undelegation. 

- [x] implement total mix stake
- [x] implement total gateway stake
- [x] tests


contracts/mixnet/src/contract.rs
```diff
@@ -201,11 +201,13 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
             gateway_identity,
             address,
         )?),
+        QueryMsg::GetTotalMixStake {} => to_binary(&queries::query_total_mix_stake(deps)),
     };
 
     Ok(query_res?)
 }
 
+// TODO: Calculate initial total stake when deploying this change
```
jstuczyn: I'd perhaps move it explicitly to migrate via `todo!()` macro, so that migration would fail to run without this change (in case somebody misses the comment and wanted to run it)

contracts/mixnet/src/queries.rs
```diff
@@ -93,6 +93,10 @@ pub(crate) fn query_layer_distribution(deps: Deps) -> LayerDistribution {
     read_layer_distribution(deps.storage)
 }
 
+pub(crate) fn query_total_mix_stake(deps: Deps) -> Uint128 {
+    total_mix_stake(deps.storage).load().unwrap()
```
jstuczyn: I'd write a comment here about why the unwrap here is fine - initially it scared me until I looked at the call stack.

contracts/mixnet/src/storage.rs
```diff
@@ -46,6 +47,32 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+
+pub fn mut_total_mix_stake(storage: &mut dyn Storage) -> Singleton<Uint128> {
+    singleton(storage, TOTAL_STAKE_KEY)
+}
+
+pub fn incr_total_mix_stake(amount: Uint128, storage: &mut dyn Storage) {
+    let stake = total_mix_stake(storage)
+        .load()
+        .unwrap()
+        .checked_add(amount)
```
jstuczyn: If we're doing a checked_add then we shouldn't be really unwraping it, should we? Either deal with failure cases or just change it into `wrapping_add` or `saturating_add`

contracts/mixnet/src/storage.rs
```diff
@@ -46,6 +47,32 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+        .unwrap();
+    mut_total_mix_stake(storage).save(&stake).unwrap();
+}
+
+pub fn decr_total_mix_stake(amount: Uint128, storage: &mut dyn Storage) {
+    let stake = total_mix_stake(storage)
+        .load()
+        .unwrap()
+        .checked_sub(amount)
```
jstuczyn: Same as above

validator-api/src/cache/mod.rs
```diff
@@ -37,6 +37,8 @@ struct ValidatorCacheInner {
 
     current_mixnode_active_set_size: AtomicUsize,
     current_gateway_active_set_size: AtomicUsize,
+
+    total_mix_stake: RwLock<Cache<u128>>,
```
jstuczyn, line 41: Does it really make sense to wrap it in cache? Also is there no AtomicU128 to get rid of the lock?

jstuczyn (changes requested):
Looks rather good to me - there are only some tiny thing I've mentioned in the code plus the additional two suggestions:

- perhaps introduce the same concept for total gateway stake for consistency and because presumably you will need it for rewards calculations?
- some tests to make sure the global value is always kept in sync in the contract, especially in the presence of the "shadow" delegations. I _think_ it's fine, but I'd feel more confident if it was actually tested

validator-api/src/cache/mod.rs
durch: Unfortunately no, `AtomicU128` is in a [weird limbo state](https://github.com/rust-lang/rust/issues/32976#issuecomment-893117019)...

contracts/mixnet/src/storage.rs
durch: This is a good point, we likely want saturating add anyway

contracts/mixnet/src/contract.rs
durch: I had no idea `todo!()` was a thing :)

contracts/mixnet/src/storage.rs
durch: Altough saturating has the potential of leading to weirdness, probably better to just handle the error

common/client-libs/validator-client/src/client.rs
```diff
@@ -182,6 +182,20 @@ impl<C> Client<C> {
...
     }
 
+    pub async fn get_total_mix_stake(&self) -> Result<u128, ValidatorClientError>
+    where
+        C: CosmWasmClient + Sync,
+    {
+        Ok(self.nymd.get_total_mix_stake().await?.u128())
+    }
+
+    pub async fn get_total_gt_stake(&self) -> Result<u128, ValidatorClientError>
```
jstuczyn: I think you should rename it to `gateway_stake` for consistency sake, I don't think we ever refer to gateways by `gt` in any public API

common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -198,6 +198,26 @@ impl<C> NymdClient<C> {
...
+    where
+        C: CosmWasmClient + Sync,
+    {
+        let request = QueryMsg::GetTotalMixStake {};
+        self.client
+            .query_contract_smart(self.contract_address()?, &request)
+            .await
+    }
+
+    pub async fn get_total_gt_stake(&self) -> Result<Uint128, NymdError>
```
jstuczyn: Same naming concern

common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -198,6 +198,26 @@ impl<C> NymdClient<C> {
...
+    where
+        C: CosmWasmClient + Sync,
+    {
+        let request = QueryMsg::GetTotalGtStake {};
```
jstuczyn: And it seems same thing for the `QueryMsg::GetTotalGtStake` if we're renaming things

contracts/mixnet/src/contract.rs
```diff
@@ -207,43 +207,17 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
...
-        let owner = delegation.owner;
-        let node_identity = delegation.node_identity;
-        let amount = delegation.delegation_data;
-        let raw_delegation = RawDelegationData::new(amount, env.block.height);
-        gateway_delegations(deps.storage, &node_identity)
-            .save(owner.as_bytes(), &raw_delegation)?;
-        reverse_gateway_delegations(deps.storage, &owner).save(node_identity.as_bytes(), &())?;
-    }
+pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+    todo!("Calculate initial total mix and gateway stake after initial deployment");
```
jstuczyn, line 218: Actually I never asked - is this  going to be in another pull request or something? Because technically after this PR is merged, any other contract changes will be blocked until the migration is implemented

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +51,72 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+}
+
+pub fn mut_total_gateway_stake(storage: &mut dyn Storage) -> Singleton<Uint128> {
+    singleton(storage, TOTAL_GT_STAKE_KEY)
+}
+
+pub fn total_mix_stake_value(storage: &dyn Storage) -> Uint128 {
+    match total_mix_stake(storage).load() {
+        Ok(value) => value,
+        Err(_e) => Uint128(0),
```
jstuczyn, line 73: Under what conditions can it produce an error? And if it produces an error isn't it technically some undefined behaviour?

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +51,72 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+    }
+}
+
+pub fn incr_total_mix_stake(
+    amount: Uint128,
+    storage: &mut dyn Storage,
+) -> Result<Uint128, ContractError> {
+    let stake = total_mix_stake_value(storage).checked_add(amount)?;
```
jstuczyn, line 88: Sorry to go back to this after we have already resolved it, but upon further thinking the original `unwrap` here is absolutely fine. The reason being is that it's impossible to have higher total mix stake than our entire token supply which is guaranteed to easily fit in an `uint128`

validator-api/src/cache/mod.rs
```diff
@@ -37,6 +37,9 @@ struct ValidatorCacheInner {
...
+    total_gt_stake: RwLock<Cache<u128>>,
```
jstuczyn: naming : )

validator-api/src/nymd_client.rs
```diff
@@ -92,6 +92,20 @@ impl<C> Client<C> {
...
     }
 
+    pub(crate) async fn get_total_mix_stake(&self) -> Result<u128, ValidatorClientError>
+    where
+        C: CosmWasmClient + Sync,
+    {
+        Ok(self.0.read().await.get_total_mix_stake().await?)
+    }
+
+    pub(crate) async fn get_total_gt_stake(&self) -> Result<u128, ValidatorClientError>
```
jstuczyn: and bit more of naming

contracts/mixnet/src/contract.rs
durch: I think I'll merge this into the tokenomics branch first, then once we're ready to get that done we can preform the migration

contracts/mixnet/src/storage.rs
durch: This handles read in case no values have yet been written to storage, primarily for tests, although it will be useful for migration as well

contracts/mixnet/src/storage.rs
durch: It resulted in much nicer code, so I think it was a good comment regardless :)

PR #799, opened 2021-10-02
Network Explorer - mix node detail page

Aid19801:
https://github.com/nymtech/nym/issues/2660

- Mixnode Detail page.
- Live Data going into each card (Uptime story, Location, Node status, stats).
- Loading and Error statuses for cards.
- Other small tweaks.


explorer/src/api/constants.ts
```diff
@@ -1,8 +1,16 @@
...
 export const VALIDATORS_API =
   'https://testnet-milhon-validator1.nymtech.net/validators';
 export const BLOCK_API = 'https://testnet-milhon-validator1.nymtech.net/block';
 export const COUNTRY_DATA_API = 'https://testnet-milhon-explorer.nymtech.net/api/countries';
+export const UPTIME_STORY_API = `https://testnet-milhon-validator1.nymtech.net/api/v1/status/mixnode`; // add ID then '/history' to this.
+export const MIXNODE_STATS_API = `https://testnet-milhon-explorer.nymtech.net/api/mix-node`; // add ID then `/stats` or `/delegations` to this.
+export const DELEGATIONS_API = `https://testnet-milhon-explorer.nymtech.net/api/mix-node`; // add ID then `/stats` or `/delegations` to this.
+
+
+export const MIXNODE_API_ERROR = 'We\'re having trouble finding that record, please try again or Contact Us.'
```
fmtabbara: Can we break this down a little so that we have something like - 
```
const BASE_URL = "http://testnet-milhon-explorer.nymtech.net"
const BASE_VALIDATOR_URL = "https://testnet-milhon-validator1.nymtech.net"

const COUNTRY_API = `${BASE_API}/block`
const UPTIME_STORY_API = `${BASE_VALIDATOR_URL}/api/v1/status/mixnode
...
```
line 12 and 13 seem to have the same url. Is that intentional?

explorer/src/api/index.ts
```diff
@@ -4,17 +4,64 @@ import {
...
+  }
+}
+
+function storeInCache(key: string, data: any) {
+  localStorage.setItem(key, data);
+}
+
+function clearCache(key: string) {
+  return localStorage.removeItem(key);
+}
```
fmtabbara, line 40: As we seem to have solved the performance issue using pagination with the data grid, we might now need any caching for the mixnode data?

Wait to see what Mark says though.

explorer/src/components/BondBreakdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+            delegations: 0,
+            pledges: 0,
+            bondsTotal: 0,
+            denom: 'PUNK',
+        })
+        const theme = useTheme();
+        const matches = useMediaQuery(theme.breakpoints.down("sm"));
+        
+        React.useEffect(() => {
+            // bonds: total, pledges, delegations
```
fmtabbara: Commented intentionally?

explorer/src/components/MixnodesTable.tsx
```diff
@@ -0,0 +1,119 @@
+import * as React from 'react';
```
fmtabbara: Do we need this component now that we have the data grid?

explorer/src/components/Mixnodes-DataGrid.tsx
```diff
@@ -0,0 +1,70 @@
...
+    field: 'identity_key',
+    headerName: 'Identity Key',
+    width: 500,
+    renderCell: (params: GridRenderCellParams) => {
+      return (
+        <Link to={`/network-components/mixnodes/${params.value}`} style={{ textDecoration: 'none', color: 'white', marginLeft: 16 }}>
+          {params.value}
+        </Link>
+      )
+    }
```
fmtabbara: :100:

explorer/src/components/UptimeChart.tsx
```diff
@@ -0,0 +1,110 @@
...
+interface ChartProps {
+    title?: string
+    xLabel: string
+    yLabel: string
+    uptimeStory: ApiState<UptimeStoryResponse>
+    loading: boolean
+}
+
+type FormattedDateRecord = [string, number, number];
+type FormattedChartHeadings = [string, string, string];
```
fmtabbara: if all the data is of type string you can do `string[]`

explorer/src/components/BondBreakdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export function BondBreakdownTable() {
+        const { id }: any = useParams();
+        const { mixnodeDetailInfo, delegations } = React.useContext(MainContext);
+        const [ bonds, setBonds ] = React.useState({
+            delegations: 0,
+            pledges: 0,
+            bondsTotal: 0,
+            denom: 'PUNK',
+        })
```
fmtabbara: Code is fine but there seems to be an inconsistency between tab spacing in some files? The spacing on this one is much more than others?

explorer/src/components/Mixnodes-DataGrid.tsx
```diff
@@ -0,0 +1,70 @@
...
+          })) || []
+        }
+        pageSize={50}
+        rowsPerPageOptions={[5]}
+        disableSelectionOnClick
+      />
+    </Box>
+  );
+  return null;
+};
```
fmtabbara: Might be an idea to make this a component that accepts the data as a props. Can the be reused for Gateways?

explorer/src/components/MixnodesTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <>
+                {mixnodes.error && <ComponentError text='There was an error retrieving this mixnode information' />}
+                <TableContainer component={Paper}>
+                    <Table sx={{ minWidth: 650 }} aria-label='mixnodes table'>
+                        <TableHead>
+                            <TableRow>
+                                {tableHeadings.map((eachHeading: TableHeading, i: number) => (
+                                    <TableCell sx={{ fontWeight: 'bold' }} key={eachHeading.id} align='left'>{eachHeading.label}</TableCell>
+                                ))}
+                            </TableRow>
```
fmtabbara: More wide spacing?

explorer/src/api/constants.ts
Aid19801: @fmtabbara yes thanks for reminding me, meant to have another look at that.

explorer/src/api/index.ts
Aid19801: Agree

explorer/src/components/BondBreakdown.tsx
Aid19801: Hangover from me working my way through requirement. Have removed.

explorer/src/components/MixnodesTable.tsx
Aid19801: Probably not, but i am still using it in the actual detail page still (right now)

explorer/src/components/BondBreakdown.tsx
Aid19801: I wonder if this was just before i moved to the UK mac. On the Swiss mac setup/config i was getting format on save perfect. On this one i've had to bring my own vscode settings file in to enact that. I worry my config isnt playing nice with Mark's _prettier_ - @mmsinclair perhaps i could borrow you for 10-15 (monday or tuesday) to check i'm not vandalising your hard work on the config.

explorer/src/components/Mixnodes-DataGrid.tsx
Aid19801: Absolutely. The intention originally was to make the MixnodesTable one table for reuse throughout app.
Now we have the datagrid looking good we can break it out further. 
Up to you and @mmsinclair really - if there's urgency to get a version of the app live quickly, deployed to staging etc - would suggest we park that for the mo (mixnodes table works). Then iterate and expand the DataGrid.

If we're doing okay for time, i'm happy to quickly sort that monday/tuesday. Won't take long.

mmsinclair:
```            id: m.owner,
            owner: m.owner,
            location: m.location?.country_name || '',
            identity_key: m.mix_node.identity_key || '',
            bond: m.bond_amount.amount || '',
            host: m.mix_node.host || '',
            layer: m.layer || '',
```

Seems odd to have to set empty strings everywhere. I think there are two options:
- a container component that does that mapping and then passes props down
- a hook that does it and watches the context state with an effect and sets its own state (to stop re-renders)

explorer/src/components/BondBreakdown.tsx
mmsinclair: These should be setting it correctly:

https://github.com/nymtech/nym/blob/feature/network-explorer-react/explorer/.editorconfig#L10-L11
https://github.com/nymtech/nym/blob/feature/network-explorer-react/explorer/.eslintrc#L84-L85

I looked through the config and what I put in prettier is too narrow (80 chars) - it should be more like 150:
https://github.com/nymtech/nym/blob/feature/network-explorer-react/explorer/.eslintrc#L84-L85

So the whole `explorer` directory probably needs a run of prettier to fix things up - this should be a separate commit.

explorer/src/components/Mixnodes-DataGrid.tsx
Aid19801: I have since broken this out into a util function, takes the api response and returns a useable formatted array of objects for the data grid. Func is then used wherever the data grid is [see latest commit]

Currently looks like this:

```
export function mixnodeToGridRow(arrayOfMixnodes: MixNodeResponse) {
  
  let arr = [];
  
  arrayOfMixnodes !== undefined && arrayOfMixnodes.forEach((eachRecord: MixNodeResponseItem) => {
    let formattedRow = {
      id: eachRecord.owner,
      owner: eachRecord.owner,
      location: eachRecord?.location?.country_name || '',
      identity_key: eachRecord.mix_node.identity_key || '',
      bond: eachRecord.bond_amount.amount || '',
      host: eachRecord.mix_node.host || '',
      layer: eachRecord.layer || '',
    }
    arr.push(formattedRow);
  })
  return arr;
}
```

Then used as, follows... let me know if you'd prefer the hook you mentioned?

```
export const PageMixnodes: React.FC = () => {
  const { mixnodes } = useContext(MainContext);
  const [rows, setRows] = React.useState();

  React.useEffect(() => {
    if (mixnodes?.data !== undefined) {
      const formattedRows = mixnodeToGridRow(mixnodes.data);
      setRows(formattedRows);
    };
  }, [mixnodes])
  return (
    <>
      <Typography sx={{ marginBottom: 1 }} variant="h5">
        Mixnodes
      </Typography>
      <UniversalDataGrid
        loading={mixnodes?.isLoading}
        columnsData={columns}
        rows={rows}
        height={1080}
      />
    </>
  );
};
```

mmsinclair (approved):
It is fine to merge in the feature branch and do some grooming there later (fix up formatting everywhere, fix up docs, etc)

Aid19801:
As approved by fmtabbara and mmsinclair

PR #802, opened 2021-10-04
Tokenomics rewards

durch:
Implementing tokenomics paper:

- [x] Determine inflation pool size
- [x] Determine default cost and profit share
- [x] Implement delegator vs owner distribution
- [x] Migration
- [x] Tests

This is ready for review, I'll make another pass through the code, but I'm pretty happy with it overall.


jstuczyn:
I'll review it properly ~~on Monday~~ when the float issue is resolved : ), in the meantime, could you link or send me the python calculation script for some reference?

durch:
> I'll review it properly ~on Monday~ when the float issue is resolved : ), in the meantime, could you link or send me the python calculation script for some reference?

+ [node rewards](https://gitlab.nymte.ch/claudia/econ-scripts/blob/master/Econ_Results_econ.py#L358)
+ [operator and delegator rewards](https://gitlab.nymte.ch/claudia/econ-scripts/blob/master/Econ_Results_econ.py#L418)

This is a giant PR, maybe it makes sense to do a review over a call, cc @futurechimp, @neacsu

contracts/mixnet/Makefile
```diff
@@ -0,0 +1,18 @@
+upload: code_id
+
+wasm:
+	RUSTFLAGS='-C link-arg=-s' cargo build --release --target wasm32-unknown-unknown
+
+code_id: 
+	echo 13 > code_id
```
jstuczyn: Does it make sense to have a hardocded 13 always saved here? : )

contracts/mixnet/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+migrate: upload
+	cat code_id | xargs -I {} validator-client-scripts --nymd-url ${QA_NYMD_URL} --mnemonic "${QA_MNEMONIC}" --contract ${QA_CONTRACT}  migrate --code-id {}
```
jstuczyn: Regarding the rest of the commands, do they make sense in the context of this public repo? Realistically nobody will be able to use it as we keep the used binary (i.e. `validator-client-scripts`) private. Alternatively we could move that to a public repo or something

common/mixnet-contract/Cargo.toml
```diff
@@ -8,10 +8,16 @@ edition = "2018"
...
+cosmwasm-std = { git = "https://github.com/jstuczyn/cosmwasm", branch = "0.14.1-updatedk256" }
 #cosmwasm-std = { version = "0.14.1" }
 
 serde = { version = "1.0", features = ["derive"] }
 serde_repr = "0.1"
 schemars = "0.8"
-ts-rs = "3.0"
+ts-rs = { version = "3.0", optional = true }
+thiserror = "1.0"
+config = { path = "../config" }
```
jstuczyn: Why do we need to depend on `config` in `mixnet-contract`? From what I can see you only use it to import `DEFAULT_OPERATOR_EPOCH_COST`, however you can access it via depending on just `network-defaults`

common/mixnet-contract/Cargo.toml
```diff
@@ -8,10 +8,16 @@ edition = "2018"
...
+num = "0.4.0"
+
+[features]
+default = ["ts-rs"]
```
jstuczyn, line 25: Should it really be default? I'd argue it should only ever be used during CI (+ maybe separate command for building wallet, etc)

common/mixnet-contract/src/mixnode.rs
```diff
@@ -1,18 +1,36 @@
...
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use std::cmp::Ordering;
 use std::fmt::Display;
-use ts_rs::TS;
 
 use crate::current_block_height;
 
-#[derive(Clone, Debug, Deserialize, PartialEq, PartialOrd, Serialize, JsonSchema, TS)]
+const DEFAULT_PROFIT_MARGIN: u8 = 10;
```
jstuczyn: Should it be here in or together with other defaults? (unless it's not a "network" default)

common/mixnet-contract/src/mixnode.rs
```diff
@@ -1,18 +1,36 @@
...
+
+fn default_alpha() -> Ratio<u128> {
+    Ratio::new(3u128, 10u128) // Sybil resistan parameter
```
jstuczyn: Why 3/10? It looks like a completely random value to me : ) I feel like this should be explained a bit more and maybe even defined as const or something (you can do `Ratio::new_raw()` in a const context. However, it might not be sufficient as it doesn't perform the reduction)

common/mixnet-contract/src/mixnode.rs
```diff
@@ -35,6 +53,100 @@ pub enum Layer {
     Three = 3,
 }
 
+#[derive(Debug, Clone, JsonSchema, PartialEq, Serialize, Deserialize, Copy)]
+pub struct NodeRewardParams {
+    income_global_mix: u128,
+    k: u128,
```
jstuczyn: I think it's what we discussed with @futurechimp , what is `k`? I feel like all those single-letter (whether English or Greek :) ) should have some explanation behind them.

common/mixnet-contract/src/mixnode.rs
```diff
@@ -35,6 +53,100 @@ pub enum Layer {
...
+            uptime,
+        }
+    }
+
+    pub fn performance(&self) -> Ratio<u128> {
+        Ratio::new(self.uptime, self.total_epoch_uptime)
+    }
+
+    pub fn operator_cost(&self) -> Ratio<u128> {
+        Ratio::new(self.uptime, (100 * DEFAULT_OPERATOR_EPOCH_COST) as u128)
```
jstuczyn: I haven't read much into the tokenomics paper, so maybe that's fine, but it feels like `operator_cost` shouldn't *always* be calculated based on `DEFAULT_OPERATOR_EPOCH_COST`. Is it not the case that individual mix operator could set it's own operator cost?

common/mixnet-contract/src/mixnode.rs
```diff
@@ -35,6 +53,100 @@ pub enum Layer {
...
+    pub fn alpha(&self) -> Ratio<u128> {
+        default_alpha()
+    }
+}
+
+#[derive(Debug)]
+pub struct NodeRewardResult {
+    reward: Ratio<u128>,
+    lambda: Ratio<u128>,
+    sigma: Ratio<u128>,
```
jstuczyn: Ah yes, lambda and sigma, we all know what they mean : )

common/mixnet-contract/src/mixnode.rs
```diff
@@ -44,6 +156,7 @@ pub struct MixNodeBond {
     #[serde(default = "current_block_height")]
```
jstuczyn: This is unrelated to your PR so feel free to ignore it, but isn't it the case we can remove the `#[serde(default = "current_block_height")]` macro at this point? From what I recall it was only added for a purposes of some migration we already performed while back.

common/mixnet-contract/src/mixnode.rs
```diff
@@ -79,6 +202,98 @@ impl MixNodeBond {
...
+
+    pub fn total_delegation(&self) -> Coin {
+        self.total_delegation.clone()
+    }
+
+    pub fn bond_to_total_stake(&self, total_stake: u128) -> Ratio<u128> {
+        Ratio::new(self.bond_amount().amount.u128(), total_stake)
+    }
+
+    pub fn stake_to_total_stake(&self, total_stake: u128) -> Ratio<u128> {
```
jstuczyn: The name feels kinda weird, but I'm not entirely sure how I would have renamed it myself either... Maybe something like `self_pledge_to_total_stake`? But I'm not sure whether that is even the correct terminology...

common/mixnet-contract/src/mixnode.rs
```diff
@@ -79,6 +202,98 @@ impl MixNodeBond {
...
+            * self.node_profit(params);
+
+        delegator_reward
+            .max(Ratio::new(0, 1))
+            .to_u128()
+            .ok_or(MixnetContractError::RatioToU128(delegator_reward))
+    }
 }
 
 impl PartialOrd for MixNodeBond {
```
jstuczyn, line 305: Should maybe the new `profit_margin_percent` field be used in PartialOrd? Unless it doesn't make any sense to compare against it.

contracts/mixnet/src/queries.rs
```diff
@@ -93,6 +94,18 @@ pub(crate) fn query_layer_distribution(deps: Deps) -> LayerDistribution {
...
 
+pub(crate) fn query_inflation_pool(deps: Deps) -> Uint128 {
+    inflation_pool_value(deps.storage)
+}
+
+pub(crate) fn query_total_mix_stake(deps: Deps) -> Uint128 {
+    total_mix_stake_value(deps.storage)
+}
+
+pub(crate) fn query_total_gt_stake(deps: Deps) -> Uint128 {
```
jstuczyn: could you rename `gt` to gateway?

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +53,107 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
     singleton_read(storage, CONFIG_KEY)
 }
 
+fn total_mix_stake(storage: &dyn Storage) -> ReadonlySingleton<Uint128> {
+    singleton_read(storage, TOTAL_MIX_STAKE_KEY)
+}
+
+pub fn mut_total_mix_stake(storage: &mut dyn Storage) -> Singleton<Uint128> {
```
jstuczyn: I kinda like the explicit `mut` in the function name. However, this is rather inconsistent with other functions that use the storage, i.e. for example we have `layer_distribution_read` for immutable access and `layer_distribution` for mutable one. Perhaps the old stuff should get renamed?

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +53,107 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+}
+
+pub fn mut_total_gateway_stake(storage: &mut dyn Storage) -> Singleton<Uint128> {
+    singleton(storage, TOTAL_GATEWAY_STAKE_KEY)
+}
+
+pub fn total_mix_stake_value(storage: &dyn Storage) -> Uint128 {
+    match total_mix_stake(storage).load() {
+        Ok(value) => value,
+        Err(_e) => Uint128(0),
```
jstuczyn: I'd argue that if we error out here, we should perhaps panic or something because isn't it the case that invalid `total_mix_stake` will mess up all rewarding?

jstuczyn: Same comment applies to `total_gateway_stake_value` and `inflation_pool_value`

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +53,107 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+    let stake = inflation_pool_value(storage).saturating_add(amount);
+    mut_inflation_pool(storage).save(&stake)?;
+    Ok(stake)
+}
+
+pub fn decr_inflation_pool(
+    amount: Uint128,
+    storage: &mut dyn Storage,
+) -> Result<Uint128, ContractError> {
+    // TODO: This could got to < 0
```
jstuczyn: So, should we deal with it somehow then?

contracts/mixnet/src/transactions.rs
```diff
@@ -111,6 +112,8 @@ pub(crate) fn try_add_mixnode(
     mixnodes(deps.storage).save(identity.as_bytes(), &bond)?;
     mixnodes_owners(deps.storage).save(sender_bytes, identity)?;
     increment_layer_count(deps.storage, bond.layer)?;
+    incr_total_mix_stake(bond.bond_amount().amount, deps.storage)?;
```
jstuczyn: Now looking at the call above it, should it rather have the full `increment_total_mix_stake` name?

contracts/mixnet/src/transactions.rs
```diff
@@ -443,6 +455,69 @@ pub(crate) fn try_reward_mixnode(
...
+
+    let reward_result = current_bond.reward(&reward_params);
+
+    // Omitting the price per packet function now, it follows that base operator reward is the node_reward
+    let operator_reward = Uint128(current_bond.operator_reward(&reward_params)?);
+
+    let total_delegation_reward =
+        increase_mix_delegated_stakes_v2(deps.storage, &current_bond, &reward_params)?;
+
+    // assert_eq!(total_delegation_reward.u128(), 1);
```
jstuczyn: Should the commented out assertion be there?

contracts/mixnet/src/storage.rs
```diff
@@ -214,6 +320,56 @@ pub(crate) fn increase_mix_delegated_stakes(
     Ok(total_rewarded)
 }
 
+pub(crate) fn increase_mix_delegated_stakes_v2(
```
jstuczyn, line 249: We definitely need to do something about this one asap...

tokenomics-py/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    k: f64,       // number of desired mixnodes - externally defined constant
+) -> f64 {
+    // set_lambda_sigma_mixnet, Network_econ:308
+    let lambda = (pledged / total_stake).min(1. / k);
+    let sigma = ((pledged + delegated) / total_stake).min(1. / k);
+
+    performance * income_global_mix * (sigma * omega_k + alpha * lambda) / (1. + alpha)
+}
+
+fn compute_mix_rewards(mix: &MixNodeBond) {
```
jstuczyn: Does it make sense to include this incomplete file?

validator-api/src/cache/mod.rs
```diff
@@ -177,6 +189,8 @@ impl ValidatorCache {
         &self,
         mut mixnodes: Vec<MixNodeBond>,
         mut gateways: Vec<GatewayBond>,
+        total_mix_stake: u128,
+        total_gt_stake: u128,
```
jstuczyn: can we use full `gateway` instead? : )

validator-api/src/rewarding/mod.rs
```diff
@@ -285,11 +312,29 @@ impl Rewarder {
                         uptime: self
                             .calculate_absolute_uptime(mix.last_day_ipv4, mix.last_day_ipv6),
                         total_delegations,
+                        params: None,
                     })
             })
             .filter(|node| node.uptime.u8() > 0)
             .collect();
 
+        let total_epoch_uptime = eligible_nodes
```
jstuczyn: Shouldn't `total_epoch_uptime` calculation also be behind the feature flag?

contracts/mixnet/src/storage.rs
```diff
@@ -24,11 +25,15 @@ use serde::Serialize;
...
 const CONFIG_KEY: &[u8] = b"config";
 const LAYER_DISTRIBUTION_KEY: &[u8] = b"layers";
+// Keeps total amount of stake towards mixnodes and gateways. Removing a bond removes all its delegations from the total, the reverse is true for adding a bond.
+const TOTAL_MIX_STAKE_KEY: &[u8] = b"total_mn";
+const TOTAL_GATEWAY_STAKE_KEY: &[u8] = b"total_gt";
+const INFLATION_POOL_PREFIX: &[u8] = b"pool";
 
 // buckets
-const PREFIX_MIXNODES: &[u8] = b"mn";
+pub const PREFIX_MIXNODES: &[u8] = b"mn";
```
jstuczyn, line 32: Any reason for those becoming public?

contracts/mixnet/Makefile
durch: Absolutley not, that has been a transition phase :)

common/mixnet-contract/src/mixnode.rs
durch: No preference, probably best to keep them all in the same place

tokenomics-py/src/lib.rs
durch: Its there for historical purposes, I'll nuke it

validator-api/src/rewarding/mod.rs
durch: Its not expensive, so I just left it there, it will become mainstream soon hopefully

contracts/mixnet/src/storage.rs
durch: Not sure, there might have been at some point in the dev cycle :)

contracts/mixnet/Makefile
durch: Thats a good point, I'll move this to the `validator-client-script` repo for reference, the .envrc is already there so it makes sense, and we can symlink it for internal usage

common/mixnet-contract/Cargo.toml
durch: Good point.

common/mixnet-contract/Cargo.toml
durch: For sure, great point, lets address it in a separate PR tough, this one is quite big already

contracts/mixnet/src/transactions.rs
durch: Nope!

contracts/mixnet/src/storage.rs
durch: We need it for the migration as its coded now.

common/mixnet-contract/src/mixnode.rs
neacsu: Yep, my bad, I'll remove it in a separate PR

common/mixnet-contract/src/mixnode.rs
durch: Not sure, don't think so, its whatever they set it is, might make sense in the network explorer tough

common/client-libs/gateway-client/src/error.rs
```diff
@@ -102,8 +102,12 @@ impl fmt::Display for GatewayClientError {
                 write!(f, "gateway returned an error response - {}", err)
             }
             GatewayClientError::UnexpectedResponse => write!(f, "received an unexpected response"),
-            GatewayClientError::NotEnoughBandwidth => {
-                write!(f, "client does not have enough bandwidth")
+            GatewayClientError::NotEnoughBandwidth((line_number, estimated, remaining)) => {
+                write!(
+                    f,
+                    "line: {} - client does not have enough bandwidth: estimated {}, remaining: {}",
```
neacsu: I do like the idea of having the estimated/remaining values in the error, but the `line!` seems more useful during debugging rather than for the final build.

common/client-libs/gateway-client/src/socket_state.rs
```diff
@@ -119,7 +119,7 @@ impl PartiallyDelegated {
             }
             .is_err()
             {
-                panic!("failed to send back `mixnet_receiver_future` result on the oneshot channel")
+                info!("failed to send back `mixnet_receiver_future` result on the oneshot channel")
```
neacsu: Maybe `warn!`?

common/mixnet-contract/src/mixnode.rs
```diff
@@ -35,6 +40,101 @@ pub enum Layer {
     Three = 3,
 }
 
+#[derive(Debug, Clone, JsonSchema, PartialEq, Serialize, Deserialize, Copy)]
+pub struct NodeRewardParams {
+    period_reward_pool: Uint128,
+    k: Uint128,
+    total_epoch_uptime: Uint128,
+    #[serde(default = "current_block_height")]
```
neacsu: FIY, I've just removed this function, as we probably only need it for migrations.

common/client-libs/gateway-client/src/error.rs
durch: You're right, we did that one in Crete, will undo :)

README.md
```diff
@@ -5,8 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 
 ## The Nym Privacy Platform
 
-This repository contains the Nym mixnet.
-
+This reposit
```
jstuczyn: You probably wanted to get rid of this entire sentence

README.md
```diff
@@ -33,6 +32,45 @@ There's a `.env.sample-dev` file provided which you can rename to `.env` if you
 
 You can chat to us in [Keybase](https://keybase.io). Download their chat app, then click **Teams -> Join a team**. Type **nymtech.friends** into the team name and hit **continue**. For general chat, hang out in the **#general** channel. Our development takes places in the **#dev** channel. Node operators should be in the **#node-operators** channel.
 
+### Rewards
+
+Node, node operator and delegator rewards are determined according to the principles laid out in the [Nym Tokenomics paper](needs-link). Below is a TLDR of the variables and formulas involved in calculating the epoch rewards. Initial reward pool is set to 250 million Nym, making the circulating supply 750 million Nym.
```
jstuczyn: I don't think we are going to be releasing the tokenomics paper anytime soon (?). But instead I think you can just say that it's laid out in section 6.5 of our whitepaper and link to that instead (https://nymtech.net/nym-whitepaper.pdf)

common/client-libs/gateway-client/src/client.rs
```diff
@@ -587,7 +590,11 @@ impl GatewayClient {
         }
         #[cfg(feature = "coconut")]
         if (mix_packet.sphinx_packet().len() as i64) > self.bandwidth_remaining {
-            return Err(GatewayClientError::NotEnoughBandwidth);
+            return Err(GatewayClientError::NotEnoughBandwidth((
+                line!(),
```
jstuczyn: Do we care about the `line!()` here in production code?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -626,7 +633,11 @@ impl GatewayClient {
         }
         #[cfg(feature = "coconut")]
         if self.bandwidth_remaining <= 0 {
-            return Err(GatewayClientError::NotEnoughBandwidth);
+            return Err(GatewayClientError::NotEnoughBandwidth((
+                line!(),
```
jstuczyn: Same thing about line

common/mixnet-contract/src/msg.rs
```diff
@@ -107,6 +114,10 @@ pub enum QueryMsg {
         address: Addr,
     },
     LayerDistribution {},
+    GetRewardPool {},
```
jstuczyn, line 85: Would it perhaps make sense to bundle all of those parameters together into a single struct (and thus single Query)?

common/mixnet-contract/src/error.rs
```diff
@@ -0,0 +1,9 @@
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum MixnetContractError {
+    #[error("Overflow Error")]
+    OverflowError(#[from] cosmwasm_std::OverflowError),
+    #[error("reward_blockstamp field not set, set_reward_blockstamp must be called before attempting to issue rewards")]
```
jstuczyn, line 7: It does not seem that `BlockstampNotSet` variant is being used anywhere

contracts/mixnet/src/contract.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+use std::u128;
```
jstuczyn, line 4: I was under impression that unless we compile with `no_std`, all of `std` things are available to us without explicit imports

contracts/mixnet/src/contract.rs
```diff
@@ -29,11 +31,19 @@ pub const INITIAL_GATEWAY_DELEGATION_REWARD_RATE: u64 = 110;
 pub const INITIAL_MIXNODE_ACTIVE_SET_SIZE: u32 = 100;
 pub const INITIAL_GATEWAY_ACTIVE_SET_SIZE: u32 = 20;
 
+pub const INITIAL_REWARD_POOL: u128 = 250_000_000_000_000;
```
jstuczyn, line 31: This comment is not really about the code, but something to remember during deployment (so that one more person is aware) - we need to actually put those tokens into the contract on initialisation

common/mixnet-contract/src/error.rs
jstuczyn: Actually the same is true for `OverflowError` and entire `MixnetContractError` enum

contracts/mixnet/src/error.rs
```diff
@@ -89,4 +89,10 @@ pub enum ContractError {
         identity: IdentityKey,
         address: Addr,
     },
+
+    #[error("Overflow error!")]
```
jstuczyn, line 80: It doesn't look like either of those variants are being constructed?

contracts/mixnet/src/storage.rs
```diff
@@ -48,6 +51,46 @@ pub fn config_read(storage: &dyn Storage) -> ReadonlySingleton<State> {
...
+}
+
+pub fn reward_pool_value(storage: &dyn Storage) -> Uint128 {
+    match reward_pool(storage).load() {
+        Ok(value) => value,
+        Err(_e) => Uint128(INITIAL_REWARD_POOL),
+    }
+}
+
+#[allow(dead_code)]
```
jstuczyn, line 67: Why is it a `dead_code`. is it not the case we should be incrementing reward pool at some points?

contracts/mixnet/src/transactions.rs
```diff
@@ -99,6 +99,7 @@ pub(crate) fn try_add_mixnode(
         layer,
         env.block.height,
         mix_node,
+        None,
```
jstuczyn, line 102: Why is profit margin always set to `None`? Shouldn't we require people to set their profit margins and just have the UI choose a default for them? (Something like the `advanced` section we currently have in the web wallet)

validator-api/src/rewarding/mod.rs
```diff
@@ -54,6 +55,19 @@ pub(crate) struct MixnodeToReward {
 
     /// Total number of individual addresses that have delegated to this particular node
     pub(crate) total_delegations: usize,
+    /// Node absolute uptime over total active set uptime
+    params: Option<NodeRewardParams>,
+}
+
+impl MixnodeToReward {
+    /// Somewhat clumsy way of feature gatting tokenomics payments. In a tokenomics scenario this will never be None at reward time. We levarage that to Into a different ExecuteMsg variant
```
jstuczyn, line 57: I'd perhaps argue that you shouldn't need to worry about `tokenomics` feature and you should just completely remove old rewards and replace it with your scheme.

README.md
jstuczyn: Or maybe just "section 6"

common/mixnet-contract/src/msg.rs
durch: I wouldn't, this way each query does one thing, makes it easier to reason about them, at least for me :)

contracts/mixnet/src/storage.rs
durch: Not sure, currently we only decrement as we distribute rewards, there are no paths that out return things to the reward pool ATM

contracts/mixnet/src/transactions.rs
durch: Its `None` for backward compatibility, so that this can run against the QA net and Milhon, without issue. Once we get to updating the mixnode bonding cli we can get rid of the Option

validator-api/src/rewarding/mod.rs
durch: Yeah, although I'd make it into a separate PR

durch: Created an issue for this

contracts/mixnet/src/transactions.rs
durch: Created an issue

PR #804, opened 2021-10-06
Map fix for colour coding nodes

Aid19801:
Issue became apparent that as number of nodes vs geolocation fluctates the scale (0-1000) that the colours were mapped to, became paler, more difficult to distinguish, little contrast between, for example "DEU": 36 and "POL": 1.

The colour scale is now dynamic with some logic checking the max/heighest number of nodes to use before defining the colours.

```
if (countryData?.data) {
      const heighestNumberOfNodes: number = countryData.data.reduce((a, b) => a.nodes > b.nodes ? a : b).nodes;
      const cs = scaleLinear()
        .domain([0, heighestNumberOfNodes]) // our data's FROM and TO
        .range(mode === 'dark' ? ['#FFD278', '#F20041'] : ['orange', 'red']) // the colours' FROM and TO that we map to.
      setColorScale(() => cs); // colorScale is stored in state and passed down to React Simple Maps.
    }
```


explorer/src/components/WorldMap.tsx
```diff
@@ -18,21 +16,25 @@ type MapProps = {
...
   React.useEffect(() => {
     if (userLocation || countryData) {
       setHasNoContent(false);
     }
+    if (countryData?.data) {
+      const heighestNumberOfNodes: number = countryData.data.reduce((a, b) => a.nodes > b.nodes ? a : b).nodes;
+      const cs: any = scaleLinear()
+        .domain([0, heighestNumberOfNodes])
+        // @ts-ignore
+        .range(mode === 'dark' ? ['#FFD278', '#F20041'] : ['orange', 'red']);
```
mmsinclair: I have to be a bit suspicious of a bit of code that has `any` and `ts-ignore` in it!

It seems like something that you're using doesn't have proper typings?

explorer/src/components/WorldMap.tsx
```diff
@@ -18,21 +16,25 @@ type MapProps = {
...
-    // @ts-ignore
-    .range(mode === 'dark' ? ['#ffedea', '#ff5233'] : ['orange', 'red']);
-
-
   React.useEffect(() => {
     if (userLocation || countryData) {
       setHasNoContent(false);
     }
+    if (countryData?.data) {
+      const heighestNumberOfNodes: number = countryData.data.reduce((a, b) => a.nodes > b.nodes ? a : b).nodes;
```
mmsinclair: What about using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max on the array of nodes per country?

```suggestion
      const heighestNumberOfNodes = Math.max(countryData.data.map(country => country.nodes));
```

explorer/src/api/constants.ts
```diff
@@ -2,6 +2,7 @@
 // master APIs
 export const MASTER_URL = 'https://testnet-milhon-explorer.nymtech.net';
 export const MASTER_VALIDATOR_URL = 'https://testnet-milhon-validator1.nymtech.net';
+export const MAP_TOPOJSON = 'https://raw.githubusercontent.com/zcreativelabs/react-simple-maps/master/topojson-maps/world-110m.json';
```
mmsinclair: Is this the outline vector for the world map?

If so, I feel it would be better to bundle (or make it a static asset) it because if that content moves / is deleted on GitHub the site could break. The values in there are constant and don't need to be dynamic.

mmsinclair (changes requested):

explorer/src/components/WorldMap.tsx
fmtabbara: If it's an array of values you're passing in you might have to do

`Math.max.apply(null, countryData.data.map(country => country.nodes))`

mmsinclair: True, I guess it could also be:

```
const heighestNumberOfNodes = Math.max(...countryData.data.map(country => country.nodes));
```

explorer/src/components/WorldMap.tsx
Aid19801: Yep, bit of googling necessary there. thanks. sorted.

explorer/src/components/WorldMap.tsx
Aid19801: Updated. thanks.

mmsinclair (approved):
lgtm

PR #805, opened 2021-10-06
Feature/coconut feature

jstuczyn:
Introduces `coconut` feature flag for:
- native client
- socks5 client
- gateway
- validator-api
that enables all coconut related functionalities. It is disabled by default.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
           args: --all
 
       - name: Run all tests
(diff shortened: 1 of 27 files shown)


neacsu (commented):
Where are all the `beta` and `nightly` tests?

jstuczyn:
And I guess the question is where is also the `stable` CI. And the answer is missing `'` in the `build.yml` that I've just introduced.

durch (approved):

PR #806, opened 2021-10-07
Bug fix: Network Explorer: Add freegeoip API key and split out tasks for country distributions 

mmsinclair:
This PR fixes up a bug in the country distribution tasks and hopefully improves the code structure by splitting out the tasks more.

- freegeoip requires an API key for 15,000 free requests per hour (and the API hostname has changed)
- country distribution is in a separate task that runs every 15 mins
- mix node cache refresh occurs every 10 minutes
- geolocation runs as a permanent task that picks off a mix node that hasn't been located (or where the location cache time has expired - 1 day) and is rate limited to run every 50ms (a variation on the suggestion by @futurechimp )
- error reporting has been improved by adding more `Error` types to the location `Result` to log when the explorer-api is being rate limited by freegoip and also when a mix node's IP address cannot be located (there are a few that can't be found)

Things that could be improved / ideas / concerns:
- the task timers (e.g. mix nodes refreshed every 10 mins, distribution every 15 mins, location cache TTL of 1 day) could be tweaked / fiddled with because I just picked some numbers out of the air
- the code structure can be improved, apologies in advance for my bad rusting
- the geolocation job could create a queue of the mix nodes to locate and pop them off, instead of the `for (...) { ... return; }` pattern (a throw back to my C++ days)

I am much happier with the location task picking things off every 50ms because it works better with the naive way we spawn tokio tasks. The service will actually stop now when a halt signal is sent.


explorer-api/src/country_statistics/distribution.rs
```diff
@@ -0,0 +1,57 @@
...
+                self.calculate_nodes_per_country().await;
+            }
+        });
+    }
+
+    /// Retrieves the current list of mixnodes from the validators and calculates how many nodes are in each country
+    async fn calculate_nodes_per_country(&mut self) {
+        let cache = self.state.inner.mix_nodes.get_location_cache().await;
+
+        let locations: Vec<&Location> = cache.values().flat_map(|i| i.location.as_ref()).collect();
```
jstuczyn: Any reason of constructing vec  of references rather than taking ownership of it? Because of it you end up having to clone `three_letter_iso_country_code` in line 39

explorer-api/src/country_statistics/distribution.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        // replace the shared distribution to be the new distribution
+        self.state
+            .inner
+            .country_node_distribution
+            .set_all(distribution)
+            .await;
+
+        info!(
```
jstuczyn, line 55: It seems like it would be a lot of bloat if left in `info!()` log statement - perhaps it should be a `trace` or `debug`?

explorer-api/src/country_statistics/geolocate.rs
```diff
@@ -0,0 +1,139 @@
...
+                    }
+
+                    self.state
+                        .inner
+                        .mix_nodes
+                        .set_location(&cache_item.mix_node.identity_key, Some(location))
+                        .await;
+
+                    // one node has been located, so break out of the loop
+                    return;
```
jstuczyn, line 70: This will not break out of the loop but return from the function. Presumably you meant to use `break` or `continue`?

explorer-api/src/country_statistics/geolocate.rs
```diff
@@ -0,0 +1,139 @@
...
+                }
+                Err(e) => match e {
+                    LocateError::ReqwestError(e) => warn!(
+                        "❌ Oh no! Location for {} failed {}",
+                        cache_item.mix_node.host, e
+                    ),
+                    LocateError::NotFound(e) => {
+                            error!(
```
jstuczyn: Is this really an `error!`? Usually `error` implies some big underlying issue with the program that might cause inevitable crash. If we failed to locate one node, it's not really that big of an issue?

explorer-api/src/mix_nodes/tasks.rs
```diff
@@ -0,0 +1,26 @@
...
+        info!("Spawning mix nodes task runner...");
+        tokio::spawn(async move {
+            let mut interval_timer = tokio::time::interval(std::time::Duration::from_secs(60 * 60)); // every hour
+            loop {
+                // wait for the next interval tick
+                interval_timer.tick().await;
+
+                info!("Updating mix node cache...");
+                self.state.inner.mix_nodes.refresh().await;
+                info!("Done");
```
jstuczyn, line 22: This is just a side note, but it looks like a perfect use case for `tracing` - https://tokio.rs/blog/2019-08-tracing. We're not really going to know what this `Done` refers to due to logs coming from different places. But I think `tracing` spans solve that issue. Anyway, just wanted to point out existence of that library (and that's something I wanted to check out myself at some point)

jstuczyn (changes requested):

explorer-api/src/country_statistics/distribution.rs
mmsinclair: I find it useful because when there have been problems I can see in the logs what has been happening over time to the distribution value. When it runs it looks like:

```
Oct 07 10:49:12 localhost network-explorer-api[1514245]:  2021-10-07T10:49:12.590Z INFO  explorer_api::country_statistics::distribution > Mixnode country distribution done: {"CHE": 36, "FRA": 92, "UKR": 49, "HKG": 15, "LTU": 8, "BGR": 7, "USA": 731, "CHN": 39, "SGP": 92, "POL": 39, "SYC": 9, "KOR": 45, "FIN": 595, "CZE": 34, "TWN": 7, "ESP": 10, "NLD": 874, "DEU": 1390, "RUS": 353, "GBR": 157, "CYP": 3, "KAZ": 3, "AUS": 18, "ITA": 1, "CAN": 58, "IND": 14, "JPN": 115, "AUT": 1, "MDA": 37, "LVA": 33, "TUR": 1}
```

I do appreciate that it is a little on the noisy side, but then maybe no more noisy than say rocket:

```
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.407Z INFO  rocket::server                                 > GET /mix-node/54p2tNwikcf5mBiy2YeDSAkjZ1PvsgHC4VNkTnjjiyn4/delegations:
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.408Z INFO  _                                              > Matched: (get_delegations) GET /mix-node/<pubkey>/delegations
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.422Z INFO  rocket::server                                 > GET /mix-node/54p2tNwikcf5mBiy2YeDSAkjZ1PvsgHC4VNkTnjjiyn4/description:
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.423Z INFO  _                                              > Matched: (get_description) GET /mix-node/<pubkey>/description
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.450Z INFO  rocket::server                                 > GET /mix-node/54p2tNwikcf5mBiy2YeDSAkjZ1PvsgHC4VNkTnjjiyn4/stats:
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.450Z INFO  _                                              > Matched: (get_stats) GET /mix-node/<pubkey>/stats
Oct 07 10:47:15 localhost network-explorer-api[1514245]:  2021-10-07T10:47:15.473Z INFO  rocket::server                                 > GET /ping/54p2tNwikcf5mBiy2YeDSAkjZ1PvsgHC4VNkTnjjiyn4:
```

explorer-api/src/country_statistics/geolocate.rs
mmsinclair: I want it to return out of the function. I'll change the comment to say that.

explorer-api/src/country_statistics/geolocate.rs
mmsinclair: Yeah, used to having `fatal` to do that. Happy to change to `warn`.

explorer-api/src/mix_nodes/tasks.rs
mmsinclair: Yup, makes sense. At the moment, these logs do not go anywhere. On a system that aggregated them, trace ids would be very useful.

explorer-api/src/country_statistics/distribution.rs
mmsinclair: I was having `Option<...>` and `flat_map` issues - have found a better way...

jstuczyn (approved):

durch (approved):

PR #807, opened 2021-10-07
Search & PageSize for DataGrid (Mixnodes)

Aid19801:
This PR seeks to add Search and Page Size selector to the Universal Data Grid component.


fmtabbara:
The page size selection is good.

I would change the approach to the way search is set up.

Keep the search functionality local to  the components involved i.e. DataGrid and TableToolBar. I don't think any of it needs to be in the context as it's not shared elsewhere in the app.

I think the architecture should be something like

1. Have a filteredMixnodes state in PageMixnodes
2. On search value change
  if value length is 0 set filteredMixNodes to mixnodes (from context)
  if  value length is greater than 0 use mixnodes.filter(...) to filter the mixnodes (from context) based on the search term
3. Before filteredMixnodes are set do your formatting of data

onChange will be something like

```
const onChange = (e) => {
 const filtered = mixnodes.filter(m => {
  if (
    m.locationName.toLowerCase().includes(e.target.value.toLowerCase()) || 
    m.identitiyKey.toLowerCase().includes(e.target.value.toLowerCase()) ||
    etc..
  ){
   return m
  }
  const formatted = filtered.map(m => format(map))
  setFiltered(formatted)
})
}

```

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,58 @@ const columns = [
...
 export const PageMixnodes: React.FC = () => {
   const { mixnodes } = useContext(MainContext);
-  const [rows, setRows] = React.useState();
+  const [filteredMixnodes, setFilteredMixnodes] = React.useState<MixNodeResponse | undefined>()
+  const [pageSize, setPageSize] = React.useState<string>("50");
+
+  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const st = event.target.value;
+    if (st === '' && mixnodes?.data) {
+      const formatted = mixnodeToGridRow(mixnodes?.data)
```
fmtabbara: We'll need a return type for this `mixnodeToGridRow` function as the shape of the data is changing. It will be something like this - 

```

type GridRowMixNode = {
  id: string;
  owner: string;
  location: string;
  identity_key: string;
  bond: number;
  host: string;
  layer: string;
};
```

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,58 @@ const columns = [
...
+      setFilteredMixnodes(formatted)
+    } else {
+      const filtered = mixnodes?.data && mixnodes?.data.filter((m) => {
```
fmtabbara: instead of && use the optional chaining operator again

`mixnodes?.data?.filter`

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,58 @@ const columns = [
 
 export const PageMixnodes: React.FC = () => {
   const { mixnodes } = useContext(MainContext);
-  const [rows, setRows] = React.useState();
+  const [filteredMixnodes, setFilteredMixnodes] = React.useState<MixNodeResponse | undefined>()
```
fmtabbara: This type will change to the one below. The shape of the data is being changed by the formatting function so no longer fits this type.

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,58 @@ const columns = [
...
+      <TableToolbar
+        onChangeSearch={handleSearch}
+        onChangePageSize={handlePageSize}
+        pageSize={pageSize}
+      />
       <UniversalDataGrid
         loading={mixnodes?.isLoading}
         columnsData={columns}
-        rows={rows}
+        rows={filteredMixnodes}
```
fmtabbara: had an idea! what if we make this
```
rows={mixnodeToGridRow(filteredMixnodes)}
```
Then you won't have to call `mixnodeToGridRow` to format the data every time you want to call `setFilteredMixnodes`
*filteredMixnodes can remain as the current type*
You can get rid of all the calls to format in your handleSearch function

explorer/src/pages/Mixnodes/index.tsx
fmtabbara: IGNORE THIS :arrow_double_up: 
Keep the type the same and see my comment below (3 comments down)

explorer/src/pages/Mixnodes/index.tsx
Aid19801: Optional chaining has been a bit playful for me. Will show you an example when I next encounter it.

explorer/src/pages/Mixnodes/index.tsx
Aid19801: Yes, bit of tweaking again, but works great. Thanks.

explorer/src/pages/Mixnodes/index.tsx
Aid19801: As discussed in priv chat, Optional Chaining issue killed; learned that u can add the O/Ch just before .filter() and .map() too. So everyday's a school day.

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,57 @@ const columns = [
...
 export const PageMixnodes: React.FC = () => {
   const { mixnodes } = useContext(MainContext);
-  const [rows, setRows] = React.useState();
+  const [filteredMixnodes, setFilteredMixnodes] = React.useState<MixNodeResponse>([])
+  const [pageSize, setPageSize] = React.useState<string>("50");
+
+  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const st = event.target.value;
+    if (st === '' && mixnodes?.data) {
+      // const formatted = mixnodeToGridRow(mixnodes?.data)
```
fmtabbara: remove commented code

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -46,24 +49,57 @@ const columns = [
...
+  };
 
   React.useEffect(() => {
-    if (mixnodes?.data !== undefined) {
-      const formattedRows = mixnodeToGridRow(mixnodes.data);
-      setRows(formattedRows);
-    };
-  }, [mixnodes])
+    if (mixnodes?.data) {
+      // const formatted = mixnodeToGridRow(mixnodes?.data)
```
fmtabbara: remove commented code

fmtabbara (approved):

PR #810, opened 2021-10-08
Feature/resending rewards on timeout

jstuczyn:


common/client-libs/validator-client/src/nymd/error.rs
```diff
@@ -98,3 +100,48 @@ pub enum NymdError {
...
+            NymdError::TendermintError(tm_err) => {
+                if tm_err.code() == Code::InternalError {
+                    // 0.34 (and earlier) versions of tendermint seemed to be using phrase "timed out waiting ..."
+                    // (https://github.com/tendermint/tendermint/blob/v0.34.13/rpc/core/mempool.go#L124)
+                    // while 0.35+ has "timeout waiting for ..."
+                    // https://github.com/tendermint/tendermint/blob/v0.35.0-rc3/internal/rpc/core/mempool.go#L99
+                    // note that as of the time of writing this comment (08.10.2021), the most recent version
+                    // of cosmos-sdk (v0.44.1) uses tendermint 0.34.13
+                    if let Some(data) = tm_err.data() {
+                        data.contains("timed out") || data.contains("timeout")
```
durch, line 116: Don't like this one bit, but looks like its the best we can do :/

durch (approved):

common/client-libs/validator-client/src/nymd/error.rs
jstuczyn: Me neither... me neither...

PR #812, opened 2021-10-09
Validators link to Big Dipper

Aid19801:
Part 1) The Content Card has an onClick which will bounce users no matter where they click on the card, this is probably the right ux (IMO)

On `blocks` we've passed down the anchor href wrapping the title, which is fine as there's no arrow/button icon there (though perhaps there should be)

On `validators`, wrapping the title would leave the button/icon outside of the clickable area.

So.... _either_ we pass down the same <a> for the icon **and** the title, or we use `onClick` for whole card - i've gone for the latter.

Part 2) Added Validators link to sidenav. Small changes to accept external link.

NOTES:

i) This solution will get a little more elegant when we bring validators api _into_ the app away from BigDipper; then we'll use `react-router-dom` to route internally rather `window.open`

ii) I have updated the _other_ title and icon/button cards (Mixnodes and Gateways) to take this approach also.


fmtabbara:
Would be good to avoid adding extra logic into the Nav if possible. An alternative way to handle the external link would be to keep things as they were, have a link to network-components/validators as usual and then let React Router pick up the rest of the work inside your network-components route component (bit of a mouthful :D).

```
    <Route
        path="/network-components/validators"
        component={(props: RouteComponentProps) => {
          window.open(BIG_DIPPER + '/validators');
          props.history.goBack();
          return null;
      }}
    />
```
This would open the Big Dipper validators page in a new window and return the user to the previous page in the network explorer.

explorer/src/icons/NetworksSVG.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const NetworkComponentsSVG = ({ mode }: WrapperProps) => {
+    const [ color, setColor ] = React.useState<string>('#000000');
+
+    React.useEffect(() => {
+        if (mode === 'dark') {
+            setColor('#FFFFFF');
+        } else {
+            setColor('#000000');
+        }
+    }, [mode])
```
fmtabbara: You could replace this with

````
const {mode} = useContext(GloablContext)
const color = mode === "dark" ? <color1> : <color2>
```

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -39,14 +39,24 @@ export const PageMixnodes: React.FC = () => {
     }
   }, [searchTerm, mixnodes?.data])
 
+  const linkStyles = {
+    color: 'inherit',
+    textDecoration: 'none',
+    marginLeft: 16
```
fmtabbara: Can we generate this margin from the mui theme? e.g. theme.spacing(2)?

Have you looked at mui's Link component? Not sure if that would work. If it does it would allow the styling to be done via sx prop?

explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -39,14 +39,24 @@ export const PageMixnodes: React.FC = () => {
...
   const columns = [
     {
       field: 'owner',
       headerName: 'Owner',
       width: 380,
       renderCell: (params: GridRenderCellParams) => {
         return (
-          <a href={`https://testnet-milhon-blocks.nymtech.net/account/${params.value}`} target='_blank' style={{ textDecoration: 'none', color: 'white', marginLeft: 16 }}>
+          <a
+            href={`https://testnet-milhon-blocks.nymtech.net/account/${params.value}`}
```
fmtabbara: Can this url go in the same place we are storing the other urls and be imported here?

explorer/src/routes/network-components.tsx
```diff
@@ -1,22 +1,33 @@
...
+      <Route path="/network-components/gateways">
+        <PageGateways />
+      </Route>
+      <Route
+        path="/network-components/validators"
+        component={(props: RouteComponentProps) => {
+          window.open(`${BIG_DIPPER}/validators`);
+          props.history.goBack();
+          return null;
+        }}
```
fmtabbara, line 26: Let's have a catch up with Mark about this part

explorer/src/components/Nav.tsx
```diff
@@ -112,51 +111,67 @@ const navOptions: navOptionType[] = [
...
 
 const ExpandableButton: React.FC<navOptionType> = ({
   nested,
   title,
-  Icon,
+  renderIcon,
   url,
 }) => {
   const [open, toggle] = React.useState(false);
+  const { mode } = React.useContext(MainContext);
```
fmtabbara: the svg icons can get this directly so no need to access here and pass as a prop

explorer/src/components/Nav.tsx
```diff
@@ -112,51 +111,67 @@ const navOptions: navOptionType[] = [
   {
     url: '/overview',
     title: 'Overview',
-    Icon: <BarChart />,
+    renderIcon: (mode) => <OverviewSVG mode={mode} />
```
fmtabbara: I would just call this Icon and pass through the component `<OverviewSVG />`. When I'm passing a component as a prop I usually uppercase the first letter of the prop name (in this can Icon). It helps to quickly identify props that are components. Just something I do and only a suggestion.

explorer/src/icons/NetworksSVG.tsx
Aid19801: Yep, much leaner thanks.

explorer/src/pages/Mixnodes/index.tsx
Aid19801: Have implemented MUI Link now. Didn't think we could get smooth routing (like RRD) if we used MUI but i can hand the RRD Link component to the MUI one. So all good now.

explorer/src/routes/network-components.tsx
Aid19801: Can do. i chatted briefly with him yesterday about it.

explorer/src/components/Nav.tsx
Aid19801: Yep. Now we're no longer passing the mode/dark etc via props (grabbing via context) my func loses its purpose. Have reverted. Cheers.

fmtabbara (approved):

PR #822, opened 2021-10-13
Change false to true, as for mixnodes

neacsu:


jstuczyn (approved):
Good catch. I already found and fixed it on my branch. It was very confusing at first 😂

PR #824, opened 2021-10-14
Flag to only run coconut-related functionalities

jstuczyn:
As requested by @mmsinclair


mmsinclair:
I'm happy with the result 😀 

@durch @neacsu ?

durch (approved):

PR #828, opened 2021-10-15
Updated CODEOWNERS

jstuczyn:


mmsinclair:
I think it is a good start 😀   

We should try it out and refine it as we find things that are annoying / unnecessary

mmsinclair (approved):

PR #831, opened 2021-10-19
update app name and icons

fmtabbara:
PR to update 

- Application name from `tauri-wallet` to `nym-wallet`
- Application icons from tauri to nym


tommyv1987 (approved):
Yep, looks good.

PR #832, opened 2021-10-19
Feature/bandwidth token

### LINKED ISSUE #2109: Integrate PoC into our current system
- [x] Bogdan talk to Dave re command line app
- [x] Integrate into socks5 client + native + network monitor
- [x] Integrate into gateway
- [x] Deploy cosmos contract on QA

neacsu:
Closes https://github.com/nymtech/nym/issues/2109

Remaining work:
- [x] Pass Ethereum endpoint and private key as arguments to clients
- [x] Pass Ethereum endpoint to gateways and check Eth events as well
- [x] Check for double spending against cosmos contract from gateway
- [x] Test against QA network


common/erc20-bridge-contract/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "erc20-bridge-contract"
+version = "0.1.0"
+edition = "2021"
```
jstuczyn: Let's keep our rust edition consistent, we're using 2018 everywhere else.

clients/client-core/src/config/mod.rs
```diff
@@ -111,6 +111,16 @@ impl<T: NymConfig> Config<T> {
         self.client.gateway_listener = gateway_listener.into();
     }
 
+    #[cfg(not(feature = "coconut"))]
+    pub fn with_eth_private_key<S: Into<String>>(&mut self, eth_private_key: S) {
```
jstuczyn, line 129: I will preface it by saying I'm going by changes from top to bottom so I might be missing the context of this change, but isn't it the case we didn't want to store user's ethereum keys? I think we just wanted to have a separate utility for obtaining the bandwidth credentials.

Also, if we really have to store them, I'd suggest storing filepaths rather than keys themselves in the config file

clients/native/src/client/mod.rs
```diff
@@ -212,7 +184,16 @@ impl NymClient {
 
         self.runtime.block_on(async {
             #[cfg(feature = "coconut")]
-            let coconut_credential = self.prepare_coconut_credential().await;
+            let bandwidth_controller = BandwidthController::new(
```
jstuczyn, line 187: This looks like a perfect use case for the `cfg!` macro

clients/native/src/commands/init.rs
```diff
@@ -54,7 +54,21 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
...
-        )
+        );
+    #[cfg(not(feature = "coconut"))]
+        let app = app
+        .arg(Arg::with_name("eth_endpoint")
+            .long("eth_endpoint")
+            .help("URL of an Ethereum full node that we want to use for getting bandwidth tokens from ERC20 tokens")
+            .takes_value(true)
+            .required(true))
+        .arg(Arg::with_name("eth_private_key")
```
jstuczyn, line 65: As I mentioned before, I'd rather not require it, but if it's strictly necessary use filepath or something because using a key directly is not obvious, especially if it's treated as a string.

common/client-libs/gateway-client/Cargo.toml
```diff
@@ -10,14 +10,19 @@ edition = "2018"
...
 # the entire crate
 futures = "0.3"
+json = "0.12.4"
 log = "0.4"
+thiserror = "1.0"
+url = "2.2"
 rand = { version = "0.7.3", features = ["wasm-bindgen"] }
 
 # internal
+credentials = { path = "../../credentials" }
```
jstuczyn, line 20: Shouldn't `credentials` be optional and feature locked behind coconut?

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    #[cfg(not(feature = "coconut"))]
+    pub async fn buy_token_credential(
+        &self,
+        verification_key: PublicKey,
+        signed_verification_key: Signature,
+    ) -> Result<(), GatewayClientError> {
+        let tokens_to_burn = BANDWIDTH_VALUE / TOKEN_BANDWIDTH_VALUE;
+        // 0 means a transaction failure, 1 means success
```
jstuczyn, line 161: So should those be defined as some constants / enums instead then?

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,160 @@
...
+        .expect("could not prepare out bandwidth credential for spending")
+    }
+
+    #[cfg(not(feature = "coconut"))]
+    pub async fn buy_token_credential(
+        &self,
+        verification_key: PublicKey,
+        signed_verification_key: Signature,
+    ) -> Result<(), GatewayClientError> {
+        let tokens_to_burn = BANDWIDTH_VALUE / TOKEN_BANDWIDTH_VALUE;
```
jstuczyn: This is an integer division, make sure rounding is happening the way you'd expect

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,160 @@
...
+        // 0 means a transaction failure, 1 means success
+        if Some(U64::from(0))
+            == self
+                .contract
+                .signed_call_with_confirmations(
+                    "burnTokenForAccessCode",
```
jstuczyn: This call (i.e. `burnTokenForAccessCode`) should definitely be defined in some constant for maintainability reasons

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,160 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use network_defaults::ETH_EVENT_NAME;
+
+    #[cfg(not(feature = "coconut"))]
+    #[test]
+    fn parse_contract() {
+        let transport =
+            Http::new("https://rinkeby.infura.io/v3/00000000000000000000000000000000").unwrap();
```
jstuczyn, line 212: That is something I just don't know, but can we make network calls in tests in our CI?

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,160 @@
...
+        json::parse(ETH_JSON_ABI)
+            .expect("Invalid json abi")
+            .dump()
+            .as_bytes(),
+    )
+    .expect("Invalid json abi")
+}
+
+#[derive(Clone)]
+pub struct BandwidthController {
```
jstuczyn, line 50: I really like this idea of `BandwidthController`. It hides all of the bandwidth-related feature juggling away

common/network-defaults/src/eth_contract.rs
```diff
@@ -0,0 +1,132 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+// This should be modified whenever an updated Ethereum contract is uploaded
+pub const ETH_JSON_ABI: &str = r#"
```
jstuczyn, line 5: This is more of a question for possible discussion: should this thing be hardcoded in the code or should it rather be a file we have in our repo?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -473,16 +459,13 @@ impl GatewayClient {
...
-        &mut self,
-        coconut_credential: Credential,
-    ) -> Result<(), GatewayClientError> {
-        if !self.authenticated {
-            return Err(GatewayClientError::NotAuthenticated);
-        }
-        if self.shared_key.is_none() {
-            return Err(GatewayClientError::NoSharedKeyAvailable);
-        }
+    async fn claim_coconut_bandwidth(&mut self) -> Result<(), GatewayClientError> {
```
jstuczyn: I think it makes more sense to pass credential as an argument to the method and actually obtaining it elsewhere.

common/client-libs/gateway-client/src/client.rs
```diff
@@ -502,7 +485,64 @@ impl GatewayClient {
         Ok(())
     }
 
-    #[cfg(feature = "coconut")]
+    #[cfg(not(feature = "coconut"))]
+    async fn claim_token_bandwidth(&mut self) -> Result<(), GatewayClientError> {
```
jstuczyn: Same thing I said about coconut credentials, I think it makes more sense if this function was only responsible for actually "claiming" bandwidth token you already have and moving responsibility of actually obtaining it elsewhere

common/client-libs/gateway-client/src/client.rs
```diff
@@ -624,10 +672,6 @@ impl GatewayClient {
         if !self.authenticated {
             return Err(GatewayClientError::NotAuthenticated);
         }
-        #[cfg(feature = "coconut")]
-        if self.bandwidth_remaining <= 0 {
```
jstuczyn, line 639: Isn't it the case that regardless of coconut we want to check for 0 bandwidth?

common/credentials/src/token/bandwidth.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use crypto::asymmetric::identity::{PublicKey, Signature, PUBLIC_KEY_LENGTH, SIGNATURE_LENGTH};
+
+use crate::error::Error;
+use std::convert::TryInto;
+
+#[cfg(not(feature = "coconut"))]
+pub struct TokenCredential {
```
jstuczyn, line 10: If I understood things correctly, this isn't a "Credential", it's just a bandwidth token

common/credentials/src/token/bandwidth.rs
```diff
@@ -0,0 +1,82 @@
...
+        self.verification_key
+    }
+
+    pub fn bandwidth(&self) -> u64 {
+        self.bandwidth
+    }
+
+    pub fn verify_signature(&self) -> bool {
+        let mut message = Vec::new();
+        message.append(&mut self.verification_key.to_bytes().to_vec());
```
jstuczyn: You don't need to do `to_vec()` conversions. And the `&mut` used is super suspicious.
You can do something like:
```rust
let message = self.verification_key.to_bytes().into_iter().chain(self.gateway_identity.to_bytes().into_iter()).collect();
```

common/credentials/src/token/bandwidth.rs
```diff
@@ -0,0 +1,82 @@
...
+        message.append(&mut self.gateway_identity.to_bytes().to_vec());
+        self.verification_key
+            .verify(&message, &self.signature)
+            .is_ok()
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut message = Vec::new();
+        message.append(&mut self.verification_key.to_bytes().to_vec());
```
jstuczyn: More `&mut` suspicion

common/credentials/src/token/bandwidth.rs
```diff
@@ -0,0 +1,82 @@
...
+        message.append(&mut self.gateway_identity.to_bytes().to_vec());
+        message.append(&mut self.bandwidth.to_be_bytes().to_vec());
+        message.append(&mut self.signature.to_bytes().to_vec());
+        message
+    }
+
+    pub fn from_bytes(b: &[u8]) -> Result<Self, Error> {
```
jstuczyn, line 73: Can you write some simple test that converts credential into bytes and then recovers it from bytes to make sure it's all fine?

common/erc20-bridge-contract/src/keys.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+// Serializable structures for what we find in common/crypto
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq, JsonSchema)]
+pub struct PublicKey([u8; 32]);
```
jstuczyn, line 9: Why are we redefining those structs?

common/network-defaults/src/lib.rs
```diff
@@ -63,6 +65,18 @@ pub fn default_api_endpoints() -> Vec<Url> {
 pub const DEFAULT_MIXNET_CONTRACT_ADDRESS: &str = "punk10pyejy66429refv3g35g2t7am0was7yalwrzen";
 pub const NETWORK_MONITOR_ADDRESS: &str = "punk1v9qauwdq5terag6uvfsdytcs2d0sdmfdy7hgk3";
 
+/// How much bandwidth (in bytes) one token can buy
+pub const TOKEN_BANDWIDTH_VALUE: u64 = 1024 * 1024 * 1024;
+/// Default bandwidth (in bytes) that we try to buy
+pub const BANDWIDTH_VALUE: u64 = 10 * TOKEN_BANDWIDTH_VALUE; // 10 GB
```
jstuczyn: What's the difference between `TOKEN_BANDWIDTH_VALUE` and `BANDWIDTH_VALUE`?

contracts/erc20-bridge/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[features]
+# for more explicit tests, cargo test --features=backtraces
+backtraces = ["cosmwasm-std/backtraces"]
+
+[dependencies]
+erc20-bridge-contract = { path = "../../common/erc20-bridge-contract" }
+
+# this branch is identical to 0.14.1 with addition of updated k256 dependency required to help poor cargo choose correct version
+cosmwasm-std = { git = "https://github.com/jstuczyn/cosmwasm", branch="0.14.1-updatedk256", features = ["iterator"] }
```
jstuczyn, line 35: Suggestion: assuming it compiles (because compiler might complain about dependency issues), since you're writing a brand new contract, try targetting cosmwasm 1.0 instead

gateway/Cargo.toml
```diff
@@ -27,22 +27,25 @@ tokio-stream = { version = "0.1", features = [ "fs" ] }
 tokio-tungstenite = "0.14"
```
jstuczyn, line 29: As discussed on keybase I'm not touching the gateway directory as "here be dragons" : )

validator-api/src/config/mod.rs
```diff
@@ -150,6 +150,14 @@ pub struct NetworkMonitor {
     /// packets before declaring nodes unreachable.
     #[serde(with = "humantime_serde")]
     packet_delivery_timeout: Duration,
+
+    /// Ethereum private key.
+    #[cfg(not(feature = "coconut"))]
+    eth_private_key: String,
```
jstuczyn, line 159: Why would validator api need ethereum key?

jstuczyn: Ah. Right. For bandwidth for testing...
Uhm... We need to discuss it with @futurechimp on how to deal with this issue

clients/client-core/src/config/mod.rs
neacsu: We _could_ not store them, but I find it to make more sense for the user to just put it there once at client initialization and not touch it again. They will still be stored on the user's machine, and we could even recommend users to use an account especially for this, but I wouldn't go as far as trying to secure this, as it's essentially an external secret.

clients/native/src/client/mod.rs
neacsu: That's what I initially thought as well, but it doesn't fit here since the `new` function is different depending on the coconut feature.

clients/native/src/commands/init.rs
neacsu: I tried to not over-complicate it by storing it in a file. Using a flow similar to how we write the other keys to `.pem` files would have meant to pollute even more the codebase with this bridging feature. Is there a specific reason on why we should be using a separate file for this?

common/client-libs/gateway-client/Cargo.toml
neacsu: I used this crate for credentials in general, adding the `token` module to it, for ERC20 credentials, so the feature lock is no longer necessary.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: I don't really see too much benefit of wrapping external library return codes in our own library, since we only check it once and don't touch it anymore. The external library (`web3`) should have used some enum wrapper for that imo, if they saw necessary.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: I'll add another constant (`TOKENS_TO_BURN`), just to avoid these ambiguities.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: Done.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: I'm not so sure this does any network call whatsoever, as the string taken by `new` seems to only do some checking that there is a base to the URL.

common/network-defaults/src/eth_contract.rs
neacsu: The reason I prefer it to being hardcoded in Rust code is that it can easily be compiled in on any platform, without having to mess around with OS file systems.

common/client-libs/gateway-client/src/client.rs
neacsu: Done

common/client-libs/gateway-client/src/client.rs
neacsu: Done

common/client-libs/gateway-client/src/client.rs
neacsu: I think this was an early error on my part, when I first introduced this check here. Mixnet messages are already taken into account on the sender part and shouldn't be counted here as well imo. This is a more philosophical question of whether bandwidth should be consumed on both upload and download, or only on one of those. Right now, I've put it for upload only (i.e. sending messages to mixnet), but we can easily change it if we want to.

common/credentials/src/token/bandwidth.rs
neacsu: Not sure what you mean by that, it is used to access the network and it is linked to the client accessing it. Do you think `BandwidthToken` would be more suitable?

common/credentials/src/token/bandwidth.rs
neacsu: Nice suggestion, I'll change it. Thanks!

common/credentials/src/token/bandwidth.rs
neacsu: Fixed it.

common/credentials/src/token/bandwidth.rs
neacsu: Sure, will do.

common/erc20-bridge-contract/src/keys.rs
neacsu: To be able to derive `Serialize`, `Deserialize`, `JsonSchema` and not have to import the `crypto` crate into the contract.

common/network-defaults/src/lib.rs
neacsu: My poor naming strikes again :) . I'll give it another shot: how does renaming `TOKEN_BANDWIDTH_VALUE` to `BYTES_PER_TOKEN` sound? In any case, I've changed things a bit, so `TOKEN_BANDWIDTH_VALUE`/`BYTES_PER_TOKEN` won't even need to be public now anyway.

contracts/erc20-bridge/Cargo.toml
neacsu: Tried it, it builds alright, but because it needs wasmd 0.20 to run, I will leave this change for the moment until we update the validators as well.

gateway/Cargo.toml
neacsu: Gateway directory is ready now as well :)

validator-api/src/config/mod.rs
neacsu: Yes, this is still an open question, so I'm just going to leave it like this until we settle on the access of "special" nodes.

.github/workflows/erc20_bridge_contract.yml
```diff
@@ -0,0 +1,44 @@
...
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          target: wasm32-unknown-unknown
+          override: true
+          components: rustfmt, clippy
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --manifest-path contracts/erc20-bridge/Cargo.toml --target wasm32-unknown-unknown
```
jstuczyn, line 30: I'd perhaps slightly change the build command to what we would actually use in production, i.e. `RUSTFLAGS='-C link-arg=-s' cargo build --release --target wasm32-unknown-unknown`
I doubt it changes anything in the context of CI though.

.github/workflows/wasm_client_build.yml
```diff
@@ -16,10 +16,11 @@ jobs:
           override: true
           components: rustfmt, clippy
 
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --manifest-path clients/webassembly/Cargo.toml --target wasm32-unknown-unknown
+# token credentials (non-coconut) don't work for wasm right now
```
jstuczyn, line 25: If we cannot even build the wasm client, I'd perhaps put big fat `panic!()` inside the wasm client code so that nobody would attempt to use it

clients/native/src/client/config/template.rs
```diff
@@ -42,6 +42,12 @@ public_encryption_key_file = '{{ client.public_encryption_key_file }}'
 # sent but not received back.
 reply_encryption_key_store_path = '{{ client.reply_encryption_key_store_path }}'
 
+# Ethereum private key.
+eth_private_key = '{{ client.eth_private_key }}'
```
jstuczyn, line 51: If those only exist in config for `not(coconut)` feature, will it not blow up if you attempt to generate and/or load config **with** coconut?

clients/native/src/commands/init.rs
jstuczyn: I just feel a bit uneasy and find it kinda weird to store directly in a config file. Also because our config is a text file, it must contain the string representation of said key. Should we use base64? Base58? Something else? It's not super obvious. And if it was stored in separate file, we could either go pure binary or use some predefined standard (as with `.pem` file). Anyway, those are my two cents. If you feel like you strongly prefer your current solution, let's just stick with it.

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let message: Vec<u8> = verification_key
+            .to_bytes()
+            .iter()
+            .chain(gateway_identity.to_bytes().iter())
+            .copied()
+            .collect();
+        let signature = kp.private_key().sign(&message);
+
+        Ok(TokenCredential::new(
```
jstuczyn, line 147: Would it be worth to also keep track of the private key we used for credential generation?
Consider the following:
- you run this function, i.e. purchase bandwidth credential and want to use it for some gateway
- however, you change your mind and dont use it yet (but credential is still valid!)
- then you come back and finally decide it's time to use the nym network!
- however, you gateway no longer exists! : (
- and now you're unable to "bind" this credential to anyone else as you threw away your private key.
- therefore my proposition is to only throw away the private key once you actually spent the credential

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,190 @@
...
+                .contract
+                .signed_call_with_confirmations(
+                    ETH_BURN_FUNCTION_NAME,
+                    (
+                        U256::from(TOKENS_TO_BURN),
+                        U256::from(&verification_key.to_bytes()),
+                        Bytes(signed_verification_key.to_bytes().to_vec()),
+                    ),
+                    Options::default(),
+                    1,
```
jstuczyn: This is something we definitely need to change. Waiting only for a single confirmation is a recipe for disaster in ethereum world. I don't remember the exact number, but I think the recommended value was somewhere around 7-12. We should perhaps also put it into some constant (or maybe into config?) to control it

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,190 @@
...
+                    &self.eth_private_key,
+                )
+                .await?
+                .status
+        {
+            Err(GatewayClientError::BurnTokenError(
+                web3::Error::InvalidResponse(String::from("Transaction status is 0 (failure)")),
```
jstuczyn: If transaction failed, maybe we could also include logs in the error message? (it's one of the fields in `TransactionReceipt`)

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,190 @@
...
+            ))
+        } else {
+            Ok(())
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
jstuczyn, line 206: Clippy doesn't like this one : ) It says it's unused

common/client-libs/gateway-client/src/bandwidth.rs
```diff
@@ -0,0 +1,190 @@
...
+    use network_defaults::ETH_EVENT_NAME;
```
jstuczyn, line 207: Same for this one

common/client-libs/gateway-client/src/client.rs
```diff
@@ -69,20 +71,19 @@ impl GatewayClient {
         mixnet_message_sender: MixnetMessageSender,
         ack_sender: AcknowledgementSender,
         response_timeout_duration: Duration,
+        bandwidth_controller: Option<BandwidthController>,
```
jstuczyn, line 74: Under what condition would we want to pass a `None` here? Isn't it the case we need it for both ERC20 and coconut bandwidths?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -517,9 +553,13 @@ impl GatewayClient {
         if !self.authenticated {
             return Err(GatewayClientError::NotAuthenticated);
         }
-        #[cfg(feature = "coconut")]
-        if self.estimate_required_bandwidth(&packets) < self.bandwidth_remaining {
-            return Err(GatewayClientError::NotEnoughBandwidth);
+        if self.estimate_required_bandwidth(&packets) > self.bandwidth_remaining {
```
jstuczyn, line 564: This is just random idea for the future. We definitely don't want to do it now, but I'm curious to know what you think about it. Rather than claiming additional bandwidth when we completely run out, maybe we should start some tokio task to claim additional bandwidth if we're below some threshold?

jstuczyn: But a thing we might want to do now is introduce some additional flag (or value) to config file indicating whether user wants to buy more bandwidth if they run out. Basically some kind of limit. Maybe they want to use at most (for argument sake) 10GB and never purchase more during single session? What do you think?

common/client-libs/validator-client/src/nymd/cosmwasm_client/signing_client.rs
```diff
@@ -485,7 +485,7 @@ pub trait SigningCosmWasmClient: CosmWasmClient {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
jstuczyn: This is kinda a warning flag for me -> under what condition would we want to clone our cosmwasm client rather than keep a single instance, i.e. we probably want to only have a single connection to the blockchain from particular binary?

common/credentials/src/token/bandwidth.rs
jstuczyn: I'm trying to remember what I meant by this when I wrote my comment haha. I think I meant that "Credential" might not be the best name for it, but naming things is hard... I'd say go with whichever you prefer more, either `TokenCredential`, `BandwidthToken` or something completely different.

common/erc20-bridge-contract/src/keys.rs
jstuczyn: I'm not the biggest fan of that, but I guess that works for now...

common/erc20-bridge-contract/src/keys.rs
```diff
@@ -0,0 +1,45 @@
...
+impl Signature {
+    pub fn new(bytes: [u8; 64]) -> Self {
+        let mut sig1 = [0u8; 32];
+        let mut sig2 = [0u8; 32];
+        sig1.copy_from_slice(&bytes[..32]);
+        sig2.copy_from_slice(&bytes[32..]);
+
+        Signature(sig1, sig2)
+    }
+    pub fn as_bytes(&self) -> [u8; 64] {
```
jstuczyn: This is a bit nit-picky, so feel free to ignore it, but usually `as_...` imply you will get a reference return type (in that case either `&[u8; 64]` or `&[u8]`) What you probably wanted to use was `to_bytes()`

common/erc20-bridge-contract/src/keys.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+// Serializable structures for what we find in common/crypto
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq, JsonSchema)]
+pub struct PublicKey([u8; 32]);
+
+impl PublicKey {
+    pub fn new(bytes: [u8; 32]) -> Self {
```
jstuczyn, line 12: We should probably do some verification of received data here, because not every set of 32 bytes is a valid ed25519 public key

contracts/erc20-bridge/src/queries.rs
```diff
@@ -0,0 +1,191 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use cosmwasm_std::{Deps, Order, StdResult};
+
+use crate::storage::payments_read;
+use erc20_bridge_contract::keys::PublicKey;
+use erc20_bridge_contract::payment::{PagedPaymentResponse, Payment};
+
+const BOND_PAGE_MAX_LIMIT: u32 = 100;
```
jstuczyn: Bond? : )

contracts/erc20-bridge/src/transactions.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    if !deps
+        .api
+        .ed25519_verify(&message, &signature, &verification_key)
+        .map_err(|_| ContractError::ParseSignatureError)?
+    {
+        return Err(ContractError::BadSignature);
+    }
+
+    status_bucket.save(&verification_key, &Status::Unchecked)?;
```
jstuczyn, line 38: Just to confirm, right now we don't have anything to change the status to either `Checked` or `Spent`, right?

gateway/src/config/template.rs
```diff
@@ -34,6 +34,9 @@ private_sphinx_key_file = '{{ gateway.private_sphinx_key_file }}'
 # Path to file containing public sphinx key.
 public_sphinx_key_file = '{{ gateway.public_sphinx_key_file }}'
 
+# Addess to an Ethereum full node.
+eth_endpoint = '{{ gateway.eth_endpoint }}'
```
jstuczyn, line 38: Similar issue as with both clients, are we sure it won't blow up due to feature locking action?

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+                Burned::from_tokens(log.params.into_iter().map(|x| x.value).collect::<Vec<_>>())?;
+            if burned_event.verify(verification_key) {
+                return Ok(());
+            }
+        }
+
+        Err(RequestHandlingError::InvalidBandwidthCredential)
+    }
+
+    pub(crate) async fn verify_double_spending(
```
jstuczyn: I'd perhaps rename it as what it's doing is actually claiming the token for itself rather than just checking for double spending

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+use validator_client::nymd::{
+    AccountId, Denom, Fee, Gas, SigningCosmWasmClient, SigningNymdClient,
+};
+
+#[derive(Clone)]
+pub(crate) struct ERC20Bridge {
+    // This is needed because web3's Contract doesn't sufficiently expose it's eth interface
+    web3: Web3<Http>,
+    contract: Contract<Http>,
+    nymd_client: SigningNymdClient,
```
jstuczyn: To make your life easier you'd probably want a `NymdClient<SigningNymdClient>` here

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+        let client_address = wallet
+            .try_derive_accounts()
+            .expect("Could not get account from wallet")
+            .into_iter()
+            .map(|account| account.address().clone())
+            .collect::<Vec<AccountId>>()
+            .get(0)
+            .expect("No account for wallet")
+            .clone();
+        let nymd_client = SigningNymdClient::connect_with_signer(nymd_url.as_str(), wallet)
```
jstuczyn: And then you wouldn't need to create wallet or try to derive address from it. You could just do `NymdClient::connect_with_mnemonic()`

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        ERC20Bridge {
+            contract: eth_contract(web3.clone()),
+            web3,
+            nymd_client,
+            client_address,
+        }
+    }
+
+    pub(crate) async fn verify_eth_events(
```
jstuczyn, line 57: There should probably be some setting to ensure events are at least X blocks old

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        Err(RequestHandlingError::InvalidBandwidthCredential)
+    }
+
+    pub(crate) async fn verify_double_spending(
+        &self,
+        credential: &TokenCredential,
+    ) -> Result<(), RequestHandlingError> {
+        // It's ok to unwrap here, as the cosmos contract and denom are set correctly
+        let contract_address = AccountId::from_str(COSMOS_CONTRACT_ADDRESS).unwrap();
```
jstuczyn: contract address is defined on the nymdclient I suggested using : )

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
```diff
@@ -0,0 +1,210 @@
...
+        let coin = Coin {
+            denom: Denom::from_str(DENOM).unwrap(),
+            amount: Decimal::from(100000u64),
+        };
+        let fee = Fee::from_amount_and_gas(coin.clone(), Gas::from(500000));
+        let req = ExecuteMsg::LinkPayment {
```
jstuczyn, line 109: Perhaps to make it slightly nicer, you could move transaction generation (and fee estimation) to nymd client?

validator-api/src/network_monitor/monitor/sender.rs
```diff
@@ -73,8 +72,7 @@ struct FreshGatewayClientData {
     // SECURITY:
     // since currently we have no double spending protection, just to get things running
```
jstuczyn: Well, now we DO have double spending protection, no?

clients/native/src/commands/init.rs
neacsu: One other solution is not storing that key at all and instead asking the user to input the private key every time the client is started, and not asking it in the `init` phase. It makes starting the client more annoying, but maybe it makes the user more conscious about the eth transactions that happen, which in the end is still a good thing.

.github/workflows/erc20_bridge_contract.yml
neacsu: The `--release` part should be used for all the other binaries anyway, so I don't think it makes much sense to include it (only) here. But the linking flag might be worth it. I'll include that in the workflow here, as well as for the mixnet contract while at it.

.github/workflows/wasm_client_build.yml
neacsu: What do you mean by putting a `panic()`? So that it doesn't work for the coconut credentials either?

clients/native/src/client/config/template.rs
neacsu: Why would you try that? It would be a different binary, with a different release version. When we'll switch from this version to the coconut version, an `update` will probably need to be written.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: Yes, this is the next step I'm working on, storing the private key somewhere such that, in case of your scenario (which I think is on some error path probably), there is still a way to recover from it. I'll do that in a separate PR though.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: I'll set it to 7 for release builds and 1 for debug builds, so that we don't wait forever to test stuff :D

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: This would be a good thing, in theory, but from what I've observed, the it looks like the logs are populated when the transaction is a success. On error, it's actually just an empty vector of `Log`... Maybe this is just for the error I've encountered (when my eth account didn't have any ERC20 tokens) so I'll include the logs in the error message, but I wouldn't get my hopes up.

common/client-libs/gateway-client/src/bandwidth.rs
neacsu: Missed that, it's fixed now. Thanks

common/client-libs/gateway-client/src/client.rs
neacsu: We do indeed. This is to handle the wasm client which uses this code too. Because the bandwidth controller is not gated by the `coconut` feature anymore, this `Option` substitutes that.

common/client-libs/gateway-client/src/client.rs
neacsu: Those are all very good points. I 100% support implementing them in the future, especially since they are valid for both types of credentials. I'll put up a user story to back it up: https://github.com/orgs/nymtech/projects/1#card-72234182

clients/native/src/client/config/template.rs
jstuczyn: What I meant is that this particular template is used regardless of `coconut` feature status. However, `client.eth_private_key` does not exist in the `coconut` case, hence it will try to include it during config generation and will fail

.github/workflows/wasm_client_build.yml
jstuczyn: Oh, if it works in the `coconut` case, then ignore it

clients/native/src/client/config/template.rs
neacsu: I understand what you mean now. `init`ing and `run`ing with `coconut` enabled works just fine with this template from what I've tested, so we should be good.

common/client-libs/validator-client/src/nymd/cosmwasm_client/signing_client.rs
neacsu: I think I overkilled it here. I've replaced the `Clone` requirement with some `Arc` wrappings, which was all I really needed anyway.

common/erc20-bridge-contract/src/keys.rs
neacsu: Renamed it, thanks.

common/erc20-bridge-contract/src/keys.rs
neacsu: I don't think this is the place to check for it though. This is just some dumb wrapper that helps sending bytes over the wire. There is a check that is done in the cosmos contract, in `link_payment`, when `ed25519_verify` is invoked. I changed a bit that error path now, so that it's more verbose.

contracts/erc20-bridge/src/queries.rs
neacsu: Ups. Renaming in progress :)

contracts/erc20-bridge/src/transactions.rs
neacsu: Nope, that should be done when we integrate into the payment system, right now we only make sure we store the data for when that happens.

gateway/src/config/template.rs
neacsu: Tested it on the gateway as well, it works just fine, these fields are just left empty.

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: Renamed it to `claim_token`

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: That simplified it a lot, thanks man.

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: Yep, just did that.

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: I've put it in place, depending on the type of build (debug or release)

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: Yep, switched to that

validator-api/src/network_monitor/monitor/sender.rs
neacsu: Just for the non-coconut bandwidth afaik. I'll update the comment.

gateway/src/node/client_handling/websocket/connection_handler/eth_events.rs
neacsu: That `NymdClient` looks more like a specific mixnet client, which is the reason I didn't try using it in the first place. But given that it simplifies the whole setup, I don't see any harm in re-using the `execute` part of it for this contract as well. However, I would prefer to avoid merging both contract APIs into it, especially since this code is meant to go away at some point.

clients/native/src/commands/init.rs
neacsu: As per an offline talk, we are going to go with the current solution until we have a proper bridge.

clients/client-core/src/config/mod.rs
```diff
@@ -22,7 +22,10 @@ const DEFAULT_MESSAGE_STREAM_AVERAGE_DELAY: Duration = Duration::from_millis(20)
 const DEFAULT_AVERAGE_PACKET_DELAY: Duration = Duration::from_millis(50);
 const DEFAULT_TOPOLOGY_REFRESH_RATE: Duration = Duration::from_secs(5 * 60); // every 5min
 const DEFAULT_TOPOLOGY_RESOLUTION_TIMEOUT: Duration = Duration::from_millis(5_000);
-const DEFAULT_GATEWAY_RESPONSE_TIMEOUT: Duration = Duration::from_millis(5_500);
+// Set this to a high value for now, so that we don't risk sporadic timeouts that might cause
+// bought bandwidth tokens to not have time to be spent; Once we remove the gateway from the
+// bandwidth bridging protocol, we can come back to a smaller timeout value
+const DEFAULT_GATEWAY_RESPONSE_TIMEOUT: Duration = Duration::from_secs(5 * 60);
```
jstuczyn, line 28: I guess the question is whether the underlying OS won't close the connection without any packets sent on it...

jstuczyn (approved):

clients/client-core/src/config/mod.rs
neacsu: If there are signs that this happens (I don't know exactly the timeouts for different OSes), we can lower the Ethereum block depth we require, from the current value of 7 to something else. On my machine, there is no timeout even in release mode, which awaits ~120s for the event.

jstuczyn: We can't really lower it as 7 is already a bare minimum for some safety.
The timeout I was talking about is something I've been experiencing with the network monitor -> the gateway connection would time out after a while if no messages were sent on it.

durch (approved):

PR #837, opened 2021-10-20
Feature/vouchers

aniampio:
* Migrate the updated coconut library into nym (called nymcoconut)
* Update the show protocol to include the serial number for the bandwidth credential 
* Create the struct of bandwidth credential with the attribute fields which we should use


mmsinclair:
@jstuczyn this is ready for you to look at again 😀

mmsinclair:
Linked to https://github.com/nymtech/nym/issues/2121

common/credentials/src/utils.rs
```diff
@@ -99,38 +112,64 @@ pub async fn obtain_aggregate_signature(
...
+            params,
+            public_attributes,
+            private_attributes,
+            &client,
+            &validator_partial_vk.key,
+        )
+        .await?;
         let share = SignatureShare::new(signature, id as u64);
         shares.push(share)
     }
```
neacsu: I might be missing something, but why are we treating the first signature separately?

common/nymcoconut/src/elgamal.rs
```diff
@@ -0,0 +1,401 @@
...
+    fn try_from(bytes: &[u8]) -> Result<Ciphertext> {
+        if bytes.len() != 96 {
+            return Err(CoconutError::Deserialization(format!(
+                "Ciphertext must be exactly 96 bytes, got {}",
+                bytes.len()
+            )));
+        }
+
+        let c1_bytes: &[u8; 48] = &bytes[..48].try_into().expect("Slice size != 48");
+        let c2_bytes: &[u8; 48] = &bytes[48..].try_into().expect("Slice size != 48");
```
neacsu: I guess `unwrap`ing would be enough, as there is a check that's done above anyway

common/nymcoconut/src/elgamal.rs
```diff
@@ -0,0 +1,401 @@
...
+    //     assert_eq!(decoded.0, ciphertext.0);
+    //     assert_eq!(decoded.1, ciphertext.1);
+    //
+    //     // it can also be deserialized directly from the raw bytes
+    //     let raw_bytes = ciphertext.to_bytes();
+    //
+    //     let decoded_raw: Ciphertext = bincode::deserialize(&raw_bytes).unwrap();
+    //     assert_eq!(decoded_raw.0, ciphertext.0);
+    //     assert_eq!(decoded_raw.1, ciphertext.1);
+    // }
```
neacsu: Will we need this test in the future? If not, I suggest we just remove it altogether.

common/nymcoconut/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+    DeserializationMinLength { min: usize, actual: usize },
+    #[error("Tried to deserialize {object} with bytes of invalid length. Expected {actual} < {} or {modulus_target} % {modulus} == 0")]
+    DeserializationInvalidLength {
+        actual: usize,
+        target: usize,
+        modulus_target: usize,
+        modulus: usize,
+        object: String,
+    },
+}
```
neacsu: Some spacing would make it easier to read:
```suggestion
    #[error("Setup error: {0}")]
    Setup(String),
    
    #[error("encountered error during keygen")]
    Keygen,

    #[error("Issuance related error: {0}")]
    Issuance(String),

    #[error("Tried to prepare blind sign request for higher than specified number of attributes (max: {}, requested: {})", max, requested)]
    IssuanceMaxAttributes { max: usize, requested: usize },

    #[error("Interpolation error: {0}")]
    Interpolation(String),
    
    #[error("Aggregation error: {0}")]
    Aggregation(String),

    #[error("Unblind error: {0}")]
    Unblind(String),
    
    #[error("Verification error: {0}")]
    Verification(String),

    #[error("Deserialization error: {0}")]
    Deserialization(String),
    
    #[error(
        "Deserailization error, expected at least {} bytes, got {}",
        min,
        actual
    )]
    DeserializationMinLength { min: usize, actual: usize },

    #[error("Tried to deserialize {object} with bytes of invalid length. Expected {actual} < {} or {modulus_target} % {modulus} == 0")]
    DeserializationInvalidLength {
        actual: usize,
        target: usize,
        modulus_target: usize,
        modulus: usize,
        object: String,
    },
}
```

common/credentials/src/bandwidth.rs
```diff
@@ -8,40 +8,84 @@
...
+pub const SERIAL_NUMBER_LEN: usize = 47;
+pub const BINDING_NUMBER_LEN: usize = 47;
+pub const VOUCHER_INFO_LEN: usize = 47;
 
-    let params = Parameters::new(TOTAL_ATTRIBUTES)?;
+pub struct BandwidthVoucherAttributes {
+    pub serial_number: PrivateAttribute,
+    pub binding_number: PrivateAttribute,
+    pub voucher_value: PublicAttribute,
+    pub voucher_info: PublicAttribute,
```
neacsu: Some documentation would be really useful here, I don't really understand what each of these values represents.

common/credentials/src/utils.rs
```diff
@@ -99,38 +112,64 @@ pub async fn obtain_aggregate_signature(
...
+        params,
+        verification_key,
+        signature,
+        serial_number,
+        binding_number,
+    )?;
 
     Ok(Credential::new(
-        (public_attributes.len() + private_attributes.len()) as u32,
+        (public_attributes.len() + 2) as u32,
```
neacsu: The reason for having the 2 value might be hard to remember in the future. Could you add some comment as to what it refers?

neacsu: Another possibility would be to keep the `private_attributes` as a vector, check that it has the length 2 in the beginning, and pass the first 2 values of the vector to `prove_bandwidth_credential`

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+
+        ProofKappaNu {
+            challenge,
+            response_serial_number,
+            response_binding_number,
+            response_blinder,
+        }
+    }
+
+    pub(crate) fn private_attributes(&self) -> usize {
```
neacsu: Maybe this could be renamed to `private_attributes_len`?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+        let mut bytes = Vec::with_capacity((attributes_len + 1) as usize * 32);
+
+        bytes.extend_from_slice(&self.challenge.to_bytes());
+        bytes.extend_from_slice(&self.response_serial_number.to_bytes());
+        bytes.extend_from_slice(&self.response_binding_number.to_bytes());
+
+        // bytes.extend_from_slice(&attributes_len.to_le_bytes());
+        // for rm in &self.response_attributes {
+        //     bytes.extend_from_slice(&rm.to_bytes());
+        // }
```
neacsu: I don't really like the idea of having commented code around.

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+        let response_binding_number = try_deserialize_scalar(
+            binding_number_bytes,
+            CoconutError::Deserialization("failed to deserialize the binding number".to_string()),
+        )?;
+        // let rm_end = 40 + rm_len as usize * 32;
+        // let response_attributes = try_deserialize_scalar_vec(
+        //     rm_len,
+        //     &bytes[40..rm_end],
+        //     CoconutError::Deserialization("Failed to deserialize attributes response".to_string()),
+        // )?;
```
neacsu: Same comment

common/nymcoconut/src/scheme/verification.rs
```diff
@@ -0,0 +1,322 @@
...
+        // let theta = prove_credential(
+        //     &mut params,
+        //     &keypair.verification_key(),
+        //     &signature,
+        //     &private_attributes,
+        // )
+        //     .unwrap();
+        //
+        // let bytes = theta.to_bytes();
+        // assert_eq!(Theta::try_from(bytes.as_slice()).unwrap(), theta);
```
neacsu: Same idea for commented code

neacsu (changes requested):

common/credentials/src/utils.rs
aniampio: If I correctly remember, this is how it was originally implemented (not sure by whom) and I only changed the parameters which go into obtain_partial_credential. Thus, I'm not sure whether there was any special reason to separate the first from the rest

clients/native/src/client/mod.rs
```diff
@@ -176,9 +181,19 @@ impl NymClient {
         .await
         .expect("could not obtain aggregate verification key of validators");
 
+        let params = Parameters::new(TOTAL_ATTRIBUTES).unwrap();
+        let bandwidth_credential_attributes = BandwidthVoucherAttributes {
+            serial_number: params.random_scalar(),
+            binding_number: params.random_scalar(),
+            voucher_value: hash_to_scalar(BANDWIDTH_VALUE.to_be_bytes()),
+            voucher_info: hash_to_scalar(String::from("BandwidthVoucher").as_bytes()),
```
jstuczyn:
```suggestion
            voucher_info: hash_to_scalar("BandwidthVoucher"),
```

clients/socks5/src/commands/init.rs
```diff
@@ -57,6 +68,40 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
...
+    let verification_key = obtain_aggregate_verification_key(validators)
+        .await
+        .expect("could not obtain aggregate verification key of validators");
+
+    let params = Parameters::new(TOTAL_ATTRIBUTES).unwrap();
+    let bandwidth_credential_attributes = BandwidthVoucherAttributes {
+        serial_number: params.random_scalar(),
+        binding_number: params.random_scalar(),
+        voucher_value: hash_to_scalar(BANDWIDTH_VALUE.to_be_bytes()),
+        voucher_info: hash_to_scalar(String::from("BandwidthVoucher").as_bytes()),
```
jstuczyn:
```suggestion
            voucher_info: hash_to_scalar("BandwidthVoucher"),
```

clients/native/src/client/mod.rs
jstuczyn: Also, perhaps `BandwidthVoucher` should be put into some constant given that it's used in multiple places and it's very specific to the bandwidth credential?

clients/native/src/commands/init.rs
```diff
@@ -57,6 +68,40 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
         )
 }
 
+// this behaviour should definitely be changed, we shouldn't
+// need to get bandwidth credential for registration
+#[cfg(feature = "coconut")]
+async fn _prepare_temporary_credential(validators: &[Url], raw_identity: &[u8]) -> Credential {
```
jstuczyn, line 88: I could not find any uses of this function anywhere - are you sure it's still required?

clients/socks5/src/commands/init.rs
```diff
@@ -57,6 +68,40 @@ pub fn command_args<'a, 'b>() -> clap::App<'a, 'b> {
         )
 }
 
+// this behaviour should definitely be changed, we shouldn't
+// need to get bandwidth credential for registration
+#[cfg(feature = "coconut")]
+async fn _prepare_temporary_credential(validators: &[Url], raw_identity: &[u8]) -> Credential {
```
jstuczyn, line 88: Same as with the native client -> I could not find the uses of this anywhere

common/credentials/src/bandwidth.rs
```diff
@@ -8,40 +8,84 @@
...
-pub const PRIVATE_ATTRIBUTES: u32 = 1;
+pub const PUBLIC_ATTRIBUTES: u32 = 2;
+pub const PRIVATE_ATTRIBUTES: u32 = 2;
 pub const TOTAL_ATTRIBUTES: u32 = PUBLIC_ATTRIBUTES + PRIVATE_ATTRIBUTES;
 
-// TODO: this definitely has to be moved somewhere else. It's just a temporary solution
-pub async fn obtain_signature(raw_identity: &[u8], validators: &[Url]) -> Result<Signature, Error> {
-    let public_attributes = vec![hash_to_scalar(BANDWIDTH_VALUE.to_be_bytes())];
-    let private_attributes = vec![hash_to_scalar(raw_identity)];
+pub const SERIAL_NUMBER_LEN: usize = 47;
```
jstuczyn: Those constants, i.e. `SERIAL_NUMBER_LEN`, `BINDING_NUMBER_LEN` and `VOUCHER_INFO_LEN` don't seem to be used anywhere -> what's the point of them? Also, why are they 47 long? Is there any explanation for them? Because if they are `Scalar` then it doesn't make sense as the scalar is, from what I recall, 32 bytes long

common/credentials/src/bandwidth.rs
jstuczyn: Agreed, some explanation would be useful for people reading this in the future. Maybe not necessarily in the code but perhaps in some `.md` file explaining how the bandwidth credentials work. Perhaps you could team up with @mfahampshire on it?

common/credentials/src/bandwidth.rs
```diff
@@ -8,40 +8,84 @@
...
+    pub voucher_value: PublicAttribute,
+    pub voucher_info: PublicAttribute,
+}
+
+impl BandwidthVoucherAttributes {
+    pub fn get_public_attributes(&self) -> Vec<PublicAttribute> {
+        let mut pub_attributes = Vec::with_capacity(PUBLIC_ATTRIBUTES as usize);
+        pub_attributes.push(self.voucher_value);
+        pub_attributes.push(self.voucher_info);
+        pub_attributes
```
jstuczyn:
```suggestion
        vec![self.voucher_value, self.voucher_info]
 ```

File: common/credentials/src/bandwidth.rs [hunk 17]
```diff
@@ -8,40 +8,84 @@
...
+    }
+
+    pub fn get_private_attributes(&self) -> Vec<PrivateAttribute> {
+        let mut priv_attributes = Vec::with_capacity(PRIVATE_ATTRIBUTES as usize);
+        priv_attributes.push(self.serial_number);
+        priv_attributes.push(self.binding_number);
+        priv_attributes
```
jstuczyn:
```suggestion
        vec![self.serial_number, self.binding_number]
 ```

File: common/credentials/src/utils.rs [hunk 18]
```diff
@@ -99,38 +112,64 @@ pub async fn obtain_aggregate_signature(
...
+    Ok(aggregate_signature_shares(
+        params,
+        verification_key,
+        &attributes,
+        &shares,
+    )?)
 }
 
 // TODO: better type flow
 pub fn prepare_credential_for_spending(
```
jstuczyn: Any particular reason for private attributes being explicitly separated as specific arguments (i.e. `serial_number` and `binding_number`) while public attributes are aggregated together? Does their order matter?

common/credentials/src/utils.rs
```diff
@@ -83,14 +86,24 @@ async fn obtain_partial_credential(
...
+    Ok(blinded_signature
+        .unblind(
+            params,
+            elgamal_keypair.private_key(),
+            validator_vk,
+            private_attributes,
+            public_attributes,
+            &blind_sign_request.commitment_hash,
+        )
+        .unwrap())
```
jstuczyn: Are we sure this `.unwrap()` here is fine? Under what conditions can `.unblind()` fail? And if none of them are applicable here, could you add a comment saying why the unwrap is fine?

common/credentials/src/utils.rs
```diff
@@ -83,14 +86,24 @@ async fn obtain_partial_credential(
...
 }
 
 pub async fn obtain_aggregate_signature(
     params: &Parameters,
     public_attributes: &[Attribute],
     private_attributes: &[Attribute],
     validators: &[Url],
+    verification_key: &VerificationKey,
```
jstuczyn: Any reason for passing this an argument while it seems we're obtaining partial vks of all validators anyway?

common/credentials/src/utils.rs
aniampio: It's 2 because each bandwidth voucher will have always 2 private attributes. I agree that this is not the most elegant way, I will replace it for now with a constant, but at some point, the code will need a bit of refactoring especially with the numeric numbers :)

common/credentials/src/utils.rs
jstuczyn: That's a good question. I remember there was some reason for that but I honestly don't remember anymore... Maybe it's no longer applicable

common/credentials/src/utils.rs
jstuczyn: If this function is specific to bandwidth credential, maybe we should be passing `BandwidthVoucherAttributes` instead of all attributes as an argument instead? However, if it's something more generic then yeah, it doesn't make sense.

common/nymcoconut/src/constants.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2021 Nym Technologies SA
```
jstuczyn: 1. This is old license
2. This file doesn't seem to have anything useful in it? : )

common/nymcoconut/src/elgamal.rs
jstuczyn: I think that's a residue from my old testing. sorry : )

common/nymcoconut/src/scheme/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+use crate::utils::try_deserialize_g1_projective;
+use crate::{elgamal, Attribute};
+
+pub mod aggregation;
+pub mod issuance;
+pub mod keygen;
+pub mod setup;
+pub mod verification;
+
+pub type SignerIndex = u64;
```
jstuczyn, line 28: If you have any ideas on how to rename it (and all uses of it), I think that'd be the perfect opportunity to do it. I used `index` because it was the first thing that came to my mind when I initially wrote it, but I hate that name so much : (
Just throwing ideas around here: maybe something like `PolynomialCoordinate`? But then `Coordinate` implies (x,y) values while this... thing is just the `x` coordinate used for the Lagrange interpolation...

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+
+use crate::elgamal::Ciphertext;
+use crate::error::{CoconutError, Result};
+use crate::scheme::setup::Parameters;
+use crate::scheme::VerificationKey;
+use crate::utils::{hash_g1, try_deserialize_scalar, try_deserialize_scalar_vec};
+use crate::{elgamal, Attribute, ElGamalKeyPair};
+
+// as per the reference python implementation
+type ChallengeDigest = Sha256;
```
jstuczyn, line 24: This is just a sanity check question, we're still fine to use `sha256`, yeah?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+    B: AsRef<[u8]>,
+{
+    let mut h = D::new();
+    for point_representation in iter {
+        h.update(point_representation);
+    }
+    let digest = h.finalize();
+
+    // TODO: I don't like the 0 padding here (though it's what we've been using before,
+    // but we never had a security audit anyway...)
```
jstuczyn, line 55: I'll use this opportunity to ask this question again :) : is the 0-pad fine here?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+    witnesses
+        .iter()
+        .zip(secrets.iter())
+        .map(|(w, x)| produce_response(w, challenge, x.borrow()))
+        .collect()
+}
+
+impl ProofCmCs {
+    /// Construct non-interactive zero-knowledge proof of correctness of the ciphertexts and the commitment
+    /// using thw Fiat-Shamir heuristic.
```
jstuczyn:
```suggestion
    /// using the Fiat-Shamir heuristic.
```

jstuczyn: That was probably my fault... sorry : )

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+    pub(crate) fn private_attributes(&self) -> usize {
+        2
+    }
+
+    pub(crate) fn verify(
+        &self,
+        params: &Parameters,
+        verification_key: &VerificationKey,
+        kappa: &G2Projective,
+        zeta: &G2Projective,
```
jstuczyn, line 447: Considering that right now we're verifying it based on `kappa` and `zeta`, is it still `ProofKappaNu`? Should it be `ProofKappaZeta` instead? Or even better, maybe something less Greek with more meaningful name : )

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+use crate::scheme::VerificationKey;
+use crate::utils::{hash_g1, try_deserialize_scalar, try_deserialize_scalar_vec};
+use crate::{elgamal, Attribute, ElGamalKeyPair};
+
+// as per the reference python implementation
+type ChallengeDigest = Sha256;
+
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct ProofCmCs {
```
jstuczyn, line 28: Related to my comment on the name of `ProofKappaNu` -> is it still appropriate to call this one `ProofCmCs`?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+                .map(|(wm_i, beta_i)| beta_i * wm_i)
+                .sum::<G2Projective>();
+
+        // zeta is the public value associated with the serial number
+        let commitment_zeta = params.gen2() * witness_serial_number;
+
+        let challenge = compute_challenge::<ChallengeDigest, _, _>(
+            std::iter::once(params.gen2().to_bytes().as_ref())
+                .chain(std::iter::once(blinded_message.to_bytes().as_ref()))
+                .chain(std::iter::once(blinded_serial_number.to_bytes().as_ref())) //kappa
```
jstuczyn: is `blinded_serial_number` kappa?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+        // Aw = g2 * wt + alpha + beta[0] * wm[0] + ... + beta[i] * wm[i]
+        let commitment_kappa = params.gen2() * witness_blinder
+            + verification_key.alpha
+            + witness_attributes
+                .iter()
+                .zip(verification_key.beta.iter())
+                .map(|(wm_i, beta_i)| beta_i * wm_i)
+                .sum::<G2Projective>();
+
+        // zeta is the public value associated with the serial number
```
jstuczyn, line 410: So maybe we could call it SerialNumber instead of zeta? : )

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+        let commitment_zeta = zeta * self.challenge + params.gen2() * self.response_serial_number;
+
+        // compute the challenge
+        let challenge = compute_challenge::<ChallengeDigest, _, _>(
+            std::iter::once(params.gen2().to_bytes().as_ref())
+                .chain(std::iter::once(kappa.to_bytes().as_ref()))
+                .chain(std::iter::once(zeta.to_bytes().as_ref())) //kappa
```
jstuczyn: I think this comment is outdated because you're referring to `zeta` as `kappa`

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+        //let attributes_len = self.response_attributes.len() as u64;
+        let attributes_len = 2;
+        let mut bytes = Vec::with_capacity((attributes_len + 1) as usize * 32);
+
+        bytes.extend_from_slice(&self.challenge.to_bytes());
+        bytes.extend_from_slice(&self.response_serial_number.to_bytes());
+        bytes.extend_from_slice(&self.response_binding_number.to_bytes());
+
+        // bytes.extend_from_slice(&attributes_len.to_le_bytes());
+        // for rm in &self.response_attributes {
```
jstuczyn: Are we fine to skip serialization of attributes responses?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,691 @@
...
+            serial_number_bytes,
+            CoconutError::Deserialization("failed to deserialize the serial number".to_string()),
+        )?;
+
+        let binding_number_bytes = &bytes[64..96].try_into().unwrap();
+        let response_binding_number = try_deserialize_scalar(
+            binding_number_bytes,
+            CoconutError::Deserialization("failed to deserialize the binding number".to_string()),
+        )?;
+        // let rm_end = 40 + rm_len as usize * 32;
```
jstuczyn: Since we don't seem to deserialize this anymore, please make sure the bounds on the byte slice are still valid, i.e. it's impossible to cause runtime panic by trying to read out of bounds

common/nymcoconut/src/scheme/issuance.rs
```diff
@@ -0,0 +1,393 @@
...
+
+// TODO NAMING: double check this one
+// Lambda
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct BlindSignRequest {
+    // cm
+    commitment: G1Projective,
+    // h
+    pub commitment_hash: G1Projective,
```
jstuczyn: why is this one field public?

common/nymcoconut/src/scheme/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+
+        let tmp = private_attributes
+            .iter()
+            .chain(public_attributes.iter())
+            .zip(partial_verification_key.beta.iter())
+            .map(|(attr, beta_i)| beta_i * attr)
+            .sum::<G2Projective>();
+
+        // Verify the signature share
+        if !check_bilinear_pairing(
```
jstuczyn, line 177: Do we really have to perform the pairing for each unblind operation?

common/nymcoconut/src/scheme/verification.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+impl TryFrom<&[u8]> for Theta {
+    type Error = CoconutError;
+
+    fn try_from(bytes: &[u8]) -> Result<Theta> {
+        if bytes.len() < 288 {
+            return Err(
+                CoconutError::Deserialization(
+                    format!("Tried to deserialize theta with insufficient number of bytes, expected >= 240, got {}", bytes.len()),
```
jstuczyn: This error message does not really seem to be updated

jstuczyn (changes requested):
I didn't have enough brain cycles today to verify the crypto stuff, I assume it's fine since the tests seem to be passing. However, if you want me to take a look at it, please do let me know and I'll find time for it.

There are few general comments apart from the things that I've left in the code:
- it seems some comments, especially related to crypto and serialization are no longer in sync with the code. I think it's very important that they are otherwise it's very confusing to figure out what refers to what exactly,
- regarding serialization/deserialization, some structures were changed, please make sure all bounds on byte slices are set correctly so it's impossible to cause an out of bound panic by passing bytes of some invalid length,
- since the coconut code is moved into the main repo, I think all license notices in `nymcoconut` files need to be updated, unless @mmsinclair or @futurechimp think it's fine to keep the old ones,

clients/native/src/client/mod.rs
aniampio: Regarding checking crypto I think you don't have to.
Regarding the 'BandwidthVoucher', this is a temporary thing until we figure out what we actually want to keep as the (public) voucher info. This can be a voucher type, or an epoch number or something else.

clients/native/src/commands/init.rs
aniampio: Not sure, we might need it once we start using actual Coconut

clients/socks5/src/commands/init.rs
aniampio: As above, this function was already in the code and I'm not sure whether we gonna need it once we move to actual Coconut :)

common/credentials/src/bandwidth.rs
aniampio: Yep, I kind of defined them for future reference. 47 is the maximum size of the attribute field we can encode in the credential (and it is specified by the prime p from our curve. It is BLS12_381, where p=381 bits, so the max byte length would be 47). Indeed we hash everything to a Scalar which is 32 bytes, but I have to think whether it is profitable to keep 47 anyway or not :)

common/nymcoconut/src/proofs/mod.rs
aniampio: it is actually done as response_serial_number, response_binding_number and response_blinder (since these are the only attributes we gonna use in the zk proof for Bandwidth Vouchers)

common/nymcoconut/src/scheme/issuance.rs
aniampio: In general we need access to this field since it's used for integrity checks at different stages of Coconut scheme. However, to make it look better I replaced it with a function to get this field instead of making it public.

common/nymcoconut/src/scheme/mod.rs
aniampio: Yes, we have to make sure that what we got back from the validator is legit

common/credentials/src/utils.rs
aniampio: Yes, public attributes are all processed in the same way, while the private attributes (serial number and the binding factor) are processed differently. And it is important to not confuse the serial number with the binding number during the show protocol (at any point)

common/credentials/src/utils.rs
aniampio: That's a good point! There is actually no need to pass it as an input argument since we can simply compute it :)

common/nymcoconut/src/scheme/mod.rs
aniampio: I actually think it is a very decent name :)

common/nymcoconut/src/proofs/mod.rs
aniampio: The audit by JP addresses that and I think we can fix it once we will be fixing everything from his audit suggestions

common/nymcoconut/src/proofs/mod.rs
aniampio: zeta is not a serial number, it is g^serial_number. At some point, I will try to rethink how to name all those things using English :)

common/nymcoconut/src/proofs/mod.rs
aniampio: No, it is not

neacsu (approved):
Lgtm, but if there are some possible out of bounds panics in there, they should be addressed before merge (I haven't seen them)

common/credentials/src/utils.rs
jstuczyn: I would feel more comfortable if this comment was addressed : )

mmsinclair:
@aniampio would you mind fixing up the conflicts with `develop`?

common/credentials/src/utils.rs
aniampio: Updated to `?` to propagate the error up

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+                ))
+                .chain(commitment_keys1_bytes.iter().map(|aw| aw.as_ref()))
+                .chain(commitment_keys2_bytes.iter().map(|bw| bw.as_ref()))
+                .chain(ciphertexts_bytes.iter().map(|c| c.as_ref())),
+        );
+
+        challenge == self.challenge
+    }
+
+    // challenge || rr || rk.len() || rk || rm.len() || rm
```
jstuczyn: I just noticed that this description is outdated since now we also put response regarding elgamal private key

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+        for rm in &self.response_attributes {
+            bytes.extend_from_slice(&rm.to_bytes());
+        }
+
+        bytes
+    }
+
+    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<Self> {
+        // at the very minimum there must be a single attribute being proven
+        if bytes.len() < 32 * 4 + 16 || (bytes.len() - 16) % 32 != 0 {
```
jstuczyn:
```suggestion
        if bytes.len() < 32 * 5 + 16 || (bytes.len() - 16) % 32 != 0 {
```
Since we require one additional response

common/nymcoconut/src/proofs/mod.rs
jstuczyn: So maybe we should remove that comment : )

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+                .chain(std::iter::once(verification_key.alpha.to_bytes().as_ref()))
+                .chain(beta_bytes.iter().map(|b| b.as_ref()))
+                .chain(std::iter::once(commitment_kappa.to_bytes().as_ref()))
+                .chain(std::iter::once(commitment_zeta.to_bytes().as_ref())),
+        );
+
+        challenge == self.challenge
+    }
+
+    // challenge || rm.len() || rm || rt
```
jstuczyn: Again, this is outdated as additional response has been added

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
+        //let attributes_len = self.response_attributes.len() as u64;
+        let attributes_len = 2;
```
jstuczyn: I thought we have 4 attributes?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+                .chain(std::iter::once(commitment_kappa.to_bytes().as_ref()))
+                .chain(std::iter::once(commitment_zeta.to_bytes().as_ref())),
+        );
+
+        challenge == self.challenge
+    }
+
+    // challenge || rm.len() || rm || rt
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
+        //let attributes_len = self.response_attributes.len() as u64;
```
jstuczyn: If the code is unused, just remove it

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+        bytes.extend_from_slice(&self.response_binding_number.to_bytes());
+
+        bytes.extend_from_slice(&self.response_blinder.to_bytes());
+
+        bytes
+    }
+
+    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<Self> {
+        // at the very minimum there must be a single attribute being proven
+        if bytes.len() < 32 * 3 || (bytes.len()) % 32 != 0 {
```
jstuczyn:
```suggestion
        if bytes.len() != 32 * 4 {
```
given that our bytes representation is just concatenation of 4 scalars

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+            return Err(CoconutError::DeserializationInvalidLength {
+                actual: bytes.len(),
+                modulus_target: bytes.len(),
+                modulus: 32,
+                object: "kappa and zeta".to_string(),
+                target: 32 * 3 + 8,
```
jstuczyn:
```suggestion
                target: 32 * 4,
```

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+            });
+        }
+
+        let challenge_bytes = bytes[..32].try_into().unwrap();
+        let challenge = try_deserialize_scalar(
+            &challenge_bytes,
+            CoconutError::Deserialization("Failed to deserialize challenge".to_string()),
+        )?;
+
+        // let rm_len = u64::from_le_bytes(bytes[32..40].try_into().unwrap());
```
jstuczyn: unused code

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+        if bytes[32..].len() != (2 + 1) as usize * 32 {
```
jstuczyn: No need for length checks as we already asserted we have 4 * 32 bytes in total

common/nymcoconut/src/scheme/issuance.rs
```diff
@@ -0,0 +1,397 @@
...
+    pi_s: ProofCmCs,
+}
+
+impl TryFrom<&[u8]> for BlindSignRequest {
+    type Error = CoconutError;
+
+    fn try_from(bytes: &[u8]) -> Result<BlindSignRequest> {
+        if bytes.len() < 48 + 48 + 8 + 96 {
+            return Err(CoconutError::DeserializationMinLength {
+                min: 48 + 48 + 8 + 9,
```
jstuczyn:
```suggestion
                min: 48 + 48 + 8 + 96,
```

common/nymcoconut/src/scheme/issuance.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    pub fn get_commitment_hash(&self) -> G1Projective {
+        self.commitment_hash
+    }
+
+    // TODO: perhaps also include pi_s.len()?
+    // to be determined once we implement serde to make sure its 1:1 compatible
+    // with bincode
+    // cm || c.len() || c || pi_s
```
jstuczyn: Another outdated comment as we seem to have added commitment hash to the blind sign request

common/nymcoconut/src/scheme/verification.rs
```diff
@@ -0,0 +1,306 @@
...
+            verification_key,
+            &self.blinded_message,
+            &self.blinded_serial_number,
+        )
+    }
+
+    // TODO: perhaps also include pi_v.len()?
+    // to be determined once we implement serde to make sure its 1:1 compatible
+    // with bincode
+    // kappa || nu || credential || pi_v
```
jstuczyn: We no longer have a `nu` here, instead we have `zeta`

jstuczyn (commented):
There is a couple of left over inconsistencies regarding serialization, but I think in general after resolving the pending comments, it could be merged

common/nymcoconut/src/proofs/mod.rs
aniampio: For bandwidth voucher, we have 4 attributes indeed, but only 2 are private and go into the zk proof

jstuczyn (approved):

PR #839, opened 2021-10-21
Linting fixes

Aid19801:
I've left 2 linting errors still in there with `// @ts-ignore` 

1. `WorldMap.tsx` - which does not come with type declarations, generates errors with the d3 scale pkg. I need to look into this a bit deeper but dont believe it should hold up progress elsewhere in app. 

2. `Nav.tsx`- which doesnt like my nested ternary (does anyone?). I think correcting that should be incorporated into a broader refactor of Nav to fix the orange line bug vs selecting and unselecting network components. Creating a branch for that now. Disabling the rule on this branch though.

### DIFF
--- a/explorer/src/components/ContentCard.tsx
+++ b/explorer/src/components/ContentCard.tsx
@@ -54,8 +54,8 @@ export const ContentCard: React.FC<ContentCardProps> = ({
 ContentCard.defaultProps = {
   title: undefined,
   subtitle: undefined,
-  Icon: () => null,
-  Action: () => null,
+  Icon: null,
+  Action: null,
   errorMsg: undefined,
   onClick: () => null,
 };
--- a/explorer/src/components/ContentCard.tsx
+++ b/explorer/src/components/ContentCard.tsx
@@ -1,10 +1,9 @@
-import { ReactJSXElement } from '@emotion/react/types/jsx-namespace';
 import { Card, CardHeader, CardContent, Typography } from '@mui/material';
 import React, { ReactEventHandler } from 'react';
 import { MainContext } from 'src/context/main';
 
 type ContentCardProps = {
-  title?: string | ReactJSXElement;
+  title?: string | React.ReactNode;
   subtitle?: string;
   Icon?: React.ReactNode;
   Action?: React.ReactNode;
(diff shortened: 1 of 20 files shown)


explorer/src/components/ContentCard.tsx
```diff
@@ -1,10 +1,9 @@
-import { ReactJSXElement } from '@emotion/react/types/jsx-namespace';
 import { Card, CardHeader, CardContent, Typography } from '@mui/material';
 import React, { ReactEventHandler } from 'react';
 import { MainContext } from 'src/context/main';
 
 type ContentCardProps = {
-  title?: string | ReactJSXElement;
+  title?: string | React.ReactNode;
```
fmtabbara, line 6: Should we make the title non-optional? It seems like the key part of the Card component

explorer/src/components/ContentCard.tsx
```diff
@@ -37,7 +36,7 @@ export const ContentCard: React.FC<ContentCardProps> = ({
           color: (theme) =>
             mode === 'dark' ? theme.palette.primary.main : 'secondary.main',
         }}
-        title={title}
+        title={title || ''}
```
fmtabbara, line 39: If Card was non-optional you could remove this

explorer/src/components/Nav.tsx
```diff
@@ -343,9 +350,9 @@ export const Nav: React.FC = ({ children }) => {
         </DrawerHeader>
 
         <List sx={{ pt: 0, pb: 0 }}>
-          {originalNavOptions.map((props, i) => (
+          {originalNavOptions.map((props) => (
```
fmtabbara, line 353: if `id` is always defined you can destructure it here. Or keep as is but maybe rename props to something else (maybe opts) generally props is used on components.

explorer/src/utils/index.ts
```diff
@@ -1,12 +1,15 @@
...
 
-export function formatNumber(num: number) {
+export function formatNumber(num: number): string {
   return new Intl.NumberFormat().format(num);
 }
 
-export function scrollToRef(ref: any) {
-  return ref.current.scrollIntoView();
+export function scrollToRef(
+  ref: MutableRefObject<HTMLDivElement | undefined>,
```
fmtabbara, line 10: :+1:

fmtabbara (approved):

explorer/src/components/Nav.tsx
Aid19801: I've left as 'props' as the ...spread feeds the rest of the object to ExpandableButton as its props, but yes i get you.

explorer/src/components/ContentCard.tsx
Aid19801: The title is not always there. We're using `<ContentCard />` to wrap data grids too, which do not require a title.

PR #838, opened 2021-10-21
Social Icons & Dark Mode selector

Aid19801:
1. Removed responsive Dark/Light on drawer SVG icons. That whole section stays same colour regardless of light/dark mode, as per designs.
2. Responsive Footer - shows Socials if mobile.
3. Socials - collection of Telegram, Github, Twitter
4. Light/Dark switch - moved from Drawer to Appbar; changes to smaller icon on mobile, as per (updated) designs. Note TS issue with onClick type. Tried defining it to reflect the `() => void | undefined` but i must be missing something.

PR #844, opened 2021-10-24
Nav changes

Aid19801:
Bug:
user could highlight two options at once. Click Network Components then click Overview and both would highlight.

Fix:
added isActive bool to the array of nav options. This is updated on click and fed back to each <ExpandableButton isActive />. Then within that, conditional rendering to apply styles dependent on whether the button isActive or isChild. This removes the awful nested ternary too.

NOTES:
I've left some comments in there but happy to remove if that's neater.
No linting errors.
At the moment the logic uses if (title === 'Network Components') {... to decide to open the nest of child routes up. You could argue this should be something like if (hasChildren) so we can use it for any future/added routes that might have children. If that's unlikely though, this would be fine.


explorer/src/icons/GitHubSVG.tsx
```diff
@@ -0,0 +1,62 @@
...
+            width="24.6"
+            height="24"
+            fill="#111826"
+            transform="translate(0.400024)"
+          />
+        </clipPath>
+      </defs>
+    </svg>
+  </>
+);
```
mmsinclair: There is a lot of duplication that can be removed by using the theme hook and changing the fill depending on if it is dark mode.

```
export const GitHubSVG: React.FC = () => {
    const { isDarkMode } = useTheme();
    return (<svg
      width="26"
      height="26"
      viewBox="0 0 26 26"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
    >
      <g clipPath="url(#clip0_661:436)">
        <path
          fillRule="evenodd"
          clipRule="evenodd"
          d="M12.7 0C5.90002 0 0.400024 5.5 0.400024 12.3C0.400024 17.7 3.90002 22.3 8.80002 24C9.40002 24.1 9.60002 23.7 9.60002 23.4C9.60002 23.1 9.60002 22.1 9.60002 21.1C6.50002 21.7 5.70002 20.3 5.50002 19.7C5.40002 19.3 4.80002 18.3 4.20002 18C3.80002 17.8 3.20002 17.2 4.20002 17.2C5.20002 17.2 5.90002 18.1 6.10002 18.5C7.20002 20.4 9.00002 19.8 9.70002 19.5C9.80002 18.7 10.1 18.2 10.5 17.9C7.80002 17.6 4.90002 16.5 4.90002 11.8C4.90002 10.5 5.40002 9.4 6.20002 8.5C6.00002 8 5.60002 6.8 6.30002 5.1C6.30002 5.1 7.30002 4.8 9.70002 6.4C10.7 6.1 11.7 6 12.8 6C13.8 6 14.9 6.1 15.9 6.4C18.3 4.8 19.3 5.1 19.3 5.1C20 6.8 19.5 8.1 19.4 8.4C20.2 9.3 20.7 10.4 20.7 11.7C20.7 16.4 17.8 17.5 15.1 17.8C15.5 18.2 15.9 18.9 15.9 20.1C15.9 21.7 15.9 23.1 15.9 23.5C15.9 23.8 16.1 24.2 16.7 24.1C21.6 22.5 25.1 17.9 25.1 12.4C25 5.5 19.5 0 12.7 0Z"
          fill={isDarkMode ? '#111826' : 'white' }
        />
      </g>
      <defs>
        <clipPath id="clip0_661:436">
          <rect
            width="24.6"
            height="24"
            fill={isDarkMode ? 'white' : '#111826' }
            transform="translate(0.400024)"
          />
        </clipPath>
      </defs>
    </svg>
  );
}
```

mmsinclair: And try to use colours from the theme rather than `white` and '#111826'. And figure out if they are really transparency values on standard themes by looking in Figma.

```
const theme = useTheme();
const fill = theme.palette.text.primary;
const bg = theme.palette.background.default;

...

<rect
  width="24.6"
  height="24"
  fill={fill}
  transform="translate(0.400024)"
/>
```

mmsinclair (changes requested):
There is a lot of duplication in the SVG icons that can be removed by using the theme hook and directly using theme colours.

explorer/src/icons/GitHubSVG.tsx
Aid19801: I don't think they (the instances of white) are transparent. I've def been getting colours from Figma and adding to the palette object. I think there's a wider piece (IMO) to do re: _expanding_ the MUI theme so it incorporates the extra colors (the black #111826, and the orange bar, for example). @fmtabbara suggested parking the colors in a palette obj & importing that for the time being. Keeps colors centralised in the theme object at least. 

The second (or ideal) step would be to get MUI theme's `.darker` or `.lighter` shading/logic working. I suppose those colors might deviate from the ones on Figma but if we did use them, it would mean fewer _manually_ added in colors...

Aid19801: Also - most importantly - the Github, Twitter, Telegram SVG files were my first approach. Should've deleted from PR. Have done so now. Instead i have the `socials.tsx` file which houses all 3 from MUI icons.

Possibly also worth noting that I had an issue with SVGs, where the `Appbar` icons should not respond to dark mode. The same icons exist in the Footer and should respond to dark mode. Hence added `disableDarkMode` bool to Socials.

Also tried Mui's native SVG wrapper component but got some sort of webpack loader issue i wasn't able to resolve. Perhaps you and i could try that in a quick 15-20 min session.

mmsinclair: > Also tried Mui's native SVG wrapper component but got some sort of webpack loader issue i wasn't able to resolve. Perhaps you and i could try that in a quick 15-20 min session.

This will be worse, because you can't substitute properties of elements inside the SVG. What is there at the moment is better.

Aid19801: sorry perhaps i should've replied to this. So i'm assuming current approach (Socials.tsx with collection of MUI icons is fine, no further changed required, shout if not)

mmsinclair:
Hopefully after https://github.com/nymtech/nym/pull/838#issuecomment-956228044 is addressed, this PR will update and can be approved

mmsinclair (approved):

PR #845, opened 2021-10-25
Mixnode Detail Alignment

Aid19801:
1. styles.css override - no offered way to adjust padding on that one specific `TitleContainer` div, which is putting the header out of alignment with the cells.
2. Added `flex: 1` (mui doc'd approach) to the _non_ word-wrapping headers in DataGrid. This helps them to space out more responsively, as per Figma designs.
3. added `height: 100%` to ContentCard so reuseable cards like those in Mixnode Detail page match the height of the grid item and thus the container row. This generates the expected alignment.
4. Adjustments to margin/padding to keep the same spacing in between Detail page cards.

### DIFF
--- a/explorer/src/styles.css
+++ b/explorer/src/styles.css
@@ -12,10 +12,15 @@ in <DataGrid /> */
     visibility: hidden;
 }
 
-/* TODO - this should be disabled somehow in MUI DataGrid
-but documentation doesnt offer an obvious way to do it. Possibly only
+/* TODO - this should be managed somehow in MUI DataGrid
+but documentation doesnt offer a way to do it. Possibly only
 included in Data Grid Pro package */
 .MuiDataGrid-header-override {
     height: 55px;
     min-height: 55px;
+}
+/* Again, no offered way to add sx to this specific div to kill the padding
+which puts it out of sync with other (sx styled) cells */
+div div.MuiDataGrid-root .MuiDataGrid-columnHeaderTitleContainer {
+    padding-left: 0;
 }
--- a/explorer/src/components/ContentCard.tsx
+++ b/explorer/src/components/ContentCard.tsx
@@ -29,6 +29,7 @@ export const ContentCard: React.FC<ContentCardProps> = ({
           mode === 'dark'
             ? theme.palette.secondary.dark
             : theme.palette.primary.light,
+        height: '100%',
       }}
     >
       <CardHeader
--- a/explorer/src/components/CustomColumnHeading.tsx
+++ b/explorer/src/components/CustomColumnHeading.tsx
@@ -22,6 +22,7 @@ export const CustomColumnHeading: React.FC<{ headingTitle: string }> = ({
               ? theme.palette.primary.main
               : theme.palette.secondary.main,
           fontSize: 14,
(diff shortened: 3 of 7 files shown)


explorer/src/styles.css
```diff
@@ -12,10 +12,15 @@ in <DataGrid /> */
...
+but documentation doesnt offer a way to do it. Possibly only
 included in Data Grid Pro package */
 .MuiDataGrid-header-override {
     height: 55px;
     min-height: 55px;
+}
+/* Again, no offered way to add sx to this specific div to kill the padding
+which puts it out of sync with other (sx styled) cells */
+div div.MuiDataGrid-root .MuiDataGrid-columnHeaderTitleContainer {
+    padding-left: 0;
```
mmsinclair, line 26: This looks to me like it would apply everywhere?

If not, then `.MuiDataGrid-override` would be best, and apply it where it is needed.

mmsinclair (changes requested):
The global theme override will affect all data grids - are you sure about this?

mmsinclair:
Are you sure this should be a data grid and not a plain table?

image

It has all the chrome to sort and select columns. Wouldn't a table component be better?

explorer/src/styles.css
Aid19801: The issue we have is we can't add a classname to that one specific header `div`. 

The div we need to kill the padding on is `.MuiDataGrid-columnHeaderTitleContainer` 

Even with their doc'd customization, ie handing the grid my `<CustomColumnHeading />` component - the div it places that _inside_ is the one we need to adjust.

So i can either override the TitleContainer myself in `styles.css` or i can counter it by styling the CustomColumnHeading with `margin-left: -6px;`. I don't normally go into minus figures though. Your call.

Aid19801:
> Are you sure this should be a data grid and not a plain table?
> 
> image
> 
> It has all the chrome to sort and select columns. Wouldn't a table component be better?

Yes, as discussed can switch that out in PR separate to the alignment work

explorer/src/styles.css
mmsinclair: This is where that style is being applied:

image

The DOM looks like this:

image

and 

image

So the selector could be `div div.MuiDataGrid-root div.MuiDataGrid-root-override .MuiDataGrid-columnHeaderTitleContainer`

And the React component (https://github.com/nymtech/nym/blob/71ffbcdb4e3bc4e62be8af102f6ee0a89c183fc2/explorer/src/pages/Mixnodes/index.tsx#L166) would now be:

```
        <Grid container className="MuiDataGrid-root-override">
```

Then those overrides **only** apply to that particular instance of the grid and not all of them.

mmsinclair: So, the question is:

1. do you want that override to apply to all `MuiDataGrid` instances?
2. is it only the one in the mix node details page?

For option 1, what you have is good. Otherwise you need something is to constrain it to only that instance.

mmsinclair (changes requested):

mmsinclair (approved):
Approving on the understanding that the MUI Data Grid header changes apply everywhere

PR #856, opened 2021-11-01
Removed gateway rewarding and delegation

jstuczyn:
As we discussed, there should be no option to delegate to gateways. Furthermore, any gateway rewards should come from fees and not from the underlying reward pool.


neacsu (approved):
Always nice to maintain less code :)

PR #857, opened 2021-11-01
Removed epoch rewarding variance

jstuczyn:
The initial idea behind this feature was to prevent all validators distributing rewards at the same time thus possibly putting additional strain on the chain. However, given that initially we will most likely only have a single entity distributing the rewards, this feature is useless. I'd even argue it's harmful as it makes things more difficult to predict and also debug.

### DIFF
--- a/validator-api/src/rewarding/mod.rs
+++ b/validator-api/src/rewarding/mod.rs
@@ -14,7 +14,6 @@ use crate::storage::models::{
 use crate::storage::NodeStatusStorage;
 use log::{error, info};
 use mixnet_contract::{ExecuteMsg, IdentityKey};
-use rand::{thread_rng, Rng};
 use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


validator-api/src/rewarding/mod.rs
```diff
@@ -45,8 +44,6 @@ pub(crate) const REWARDING_GAS_LIMIT_MULTIPLIER: f64 = 1.05;
 
 pub(crate) const MAX_TO_REWARD_AT_ONCE: usize = 50;
 
-pub(crate) const REWARDING_TIME_VARIANCE: f32 = 0.05; // 5% (so for example +/-1.2h for 24h epoch)
```
neacsu, line 48: Why not just put a very small value here (not sure if 0 works, but something close to that)? That way we have an easy way of re-enabling it in the future?

jstuczyn: Because in hindsight I'm not entirely sure whether it was a good idea to begin with. It could bring more harm than benefit by being nondeterministic. And the ranges of values change drastically if we go from 1 day rewarding interval into 1 month. I think we should try the problem in a different way once we get to that point.

neacsu (approved):

PR #860, opened 2021-11-01
Made daily uptime calculation be independent of epoch rewarding

jstuczyn:
Currently the `/mixnode/<pubkey>/history` endpoint is populated with data produced every 24h during rewarding. However, it only works due to the fact the rewarding interval is set to 24h. But as we discussed this assumption might not be true for much longer thus making this endpoint useless.

This pull request moves the calculation of the historical uptime into a separate timer that is completely independent of the rewarding interval.


neacsu (approved):
lgtm

PR #867, opened 2021-11-02
Hamburger Menu & Fixed AppBar

Aid19801:
This PR seeks to implement a *fixed Appbar at the top of the screen*, adding an external link to the NYM logo (taking the user to the main NymTech.net website & adds a 'hamburger' menu to the sidebar, to match Figma designs.

Logic is essentially that i've moved the sidebar down by the same height as the AppBar, and added the Hamburger to it.

`<Appbar />` is a much leaner (but still native MUI) version. Previously popping the drawer pushed the NYM logo / "Network Explorer" to the side. Now that AppBar sits outside of that logic and no longer requires transition etc, hence killing that code.

### DIFF
--- a/explorer/src/components/Nav.tsx
+++ b/explorer/src/components/Nav.tsx
@@ -1,12 +1,12 @@
 /* eslint-disable no-nested-ternary */
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import { ChevronLeft, ExpandLess, ExpandMore } from '@mui/icons-material';
+import { ChevronLeft, ExpandLess, ExpandMore, Menu } from '@mui/icons-material';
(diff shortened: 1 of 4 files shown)


explorer/src/components/Nav.tsx
```diff
@@ -179,20 +155,14 @@ const ExpandableButton: React.FC<ExpandableButtonType> = ({
 
   const handleClick = () => {
     openDrawer();
-    // 1. if it's NetworkComponents parent...
     if (title === 'Network Components') {
       if (nested) {
```
mmsinclair:
```suggestion
    if (title === 'Network Components' && nested) {
```

mmsinclair (approved):
One small syntax thing to fix and then good to merge

PR #871, opened 2021-11-03
add native window items (copy/paste) via tauri

fmtabbara:


tommyv1987 (approved):
We don't need any other actions. I'm happy with this. This should satisfy pasting.

PR #873, opened 2021-11-03
remove gateway selection on delegation and undelegation pages

fmtabbara:


tommyv1987 (approved):
Checked out the branch, built and ran the wallet, and the areas in which gateway delegation was applied has now been successfully removed.

PR #875, opened 2021-11-04
Put client_address and id in the correct order

neacsu:


jstuczyn (approved):
Oh damn, I'm sorry about that one, good fix!

PR #880, opened 2021-11-05
Feature/rewarding interval updates

jstuczyn:
This PR further builds on https://github.com/nymtech/nym/pull/864 and thus requires the mentioned PR to be merged first.

This PR introduces mechanism to update `rewarding_interval_starting_block` in the contract state and thus allow for new seed for the PRNG for the selection of the "demanded" set of mixnodes. It does so by introducing two new transactions that explicitly begin and finish mixnode rewarding: `ExecuteMsg::BeginMixnodeRewarding` and `ExecuteMsg::FinishMixnodeRewarding` respectively. `ExecuteMsg::BeginMixnodeRewarding` sets new `rewarding_interval_starting_block` while `ExecuteMsg::FinishMixnodeRewarding` finishes rewarding procedure thus indicating to the validator API that the new set should be determined.
Furthermore, no `ExecuteMsg::RewardMixnode` transactions can be sent outside the block defined by the begin and finish transactions.  Creation of the block also allowed for easy introduction of protection against rewarding any mixnode more than once during a rewarding interval. We are simply storing a list of which node has been already rewarded during particular interval. Since the rewarding interval is expected to last in order of months, it shouldn't really bloat our storage.

Remaining work:
- [ ] measure changes to `MIXNODE_REWARD_OP_BASE_GAS_LIMIT`


neacsu (approved):
Small suggestion: I think you could have targeted for merge the `feature/working-rewarded-sets` branch, until that one was merged, so that this PR only showed its own diffs and not those from the other PR

jstuczyn:
Yep... I definitely should have. My bad!

durch (approved):

PR #883, opened 2021-11-08
Updated contract code to be compatible with cosmwasm 1.0-beta2

### LINKED ISSUE #2110: Upgrade to cosmwasm 1.0

jstuczyn:
Probably we don't want to merge this sucker until we want to upgrade all validators as this requires wasmd upgrade.

TODOs:
- [x] rebase + update erc20 bridge code

Closes https://github.com/nymtech/nym/issues/2110


durch (approved):

common/client-libs/validator-client/Cargo.toml
```diff
@@ -31,7 +31,7 @@ prost = { version = "0.9", default-features = false, optional = true }
 flate2 = { version = "1.0.20", optional = true }
 sha2 = { version = "0.9.5", optional = true }
 itertools = { version = "0.10", optional = true }
-cosmwasm-std = { git = "https://github.com/jstuczyn/cosmwasm", branch="0.14.1-updatedk256", optional = true }
+cosmwasm-std = { git = "https://github.com/CosmWasm/cosmwasm", rev="v1.0.0-beta2", optional = true }
```
neacsu: Isn't `v1.0.0-beta2` more of a tag then a rev? Also, why not get it from crates.io instead?

neacsu (commented):
Would it make sense to include the `erc20-bridge` contract into this update too? I can write the commit for that, if you want

common/client-libs/validator-client/Cargo.toml
jstuczyn: `v1.0.0-beta2` is indeed a tag, but to pull tags in `Cargo.toml`, you use keyword `rev` : )
Also I couldn't get it from crates.io because they haven't released it on crates.io

jstuczyn:
> Would it make sense to include the `erc20-bridge` contract into this update too? I can write the commit for that, if you want

Ah, it would make perfect sense to do it. When I created this PR, the `erc20-bridge` wasn't merged yet : )

common/client-libs/validator-client/Cargo.toml
neacsu: There is also a `tag` keyword, but if it works with `rev` as well it's all good. There is also the official release and it seems to work alright: https://crates.io/crates/cosmwasm-std/1.0.0-beta2

jstuczyn: Huh, For some reason I didn't see it last time... Let me fix that then, good catch!

neacsu (approved):

jstuczyn:
Due to changes in https://github.com/nymtech/nym/pull/910 it's going to be a pain to merge this sucker, so it will be superseded by another branch.

PR #884, opened 2021-11-08
Fix path for github action running tauri-wallet-tests

tommyv1987:
. Applied prettier to the code base
. Removed delegation to gateway tests

. Add future tasks in product board for TS, And unbonding nodes / gateways.


durch (approved):

fmtabbara (approved):

tauri-wallet/webdriver/.prettierrc.json
```diff
@@ -0,0 +1 @@
+{}
```
mmsinclair, line 1: Are you sure this should be an empty config?

Also, be careful only using `prettier` because the root of the project has `eslint` (and no prettier). They will end up flip-flopping unless configured to work together. Look at https://prettier.io/docs/en/integrating-with-linters.html and use https://github.com/prettier/eslint-plugin-prettier to drive prettier.

I'm wondering if we should maybe do that work in another PR? Because it will likely mean a lot of formatting changes to the wallet JS/TS code.

mmsinclair (approved):
Other than my comment about prettier/eslint, I think we should just merge this and come back to sorting out linting/formatting later.

PR #885, opened 2021-11-09
Adding data-test-ids for the explorer

tommyv1987:
Test ID's for doing automation on explorer

### DIFF
--- a/explorer/src/components/TableToolbar.tsx
+++ b/explorer/src/components/TableToolbar.tsx
@@ -39,10 +39,18 @@ export const TableToolbar: React.FC<TableToolBarProps> = ({
             marginBottom: matches ? 2 : 0,
           }}
         >
-          <MenuItem value={10} data-testid="ten">10</MenuItem>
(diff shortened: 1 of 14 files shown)


explorer/src/components/TableToolbar.tsx
```diff
@@ -30,23 +30,32 @@ export const TableToolbar: React.FC<TableToolBarProps> = ({
         }}
       >
         <Select
-          labelId="demo-simple-select-label"
-          id="demo-simple-select"
+          labelId="simple-select-label"
+          id="simple-select"
```
Aid19801, line 34: OOC, Any reason this isn't a data-testid?

Aid19801 (approved):

explorer/src/components/TableToolbar.tsx
tommyv1987: This was already baked into the framework from the looks of things, I just dropped the `demo` from the naming, because it isn't a demo anymore 😉

PR #891, opened 2021-11-10
remove delegation and undelegation from gateways

fmtabbara:


jstuczyn:
I'd argue that this one doesn't really matter, as web wallet only exists for the purposes of 0.11.0 milhon, where you can actually do gateway delegation stuff. In future release, where gateway delegation is not going to exist, we're only going to be supporting the desktop wallet anyway.

tommyv1987:
Totally agree with @jstuczyn here. Support will be done via the wallet only. I will check the product board, to see if there are tickets to make the wallet environments configurable. In the meantime, I'm going to merge this.

PR #898, opened 2021-11-16
Feature/batch delegator rewarding

jstuczyn:
At first glance this PR might seem like it's super bloated, but if you take closer look, a lot of LoC changes come from the fact that I split `transactions.rs` into `/transactions/mod.rs` and `/transactions/rewarding.rs`.

As what it PR does:
- removed `RewardMixnode` transaction message,
- changes `RewardMixnodeV2`, so that if node had more than a single page of delegators, rewarding parameters are saved and `RewardNextMixDelegators` can be called
- introduces `RewardNextMixDelegators` transaction that allows rewarding of next page of delegators. It can be called until exhaustion. Upon being finished statistics regarding this particular nodes are saved permanently -> this might be of interest to @durch . Currently there is no query to retrieve it, but it shouldn't be difficult to make one.
- Updated uptime calculation to use entire epoch of data rather than just 24h, however, I'm not 100% sure whether I didn't miss something in this one, some QA on it would be amazing.

What's needed:
- [ ] extensive QA. While unit tests pass, I don't feel super confident about deploying this to testnet / mainnet at this point. To properly test it, we would need bunch of nodes with high number of delegators,
- [ ] re-determine values for `MIXNODE_REWARD_OP_BASE_GAS_LIMIT` and `PER_MIXNODE_DELEGATION_GAS_INCREASE` for gas calculation
- [x] not sure if this is a problem, but I just came up with the potential issue => will the paging break if somebody decides to create fresh delegation in the middle of rewarding? Update: if page has exactly `MIXNODE_DELEGATORS_PAGE_LIMIT` delegators to reward and a fresh one is added, then it will push "old" delegator to new page that the validator API is not aware off. Solution: prevent fresh delegations whilst rewarding is in progress or, somehow..., retrieve event data from the executed transaction to figure out whether there exists next page.


common/mixnet-contract/src/mixnode.rs
```diff
@@ -110,14 +114,98 @@ impl NodeRewardParams {
     }
 
     pub fn one_over_k(&self) -> U128 {
-        U128::from_num(1) / U128::from_num(self.k.u128())
+        ONE / U128::from_num(self.k.u128())
```
futurechimp, line 117: Check on any possibility of division by zero here to ensure we can't panic

contracts/mixnet/src/transactions/rewarding.rs
```diff
@@ -0,0 +1,2053 @@
...
+                deps.as_mut(),
+                env.clone(),
+                mock_info(rewarding_validator_address.as_ref(), &[]),
+                1,
+            );
+            assert!(res.is_ok())
+        }
+
+        #[test]
+        fn cannot_be_called_if_rewarding_is_already_in_progress_with_little_day() {
```
futurechimp: Typo here

validator-api/src/rewarding/mod.rs
```diff
@@ -149,132 +139,117 @@ impl Rewarder {
...
+        current_rewarding_nonce: u32,
+    ) -> (Vec<MixnodeToReward>, Vec<MixnodeToReward>) {
+        let mut unrewarded = Vec::new();
+        let mut further_delegators_present = Vec::new();
+        for mix in eligible_mixnodes {
+            let rewarding_status = self
+                .nymd_client
+                .get_rewarding_status(mix.identity.clone(), current_rewarding_nonce)
+                .await
+                .expect("error handling");
```
futurechimp: Heheheh

validator-api/src/rewarding/mod.rs
```diff
@@ -149,132 +139,117 @@ impl Rewarder {
...
+            match rewarding_status.status {
+                // that case is super weird, it implies the node hasn't been rewarded at all!
+                // maybe the transaction timed out twice or something? In any case, we should attempt
+                // the reward for the final time!
+                None => unrewarded.push(mix.clone()),
```
futurechimp: Double check this to figure out whether there's any legitimate case that this can happen in, or whether we should error out of this.

futurechimp (approved):
Have a look at minor comments, but looks good!

PR #900, opened 2021-11-16
Vesting contract

durch:
TODO:

- [x] Spec out interface
- [x] Special undelegation/unbonding handling due to rewards
- [x] Node bonding/unbonding
- [x] Unit tests
- [ ] QAnet tests


contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+            .periods
+            .iter()
+            .map(|period| period.start_time)
+            .collect::<Vec<u64>>()
+            .binary_search(&block_time.seconds())
+        {
+            Ok(u) => u,
+            Err(u) => u,
+        }
+    }
```
neacsu: Will you be needing the next vesting period as well, as a separate function? Now it seems this is unnecessary, as only the current vesting period is actually used. Also, in the current form, there is a small chance this function returns the wrong value i.e. when the `block_time` is the same as the beginning of a period, so it returns the current period instead of the next one, and as such the function bellow returns the previous period instead of the current one. To solve that we could return `u + 1` on the `Ok` case.

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,369 @@
...
+                ));
+            }
+
+            let messages = vec![BankMsg::Send {
+                to_address: address.as_str().to_string(),
+                amount: vec![amount],
+            }
+            .into()];
+
+            let attributes = vec![attr("action", "whitdraw")];
```
neacsu:
```suggestion
            let attributes = vec![attr("action", "withdraw")];
```

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,369 @@
...
+    mix_identity: IdentityKey,
+    delegate_addr: String,
+    amount: Coin,
+    info: MessageInfo,
+    env: Env,
+    deps: DepsMut,
+) -> Result<Response, ContractError> {
+    if info.sender != ADMIN_ADDRESS {
+        return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
+    }
```
neacsu: I may not be getting the full picture of the story, but I don't really understand why the admin is the one that has to intermediate the delegation as well? Shouldn't it be the vesting account that can do delegations/undelegations, with sender control just like in `try_withdraw_vested_coins`?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        querier: Option<QuerierWrapper>,
+    ) -> Result<(), ContractError> {
+        if coin.amount < self.get_balance(storage) {
+            if let Some(querier) = querier {
+                let msg = MixnetExecuteMsg::DelegateToMixnodeOnBehalf {
+                    mix_identity: mix_identity.clone(),
+                    delegate_addr: self.address.clone(),
+                    coin: coin.clone(),
+                };
+                querier.query_wasm_smart(DEFAULT_MIXNET_CONTRACT_ADDRESS, &msg)?;
```
neacsu: This is more of a question than a comment: even though it's called `query_...`, it can still do `execute` calls, not only queries, right?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+                delegations
+            } else {
+                Vec::new()
+            };
+        delegations.push(DelegationData {
+            mix_identity,
+            amount: delegation_amount.amount,
+            block_time,
+        });
+        set_account_delegations(storage, &self.address, delegations)?;
```
neacsu: This looks a bit suboptimal, to re-write the entire vector every time. I think this could be done better with a multilevel bucket, otherwise costs would go up constantly for this function.

contracts/vesting/src/vesting.rs
durch: `next` was just a crutch to get `current`, its not super great ATM, I think I'll rework it and probably get rid of this one

.gitignore
```diff
@@ -33,4 +33,5 @@ contracts/mixnet/code_id
 contracts/mixnet/Justfile
 contracts/mixnet/Makefile
 validator-config
-*.patch
\ No newline at end of file
+*.patch
+validator-api-config.toml
```
jstuczyn, line 38: Why would the api config even exist here?

contracts/vesting/src/contract.rs
durch: You're right, at one point I was under the mistaken assumption that the transactions will be coming in from the admin address, and not from the vesting contract address, great catch

common/mixnet-contract/src/msg.rs
```diff
@@ -59,6 +58,15 @@ pub enum ExecuteMsg {
         // nonce of the current rewarding interval
         rewarding_interval_nonce: u32,
     },
+    DelegateToMixnodeOnBehalf {
+        mix_identity: IdentityKey,
+        delegate_addr: Addr,
+        coin: Coin,
```
jstuczyn: Why do we need the coin as a separate argument? When we send transaction itself, we attach some funds - I think that's a better way of doing it. Unless there's some very specific reason for this change

contracts/vesting/src/vesting.rs
durch: Thats an excellent question, and I'm not really sure, the docs don't say exactly, I'll have to run that by the `cosmwasm` folks

contracts/vesting/src/vesting.rs
durch: Yeah, it very much is, I just really don't like multilevel buckets, but you're right, this has the potential to really slow us down

contracts/mixnet/src/contract.rs
```diff
@@ -36,6 +36,8 @@ pub const DEFAULT_SYBIL_RESISTANCE_PERCENT: u8 = 30;
 // We'll be assuming a few more things, profit margin and cost function. Since we don't have relialable package measurement, we'll be using uptime. We'll also set the value of 1 Nym to 1 $, to be able to translate epoch costs to Nyms. We'll also assume a cost of 40$ per epoch(month), converting that to Nym at our 1$ rate translates to 40_000_000 uNyms
 pub const DEFAULT_COST_PER_EPOCH: u32 = 40_000_000;
 
+pub const VESTING_CONTRACT_ADDR: &str = "vesting-contract-address";
```
jstuczyn: Rather than being hardcoded in the code, should it perhaps be set on initilisation in the InitMsg?

contracts/mixnet/src/transactions.rs
```diff
@@ -656,6 +658,38 @@ pub(crate) fn try_delegate_to_mixnode(
...
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    mix_identity: IdentityKey,
+    delegate_addr: Addr,
+    coin: Coin,
+) -> Result<Response, ContractError> {
+    // check if the delegation contains any funds of the appropriate denomination
+    validate_delegation_stake(&[coin.clone()])?;
+    let amount = coin.amount;
```
jstuczyn: As I was suggesting above, why can't we just use data from the MessageInfo?

jstuczyn: Oh. Unless because it's coming from the vesting contract, we're not actually moving any coins... But then we end up in a weird state in the mixnet contract that we actually have more coins delegated than the actual funds in the contract....

contracts/mixnet/src/transactions.rs
```diff
@@ -693,9 +725,32 @@ pub(crate) fn try_remove_delegation_from_mixnode(
...
+    delegate_addr: Addr,
+) -> Result<Response, ContractError> {
+    if info.sender != VESTING_CONTRACT_ADDR {
+        return Err(ContractError::Unauthorized);
+    }
+
+    _try_remove_delegation_from_mixnode(deps, info, mix_identity, &delegate_addr)
+}
+
+fn _try_remove_delegation_from_mixnode(
```
jstuczyn: Is it not the case that if the vesting contract delegated on behalf of somebody else, that somebody will now be able to undelegate himself directly since he's going to be the correct "sender"?

Perhaps the vesting-related delegation should be stored elsewhere? (i.e. different bucket or something)

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,369 @@
...
+use config::defaults::DENOM;
+use cosmwasm_std::{
+    attr, entry_point, to_binary, BankMsg, Coin, Deps, DepsMut, Env, MessageInfo, QueryResponse,
+    Response, Timestamp, Uint128,
+};
+use mixnet_contract::IdentityKey;
+
+pub const NUM_VESTING_PERIODS: u64 = 8;
+pub const VESTING_PERIOD: u64 = 3 * 30 * 86400;
+pub const ADMIN_ADDRESS: &str = "admin";
```
jstuczyn, line 19: If it's admin for the contract, it's set implicitly when you initialise it.

contracts/vesting/src/errors.rs
```diff
@@ -0,0 +1,18 @@
...
+    #[error("Account does not exist - {0}")]
+    NoAccountForAddress(String),
+    #[error("Only admin can perform this action, {0} is not admin")]
+    NotAdmin(String),
+    #[error("Balance not found for existing account ({0}), this is a bug")]
+    NoBalanceForAddress(String),
+    #[error("Insufficient balance")]
+    InsufficientBalance(String, u128),
+    #[error("Insufficient spendable balance")]
+    InsufficientSpendable(String, u128),
```
jstuczyn, line 18: Since the errors have additional information stored, shouldn't we also print them?

contracts/vesting/src/storage.rs
```diff
@@ -0,0 +1,90 @@
...
+    bucket_read(storage, PREFIX_ACCOUNT_BALANCE)
+}
+
+fn account_balance_mut(storage: &mut dyn Storage) -> Bucket<Uint128> {
+    bucket(storage, PREFIX_ACCOUNT_BALANCE)
+}
+
+pub fn get_account(storage: &dyn Storage, address: &Addr) -> Option<PeriodicVestingAccount> {
+    // Due to using may_load this should be safe to unwrap
+    accounts(storage).may_load(address.as_bytes()).unwrap()
```
jstuczyn: Are we sure the unwrap here is fine? Shouldn't we instead wrap it all in a `StdResult`?

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,369 @@
...
+) -> Result<Response, ContractError> {
+    let address = info.sender;
+    if let Some(account) = get_account(deps.storage, &address) {
+        let spendable_coins = account.spendable_coins(None, &env, deps.storage);
+        if amount.amount < spendable_coins.amount {
+            if let Some(balance) = get_account_balance(deps.storage, &address) {
+                let new_balance = balance.u128().saturating_sub(amount.amount.u128());
+                set_account_balance(deps.storage, &address, Uint128(new_balance))?;
+            } else {
+                return Err(ContractError::NoBalanceForAddress(
```
jstuczyn: Is this error even possible? Isn't it the case we just checked that we have more spendable_coins than the requested amount and hence we must have some balance?

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,369 @@
...
+    env: Env,
+    info: MessageInfo,
+    deps: DepsMut,
+) -> Result<Response, ContractError> {
+    let address = info.sender;
+    if let Some(account) = get_account(deps.storage, &address) {
+        let spendable_coins = account.spendable_coins(None, &env, deps.storage);
+        if amount.amount < spendable_coins.amount {
+            if let Some(balance) = get_account_balance(deps.storage, &address) {
+                let new_balance = balance.u128().saturating_sub(amount.amount.u128());
```
jstuczyn: Are we sure `saturating_sub` is the way to go? Isn't it the case we must ensure overflow is IMPOSSIBLE?

contracts/vesting/src/contract.rs
jstuczyn: That's a very good question.

contracts/vesting/src/vesting.rs
durch: I have a thread there, there is also `wasm_execute` that is probably a better candidate for the job here...

contracts/vesting/src/storage.rs
```diff
@@ -0,0 +1,90 @@
...
+    storage: &mut dyn Storage,
+    account: PeriodicVestingAccount,
+) -> Result<(), ContractError> {
+    Ok(accounts_mut(storage).save(account.address().as_bytes(), &account)?)
+}
+
+pub fn get_account_delegations(
+    storage: &dyn Storage,
+    address: &Addr,
+) -> Option<Vec<DelegationData>> {
```
jstuczyn: This Vec here is a big red flag for me, because it kinda indicates the amount of data here is unbounded and this may blow up.

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+pub trait VestingAccount {
+    // locked_coins returns the set of coins that are not spendable (can still be delegated tough) (i.e. locked),
+    // defined as the vesting coins that are not delegated.
+    //
+    // To get spendable coins of a vesting account, first the total balance must
+    // be retrieved and the locked tokens can be subtracted from the total balance.
+    // Note, the spendable balance can be negative.
```
jstuczyn: Why could the spendable balance be negative?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+}
+
+pub trait DelegationAccount {
+    fn try_delegate_to_mixnode(
+        &self,
+        mix_identity: IdentityKey,
+        amount: Coin,
+        env: &Env,
+        storage: &mut dyn Storage,
+        // There isn't a trait that unifies Queriers so we clutch by passing in the option here, so we'll query
```
jstuczyn: I think there was something nicer in cosmwasm 1.0. Maybe we should just update and use that?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        Ok(account)
+    }
+
+    pub fn address(&self) -> Addr {
+        self.address.clone()
+    }
+
+    pub fn tokens_per_period(&self) -> u128 {
+        // Remainder tokens will be lumped into the last period.
+        self.coin.amount.u128() / NUM_VESTING_PERIODS as u128
```
jstuczyn: Just a sanity check -> we're fine that this is an integer division?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+    pub fn end_time(&self) -> Timestamp {
+        Timestamp::from_seconds(self.start_time + VESTING_PERIOD)
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct PeriodicVestingAccount {
+    address: Addr,
+    start_time: Timestamp,
+    periods: Vec<VestingPeriod>,
```
jstuczyn: This makes me slightly nervous. In terms of orders of magnitude, how much of data are we looking at here?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+            };
+        }
+
+        let end_time = if period >= NUM_VESTING_PERIODS as usize {
+            u64::MAX
+        } else {
+            self.periods[period].end_time().seconds()
+        };
+
+        if let Some(delegations) = get_account_delegations(storage, &self.address) {
```
jstuczyn: A big red-flag from me -> what if the account has A LOT of delegations? Say in the order of hundreds, which is possible. This function will blow up then.

contracts/vesting/src/vesting.rs
jstuczyn: `query_wasm_smart` CAN'T do execute calls. Only queries.
Instead, you can do execute calls via `Msg` and `SubMsg`. And with the latter you can catch errors from the contract. Though I think you need to add additional method to the contract to do it. I don't remember the details.

neacsu: I think so, yeah. This can't really work as you're not sending funds anywhere for the delegation to happen. And in order to send those funds, there is going to be a hard problem of where do these funds come from. From the vested or the unvested tokens? What happens when you enter a new period and other tokens become liquid as well?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+}
+
+impl VestingAccount for PeriodicVestingAccount {
+    fn locked_coins(
+        &self,
+        block_time: Option<Timestamp>,
+        env: &Env,
+        storage: &dyn Storage,
+    ) -> Coin {
+        // Returns 0 in case of underflow.
```
jstuczyn: Under what conditions could it possible underflow? Shouldn't we be accounting for every single token?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        &self,
+        block_time: Option<Timestamp>,
+        env: &Env,
+        storage: &dyn Storage,
+    ) -> Coin {
+        Coin {
+            amount: Uint128(
+                self.get_balance(storage)
+                    .u128()
+                    .saturating_sub(self.locked_coins(block_time, env, storage).amount.u128()),
```
jstuczyn: Again, why is it a `saturating_sub`. Could it ever fail?

jstuczyn: Also, in some other location you said that spendable balance could be negative. However, this code seems to contradict it as you can't store negative values in the `Coin`

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        let block_time = block_time.unwrap_or(env.block.time);
+        let period = self.get_current_vesting_period(block_time);
+
+        match period {
+            // We're in the first period, or the vesting has not started yet.
+            0 => Coin {
+                amount: Uint128(0),
+                denom: DENOM.to_string(),
+            },
+            1..=7 => Coin {
```
jstuczyn: Why are period 1-7 treated differently? Do we have only 8 of them in total or something?
If that's the case, perhaps this value should be defined as a constant / state value?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        &self,
+        block_time: Option<Timestamp>,
+        env: &Env,
+        storage: &dyn Storage,
+    ) -> Coin {
+        self.get_delegated_with_op(&ge, block_time, env, storage)
+    }
+}
+
+fn lt(x: u64, y: u64) -> bool {
```
jstuczyn: What's wrong with the lt and ge from the standard library?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+        let end_time = if period >= NUM_VESTING_PERIODS as usize {
+            u64::MAX
+        } else {
+            self.periods[period].end_time().seconds()
+        };
+
+        if let Some(delegations) = get_account_delegations(storage, &self.address) {
+            delegations
+                .iter()
+                .filter(|d| op(d.block_time.seconds(), end_time))
```
jstuczyn: I don't fully understand the purpose of passing the `lt` / `ge` op here. Couldn't we put it directly in the code here instead?

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+pub fn populate_vesting_periods(start_time: u64, n: u64) -> Vec<VestingPeriod> {
+    let mut periods = Vec::new();
+    // There are eight 3 month periods in two years
+    for i in 0..n {
+        let period = VestingPeriod {
+            start_time: start_time + i * VESTING_PERIOD,
+        };
+        periods.push(period);
+    }
+    periods.shrink_to_fit();
```
jstuczyn: Rather than having to shrink to fit, why not create the `periods` with appropriate capacity from the start, i.e. `Vec::with_capacity(n as usize)`? Alternatively create the vector from Iterator

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+    use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
+    use cosmwasm_std::{Addr, Coin, Timestamp, Uint128};
+
+    use super::DelegationAccount;
+
+    #[test]
+    fn test_account_creation() {
+        let mut deps = init_contract();
+        let env = mock_env();
+        let account = vesting_account_fixture(&mut deps.storage, &env);
```
jstuczyn: This test function kinda feels as it should be testing the actual constructor for the vesting account rather than using the fixture

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+            Uint128(5 * account.get_original_vesting().amount.u128() / NUM_VESTING_PERIODS as u128)
+        );
+        assert_eq!(
+            vesting_coins.amount,
+            Uint128(
+                account.get_original_vesting().amount.u128()
+                    - 5 * account.get_original_vesting().amount.u128()
+                        / NUM_VESTING_PERIODS as u128
+            )
+        );
```
jstuczyn: It feels like it could be useful to have one additional case here of when we reach the final period (or maybe also for one before it)

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    #[test]
+    fn test_toodelegations() {
```
jstuczyn: `too`?

jstuczyn (commented):
Thing that scares me the most is my arch nemesis - unbounded call for the delegations...

.gitignore
durch: I have a symlink to the config locally for easier editing when I'm tinkering with it

contracts/mixnet/src/transactions.rs
durch: Yup, I've just corrected those

contracts/vesting/src/vesting.rs
durch: The thing is that the function logic is identical except for the comparison operator, so this puts all that code in one place, so maintenance should be simpler

contracts/vesting/src/vesting.rs
durch: Good call

contracts/vesting/src/vesting.rs
durch: Yup, I just figured that out, as I broke the constructor msg, but my tests were still passing :)

contracts/mixnet/src/transactions.rs
durch: I've been thinking through that, undelegations also have to go back to the vesting contract, as ppl can delegate unvested coins, so we can't let them withdraw them. So the flow is for folks to execute against the vesting contract and then withdraw what they can once the funds get back into it.

contracts/mixnet/src/contract.rs
durch: Yes! Great call

common/mixnet-contract/src/msg.rs
durch: We don't, corrected

contracts/vesting/src/vesting.rs
durch: toooooooo ;)

contracts/vesting/src/contract.rs
durch: Totally forgot about that, although we once neglected to do that, would it make sense to check if its set at Init and panic if it was not set?

contracts/vesting/src/errors.rs
durch: We really should :)

contracts/vesting/src/storage.rs
durch: `may_load` should not fail unless storage is somehow broken, I'll add an error case for that, good call

contracts/vesting/src/contract.rs
durch: It shouldn't be, but I'd rather handle it then leave it dangling

contracts/vesting/src/contract.rs
jstuczyn: Now that I'm looking slightly deeper into the cosmwasm, I don't think we can easily check whether the admin was set and who it is. I guess in our case it's probably going the be the governance account

contracts/vesting/src/contract.rs
durch: spendable_coins are guaranteed to be <= balance, so that sub should always be >= 0, I'll add a comment there

contracts/vesting/src/storage.rs
durch: I'll refactor that to use the multilevel bucket

contracts/vesting/src/vesting.rs
durch: That is from cosmos interface spec, my interpretation is that since delegations subtract from the balance but still count in the locked tokens that can go <= 0. I plan to double check that in their Go code

contracts/vesting/src/vesting.rs
durch: `wasm_execute` solves this one, I don't wanna go to 1.0 just yet, as I'd like to deploy this to QAnet

contracts/vesting/src/vesting.rs
durch: I think so, `NUM_VESTING_PERIOD` will always be a lot smaller then the vesting amount, and the remainders will get freed in the last vesting period. I'll add a check to verify this assumption

contracts/vesting/src/vesting.rs
durch: There will always be only 8 `VestingPeriods`, maybe I can just use an Array there

contracts/vesting/src/vesting.rs
durch: Your arch-nemesis :), I'll switch to a paged implementation, like we have in the mixnet contract

contracts/vesting/src/vesting.rs
durch: Thats just a comment to draw attention to`saturating_sub`, in practice `delegated_vesting` can never be greater then `total_vesting`, I'll add a test around this

contracts/vesting/src/vesting.rs
durch: `saturating_sub` will never fail, in case of an underflow it goes to `0`. The balance itself can be negative, but I don't think we care if its negative or 0, in either case you have nothing to spend.

contracts/vesting/src/vesting.rs
durch: Yeah, we always have 8 vesting period, there is a `NUM_VESTING_PERIOD` constant, I'll substitute it here.

contracts/vesting/src/vesting.rs
durch: Nothing, but its a method on a type, so this is only a proxy to call it as a function

contracts/mixnet/src/transactions.rs
jstuczyn: But, unless I misunderstood it, the delegation information [inside the mixnet contract] is saved using the "delegator's" address. So in principle nothing is preventing the following to happen:
- user delegates bunch of tokens to some mixnode using the vesting contract
- now he calls the mixnet contract directly to undelegate it all
- he skipped all the vesting

contracts/vesting/src/vesting.rs
durch: We could write this differently, parametrize a boolean, and then decide on which op to use based on that param

contracts/vesting/src/vesting.rs
durch: Turns out I can't, adding a comment instead

contracts/vesting/src/vesting.rs
durch: I forgot how unflexible Arrays are, going with Vec

contracts/vesting/src/contract.rs
durch: Interesting, so you're saying leave this as is, maybe put it in the InitMsg...

durch: Altough that makes it more difficult to change later, if its a const a simple governance vote can change, if its in a State we need to worry about migrations as well...

contracts/vesting/src/vesting.rs
durch: vested vs unvested is not important here, everything can be delegated, [this](https://github.com/nymtech/nym/tree/vesting-staking-cointract/contracts/vesting#vesting-coin-delegation-flow) is how the flow should go

contracts/mixnet/src/transactions.rs
durch: Thats a beautiful workaround, yeah, I think the best is to store it in a different bucket, love that paranoid mind of yours :)

durch: @jstuczyn Thinking it through some more, saving them to different buckets would introduce complexity throughout the stack, as we'd have to double everything we're doing for rewarding as well, I'm thinking of putting in a flag into the RawDelegationData that specifies if the delegation can be withdrawn by the owner...

contracts/vesting/src/vesting.rs
neacsu: But how do you know from what pool to send tokens do be delegated? In the current PR, I don't see there any funds being sent on `DelegateToMixnodeOnBehalf`, but I don't think that call would work, as the mixnet contract expects the funds that will be delegated to be specified in the execute message.

contracts/mixnet/src/transactions.rs
durch: Thinking it through some more, that would entail a migration, and a lot of updates, still not sure which path to take

contracts/vesting/src/vesting.rs
durch: [Here](https://github.com/nymtech/nym/blob/vesting-staking-cointract/contracts/vesting/src/vesting.rs#L243) is where we add funds to the message and then return it as a `Response`, which as far as I understand should send the message out

contracts/mixnet/src/transactions.rs
durch: Since an additional problem of the undelegation procedure is that we have to know where to send the funds, and in the case of `proxy` delegations this is the vesting account, I think it makes most sense to add an `Option<Addr>` field to the `RawDelegationData` struct. That way we can both know if the delegation is a proxy delegation, and where to send the funds on undelegate...

durch: @jstuczyn I'm pretty happy with how this looks now, let me know what you think

contracts/mixnet/src/contract.rs
durch: Altough again, do we maybe want to leave this up to the governance, in which case it might be better to have a `const` in the contract

contracts/vesting/src/vesting.rs
durch: @neacsu, @jstuczyn I've been noodling on this some more, and the most common case here is iterating through all delegations to find which are vesting/vested, for that one vector is the ideal data structure. As far as a performance hit, even hundreds of delegations should not take a lot of storage space, so read/write ops should be fairly fast. I'd like to benchmark this somehow, maybe implement pages as vectors instead a multilevel bucket, which would be pretty nasty to work with, and would bring along at lest N reads for every time we want to figure out vested/vesting tokens, as well as any derived function, such as `delegated_vesting/delegated_free/locked/spendable`. I'll bring this to cosmwasm folks before doing any changes, as I'd really like to understand the constraints involved... Dissenting opinions more the welcome of coourse :)

contracts/mixnet/src/mixnodes/bonding_transactions.rs
```diff
@@ -2,34 +2,73 @@
...
+
+    if proxy != VESTING_CONTRACT_ADDR {
+        return Err(ContractError::Unauthorized);
+    }
+
+    let minimum_bond =
+        mixnet_params_storage::read_contract_settings_params(deps.storage).minimum_mixnode_bond;
+    validate_mixnode_bond(&info.funds, minimum_bond)?;
+
+    _try_add_mixnode(deps, env, mix_node, info.funds[0].clone(), owner, None)
```
neacsu:
```suggestion
    _try_add_mixnode(deps, env, mix_node, info.funds[0].clone(), owner, Some(proxy))
```

contracts/mixnet/src/mixnodes/storage.rs
```diff
@@ -196,8 +214,9 @@ pub fn mix_delegations_read<'a>(
 }
 
 // TODO: note for JS when doing a deep review for the contract. Don't store it as (), instead do it as u8
-pub fn reverse_mix_delegations<'a>(storage: &'a mut dyn Storage, owner: &Addr) -> Bucket<'a, ()> {
-    Bucket::multilevel(storage, &[PREFIX_REVERSE_MIX_DELEGATION, owner.as_bytes()])
+// Using bytes here so that we can handle proxy delegations more easily as their key is formed as proxy_addr ^ delegation_addr
```
neacsu: How are we going to guard against collisions? e.g. two different pairs (proxy1, delegation1) and (proxy2, delegation2) that give the same result when XORing

contracts/mixnet/src/mixnodes/storage.rs
```diff
@@ -91,6 +95,19 @@ impl Display for StoredMixnodeBond {
     }
 }
 
+pub fn generate_storage_key(address: &Addr, proxy: Option<&Addr>) -> Vec<u8> {
+    if let Some(proxy) = &proxy {
+        address
+            .as_bytes()
+            .iter()
+            .zip(proxy.as_bytes())
+            .map(|(x, y)| x ^ y)
```
neacsu: What if the two addresses are not the same length? Is there any check that makes sure they have the bech32 specific length? Even so, it might make some sense to check that the two are equal in length (although I don't really understand why derive the storage key by XORing)

contracts/mixnet/src/mixnodes/delegation_transactions.rs
```diff
@@ -89,42 +130,88 @@ pub(crate) fn try_remove_delegation_from_mixnode(
...
+            let mut response = Response::new().add_message(return_tokens);
+
+            if let Some(proxy) = &proxy {
+                let msg = Some(VestingContractExecuteMsg::TrackUndelegation {
+                    owner: delegate.to_owned(),
+                    mix_identity: mix_identity.clone(),
+                    amount: coins(delegation.amount.u128(), DENOM)[0].clone(),
+                });
+
+                response = response.add_message(wasm_execute(proxy, &msg, coins(0, DENOM))?);
```
neacsu: Small nit picking: Maybe do the wasm_execute separately, like you did on bonding?

contracts/vesting/src/support/tests.rs
```diff
@@ -0,0 +1,47 @@
...
+    use cosmwasm_std::testing::mock_dependencies;
+    use cosmwasm_std::testing::mock_env;
+    use cosmwasm_std::testing::mock_info;
+    use cosmwasm_std::testing::MockApi;
+    use cosmwasm_std::testing::MockQuerier;
+    use cosmwasm_std::Addr;
+    use cosmwasm_std::Coin;
+    use cosmwasm_std::OwnedDeps;
+    use cosmwasm_std::Uint128;
+    use cosmwasm_std::{Empty, Env, MemoryStorage, Storage};
```
neacsu: nit: Most of this can be collapsed into fewer `use`s

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,765 @@
...
+                self.address.as_str().to_string(),
+            ));
+        } else {
+            BondData {
+                block_time,
+                amount: bond.amount,
+            }
+        };
+
+        let new_balance = Uint128::new(self.load_balance(storage)?.u128() - bond.amount.u128());
```
neacsu: `load_balance` is already invoked in the caller function? Maybe it would make sense to pass that as an argument to this function, to diminish the costs? (I'm not sure if any caching is done that already does that)

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+use config::defaults::{DEFAULT_MIXNET_CONTRACT_ADDRESS, DENOM};
+use cosmwasm_std::{
+    entry_point, to_binary, Addr, BankMsg, Coin, Deps, DepsMut, Env, MessageInfo, QueryResponse,
+    Response, Storage, Timestamp, Uint128,
+};
+use cw_storage_plus::Map;
+use mixnet_contract::{IdentityKey, MixNode};
+
+pub const NUM_VESTING_PERIODS: usize = 8;
+pub const VESTING_PERIOD: u64 = 3 * 30 * 86400;
```
neacsu, line 17: I know what this is, but a comment on the values would be nice anyway.

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,765 @@
...
+            existing_delegation + delegation.amount
+        } else {
+            delegation.amount
+        };
+
+        self.delegations()
+            .save(storage, delegation_key, &new_delegation)?;
+
+        let new_balance =
+            Uint128::new(self.load_balance(storage)?.u128() - delegation.amount.u128());
```
neacsu: Same idea for `load_balance` as in `track_bond`

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+fn try_delegate_to_mixnode(
+    mix_identity: IdentityKey,
+    info: MessageInfo,
+    env: Env,
+    deps: DepsMut,
+) -> Result<Response, ContractError> {
+    let delegate_addr = info.sender;
+    let amount = validate_funds(&info.funds)?;
+    let address = deps.api.addr_validate(delegate_addr.as_str())?;
+    if let Some(account) = load_account(&address, deps.storage)? {
```
neacsu: All of the `if let Some(account) = ...` could be converted to `let account = ... .ok_or(err)?` and lose a level of imbrication (in other functions like `try_withdraw_vested_coins`, the advantage is even more visible)

neacsu (changes requested):

contracts/vesting/src/support/tests.rs
durch: thats a good nit, we should keep the code tidy

contracts/vesting/src/vesting.rs
durch: Good call

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,385 @@
...
+            .saturating_sub(amount.amount.u128());
+        account.save_balance(Uint128::new(new_balance), deps.storage)?;
+
+        let send_tokens = BankMsg::Send {
+            to_address: address.as_str().to_string(),
+            amount: vec![amount],
+        };
+
+        Ok(Response::new()
+            .add_attribute("action", "whitdraw")
```
neacsu, line 115: Something got lost in the rebase :D

contracts/vesting/src/vesting.rs
```diff
@@ -0,0 +1,894 @@
...
+        let block_time = block_time.unwrap_or(env.block.time);
+        let period = self.get_current_vesting_period(block_time);
+        let max_vested = self.tokens_per_period()? * period as u128;
+        let start_time = self.periods[period].start_time;
+
+        let delegations_keys = self
+            .delegations()
+            .keys_de(storage, None, None, Order::Ascending)
+            .scan((), |_, x| x.ok())
+            .filter(|(_mix, block_time)| block_time < &start_time)
```
neacsu: Instead of adding a reference and letting the compiler auto-dereference, you could directly dereference it yourself.
```suggestion
            .filter(|(_mix, block_time)| *block_time < start_time)
```

contracts/mixnet/src/mixnodes/storage.rs
durch: I think the chances of it are quite small, but I'm open to ideas :/

contracts/mixnet/src/mixnodes/storage.rs
durch: The `^` will work on the shortest length of the two, so resulting byte slice will have the length of the shorter slice

contracts/mixnet/src/mixnodes/bonding_transactions.rs
durch: Great catch!

contracts/mixnet/src/mixnodes/storage.rs
durch: To quantify small, birthday paradox says upper bound of collision risk is `1/2^(N/2)`. Address I have laying around has 42 bytes, so assuming a 42 byte output we'd get a collision once every 2.1 million encodings, not great, no terrible :)

neacsu: Yeah, I think you're right, I haven't realized that the proxy address is actually kind of fixed (either `None` or that fixed contract address), which I think limits the potential of collision quite much.

neacsu (approved):

common/mixnet-contract/src/msg.rs
```diff
@@ -50,12 +49,26 @@ pub enum ExecuteMsg {
...
     },
-
     RewardNextMixDelegators {
         mix_identity: IdentityKey,
         // nonce of the current rewarding interval
         rewarding_interval_nonce: u32,
     },
+    DelegateToMixnodeOnBehalf {
+        mix_identity: IdentityKey,
+        delegate: Addr,
```
jstuczyn: Since those values are coming directly from user input and are not validated in any way, we want to use a simple `String` rather than `Addr` (this is also recommended by cosmwasm). If we really need an `Addr`, we can obtain it via validation with `deps.api.addr_validate(...)`

This is also something I've done in the most recent changes to the mixnet contract.

contracts/mixnet/src/contract.rs
```diff
@@ -42,6 +42,8 @@ pub const DEFAULT_SYBIL_RESISTANCE_PERCENT: u8 = 30;
 // We'll be assuming a few more things, profit margin and cost function. Since we don't have reliable package measurement, we'll be using uptime. We'll also set the value of 1 Nym to 1 $, to be able to translate epoch costs to Nyms. We'll also assume a cost of 40$ per epoch(month), converting that to Nym at our 1$ rate translates to 40_000_000 uNyms
 pub const DEFAULT_COST_PER_EPOCH: u32 = 40_000_000;
 
+pub const VESTING_CONTRACT_ADDR: &str = "";
```
jstuczyn: What do you think about passing it in an `InitMsg` so that we wouldn't need to remember to update this constant before deployment?

jstuczyn: But then... Isn't it a chicken and an egg problem? Vesting contract needs to know about mixnet contract and mixnet contract needs to know about the vesting contract...

contracts/mixnet/src/mixnodes/bonding_transactions.rs
```diff
@@ -39,73 +85,110 @@ pub(crate) fn try_add_mixnode(
...
-pub(crate) fn try_remove_mixnode(
+pub fn try_remove_mixnode_on_behalf(
     deps: DepsMut,
     info: MessageInfo,
+    owner: Addr,
 ) -> Result<Response, ContractError> {
-    let sender_bytes = info.sender.as_bytes();
+    let proxy = info.sender;
+
+    if proxy != VESTING_CONTRACT_ADDR {
```
jstuczyn: This is something I've discovered recently, but `cosmwasm` exposes 3 nice macros: `ensure!`, `ensure_eq!` and `ensure_ne!. This is basically me sharing this discovery and if you wanted, you could rewrite this function into
```rust
    ensure!(proxy == VESTING_CONTRACT_ADDR, ContractError::Unauthorized);
    _try_remove_mixnode(deps, owner, Some(proxy))
```

contracts/mixnet/src/mixnodes/bonding_transactions.rs
```diff
@@ -39,73 +85,110 @@ pub(crate) fn try_add_mixnode(
...
+                owner,
+                amount: coins(mixnode_bond.bond_amount.amount.u128(), DENOM)[0].clone(),
+            };
+
+            let track_unbond_message = wasm_execute(proxy, &msg, coins(0, DENOM))?;
+            response = response.add_message(track_unbond_message);
+        }
+
+        Ok(response)
+    } else {
```
jstuczyn: Would it make sense to collapse this entire if statement by moving the `else` branch into its separate thing, i.e.
```rust
if proxy != mixnode_bond.proxy {
  return Err(...)
}
```
and then continue with the happy path

contracts/mixnet/src/mixnodes/bonding_transactions.rs
```diff
@@ -39,73 +85,110 @@ pub(crate) fn try_add_mixnode(
...
+
+        let mut response = Response::new()
+            .add_message(return_tokens)
+            .add_attribute("action", "unbond")
+            .add_attribute("mixnode_bond", mixnode_bond.to_string());
+
+        if let Some(proxy) = &proxy {
+            let msg = VestingContractExecuteMsg::TrackUnbond {
+                owner,
+                amount: coins(mixnode_bond.bond_amount.amount.u128(), DENOM)[0].clone(),
```
jstuczyn: Even though in theory we should be using `nym` currency everywhere, I still feel slightly uneasy with that `DENOM` constant here. Couldn't we instead use something like:
```rust
amount: mixnode_bond.bound_amount.clone()
```
?

contracts/mixnet/src/mixnodes/bonding_transactions.rs
```diff
@@ -39,73 +85,110 @@ pub(crate) fn try_add_mixnode(
...
+            };
+
+            let track_unbond_message = wasm_execute(proxy, &msg, coins(0, DENOM))?;
```
jstuczyn: Are we sure we don't want to be sending some tokens back to the vesting contract (i.e. presumably the difference between initial bond and reward)?
Say some somebody bonded its mixnode via vesting contract. It was running for a while **and** got rewarded.
Those "new" tokens don't exist anymore on the vesting contract side, but they were removed from the mixnet reward pool.

jstuczyn: The problem I'm trying to solve is what happens after vesting finishes and everyone reclaims their  tokens -> there won't be enough of them.

contracts/mixnet/src/mixnodes/delegation_transactions.rs
```diff
@@ -89,42 +130,90 @@ pub(crate) fn try_remove_delegation_from_mixnode(
...
+            let mut response = Response::new().add_message(return_tokens);
+
+            if let Some(proxy) = &proxy {
+                let msg = Some(VestingContractExecuteMsg::TrackUndelegation {
+                    owner: delegate.to_owned(),
+                    mix_identity: mix_identity.clone(),
+                    amount: coins(delegation.amount.u128(), DENOM)[0].clone(),
+                });
+
+                let track_undelegation_msg = wasm_execute(proxy, &msg, coins(0, DENOM))?;
```
jstuczyn: Basically the same concern as with unbonding -> don't we need to transfer "rewarded" tokens?

contracts/mixnet/src/mixnodes/delegation_transactions.rs
```diff
@@ -50,9 +88,11 @@ pub(crate) fn try_delegate_to_mixnode(
         });
     }
 
+    let storage_key = storage::generate_storage_key(delegate, proxy.as_ref());
```
jstuczyn: Why can't we use the `delegate` directly?

contracts/mixnet/src/mixnodes/storage.rs
```diff
@@ -91,6 +95,19 @@ impl Display for StoredMixnodeBond {
     }
 }
 
+pub fn generate_storage_key(address: &Addr, proxy: Option<&Addr>) -> Vec<u8> {
```
jstuczyn: cosmwasm-plus storage will make this function redundant : )

contracts/vesting/src/contract.rs
```diff
@@ -0,0 +1,352 @@
...
+            mix_identity,
+            amount,
+        } => try_track_undelegation(owner, mix_identity, amount, info, deps),
+        ExecuteMsg::BondMixnode { mix_node } => try_bond_mixnode(mix_node, info, env, deps),
+        ExecuteMsg::UnbondMixnode {} => try_unbond_mixnode(info, deps),
+        ExecuteMsg::TrackUnbond { owner, amount } => try_track_unbond(owner, amount, deps),
+    }
+}
+
+pub fn try_bond_mixnode(
```
jstuczyn, line 63: Would it perhaps make sense to move all transactions into a separate file as `contract.rs` is already quite lengthy?

contracts/vesting/src/storage.rs
```diff
@@ -0,0 +1,30 @@
...
+    load_account(address, storage)?
+        .ok_or_else(|| ContractError::NoAccountForAddress(address.as_str().to_string()))
+}
+
+pub fn account_from_address(
+    address: &Addr,
+    storage: &dyn Storage,
+    api: &dyn Api,
+) -> Result<Account, ContractError> {
+    validate_account(&api.addr_validate(address.as_str())?, storage)
```
jstuczyn: If you're using an `Addr` type, you don't need to perform `.addr_validate`, as the address is already validated. (this was performed before your transaction hit the `execute`)

contracts/vesting/src/messages.rs
```diff
@@ -0,0 +1,78 @@
...
+pub enum ExecuteMsg {
+    DelegateToMixnode {
+        mix_identity: IdentityKey,
+    },
+    UndelegateFromMixnode {
+        mix_identity: IdentityKey,
+    },
+    CreateAccount {
+        address: String,
+        start_time: Option<u64>,
```
jstuczyn, line 22: Shouldn't this be perhaps an `Option<Timestamp>`?

contracts/vesting/src/traits/vesting_account.rs
```diff
@@ -0,0 +1,65 @@
+use crate::errors::ContractError;
+use cosmwasm_std::{Coin, Env, Storage, Timestamp};
+
+pub trait VestingAccount {
+    // locked_coins returns the set of coins that are not spendable (can still be delegated tough) (i.e. locked),
+    // defined as the vesting coins that are not delegated.
+    //
+    // To get spendable coins of a vesting account, first the total balance must
+    // be retrieved and the locked tokens can be subtracted from the total balance.
+    // Note, the spendable balance can be negative.
```
jstuczyn, line 10: I feel like I've asked this question before, but I forgot the answer to it -> under what condition can the spendable balance be negative, especially if it's represented by an **unsigned int128**?

contracts/vesting/src/vesting/account/bonding_account.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.save_balance(new_balance, storage)?;
+        self.save_bond(bond_data, storage)?;
+
+        Ok(Response::new()
+            .add_attribute("action", "bond mixnode on behalf")
+            .add_message(bond_mixnode_mag))
+    }
+
+    fn try_unbond_mixnode(&self) -> Result<Response, ContractError> {
+        let msg = MixnetExecuteMsg::UnbondMixnodeOnBehalf {
```
jstuczyn, line 56: We could perform a slight optimisation: before calling the mixnet contract, check whether this account even has a node bonded on our behalf. If they don't, we don't have to do anything.

contracts/vesting/src/vesting/account/delegating_account.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(Response::new()
+            .add_attribute("action", "delegate to mixnode on behalf")
+            .add_message(delegate_to_mixnode))
+    }
+
+    fn try_undelegate_from_mixnode(
+        &self,
+        mix_identity: IdentityKey,
+    ) -> Result<Response, ContractError> {
+        let msg = MixnetExecuteMsg::UndelegateFromMixnodeOnBehalf {
```
jstuczyn, line 52: Same thing as with bonding, first check whether any delegation even exists.

contracts/vesting/src/vesting/account/bonding_account.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fn track_unbond(&self, amount: Coin, storage: &mut dyn Storage) -> Result<(), ContractError> {
+        let new_balance = Uint128::new(self.load_balance(storage)?.u128() + amount.amount.u128());
+        self.save_balance(new_balance, storage)?;
+
+        self.remove_bond(storage)?;
+        Ok(())
+    }
+}
```
jstuczyn, line 80: Maybe I missed something, but don't we need a `track_bond` function considering we do have a `track_delegation` and `track_undelegation`?

contracts/vesting/src/vesting/account/delegating_account.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn track_delegation(
+        &self,
+        block_time: Timestamp,
+        mix_identity: IdentityKey,
+        current_balance: Uint128,
+        delegation: Coin,
+        storage: &mut dyn Storage,
+    ) -> Result<(), ContractError> {
+        let delegation_key = (mix_identity.as_bytes(), block_time.seconds());
```
jstuczyn, line 78: You perhaps want to call a `.joined_key()` on this sucker if it's really a composite key to your map

contracts/vesting/src/vesting/account/delegating_account.rs
```diff
@@ -0,0 +1,117 @@
...
+        amount: Coin,
+        storage: &mut dyn Storage,
+    ) -> Result<(), ContractError> {
+        let mix_bytes = mix_identity.as_bytes();
+
+        // Iterate over keys matching the prefix and remove them from the map
+        let block_times = self
+            .delegations()
+            .prefix_de(mix_bytes)
+            .keys_de(storage, None, None, Order::Ascending)
```
jstuczyn, line 110: 1. 🟥 unbounded call -> we go through ALL delegations towards this mixnode
2. (this is a legit question): do we want to iterate via `keys()` or via `range()` on this  prefix?

contracts/vesting/src/vesting/account/delegating_account.rs
```diff
@@ -0,0 +1,117 @@
...
+            // Scan will blow up on first error
+            .scan((), |_, x| x.ok())
```
jstuczyn, line 112: Do we want to blow up? Why not just collect to `StdResult<Vec<u64>>` instead?

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        Ok(self
+            .delegations_for_mix(mix, storage)?
+            .iter()
+            .fold(Uint128::zero(), |acc, x| acc + *x))
+    }
+
+    pub fn total_delegations(&self, storage: &dyn Storage) -> Result<Uint128, ContractError> {
+        Ok(self
+            .delegations()
+            .range(storage, None, None, Order::Ascending)
```
jstuczyn, line 187: 🟥  unbounded call

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn bond(&self) -> Item<BondData> {
+        Item::new(self.bond_key.as_ref())
+    }
+
+    fn delegations(&self) -> Map<(&[u8], u64), Uint128> {
+        Map::new(self.delegations_key.as_ref())
+    }
+
+    // Returns block_time part of the delegation key
+    pub fn delegation_keys_for_mix(&self, mix: &str, storage: &dyn Storage) -> Vec<u64> {
```
jstuczyn, line 143: Do we even have to scan the entire storage? Why not use `IndexedMap` instead and grab the data with `idx` instead?

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+use super::Account;
+
+impl VestingAccount for Account {
+    fn locked_coins(
+        &self,
+        block_time: Option<Timestamp>,
+        env: &Env,
+        storage: &dyn Storage,
+    ) -> Result<Coin, ContractError> {
+        // Returns 0 in case of underflow.
```
jstuczyn, line 16: But should underflow be EVER possible?

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+        let period = self.get_current_vesting_period(block_time);
+
+        let amount = match period {
+            // We're in the first period, or the vesting has not started yet.
+            0 => Coin {
+                amount: Uint128::new(0),
+                denom: DENOM.to_string(),
+            },
+            // We always have 8 vesting periods, so periods 1-7 are special
+            1..=7 => Coin {
```
jstuczyn, line 70: considering we have defined `NUM_VESTING_PERIODS` as a constant, shouldn't this bound be instead `1..=(NUM_VESTING_PERIODS-1)`?

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+        storage: &dyn Storage,
+    ) -> Result<Coin, ContractError> {
+        let block_time = block_time.unwrap_or(env.block.time);
+        let period = self.get_current_vesting_period(block_time);
+        let max_vested = self.tokens_per_period()? * period as u128;
+        let start_time = self.periods[period].start_time;
+
+        let delegations_keys = self
+            .delegations()
+            .keys_de(storage, None, None, Order::Ascending)
```
jstuczyn, line 119: Yeah... about that... :)

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+            .scan((), |_, x| x.ok())
+            .filter(|(_mix, block_time)| block_time < &start_time)
+            .map(|(mix, block_time)| (mix, block_time))
+            .collect::<Vec<(Vec<u8>, u64)>>();
+
+        let mut amount = Uint128::zero();
+        for (b, t) in delegations_keys {
```
jstuczyn: could we use slightly better names for those? I'm not entirely sure what `b` and `t` are supposed to represent here

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    fn get_vesting_coins(
+        &self,
+        block_time: Option<Timestamp>,
+        env: &Env,
+    ) -> Result<Coin, ContractError> {
+        Ok(Coin {
+            amount: Uint128::new(
+                self.get_original_vesting().amount.u128()
```
jstuczyn: Why are we converting those from `Uint128` to `u128`? Theres a `Sub` trait defined on `Uint128`  directly

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -0,0 +1,206 @@
...
+        storage: &dyn Storage,
+    ) -> Result<Coin, ContractError> {
+        // Returns 0 in case of underflow.
+        Ok(Coin {
+            amount: Uint128::new(
+                self.get_vesting_coins(block_time, env)?
+                    .amount
+                    .u128()
+                    .saturating_sub(
+                        self.get_delegated_vesting(block_time, env, storage)?
```
jstuczyn, line 23: Aren't the coins we have delegated/bonded still technically locked though?

jstuczyn:
After everything is merged, I'll definitely have another look at it all

contracts/mixnet/src/contract.rs
durch: Yeah, I don't have a good take on here, we should probably do what ever feels more ergonomic, we can also put it in contract state, so that we can update it more easily

contracts/vesting/src/vesting/account/vesting_account.rs
durch: I've tried that, but you can't use expressions in match statements, as the compiler can't ensure the match is exhaustive

contracts/vesting/src/vesting/account/vesting_account.rs
durch: Its probably pre cosmwasm 1.0 code, I think you could not do it directly before, will take a look

contracts/vesting/src/vesting/account/vesting_account.rs
durch: Its a weird definition of `locked`, we might want to change it...

```
locked_coins returns the set of coins that are not spendable (can still be delegated tough) (i.e. locked), defined as the vesting coins that are not delegated or bonded.
```

contracts/vesting/src/vesting/account/vesting_account.rs
durch: Thats a good point, no it should never be possible

contracts/vesting/src/vesting/account/bonding_account.rs
durch: I've inlined track_delegation as well, its just accounting unbond/undelegate needs its own tough

contracts/vesting/src/vesting/account/mod.rs
durch: I don't think we scan the entire storage, we just get those with a mix_identity prefix

contracts/mixnet/src/contract.rs
durch: I don't think we even need a vesting address here, since we control on the proxy, noone can undelegate/unbond except the proxy that did the original bonding/delegation, so there is no specific need to further gate on the vesting contract, as far as we care anyone can delegate on behalf of someone else, as long as they pay the bonding/delegation fee... This way we can have multiple vesting contracts, that implement the same interface...

contracts/vesting/src/traits/vesting_account.rs
durch: It can be negative if more tokens are delegated/bonded or vesting then there are vested tokens, so when calculating it we go with the saturating sub, as we don't really care if its negative or zero, we just want to block a withdraw

contracts/vesting/src/vesting/account/delegating_account.rs
durch: It is, but I want to be able to search on a prefix, so not sure if a `joined_key()` will allow for that

contracts/vesting/src/vesting/account/delegating_account.rs
durch: Well, we want all the keys for with a specific prefix, and keys are `(mix_identity, block_time)`, since we only want to remove the delegation we don't care about specific values, although we might care about the legitimacy of the message that asked for this...

contracts/vesting/src/vesting/account/delegating_account.rs
durch: Blow up in case there is an error in the storage, maybe it would be fine to just blow through it

durch:
I'm mostly happy here, we can probably optimize storage some more tough, although lets defer that until the testnet

contracts/vesting/src/vesting/account/delegating_account.rs
jstuczyn: there's a `.prefix()` : )

PR #910, opened 2021-11-20
Feature/mixnet contract refactor

futurechimp:
I was pondering a whole set of questions around how to make payments a bit simpler on Friday. As a bit of a preliminary, I've refactored the contract to see what all the moving parts look like.


futurechimp:
Note: if you check out the branch, it'll probably make more sense than looking at individual file changes on Github. There are no functional changes, I've just pushed things into modules.

neacsu (approved):
I like this split, it makes it easier to find stuff. Would it make sense to also group together the `bonding_` and `delegation_` (and possibly `layer_`) into their respective modules, with separate directories? I think it would keep things even more consistent.

jstuczyn:
I like this idea, I was already thinking about doing something similar at some point, but never had the time to go for it. However, after glancing at the changes, there is a couple of things I noticed:
- the tauri wallet no longer compiles due to changed types,
- even though types were renamed, the functions using them still refer to the old names. For example, `StateParams` was changed to `ContractSettingsParams`, however, the function to update it is still called `try_update_state_params`. There's also a lot of such occurrences in the client code.
- functions using the storage are inconsistent. We seem to be using two different conventions: `*` and `*_read` for mutable and immutable access, for example `contract_settings` and `contract_settings_read` as well as `mut_*` and `*`, for example `reward_pool` and `mut_reward_poll`. Personally I prefer using the `mut` phrase, but it's a personal preference. Though I'd probably would have put it at the end of the name, i.e. `reward_pool_mut`
- not sure how relevant that is, but some files are missing license notices,
- one of us will have to deal with so many merge conflicts due to https://github.com/nymtech/nym/pull/898 : (

futurechimp:
> I like this split, it makes it easier to find stuff. Would it make sense to also group together the `bonding_` and `delegation_` (and possibly `layer_`) into their respective modules, with separate directories? I think it would keep things even more consistent.

I initially started to split bonding and delegation, and found that there were enough cross-dependencies between the two that, at least as they're coded now, they want to live in the same home together. If anybody wants to take a crack at splitting them out in a future PR that would be fine. The layer stuff I didn't understand very well, it feels like something that's not a standalone thing. 

The question I asked myself when doing the refactoring wasn't "how can I split this up to the smallest degree?", it was "if I were to deploy things as separate cosmwasm contracts, where would I put the boundaries between those contracts". So far, this is my best answer.

futurechimp:
> the tauri wallet no longer compiles due to changed types

Good catch, I'll figure out where those types come from and fix it.

futurechimp:
> even though types were renamed, the functions using them still refer to the old names. For example, StateParams was changed to ContractSettingsParams, however, the function to update it is still called try_update_state_params. There's also a lot of such occurrences in the client code.

I'll find and fix the ones I can see, I think that was the only major rename so should be pretty quick. 

> functions using the storage are inconsistent. We seem to be using two different conventions: * and *_read for mutable and immutable access, for example contract_settings and contract_settings_read as well as mut_* and *, for example reward_pool and mut_reward_poll. Personally I prefer using the mut phrase, but it's a personal preference. Though I'd probably would have put it at the end of the name, i.e. reward_pool_mut

This is not an artefact of the restructuring, I haven't touched those. Please put it in a future PR if you like!

> one of us will have to deal with so many merge conflicts due to Feature/batch delegator rewarding #898 : (

Let's get both of these through the mill as quick as we can or the pain will intensify even further.

PR #915, opened 2021-11-23
Fix Mobile View for MUI data-grid (CARD 108)

Aid19801:
This PR seeks to fix the following issues, as raised in card#108 [https://github.com/nymtech/nym/issues/2540].

✅  The text rendered in screen does not fit adeptly in the grid 
✅  Cells were disappearing on-scroll left / right, would look blank (see images in original issue).
✅  All grids to render data grid text in a visible manner for the end user
✅  All SVGs aligned on Mobile overview page (Chevrons / Arrows).


explorer/src/components/CustomColumnHeading.tsx
```diff
@@ -17,6 +17,8 @@ export const CustomColumnHeading: React.FC<{ headingTitle: string }> = ({
           fontWeight: 'bold',
           fontSize: 14,
           padding: 0,
+          // border: '1px solid red',
```
fmtabbara, line 20: Remove commented code

explorer/src/utils/index.ts
```diff
@@ -65,26 +65,23 @@ export function countryDataToGridRow(
...
           owner: mn.owner,
-          location: mn?.location?.country_name || '',
           identity_key: mn.mix_node.identity_key || '',
           bond: totalBond || 0,
+          location: mn?.location?.country_name || '',
           self_percentage: selfPercentage,
-          host: mn.mix_node.host || '',
-          layer: mn.layer || '',
+          host: mn?.mix_node?.host || '',
+          layer: mn?.layer || '',
```
fmtabbara, line 84: Just curious as to why some `mn` don't have the optional chaining operator and some do? I would have thought either they'd all need it or none of them would.

fmtabbara (approved):

explorer/src/utils/index.ts
Aid19801: From memory i think it was that ALL of them have an identity key, owner etc. But there are records that have blank location/null. I will add to my to-do list though to check..

PR #929, opened 2021-12-01
Move cleaned up smart contracts to main code repo 

mfahampshire:


contracts/basic-bandwidth-generation/contracts/BandwidthGenerator.sol
```diff
@@ -0,0 +1,56 @@
+pragma solidity 0.6.6;
+
+import "./CosmosToken.sol";
+import "./Gravity.sol";
+
+/** 
+ * @title BandwidthGenerator
+ * @dev   Contract for generating basic bandwidth credentials on the Nym cosmos blockchain, 
+ *        using ERC20 representations of CW20 NYM as payment.
```
futurechimp: The nym token is native cosmos, not a CW20 token, so you can remove CW20 from here.

futurechimp (approved):
One minor docs problem (we're not using CW20) but other than that, looks good to me so that we can get it in place and start trying it out.

mfahampshire:
Incorporating suggestions from Jelle

contracts/basic-bandwidth-generation/contracts/BandwidthGenerator.sol
```diff
@@ -0,0 +1,91 @@
...
+ *        `(Token amount in 'wei' / 10**18) * MBPerToken`
+ */ 
+contract BandwidthGenerator is Ownable {
+    
+    CosmosERC20 public erc20;
+    Gravity     public gravityBridge; 
+    uint256     public MBPerToken; 
+    
+    event BBCredentialPurchased(
+        uint256 indexed Bandwidth,
```
neacsu: I don't think we need to index by `Bandwidth` as the `VerificationKey` should be enough to provide unicity. Also, the client expects we only index by the `VerificationKey` right now.

contracts/basic-bandwidth-generation/contracts/BandwidthGenerator.sol
```diff
@@ -0,0 +1,91 @@
...
+        uint256 indexed VerificationKey,
+        bytes   SignedVerificationKey, 
+        bytes32 indexed CosmosRecipient
```
neacsu, line 31: This field isn't expected by the client either, so it won't work with it in the current form. Initially, I considered that the person purchasing the bandwidth might not know at the time of buying it with what gateway it will use it, and so it would have some flexibility. In order to simplify the payment flow, we might consider that it does know all of this at the moment of purchase.
But even if it does, the client binary (the one calling to emit this event) is only aware of the gateway identity, not of the gateway's owner address, which is what would need to be put in this field. Some additional work would need to be done for the client to obtain that information (query the contract, filter the gateways, extract the owner)

contracts/basic-bandwidth-generation/contracts/BandwidthGenerator.sol
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @param _erc20          Address of the erc20NYM deployed through the Gravity Bridge.
+     * @param _gravityBridge  Address of the deployed Gravity Bridge. 
+     */
+    constructor(CosmosERC20 _erc20, Gravity _gravityBridge) public {
+        require(address(_erc20) != address(0),         "BandwidthGenerator: erc20 address cannot be null"); 
+        require(address(_gravityBridge) != address(0), "BandwidthGenerator: gravity bridge address cannot be null"); 
+        erc20 = _erc20;
+        gravityBridge = _gravityBridge; 
+        MBPerToken = 1024; // default amount set at deployment: 1 erc20NYM = 1024MB = 1GB
```
neacsu: This is making the assumption that we operate on MB level, whereas the client and gateway currently count bandwidth at byte level. I think this should be `BytesPerToken = 1024 * 1024 * 1024` so that the contract counts at byte level as well. That would be better than working with floats in the client/gateway, as the Sphinx packets are smaller than MB level from what I recall.

neacsu (commented):
I've only looked at the `BandwidthGenerator.sol`, as discussed offline. Let me know if I should review other parts of the code as well.

contracts/basic-bandwidth-generation/contracts/BandwidthGenerator.sol
mfahampshire: As per offline chat, going to leave this in and we can tweak the client

mfahampshire:
@futurechimp if you want to give this another look over now i've incorporated requested changes that'd be great

PR #932, opened 2021-12-01
Adding tx_hash to wallet response

futurechimp:
This will allow the wallet ui, later, to  provide a link to Big Dipper.

 The user can then see which block the transaction happened in, and other associated information. cc @jessgess


tommyv1987 (approved):
Built the app with the changes, completed transaction, all behaving nicely.

PR #933, opened 2021-12-01
Webpack wallet prod configuration

tommyv1987:
- In `production` mode in the webpack config, there's an unmet dependency that requires favicons to be imported. If trying to supply a `.ico` file it complains of a MIME type error, therefore, adding a `.png` to replace the `.ico` to the same spec works.
- Added the references to rebuild on macOS for `MenuItems`
- Built on all platforms - Windows, Mac, Linux


nym-wallet/package.json
```diff
@@ -5,11 +5,11 @@
   "license": "MIT",
   "scripts": {
     "webpack:dev": "yarn webpack serve",
-    "webpack:build": "yarn webpack",
+    "webpack:prod": "yarn webpack --progress --config webpack.prod.config.js",
     "tauri:dev": "yarn tauri dev",
     "tauri:build": "yarn tauri build",
     "dev": "yarn run webpack:dev & yarn run tauri:dev",
-    "build": "yarn run webpack:build & yarn run tauri:build"
+    "build:prod": "run-s webpack:prod tauri:build"
```
fmtabbara, line 12: Nice :+1: I didn't know about this little npm command. Just what we need

fmtabbara (approved):

nym-wallet/package.json
tommyv1987: :rocket:

PR #946, opened 2021-12-07
Docker updates

tommyv1987:
- Remove web-wallet
- Update wasmd - 0.21.0
- Add vesting
- Update explorer  (still needs fine tuning)


docker/validator/setup.sh
```diff
@@ -10,7 +10,7 @@ go build \
     -X github.com/cosmos/cosmos-sdk/version.AppName=nymd \
     -X github.com/CosmWasm/wasmd/app.NodeDir=.nymd \
     -X github.com/cosmos/cosmos-sdk/version.Version=${WASMD_VERSION} \
-    -X github.com/cosmos/cosmos-sdk/version.Commit=1920f80d181adbeaedac1eeea1c1c6e1704d3e49 \
+    -X github.com/cosmos/cosmos-sdk/version.Commit=1d436638af7cacb5aeeb7248b57b085c64f3ae35 \
```
jstuczyn: Perhaps this hash should be defined alongside `WASMD_VERSION` since it really depends on it?

docker/vesting_contract/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM rust
+RUN rustup target add wasm32-unknown-unknown
+CMD cd nym/contracts/vesting && RUSTFLAGS='-C link-arg=-s' cargo build --release --target wasm32-unknown-unknown
```
jstuczyn, line 3: If you're already using docker, you might want to take a look at cosmwasm' image for producing smallest wasm blobs: https://github.com/cosmwasm/rust-optimizer
Unless it's a quick change, I don't think you really need to do it now

jstuczyn (approved):

docker/vesting_contract/Dockerfile
tommyv1987: I'll do this in a separate change for both vesting and mixnet :+1:

tommyv1987: I'll need to have a quick glance over it.

docker/validator/setup.sh
tommyv1987: Done :+1:

PR #949, opened 2021-12-07
Feature/profit margin percent config

durch:


jstuczyn (approved):
I wonder if perhaps there should be some check at bonding time to ensure profit margin is within 0-100 range

PR #950, opened 2021-12-08
Feature/simulate

jstuczyn:
This PR introduces option to simulate transaction execution and as a result ability to set an "auto" gas estimation.

Addresses https://github.com/nymtech/nym/issues/2076


common/client-libs/validator-client/src/nymd/cosmwasm_client/types.rs
```diff
@@ -215,6 +219,117 @@ impl TryFrom<ProtoContractCodeHistoryEntry> for ContractCodeHistoryEntry {
...
+    pub events: Vec<abci::Event>,
+}
+
+impl TryFrom<ProtoAbciResult> for AbciResult {
+    type Error = NymdError;
+
+    fn try_from(value: ProtoAbciResult) -> Result<Self, Self::Error> {
+        // annoyingly parsing of the event could not be moved to a separate function as protobuf
+        // representation of the event is not exposed by the existing libraries
+        Ok(AbciResult {
```
durch, line 292: I'd maybe extract the `events` construction into its own binding, make it a bit more readable.

durch (approved):
Looks good in general, I'd just write out that events mapping and remapping.

common/client-libs/validator-client/src/nymd/cosmwasm_client/types.rs
jstuczyn: I really wanted to do it, unfortunately that would require pulling separate tendermint-proto dependency just for that single type definition (it's not publicly re-exposed by cosmrs)

durch:
`nym-wallet` also needs the `connect_with_signer()` call updated

neacsu (approved):

PR #951, opened 2021-12-08
Different workshare calculations for rewarded vs active set

### LINKED ISSUE #2071: Integrate rewarded set into `tokenomics` rewarding

durch:
Closes nymtech/nym#2071 and nymtech/nym#2070

### DIFF
--- a/validator-api/src/rewarding/mod.rs
+++ b/validator-api/src/rewarding/mod.rs
@@ -50,7 +50,6 @@ pub(crate) struct MixnodeToReward {
 }
 
 impl MixnodeToReward {
-    /// Somewhat clumsy way of feature gatting tokenomics payments. In a tokenomics scenario this will never be None at reward time. We leverage that to Into a different ExecuteMsg variant
     fn params(&self) -> NodeRewardParams {
         self.params
     }
@@ -166,7 +165,7 @@ impl Rewarder {
         let epoch_reward_percent = self.nymd_client.get_epoch_reward_percent().await?;
 
         // TODO: question to @durch: is k active set or 'rewarded' set?
-        let k = state.mixnode_active_set_size;
+        let k = state.mixnode_rewarded_set_size;
         let period_reward_pool = (reward_pool / 100) * epoch_reward_percent as u128;
 
         info!("Rewarding pool stats");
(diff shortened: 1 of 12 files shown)


jstuczyn (commented):
Conceptually looks good to me. However, I've got one question - how does this calculation update affect delegators rewarding? Should they also be adjusted?

durch:
> Conceptually looks good to me. However, I've got one question - how does this calculation update affect delegators rewarding? Should they also be adjusted?

Overall the node reward will be lower, so everyone will get less, owner, and delegators

jstuczyn:
> > Conceptually looks good to me. However, I've got one question - how does this calculation update affect delegators rewarding? Should they also be adjusted?
> 
> Overall the node reward will be lower, so everyone will get less, owner, and delegators

I might be misremembering, but wasn't it the case that delegators of nodes from the inactive set aren't getting anything?

durch:
> > > Conceptually looks good to me. However, I've got one question - how does this calculation update affect delegators rewarding? Should they also be adjusted?
> > 
> > 
> > Overall the node reward will be lower, so everyone will get less, owner, and delegators
> 
> I might be misremembering, but wasn't it the case that delegators of nodes from the inactive set aren't getting anything?

I'm not sure, will tap Claudia

durch:
> > > > Conceptually looks good to me. However, I've got one question - how does this calculation update affect delegators rewarding? Should they also be adjusted?
> > > 
> > > 
> > > Overall the node reward will be lower, so everyone will get less, owner, and delegators
> > 
> > 
> > I might be misremembering, but wasn't it the case that delegators of nodes from the inactive set aren't getting anything?
> 
> I'm not sure, will tap Claudia

Good thing I did, it slipped my mind completely that this has changed, and we did not capture the change with the ticket. Double checked with Claudia, and we should be good

jstuczyn:
One more thing I forgot to mention - we probably also need to update validator API and how it calculates all those values as currently it only sends rewards to the **CURRENT** rewarded set

durch:
> One more thing I forgot to mention - we probably also need to update validator API and how it calculates all those values as currently it only sends rewards to the **CURRENT** rewarded set

I know what you mean, we need to ticket that up, but we also need to discuss how we'll go about actually updating the rewarded set at required intervals

validator-api/src/rewarding/mod.rs
```diff
@@ -158,21 +185,18 @@ impl Rewarder {
...
-        info!("-- Reward pool: {} unym", reward_pool);
-        info!("---- Epoch reward pool: {} unym", period_reward_pool);
-        info!("-- Circulating supply: {} unym", circulating_supply);
+        info!("-- Reward pool: {} unym", epoch_reward_params.reward_pool);
+        info!(
+            "---- Epoch reward pool: {} unym",
+            epoch_reward_params.period_reward_pool
+        );
+        info!(
+            "-- Circulating supply: {} unym",
```
neacsu: I think it would be better to use `DENOM` for those as well.

PR #954, opened 2021-12-09
Feature/vesting to wallet

durch:
+ Adds vesting contract calls to nym-wallet
+ Refactoring and simplifying of error handling


common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -44,7 +44,7 @@ pub mod wallet;
 
 pub use traits::{VestingQueryClient, VestingSigningClient};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
jstuczyn: Why would we ever need/want to clone our client? I think we only ever want to have a single connection to the blockchain from our application, no?

common/client-libs/validator-client/src/nymd/traits/vesting_signing_client.rs
```diff
@@ -230,28 +230,32 @@ impl<C: SigningCosmWasmClient + Sync + Send> VestingSigningClient for NymdClient
     }
     async fn vesting_delegate_to_mixnode(
         &self,
-        mix_identity: IdentityKey,
-        amount: Coin,
+        mix_identity: &str,
```
jstuczyn: There's an `IdentityKeyRef` type which is basically an alias to `&str`

common/client-libs/validator-client/src/nymd/mod.rs
durch: I wanted to be able to hand out clients as needed from the application state, [like so](https://github.com/nymtech/nym/blob/feature/vesting-to-wallet/nym-wallet/src-tauri/src/state.rs#L12). That used to pass references around, but in that case we needed two lines of code to pass it out, one to get the state, and one to hand out the client. This way we can do it all in one line. I like your conceptual point tough, I'll think this one through some more

PR #957, opened 2021-12-13
Add custom denom balance query

neacsu:
This is needed so that we can query other denom balances, which is important especially for testing purposes at this moment.


jstuczyn:
What do you think about doing it the other way around, i.e. have the `get_balance` take a `Denom` argument and create something like `get_mixnet_balance` that would call `get_balance` with the denom associated with the mixnet contract?

neacsu:
> What do you think about doing it the other way around, i.e. have the `get_balance` take a `Denom` argument and create something like `get_mixnet_balance` that would call `get_balance` with the denom associated with the mixnet contract?

Sounds good, I've pushed the patch

durch (approved):

PR #965, opened 2021-12-15
Feature/optional bandwidth bypass

jstuczyn:
Allows for starting gateways/clients in a `testnet-mode` that will expose additional message to claim bandwidth without presenting a proper credential.


durch (approved):
+ wasm client is complaining about a move value, looks good otherwise

neacsu (approved):

PR #966, opened 2021-12-15
Fixes to Network Explorer for new testnet

mmsinclair:


explorer/.env.prod
```diff
@@ -0,0 +1,5 @@
+EXPLORER_API_URL=https://sandbox-explorer.nymtech.net/api/v1
+VALIDATOR_API_URL=https://sandbox-validator1.nymtech.net
```
tommyv1987: @mmsinclair I think you need to remove the `validator1.`
Currently the validator API resolves to `https://sandbox-validator.nymtech.net`

.github/workflows/network-explorer.yml
```diff
@@ -22,6 +22,8 @@ jobs:
         node-version: '14'
```
tommyv1987, line 22: Is there any reason why we're running version 14 as well?

tommyv1987, line 22: Is there any reason why we're using node v14 on the workflow?

explorer/.env.prod
tommyv1987: Network defaults doesn't have this in it's vec of validators-details like milhon did.

.github/workflows/network-explorer.yml
mmsinclair: No reason, it could be changed to Node 16 some time

explorer/.env.prod
mmsinclair: Ok I've changed this to `https://sandbox-validator.nymtech.net`

mfahampshire (approved):

PR #968, opened 2021-12-16
Feature/new testnet wallet updates

fmtabbara:
- Adds env variables to configure the currency, admin address (possibly not required) and the testnet name.
- Adds new fields to the bond form which will work once the wallet is transferred to the new testnet. 
- Update to webpack config (introduce common file)


fmtabbara:
I think my prettier updates may make it look like there are more changes than there actually are. Some of these changes are just formatting!

tommyv1987 (approved):
I'm happy with the functionality, I've checked out the branch put it through it's paces, and there's another branch to be merged on top of this. I will then give all 3 OS's a spin, currently tested on a Linux based box.

PR #972, opened 2021-12-16
Feature/node info command

jstuczyn:
Introduces `node-details` command that basically prints out all information required for node bonding. Also removed some duplicate code.


tommyv1987 (approved):
Tested the CLI against both gateway and node, all good.

PR #973, opened 2021-12-16
Feature/implicit runtime

jstuczyn:
Removes explicit construction of tokio runtimes in favour of implicit `#[tokio::main]`


futurechimp:
Is it normal that this would fail tests on Windows? It seems to be failing a format check, but I'd have expected lots of other things to break.

futurechimp (changes requested):
I'm not sure what's up with the Windows format checks, but as this one is just a bit of code grooming I'd rather leave it until after we cut 0.12.0 and merge it after that.

jstuczyn:
> Is it normal that this would fail tests on Windows? It seems to be failing a format check, but I'd have expected lots of other things to break.

Those have already been broken for a while now - I have absolutely no idea why. I think @durch managed to fix those in one of his other PRs but I don't remember exactly how

neacsu:
I've fixed it by removing a dangling rustfmt file: https://github.com/nymtech/nym/pull/975

neacsu (approved):

PR #997, opened 2021-12-23
Fix wallet build instructions

tommyv1987:
- Supplying the .env file requirement


nym-wallet/.env.sample
```diff
@@ -1,4 +1,4 @@
 MAJOR_CURRENCY=nymt
 MINOR_CURRENCY=unymt
-ADMIN_ADDRESS=nymt1a586lvexy7ahva8h27fp6h8ds8taea6p95w3vx
```
neacsu, line 3: Why do we want to remove the admin address too?

neacsu (approved):

PR #1004, opened 2022-01-03
Feature/fix gateway sign

neacsu:
Remove the implicit derivation of the wallet address from the stored mnemonic for gateways. The reason for this is that the cosmos mnemonic that is asked at init is for the address that gets rewarded for gateway usage. Since that address is not necessarily set now and it can take a default value (see the `eth` feature flag), we won't be using that to derive the address in the signing process, but rather go with the same flow as for the mixnode.

A slightly different approach would be to use the wallet address that we already set at `init`, for both gateway and mixnode, by moving some code around, if we want that functionality. I personally think that would make the `sign` command less flexible, but I'm open to suggestions.

### DIFF
--- a/gateway/src/commands/mod.rs
+++ b/gateway/src/commands/mod.rs
@@ -148,3 +148,22 @@ pub(crate) fn validate_bech32_address_or_exit(address: &str) {
         process::exit(1);
     }
 }
+
+// this only checks compatibility between config the binary. It does not take into consideration
+// network version. It might do so in the future.
+pub(crate) fn version_check(cfg: &Config) -> bool {
+    let binary_version = env!("CARGO_PKG_VERSION");
+    let config_version = cfg.get_version();
+    if binary_version != config_version {
+        log::warn!("The gateway binary has different version than what is specified in config file! {} and {}", binary_version, config_version);
+        if version_checker::is_minor_version_compatible(binary_version, config_version) {
+            log::info!("but they are still semver compatible. However, consider running the `upgrade` command");
+            true
+        } else {
+            log::error!("and they are semver incompatible! - please run the `upgrade` command before attempting `run` again");
+            false
+        }
+    } else {
+        true
+    }
+}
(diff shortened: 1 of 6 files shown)


jstuczyn:
> The reason for this is that the cosmos mnemonic that is asked at init is for the address that gets rewarded for gateway usage.
Isn't it the case that this address should be the same as the one set for bonding?

jstuczyn (approved):

neacsu:
> The reason for this is that the cosmos mnemonic that is asked at init is for the address that gets rewarded for gateway usage.
>> Isn't it the case that this address should be the same as the one set for bonding?

In theory, yes, but since we have a workaround for getting bandwidth now, it's not :) When we remove the workaround (probably at the same time when we go full coconut), there won't be any reason for them not to be the same I think.

jstuczyn:
> > The reason for this is that the cosmos mnemonic that is asked at init is for the address that gets rewarded for gateway usage.
> > > Isn't it the case that this address should be the same as the one set for bonding?
> 
> In theory, yes, but since we have a workaround for getting bandwidth now, it's not :) When we remove the workaround (probably at the same time when we go full coconut), there won't be any reason for them not to be the same I think.

I see. But isn't it the case that (in theory), when somebody decides to run gateway in non-testnet mode and without coconut, he's asked for the mnemonic and at this point we can actually derive the address to verify whether it matches the other argument (i.e. address used for bonding)? Note: this is just a hypothetical question, not something we really have to worry about : )

neacsu:
> > > The reason for this is that the cosmos mnemonic that is asked at init is for the address that gets rewarded for gateway usage.
> > > > Isn't it the case that this address should be the same as the one set for bonding?
> > 
> > 
> > In theory, yes, but since we have a workaround for getting bandwidth now, it's not :) When we remove the workaround (probably at the same time when we go full coconut), there won't be any reason for them not to be the same I think.
> 
> I see. But isn't it the case that (in theory), when somebody decides to run gateway in non-testnet mode and without coconut, he's asked for the mnemonic and at this point we can actually derive the address to verify whether it matches the other argument (i.e. address used for bonding)? Note: this is just a hypothetical question, not something we really have to worry about : )

Yes, that would be a nice double check to have in place

PR #1007, opened 2022-01-04
Feature/new example readmes

mfahampshire:


clients/native/examples/README.md
```diff
@@ -0,0 +1,48 @@
+# Nym Desktop Client Code Examples 
+
+This directory contains example code for javascript, python, and rust. Please note that **none of these examples are production-ready**, and are included really just to show how the client interacts with the Nym system. 
+
+> Make sure that you have an instance of the `nym-client` set up and running locally on port 1977 before trying to run any of these examples.
```
jstuczyn, line 5: Could it be worth to mention that port `1977` is the default one, so that people wouldn't be confused trying to figure out how to make the client run on that particular port?

clients/native/examples/README.md
```diff
@@ -0,0 +1,48 @@
+# Nym Desktop Client Code Examples 
+
+This directory contains example code for javascript, python, and rust. Please note that **none of these examples are production-ready**, and are included really just to show how the client interacts with the Nym system. 
+
+> Make sure that you have an instance of the `nym-client` set up and running locally on port 1977 before trying to run any of these examples. 
+
+## Go
```
jstuczyn, line 7: I think when I was writing those example way back, my idea was to have a `binarysend` and `textsend` files that would be responsible for sending binary and text data respectively for each of the languages, i.e. Python, Go and Rust.  One important note is that if you look into each file, each of those examples has 2 methods in its `main`: `sendWithReply` and `sendWithoutReply` which basically also demonstrate use of SURBs. However, in each case one of them is commented out. I can't remember the exact reason for it, I think I wanted to avoid executing the same thing twice? Not sure.
But what I'm getting to is that we could have exactly the same README bit for all those 3 languages since the behaviour should be identical.

Javascript works differently, because of course it does....

jstuczyn (changes requested):
Few more notes:
- you accidentally included the go's binary to the version control, i.e. `clients/native/examples/go-examples/websocket/binary/binary`
- if you can be bothered (and honestly it's a super low priority), perhaps we could use the same `dummy_file` for all examples?
- again, assuming you have nothing else to do, perhaps those examples could be further split into `text_send_with_reply`, `text_send_without_reply`, `binary_send_with_reply` and `binary_send_without_reply`?

PR #1008, opened 2022-01-05
Feature/configure profit

neacsu:


common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -773,6 +773,31 @@ impl<C> NymdClient<C> {
             .await
     }
 
+    /// Update the configuration of a mixnode. Right now, only possible for profit margin.
```
jstuczyn, line 776: So since you can only update profit margin, perhaps the operation itself should be called `UpdateProfitMargin`?

contracts/mixnet/src/mixnodes/transactions.rs
```diff
@@ -204,6 +204,40 @@ pub(crate) fn _try_remove_mixnode(
...
+    // We don't have to check lower bound as its an u8
+    if profit_margin_percent > 100 {
+        return Err(ContractError::InvalidProfitMarginPercent(
+            profit_margin_percent,
+        ));
+    }
+
+    storage::mixnodes().update(deps.storage, &mix_identity, |mixnode_bond_opt| {
+        mixnode_bond_opt
+            .map(|mut mixnode_bond| {
```
jstuczyn, line 232: Shouldn't we also update the mixnode `bond_height` so that somebody wouldn't attempt to change their profit margin just before rewards are distributed and thus scam all of their delegators?

common/client-libs/validator-client/src/nymd/mod.rs
neacsu: I've named it like that so that we can update other stuff as well in the future, so it should be more generic then just profit margin.

contracts/mixnet/src/mixnodes/transactions.rs
neacsu: I think you're right, I'll modify that as well.

jstuczyn (approved):

PR #1009, opened 2022-01-05
Use serial integer instead of random

durch:
Remove random, it might introduce additional problems, so treat it as a clutch until we verify it works in QAnet


contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -20,17 +18,10 @@ const BALANCE_SUFFIX: &str = "ba";
...
-        let balance_key = format!("{}{}", key, BALANCE_SUFFIX);
-        let balance: Item<Uint128> = Item::new(&balance_key);
-        if balance.may_load(storage)?.is_none() {
-            return Ok(key);
-        }
-    }
+fn generate_storage_key(storage: &mut dyn Storage) -> Result<String, ContractError> {
+    let key = KEY.may_load(storage)?.unwrap_or(0) + 1;
+    KEY.save(storage, &key)?;
+    Ok(key.to_string())
```
neacsu: I get this is a temporary workaround, but why is the key saved now and previously was only checked for existence?

neacsu (commented):
Perhaps a tracking issue should be created for the proper fix?

neacsu (approved):

contracts/vesting/src/vesting/account/mod.rs
durch: It was checked, as it was random, so the check was supposed to check for collisions, since we're only ever incrementing this it should never collide

PR #1011, opened 2022-01-05
Feature/remove unused profit margin

neacsu:
Do this as a separate PR, dependent on https://github.com/nymtech/nym/pull/1008 , to avoid having to do rebases. I'll remove it from draft once #1008 is merged

I successfully tested the migration locally with a mixnet containing aprox. 5000 bonded mixnodes.

### DIFF
--- a/contracts/mixnet/src/contract.rs
+++ b/contracts/mixnet/src/contract.rs
@@ -289,7 +289,63 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> Result<QueryResponse, Cont
     Ok(query_res?)
 }
 #[entry_point]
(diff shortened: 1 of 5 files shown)


jstuczyn (changes requested):
Could you merge/rebase(?) with develop so that it would be more obvious which changes were introduced in this PR? because for example it says `pub async fn update_mixnode_config(` was added now, while we've already merged it as part of https://github.com/nymtech/nym/pull/1008

neacsu:
> Could you merge/rebase(?) with develop so that it would be more obvious which changes were introduced in this PR? because for example it says `pub async fn update_mixnode_config(` was added now, while we've already merged it as part of #1008

Looks like my plan to draft on top of another PR until that one is merged backfired. I've rebased with develop

jstuczyn (approved):

PR #1012, opened 2022-01-05
Feature/hourly set updates

durch:
---
edit by @jstuczyn:
At the time of writing this comment, all the code changes should already be implemented. The only remaining things include writing new unit tests for the changed code in the mixnet smart contract and fix inevitable bugs, however, I feel the PR could already be opened for review while those are being written, so that it could go through QA more quickly.

Speaking of QA, there's couple of significant changes/additions:
- the rewarding procedure has been simplified. The old steps included sending `begin_mixnode_rewarding` transaction, followed by actual rewarding txs and finally finished with `finish_mixnode_rewarding`. This has gone away in favour of simply just having the rewarding txs, since it should no longer be possible to abuse as the contract itself stores the information about the current epoch. Plus we don't really need the block height anymore. However, one note is that there's a new `advance_epoch` tx that should be called after each reward cycle that advances the current epoch stored in the contract. Note: it cannot be called (well, it can but it will fail), until the epoch has actually finished (according to the BFT tendermint time). This is to prevent a malicious validator from quickly advancing epochs in order to reward particular node many times in a short span of time. 
- note: on startup, the validator API (in theory, I haven't tested it locally yet), should grab the currently stored epoch [in the smart contract], compare it with what should be the actual value [according to tendermint BFT time] and successively advance it and, if possible, i.e. if it has available measurements, distribute rewards for those past epochs. 
- there's a new `WriteRewardedSet` tx (which was the main purpose of this PR), that allows the permitted validator to write the rewarded set to the smart contract. Note that again, it cannot be called too often or it will be rejected. Perhaps for testing sake those constants should temporarily be set to some rather low values.

Mental note to check before merging/during QA:
- if validator API gets restarted, does it serve exactly the same active/rewarded sets as before the restart?


common/cosmwasm-smart-contracts/mixnet-contract/src/epoch.rs
```diff
@@ -33,6 +34,8 @@ impl Epoch {
     /// Returns the last epoch.
     pub fn previous_epoch(&self) -> Self {
         Epoch {
+            // TODO: is saturating_sub what we really want? perhaps we want to return an Option instead?
```
durch: I agree, returning an Option here would be nicer...

common/cosmwasm-smart-contracts/mixnet-contract/src/msg.rs
```diff
@@ -12,6 +12,7 @@ pub struct InstantiateMsg {
     pub rewarding_validator_address: String,
 }
 
+// TODO: We've removed JsonSchema derive from here due to Epoch, it needs to be put back
```
durch: Can we remove this TODO?
```suggestion
```

common/cosmwasm-smart-contracts/mixnet-contract/src/msg.rs
```diff
@@ -85,8 +76,14 @@ pub enum ExecuteMsg {
...
         owner: String,
     },
+    WriteRewardedSet {
+        rewarded_set: Vec<IdentityKey>,
+        expected_active_set_size: u32,
+    },
+    AdvanceCurrentEpoch {},
 }
 
+// TODO: See comment above for JsonSchema derive
```
durch: Same, as above
```suggestion
```

contracts/mixnet/src/epoch/queries.rs
```diff
@@ -0,0 +1,422 @@
...
+
+pub fn query_current_rewarded_set_height(storage: &dyn Storage) -> Result<u64, ContractError> {
+    Ok(storage::CURRENT_REWARDED_SET_HEIGHT.load(storage)?)
+}
+
+fn query_rewarded_set_at_height(
+    storage: &dyn Storage,
+    height: u64,
+    start_after: Option<IdentityKey>,
+    limit: u32,
```
durch: Should there be a default limit  implementation here, like we have below?

contracts/mixnet/src/epoch/storage.rs
```diff
@@ -0,0 +1,114 @@
...
+use cw_storage_plus::{Item, Map};
+use mixnet_contract_common::{Epoch, IdentityKey, RewardedSetNodeStatus};
+
+// type aliases for better reasoning for storage keys
+// (I found it helpful)
+type BlockHeight = u64;
+type EpochId = u32;
+
+// TODO: those values need to be verified
+pub(crate) const REWARDED_NODE_DEFAULT_LIMIT: u32 = 1000;
```
durch:
```suggestion
pub(crate) const REWARDED_NODE_DEFAULT_PAGE_LIMIT: u32 = 1000;
```

contracts/mixnet/src/epoch/storage.rs
```diff
@@ -0,0 +1,114 @@
...
+pub(crate) const REWARDED_NODE_MAX_LIMIT: u32 = 1500;
```
durch:
```suggestion
pub(crate) const REWARDED_NODE_MAX_PAGE_LIMIT: u32 = 1500;
```

durch (commented):
Code looks good overall, I need to spend some more time thinking through the conceptual rewarding, so I might come back with comments :)

contracts/mixnet/src/epoch/queries.rs
```diff
@@ -0,0 +1,422 @@
...
+    height: Option<u64>,
+    start_after: Option<IdentityKey>,
+    limit: Option<u32>,
+) -> Result<PagedRewardedSetResponse, ContractError> {
+    let height = match height {
+        Some(height) => height,
+        None => query_current_rewarded_set_height(storage)?,
+    };
+    let limit = limit
+        .unwrap_or(storage::REWARDED_NODE_DEFAULT_LIMIT)
```
durch:
```suggestion
        .unwrap_or(storage::REWARDED_NODE_DEFAULT_PAGE_LIMIT)
```

contracts/mixnet/src/epoch/queries.rs
```diff
@@ -0,0 +1,422 @@
...
+        .min(storage::REWARDED_NODE_MAX_LIMIT);
```
durch:
```suggestion
        .min(storage::REWARDED_NODE_MAX_PAGE_LIMIT);
```

contracts/mixnet/src/epoch/queries.rs
jstuczyn: So this function is kinda an artifact of experimentation with different kinds of queries and trying to extract common code. If you notice, this function is not public and is only used internally by `query_rewarded_set`. So I guess the question is: should those be combined together or left separated as they are now

jstuczyn:
> Code looks good overall, I need to spend some more time thinking through the conceptual rewarding, so I might come back with comments :)

Yeah, I'm scared whenever I have to touch anything to do with rewarding, however, I think this simplification shouldn't have any negative consequences, though I can't really prove it empirically

contracts/mixnet/src/epoch/queries.rs
durch: I'm ok with it being separate, I'm thinking that someone will start using it in a different place, and then the lack of a default limit will bite us :)

PR #1014, opened 2022-01-06
update frontend to use new profit update api

fmtabbara:


tommyv1987:
I'm happy with these changes, as mentioned to @fmtabbara - we need to add / display the txs cost when performing the action in the settings page to update the profit percent, apart from that everything works great.

tommyv1987 (approved):
Looks good.

PR #1015, opened 2022-01-06
Feature/expanded events

### LINKED ISSUE #2069: Emit cosmos events for reward distribution, so that we can do rewards distribution analytics

jstuczyn:
This PR consolidates events emitted by transactions in our smart contract(s). Each transaction (where applicable) gets its own event type. 

If @durch thinks sufficient amount of information is emitted during rewarding (`node_reward`, `lambda`, `sigma`), this PR might close https://github.com/nymtech/nym/issues/2069

Note, probably to @tommyv1987 or @mfahampshire , there might be one breaking change here as I've renamed our mixnet contract crate from `mixnet-contracts` to `mixnet-contract` (as that plural was bothering me for a while now). As a result, the compiled wasm blob will change its name accordingly to `mixnet_contract.wasm` (as opposed to `mixnet_contracts.wasm`)

Remaining work includes:
- [x] fixing up tests due to new responses returned
- [x] possibly introducing the same revamp to the vesting contract


common/mixnet-contract/src/events.rs
```diff
@@ -350,3 +349,29 @@ pub fn new_mix_delegators_rewarding_event(
...
+/// * `event`: event to search through.
+/// * `key`: key associated with the particular attribute
+pub fn may_find_attribute(event: &Event, key: &str) -> Option<String> {
+    for attr in &event.attributes {
+        if attr.key == key {
+            return Some(attr.value.clone());
+        }
+    }
+    None
+}
```
neacsu: Why not just `may_find_attribute(event, key).ok()`?

Cargo.toml
```diff
@@ -26,7 +26,9 @@ members = [
     "common/credentials",
     "common/crypto",
     "common/bandwidth-claim-contract",
-    "common/mixnet-contract",
+    "common/cosmwasm-smart-contracts/contract-common",
+    "common/cosmwasm-smart-contracts/mixnet-contract",
+    "common/cosmwasm-smart-contracts/vesting-contract",
```
neacsu: Nice

common/cosmwasm-smart-contracts/contracts-common/src/events.rs
```diff
@@ -0,0 +1,30 @@
...
+///
+/// * `event`: event to search through.
+/// * `key`: key associated with the particular attribute
+pub fn may_find_attribute(event: &Event, key: &str) -> Option<String> {
+    for attr in &event.attributes {
+        if attr.key == key {
+            return Some(attr.value.clone());
+        }
+    }
+    None
```
neacsu, line 29: Same thing here

neacsu (approved):

common/mixnet-contract/src/events.rs
jstuczyn: I think you were looking at an old revision because this file doesn't exist anymore : )

common/cosmwasm-smart-contracts/contracts-common/src/events.rs
jstuczyn: I'm not entirely sure what do you mean. I think you thought there was code duplication?

common/mixnet-contract/src/events.rs
neacsu: Disregard this comment an the other one, I've misread the code

PR #1019, opened 2022-01-07
Feature/additional mixnode endpoints

jstuczyn:
This pull request introduces two additional endpoints on our validator API: 
- `/mixnode/<identity>/reward_estimation` that tries to estimate reward of the particular mixnode if nothing changed until the end of the epoch (i.e. its uptime and distribution of stake). I've got one question to @durch here, I actually don't remember whether in the end we updated calculation for situations where given node was only sometimes in the "active"/"rewarded" set (i.e say it it was "active" for first hour, then was on "standby" for another few, then it wasn't chosen at all, etc.), do you happen to know that? Or is it perhaps something that will be introduced in https://github.com/nymtech/nym/pull/1012 ?
- `/mixnode/<identity>/stake_saturation` that determines current stake saturation of particular node

### DIFF
--- a/common/mixnet-contract/src/mixnode.rs
+++ b/common/mixnet-contract/src/mixnode.rs
@@ -327,6 +327,11 @@ impl MixNodeBond {
         self.total_delegation.clone()
     }
 
+    pub fn stake_saturation(&self, circulating_supply: u128, rewarded_set_size: u32) -> U128 {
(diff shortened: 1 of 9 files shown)


mmsinclair (approved):

durch:
@jstuczyn my plan is to have it introduced in https://github.com/nymtech/nym/pull/1012

durch (approved):

PR #1024, opened 2022-01-10
Feature/validator api client endpoints

jstuczyn:
It depends on https://github.com/nymtech/nym/pull/1019 and thus remains as a draft until the aforementioned PR is merged.

### DIFF
--- a/common/client-libs/validator-client/src/validator_api/mod.rs
+++ b/common/client-libs/validator-client/src/validator_api/mod.rs
@@ -1,16 +1,24 @@
-// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::validator_api::error::ValidatorAPIError;
+use crate::validator_api::routes::{CORE_STATUS_COUNT, SINCE_ARG};
 use coconut_interface::{BlindSignRequestBody, BlindedSignatureResponse, VerificationKeyResponse};
-use mixnet_contract::{GatewayBond, MixNodeBond};
+use mixnet_contract::{GatewayBond, IdentityKeyRef, MixNodeBond};
 use serde::{Deserialize, Serialize};
 use url::Url;
+use validator_api_requests::models::{
+    CoreNodeStatusResponse, MixnodeStatusResponse, RewardEstimationResponse,
+    StakeSaturationResponse,
+};
 
 pub mod error;
 pub(crate) mod routes;
 
 type PathSegments<'a> = &'a [&'a str];
+type Params<'a, K, V> = &'a [(K, V)];
+
+const NO_PARAMS: Params<'_, &'_ str, &'_ str> = &[];
 
 pub struct Client {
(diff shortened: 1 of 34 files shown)


nym-wallet/src-tauri/Cargo.toml
```diff
@@ -49,7 +49,7 @@ features = ["ts-rs"]
 
 [dev-dependencies.validator-client]
 path = "../../common/client-libs/validator-client"
-features = ["ts-rs"]
+features = ["typescript-types"]
```
mmsinclair, line 52: Is there still a GitHub Action that builds the types? And would this affect it?

validator-api/validator-api-requests/src/models.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
```
mmsinclair, line 7: Is the feature = `ts-rs` correct still?

mmsinclair (approved):
Looks good - only a couple of questions about the typescript feature flag

nym-wallet/src-tauri/Cargo.toml
jstuczyn: The type building is done via simple `cargo test` command which is still done as part of our CI pipeline and hence it should be unaffected.

validator-api/validator-api-requests/src/models.rs
jstuczyn: Correct. If you notice the `typescript-types` was defined for the client crate and this refers to the `validator-api-requests` crate. This crate was introduced to remove duplication of type definitions. And as for the new "feature", it was defined due to name clash of `ts-rs`

common/client-libs/validator-client/src/validator_api/mod.rs
```diff
@@ -1,16 +1,24 @@
-// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
```
neacsu: The correct copyright would be 2021-2022 I think, but if that's not how we did it in the rest of the codebase then I guess it doesn't matter that much. Or we could start now :?

neacsu (approved):

common/client-libs/validator-client/src/validator_api/mod.rs
jstuczyn: Good question. @mmsinclair or @futurechimp any opinions on that?

mmsinclair: My preference would be `2021-2022`

jstuczyn: As discussed on keybase, @mmsinclair aggress with @neacsu and I'm gonna update that header accordingly.

PR #1028, opened 2022-01-11
Wallet UI updates

fmtabbara:
PR includes refinement to the layout and styling of card components.

**NOTE** The "profit percentage update" in the settings area will not work in this branch as the updates to the api are in a separate PR - https://github.com/nymtech/nym/pull/1014


tommyv1987 (approved):
Happy with the changes. Checked out the branch and gave it a spin.

PR #1031, opened 2022-01-13
Feature/downcast reward estimation

### LINKED ISSUE #1029: Change `u128` fields in `RewardEstimationResponse` to `u64`
It should be safe to do so as the maximum theoretical value equals to 1 billion * 1 million (i.e. the entire token supply), which is smaller than `u64::MAX`.

This change has to be introduced as it looks like typescript doesn't like `u128` type and it fails on responses containing it.

durch:
Closes #1029


jstuczyn (approved):

jstuczyn:
You haven't run `cargo fmt` after updating status code : )

PR #1035, opened 2022-01-13
Migrate to cw-storage-plus 0.11.1

### LINKED ISSUE #1032: Update cw-storage-plus to 0.11
https://github.com/CosmWasm/cw-plus/blob/main/MIGRATING.md

durch:
There is an interesting change where the `U32Key` is deprecated in favour of `u32`. I'm not sure if we need a migration for this

Closes #1032

### DIFF
--- a/contracts/vesting/src/vesting/account/delegating_account.rs
+++ b/contracts/vesting/src/vesting/account/delegating_account.rs
@@ -111,8 +111,8 @@ impl DelegatingAccount for Account {
 
         // Iterate over keys matching the prefix and remove them from the map
         let block_times = delegations
-            .prefix_de(mix_bytes)
-            .keys_de(storage, None, None, Order::Ascending)
+            .prefix(mix_bytes)
+            .keys(storage, None, None, Order::Ascending)
             // Scan will blow up on first error
             .scan((), |_, x| x.ok())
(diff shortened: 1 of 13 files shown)


jstuczyn:
On a quick glance it looks like we might not need a migration, as it's still being serialized to big endian bytes representation

contracts/vesting/src/vesting/account/delegating_account.rs
```diff
@@ -111,8 +111,8 @@ impl DelegatingAccount for Account {
 
         // Iterate over keys matching the prefix and remove them from the map
         let block_times = delegations
-            .prefix_de(mix_bytes)
-            .keys_de(storage, None, None, Order::Ascending)
+            .prefix(mix_bytes)
```
jstuczyn, line 114: Just a question,has the behaviour of `prefix_de` (and `keys_de`) changed?

jstuczyn (approved):

contracts/vesting/src/vesting/account/delegating_account.rs
durch: No, `_de` suffix got removed, and `_raw` suffix got added to previously suffixless methods

PR #1036, opened 2022-01-13
Feature/node settings update

fmtabbara:


tommyv1987 (approved):
I've had a good test of all of these, and everything is looking good.

PR #1038, opened 2022-01-14
Network Explorer API: active set and overview summary

mmsinclair:
Adds new resources to the Network Explorer API:

- get mix node by id
- get mixnodes by active / standby / inactive status
- summary overview to return gateway count, mixnode count, validator count and block height in a single call

Various clean up of code and renaming / moving of things to better places


explorer-api/src/cache/mod.rs
```diff
@@ -13,19 +13,33 @@ impl<T: Clone> Cache<T> {
...
     }
 
-    pub(crate) fn get(&self, identity_key: &str) -> Option<T>
+    pub(crate) fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    pub(crate) fn get_all(&self) -> Vec<T>
+    where
+        T: Clone,
```
jstuczyn: `T: Clone` here (as well as line 32) is redundant as you've already declared `T: Clone` on the entire impl block (line 9)

explorer-api/src/gateways/models.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    pub(crate) async fn update_cache(&self, gateways: Vec<GatewayBond>) {
+        let mut guard = self.inner.write().await;
+
+        for gateway in gateways {
+            guard
+                .gateways
+                .set(gateway.gateway.identity_key.clone().as_str(), gateway)
```
jstuczyn: You're essentially doing the following: `&String` -> `String` -> `&str` : )
Instead you could just do the following:
```suggestion
                .set(&gateway.gateway.identity_key, gateway)
```

explorer-api/src/mix_node/http.rs
```diff
@@ -1,26 +1,46 @@
...
     ]
 }
 
+#[openapi(tag = "mix_nodes")]
+#[get("/<pubkey>")]
+pub(crate) async fn get_by_id(
+    pubkey: &str,
+    state: &State<ExplorerApiStateContext>,
+) -> Result<Json<PrettyDetailedMixNodeBond>, NotFound<String>> {
+    let mixnodes = state.inner.mixnodes.get_detailed_mixnodes().await;
```
jstuczyn: This seems rather inefficient. With the `get_detailed_mixnodes` you're making the API construct the `Vec<PrettyDetailedMixNodeBond>` for ALL mixnodes (which requires grabbing all locations, grabbing all mixnode data and then determining node status (i.e. active/inactive/etc)) only to discard everything but a single result. Perhaps you could create a function working the same way, but only for a single mixnode, i.e. something like `    pub(crate) async fn get_detailed_mixnodes(&self, identity_key: &str) -> Option<PrettyDetailedMixNodeBond>` ?

explorer-api/src/mix_node/models.rs
```diff
@@ -10,7 +10,7 @@ use std::sync::Arc;
 use std::time::SystemTime;
 use tokio::sync::RwLock;
 
-#[derive(Clone, Debug, Serialize, JsonSchema)]
+#[derive(Clone, Debug, Serialize, JsonSchema, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum MixnodeStatus {
```
jstuczyn, line 15: This is not a comment regarding this PR, but more of a mental note that currently we have enums in 3 different places doing essentially the same thing and we should unify them at some point...

explorer-api/src/mix_nodes/http.rs
```diff
@@ -7,13 +8,83 @@ use rocket_okapi::openapi_get_routes_spec;
...
+
+#[openapi(tag = "mix_nodes")]
+#[get("/summary")]
+pub(crate) async fn summary(state: &State<ExplorerApiStateContext>) -> Json<MixNodeSummary> {
+    Json(get_mixnode_summary(state).await)
+}
+
+pub(crate) async fn get_mixnode_summary(state: &State<ExplorerApiStateContext>) -> MixNodeSummary {
+    let mixnodes = state.inner.mixnodes.get_detailed_mixnodes().await;
+    let active = get_mixnodes_by_status(mixnodes.clone(), MixnodeStatus::Active).len();
```
jstuczyn, line 69: I guess we could have sorted it into categories with a single loop iteration (rather than 3 of them), but this is so minor that unless this call is taking ages to compute, it's not really worth doing.

explorer-api/src/tasks.rs
```diff
@@ -0,0 +1,146 @@
...
+            Err(e) => {
+                error!("Failed to get validators: {:?}", e)
+            }
+        }
+    }
+
+    pub(crate) fn start(self) {
+        info!("Spawning mix nodes task runner...");
+        tokio::spawn(async move {
+            let mut interval_timer = tokio::time::interval(MIXNODES_CACHE_REFRESH_RATE);
```
jstuczyn: Just nitpicking here: since we're also refreshing gateways and validators caches, is it really still a `MIXNODES_CACHE_REFRESH_RATE`?

explorer-api/src/tasks.rs
```diff
@@ -0,0 +1,146 @@
...
+                error!("Failed to get validators: {:?}", e)
+            }
+        }
+    }
+
+    async fn update_gateways_cache(&self) {
+        match self.retrieve_all_gateways().await {
+            Ok(response) => self.state.inner.gateways.update_cache(response).await,
+            Err(e) => {
+                error!("Failed to get validators: {:?}", e)
```
jstuczyn: Don't you mean gateways? : )

jstuczyn (commented):
Apart from some minor things, looks good to me!

explorer-api/src/mix_nodes/http.rs
```diff
@@ -7,13 +8,83 @@ use rocket_okapi::openapi_get_routes_spec;
...
+    state: &State<ExplorerApiStateContext>,
+) -> Json<Vec<PrettyDetailedMixNodeBond>> {
+    Json(get_mixnodes_by_status(
+        state.inner.mixnodes.get_detailed_mixnodes().await,
+        MixnodeStatus::Active,
+    ))
+}
+
+#[openapi(tag = "mix_nodes")]
+#[get("/active-set/inactive")]
```
neacsu, line 40: This sounds a bit strange, that the inactive set is on a path named `active-set`. I don't have a good alternative proposal though... Maybe just `set`?

neacsu (approved):

explorer-api/src/mix_node/models.rs
mmsinclair: Yeah, I tried to re-use the other ones, they were just lacking `JsonSchema` :cry:

explorer-api/src/tasks.rs
mmsinclair: Renamed to just `CACHE_...`

explorer-api/src/tasks.rs
mmsinclair: ✅

explorer-api/src/mix_nodes/http.rs
mmsinclair: Yeah, not the best I agree. Originally I was thinking `/status/inactive` but `status` is also really overloaded. We can bug Claudia for the right terminology some time...

explorer-api/src/mix_nodes/http.rs
mmsinclair: Think it is ok for now

explorer-api/src/mix_node/http.rs
mmsinclair: Good idea - changed this

explorer-api/src/gateways/models.rs
mmsinclair: Fixed this up

explorer-api/src/cache/mod.rs
mmsinclair: Done

PR #1042, opened 2022-01-14
Endpoint for rewarded set inclusion probabilities

### LINKED ISSUE #1037: Rewarded set inclusion probability API endpoint
Introduce additional endpoint on the validator API for determining probability of particular node being included in the rewarded/active set

durch:
Closes #1037

### DIFF
--- a/common/client-libs/validator-client/src/validator_api/routes.rs
+++ b/common/client-libs/validator-client/src/validator_api/routes.rs
@@ -23,4 +23,4 @@ pub const SINCE_ARG: &str = "since";
 pub const STATUS: &str = "status";
 pub const REWARD_ESTIMATION: &str = "reward-estimation";
 pub const STAKE_SATURATION: &str = "stake-saturation";
-pub const INCLUSION_CHANCE: &str = "inclusion-prob";
+pub const INCLUSION_CHANCE: &str = "inclusion-probability";
--- a/common/client-libs/validator-client/src/validator_api/routes.rs
+++ b/common/client-libs/validator-client/src/validator_api/routes.rs
@@ -23,3 +23,4 @@ pub const SINCE_ARG: &str = "since";
 pub const STATUS: &str = "status";
 pub const REWARD_ESTIMATION: &str = "reward-estimation";
 pub const STAKE_SATURATION: &str = "stake-saturation";
+pub const INCLUSION_CHANCE: &str = "inclusion-prob";
--- a/validator-api/src/cache/mod.rs
+++ b/validator-api/src/cache/mod.rs
@@ -312,8 +312,16 @@ impl ValidatorCache {
         // (rewarded_set_size - active_set_size) * prob_one_draw * (1. - prob_active_set);
 
         Some(InclusionProbabilityResponse {
-            in_active: prob_active_set as f32,
-            in_reserve: prob_reserve_set as f32,
+            in_active: if prob_active_set > 1. {
+                1.
+            } else {
+                prob_active_set
+            } as f32,
+            in_reserve: if prob_reserve_set > 1. {
+                1.
+            } else {
+                prob_reserve_set
+            } as f32,
         })
     }
 
--- a/validator-api/src/cache/mod.rs
+++ b/validator-api/src/cache/mod.rs
@@ -14,8 +14,9 @@ use rand::prelude::SliceRandom;
 use rand_chacha::rand_core::SeedableRng;
 use rand_chacha::ChaCha20Rng;
 use rocket::fairing::AdHoc;
-use serde::Serialize;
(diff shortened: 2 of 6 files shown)


common/client-libs/validator-client/src/validator_api/routes.rs
```diff
@@ -23,3 +23,4 @@ pub const SINCE_ARG: &str = "since";
 pub const STATUS: &str = "status";
 pub const REWARD_ESTIMATION: &str = "reward-estimation";
 pub const STAKE_SATURATION: &str = "stake-saturation";
+pub const INCLUSION_CHANCE: &str = "inclusion-prob";
```
jstuczyn: since it's an http route, would it make sense to just use the full name, i.e. `"inclusion-probability"`? what do people more familiar with web think? @mmsinclair @fmtabbara ?

validator-api/src/cache/routes.rs
```diff
@@ -25,3 +26,11 @@ pub(crate) async fn get_rewarded_mixnodes(cache: &State<ValidatorCache>) -> Json
 pub(crate) async fn get_active_mixnodes(cache: &State<ValidatorCache>) -> Json<Vec<MixNodeBond>> {
     Json(cache.active_mixnodes().await.value)
 }
+
+#[get("/mixnodes/rewarded/inclusion-prob/<mixnode_id>")]
+pub(crate) async fn get_probs_mixnode_rewarded(
+    cache: &State<ValidatorCache>,
+    mixnode_id: String,
+) -> Json<HashMap<String, f32>> {
```
jstuczyn: I think that instead it should be returning a concrete type with explicit fields. Something like
```rust
pub struct InclusionProbabilityResponse {
  rewarded_set: f32,
  active_set: f32,
}
```

validator-api/src/cache/mod.rs
```diff
@@ -214,31 +216,87 @@ impl ValidatorCache {
...
+                // note that the theoretical maximum possible stake is equal to the total
+                // supply of all tokens, i.e. 1B (which is 1 quadrillion of native tokens, i.e. 10^15 ~ 2^50)
+                // which is way below maximum value of f64, so the cast is fine
+                let total_stake = mix.total_stake().unwrap_or_default() as f64;
+                (mix.clone(), total_stake)
+            }) // if for some reason node is invalid, treat it as 0 stake/weight
+            .collect()
+    }
+
+    // Estimate probability that a node will end up in the rewarded set, by running the selection process 100 times and aggregating the results.
```
jstuczyn, line 273: Why should we run the selection process N times and then aggregate the result? The node's probability is just a ratio of its total bond over all bonded tokens and with probability being 1 if number of bonded nodes <= K. And I guess ... somehow...? taking into consideration the active/rewarded set sizes. uhm...
My point being, if the network doesn't change, the probability should always be the same for given node

jstuczyn (changes requested):

validator-api/src/cache/mod.rs
jstuczyn: Aha. Found something possibly useful: https://math.stackexchange.com/questions/316175/probability-to-choose-specific-item-in-a-weighted-sampling-without-replacement

neacsu: +1 to that, I don't think we should simulate it

sveta074:
hi

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Без
вирусов. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

вт, 18 янв. 2022 г. в 12:58, Bogdan-Ștefan Neacşu ***@***.***

validator-api/src/cache/mod.rs
durch: This is an excellent point, I think I was just being lazy and not wanting to do the math for active vs reserve set, although its not much of a math, its just stacking probabilities, let me revisit that

jstuczyn: Though it kinda seems like a possibly expensive computation. Since it doesn't really depend on validator's state, perhaps it might be worth to compute it client's side so that validator-api couldn't be overwhelmed with those?

durch: I'm hoping Rust crush those calculations, I guess it depends on how it's used, if the wallet will only make one request at a time I think we'll be fine, but worth testing for sure

durch: Ok, going with straightforward probability calculations, we're approximating some, but not a lot.

PR #1044, opened 2022-01-17
Fix clippy on relevant lints

neacsu:
return_self_not_must_use still produces errors, but that will be
auto-fixed once the change to move it to pedantic is released to beta
channel


jstuczyn:
Can you run `cargo fmt` on it? : )

durch (approved):

PR #1045, opened 2022-01-17
Add migrate function to vesting contract

durch:


tommyv1987 (approved):

contracts/vesting/src/contract.rs
```diff
@@ -1,14 +1,14 @@
...
 use crate::storage::account_from_address;
 use crate::traits::{
     DelegatingAccount, GatewayBondingAccount, MixnodeBondingAccount, VestingAccount,
 };
 use crate::vesting::{populate_vesting_periods, Account};
 use config::defaults::{DEFAULT_MIXNET_CONTRACT_ADDRESS, DENOM};
 use cosmwasm_std::{
-    coin, entry_point, to_binary, BankMsg, Coin, Deps, DepsMut, Env, MessageInfo, QueryResponse,
-    Response, Timestamp, Uint128,
+    coin, entry_point, to_binary, BankMsg, Coin, Deps, DepsMut, Env, Event, MessageInfo,
```
neacsu, line 10: Instead of including those here, I suggest putting as much as possible directly in the migrate function i.e. `cosmwasm_std::Event::new("test")...`, so that we only need to remove code from one place. Reverting commits backfired a while ago, and I forgot some residual code in our codebase.

neacsu (approved):

contracts/vesting/src/contract.rs
jstuczyn: We already have a central place for event creation: https://github.com/nymtech/nym/blob/develop/common/cosmwasm-smart-contracts/vesting-contract/src/events.rs

PR #1047, opened 2022-01-18
Upgrade Clap and use declarative argument parsing for nym-mixnode

### LINKED ISSUE #2061: Upgrade `nym-mixnode` to latest clap and switch to declarative argument parsing
Upgrade `nym-mixnode` to latest `clap`, which supports declarative argument parsing. The idea is that it should make it easier to read and more type safe.

octol:
Upgrade Clap to the latest version, which supports declarative argument parsing.

Some notable changes:
- Full version information is now part of `--help` instead of `--version` which is as far as I can see due to a change in clap. Adding an additional `--full-version` would be nice but that doesn't seem to play nicely with subcommands.
- Formatting and colors are a bit different

Closes https://github.com/nymtech/nym/issues/2061


mixnode/src/commands/describe.rs
```diff
@@ -1,35 +1,23 @@
...
-                .long(ID_ARG_NAME)
-                .help("The id of the mixnode you want to describe")
-                .takes_value(true)
-                .required(true),
-        )
+#[derive(Args)]
+pub(crate) struct Describe {
+    /// The id of the mixnode you want to describe
+    #[clap(long)]
+    id: String,
```
jstuczyn, line 12: This comment really applies to all argument parsing, but I don't feel like I have to copy-paste it everywhere : ) 
would it make sense to "restore" the `about`? In `clap` 3.0 it's applied via `help` macro, so in that case it'd be something like: `[clap(long, help = "Describe your mixnode....")]`

mixnode/src/commands/init.rs
```diff
@@ -5,70 +5,64 @@ use crate::commands::*;
...
+            announce_host: init_config.announce_host,
+            validators: init_config.validators,
+            wallet_address: init_config.wallet_address,
+        }
+    }
 }
 
-pub async fn execute(matches: ArgMatches<'static>) {
+pub(crate) async fn execute(args: &Init) {
     // TODO: this should probably be made implicit by slapping `#[tokio::main]` on our main method
```
jstuczyn: When you're doing further refactoring, could you remove that comment? It seems I forgot to get rid of it when I was removing the explicit `tokio::Runtime`

mixnode/src/commands/mod.rs
```diff
@@ -16,14 +16,38 @@ pub(crate) mod run;
...
+
+    /// Try to upgrade the mixnode
+    Upgrade(upgrade::Upgrade),
+
+    /// Show details of this mixnode
+    NodeDetails(node_details::NodeDetails),
+}
+
+// Configuration that can be overridden.
+struct OverrideConfig {
```
jstuczyn, line 41: Ah, clever, I like it!

mixnode/src/commands/run.rs
```diff
@@ -4,56 +4,52 @@
...
+    fn from(run_config: Run) -> Self {
+        OverrideConfig {
+            id: run_config.id,
+            host: run_config.host,
+            mix_port: run_config.mix_port,
+            verloc_port: run_config.verloc_port,
+            http_api_port: run_config.http_api_port,
+            announce_host: run_config.announce_host,
+            validators: run_config.validators,
+            wallet_address: None,
```
jstuczyn: I think this might be an omission from the previous code, because if you can set a particular value during `init`, you should be able to override it during `run` and hence the `wallet_address` field should also exist here.

mixnode/src/commands/sign.rs
```diff
@@ -3,39 +3,44 @@
...
+}
+
+impl From<Sign> for SignedTarget {
+    fn from(args: Sign) -> Self {
+        if let Some(text) = args.text {
+            SignedTarget::Text(text)
+        } else if let Some(address) = args.address {
+            SignedTarget::Address(address)
+        } else {
+            // Clap should guarantee this by using ArgGroup. Perhaps support for outputting the enum
```
jstuczyn: I only had a very brief look at the clap's macro stuff, so my suggestion might be completely wrong. But perhaps `ArgEnum` might solve this issue here?

jstuczyn (changes requested):
Generally looks good to me, the only thing I'd really want changed is to re-introduce the `help` text for each argument to describe its purpose

mixnode/src/commands/describe.rs
octol: Isn't this picked up from the doc comment? The string you're referring to is on the subcommand enum

mixnode/src/commands/init.rs
octol: Ah yes, good point!

neacsu (approved):
Lgtm

jstuczyn (approved):

PR #1048, opened 2022-01-19
feature/pedersen-commitments

### LINKED ISSUE #901: [Coconut]: Replace ElGamal with Pedersen commitments
Coconut scheme uses ElGamal to hide the private attributes. While working on the security analysis, Alfredo suggested a great idea to swap ElGamal to Pedersen commitments, which makes the protocol more efficient.

danielementary:
Replace original ElGamal encryption with Pedersen commitments for blind signatures.


common/nymcoconut/src/elgamal.rs
```diff
@@ -222,6 +223,18 @@ pub fn elgamal_keygen(params: &Parameters) -> ElGamalKeyPair {
     }
 }
 
+pub fn compute_attribute_encryption(
```
jstuczyn, line 226: This function doesn't seem to be used anywhere?

jstuczyn: As a matter of fact, are we still using `elgamal` anywhere at all?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -88,23 +86,21 @@ impl ProofCmCs {
     /// using the Fiat-Shamir heuristic.
```
jstuczyn, line 86: Just a sanity check, since I have absolutely no idea, does this comment still hold true?

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -550,61 +506,44 @@ mod tests {
...
     use rand::thread_rng;
 
-    use crate::scheme::issuance::{compute_attribute_encryption, compute_commitment_hash};
     use crate::scheme::keygen::keygen;
     use crate::scheme::setup::setup;
     use crate::scheme::verification::{compute_kappa, compute_zeta};
 
     use super::*;
 
+    // TODO check this test
```
jstuczyn: So was it checked? : )

common/nymcoconut/src/scheme/issuance.rs
```diff
@@ -343,37 +349,39 @@ mod tests {
...
-            &private_attributes,
-            &public_attributes,
-        )
-        .unwrap();
+        let public_attributes = params.n_random_scalars(2);
+        let (_commitments_openings, lambda) =
+            prepare_blind_sign(&mut params, &private_attributes, &public_attributes).unwrap();
 
         let bytes = lambda.to_bytes();
+        println!(
```
jstuczyn: It doesn't really make sense to print anything to `stdout` during tests. If those values are relevant, can we somehow compare them against "good" values or something instead?

common/nymcoconut/src/scheme/keygen.rs
```diff
@@ -219,7 +265,7 @@ where
     {
         let mut peekable = iter.peekable();
         let head_attributes = match peekable.peek() {
-            Some(head) => head.borrow().beta.len(),
+            Some(head) => head.borrow().beta_g2.len(),
```
jstuczyn, line 275: If I understood it correctly, I think we might need an additional check here to make sure `beta_g2.len() == beta_g1.len()`

common/nymcoconut/src/scheme/verification.rs
```diff
@@ -291,6 +293,7 @@ mod tests {
         .unwrap();
 
         let bytes = theta.to_bytes();
+        println!("Theta (Prove credential) containing blinded message, blinded serial number, credential, zkproof (with 2 private attributes) {:?}", bytes.len());
```
jstuczyn: Possibly redundant print?

jstuczyn (commented):
I can't really say anything about the correctness of the crypto itself, but I assume you and @aniampio verified it's correct. 

There's couple of tiny changes that you might (or not) want to introduce.

Also, since it changes how our coconut works, do we want to be merging that into develop or should we keep it in separate branch? @futurechimp @mmsinclair ? Perhaps we could still merge it since coconut is not really used anywhere just yet?

common/nymcoconut/src/proofs/mod.rs
aniampio: Yes, we still use the Fiat-Shamir heuristic in our zk proofs

common/nymcoconut/src/elgamal.rs
aniampio: No, we replaced ElGamal with Pedersen commitments. Should we then remove the whole elgamal package?

common/nymcoconut/src/proofs/mod.rs
danielementary: Yes :)

common/nymcoconut/src/scheme/issuance.rs
danielementary: Removed. Checks are already in place, so this should be OK.

common/nymcoconut/src/scheme/keygen.rs
danielementary: For a given VerificationKey, was it [freshly derived from a SecretKey](https://github.com/danielementary/nym/blob/e32301783b3fbd52e845a68631996dad084523a9/common/nymcoconut/src/scheme/keygen.rs#L74-L82) or [deserialized from bytes](https://github.com/danielementary/nym/blob/e32301783b3fbd52e845a68631996dad084523a9/common/nymcoconut/src/scheme/keygen.rs#L127-L197), it is enforced that betas_g1 and betas_g2 are of the same length. Does this additional check still be needed?

common/nymcoconut/src/scheme/verification.rs
danielementary: Removed. Checks are already in place, so this should be OK.

common/nymcoconut/src/proofs/mod.rs
```diff
@@ -213,32 +196,14 @@ impl ProofCmCs {
             .collect::<Vec<_>>();
 
         // recompute witnesses commitments
```
neacsu, line 198: Are those still witnesses commitments?

common/nymcoconut/src/scheme/mod.rs
```diff
@@ -641,18 +610,14 @@ mod tests {
...
         let r = params.random_scalar();
         let s = params.random_scalar();
-        let t = params.random_scalar();
-        let blinded_sig = BlindedSignature(
-            params.gen1() * t,
-            Ciphertext(params.gen1() * r, params.gen1() * s),
-        );
+        let blinded_sig = BlindedSignature(params.gen1() * r, params.gen1() * s);
         let bytes = blinded_sig.to_bytes();
+        println!("BlindedSignature: {:?}", bytes.len());
```
neacsu: I guess this `println!` is not needed either
```suggestion
```

common/nymcoconut/src/elgamal.rs
aniampio: I will keep this ElGamal create as it seems it is used by some external packages

common/nymcoconut/src/scheme/keygen.rs
```diff
@@ -316,7 +317,7 @@ impl VerificationKey {
         let beta_g1_len = self.beta_g1.len() as u64;
         let beta_g2_len = self.beta_g2.len() as u64;
         let mut bytes = Vec::with_capacity(
-            8 + (beta_g1_len + 1) as usize * 48 + (beta_g2_len + 1) as usize + 96,
+            96 + 8 + beta_g1_len as usize * 48 + beta_g2_len as usize * 96,
         );
```
neacsu: Why convert to `u64` and then later return to `usize`? It makes the operation harder to read and it's not really needed.

common/nymcoconut/src/scheme/keygen.rs
```diff
@@ -316,9 +316,8 @@ impl VerificationKey {
     pub fn to_bytes(&self) -> Vec<u8> {
         let beta_g1_len = self.beta_g1.len() as u64;
         let beta_g2_len = self.beta_g2.len() as u64;
-        let mut bytes = Vec::with_capacity(
-            96 + 8 + beta_g1_len as usize * 48 + beta_g2_len as usize * 96,
-        );
+        let mut bytes =
+            Vec::with_capacity(96 + 8 + beta_g1_len as usize * 48 + beta_g2_len as usize * 96);
```
neacsu: Same comment for type conversions

neacsu (approved):
Looks good overall. Is there something specific that needs to be done when generating keys, using for example https://github.com/nymtech/coconut/tree/develop/keygen-cli ?

PR #1057, opened 2022-01-21
Add mnemonic just on creation, to display it

neacsu:


tommyv1987 (approved):
LGTM, tested this out, a change is required on the FE, which will be done in a separate branch.

PR #1060, opened 2022-01-21
Mixnet Contract constants extraction

jstuczyn:
This PR basically extracts relevant constants, i.e. those that could realistically be ever controlled by governance, such as reward percent per interval to a `constants.rs` file. 

There is, however, one question that I'd like to have @durch 's opinion on:
should `INITIAL_ACTIVE_SET_WORK_FACTOR` be treated similarly to `SYBIL_RESISTANCE_PERCENT`, ie. should only ever (if at all) be changed through governance rather than by permitted owner address? To remind, currently `active_set_work_factor` lives as field on our `ContractStateParams` that can be modified by the contract owner at any time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3625,7 +3625,6 @@ dependencies = [
  "cfg-if 1.0.0",
  "hex-literal",
  "serde",
- "time 0.3.5",
  "url",
 ]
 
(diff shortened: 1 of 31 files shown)


neacsu (approved):

jstuczyn:
As per discussion with @durch , I'm going to update handling of `INITIAL_ACTIVE_SET_WORK_FACTOR`

PR #1061, opened 2022-01-21
Rework vesting contract storage

durch:
Apparently you can't just make up storage keys on the fly, it needs to be `const`

### DIFF
--- a/contracts/vesting/src/contract.rs
+++ b/contracts/vesting/src/contract.rs
@@ -288,16 +288,8 @@ fn try_create_periodic_vesting_account(
     env: Env,
     deps: DepsMut,
 ) -> Result<Response, ContractError> {
-    cfg_if::cfg_if! {
-        if #[cfg(test)] {
-            if info.sender != "admin" {
-                return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
-            }
-        } else {
-            if info.sender != ADMIN.load(deps.storage)? {
-                return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
-            }
-        }
+    if info.sender != ADMIN.load(deps.storage)? {
+        return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
     }
 
     let coin = validate_funds(&info.funds)?;
--- a/contracts/vesting/src/contract.rs
+++ b/contracts/vesting/src/contract.rs
@@ -48,9 +48,10 @@ pub fn execute(
     msg: ExecuteMsg,
 ) -> Result<Response, ContractError> {
     match msg {
-        ExecuteMsg::DelegateToMixnode { mix_identity } => {
-            try_delegate_to_mixnode(mix_identity, info, env, deps)
-        }
+        ExecuteMsg::DelegateToMixnode {
+            mix_identity,
+            amount,
+        } => try_delegate_to_mixnode(mix_identity, amount, info, env, deps),
         ExecuteMsg::UndelegateFromMixnode { mix_identity } => {
             try_undelegate_from_mixnode(mix_identity, info, deps)
         }
@@ -77,7 +78,8 @@ pub fn execute(
         ExecuteMsg::BondMixnode {
             mix_node,
             owner_signature,
-        } => try_bond_mixnode(mix_node, owner_signature, info, env, deps),
+            amount,
+        } => try_bond_mixnode(mix_node, owner_signature, amount, info, env, deps),
         ExecuteMsg::UnbondMixnode {} => try_unbond_mixnode(info, deps),
         ExecuteMsg::TrackUnbondMixnode { owner, amount } => {
             try_track_unbond_mixnode(&owner, amount, info, deps)
@@ -85,7 +87,8 @@ pub fn execute(
         ExecuteMsg::BondGateway {
             gateway,
             owner_signature,
-        } => try_bond_gateway(gateway, owner_signature, info, env, deps),
+            amount,
+        } => try_bond_gateway(gateway, owner_signature, amount, info, env, deps),
         ExecuteMsg::UnbondGateway {} => try_unbond_gateway(info, deps),
         ExecuteMsg::TrackUnbondGateway { owner, amount } => {
             try_track_unbond_gateway(&owner, amount, info, deps)
@@ -180,11 +183,12 @@ fn try_update_staking_address(
 pub fn try_bond_gateway(
     gateway: Gateway,
     owner_signature: String,
+    amount: Coin,
     info: MessageInfo,
     env: Env,
     deps: DepsMut,
 ) -> Result<Response, ContractError> {
-    let pledge = validate_funds(&info.funds)?;
+    let pledge = validate_funds(&[amount])?;
     let account = account_from_address(info.sender.as_str(), deps.storage, deps.api)?;
     account.try_bond_gateway(gateway, owner_signature, pledge, &env, deps.storage)
 }
@@ -211,11 +215,12 @@ pub fn try_track_unbond_gateway(
 pub fn try_bond_mixnode(
     mix_node: MixNode,
     owner_signature: String,
+    amount: Coin,
     info: MessageInfo,
     env: Env,
     deps: DepsMut,
 ) -> Result<Response, ContractError> {
-    let pledge = validate_funds(&info.funds)?;
+    let pledge = validate_funds(&[amount])?;
     let account = account_from_address(info.sender.as_str(), deps.storage, deps.api)?;
     account.try_bond_mixnode(mix_node, owner_signature, pledge, &env, deps.storage)
 }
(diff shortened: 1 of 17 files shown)


contracts/vesting/src/storage.rs
```diff
@@ -1,16 +1,121 @@
...
+use cosmwasm_std::Order;
+use cosmwasm_std::{Addr, Api, Storage, Uint128};
 use cw_storage_plus::{Item, Map};
 
 pub const KEY: Item<u32> = Item::new("key");
 const ACCOUNTS: Map<Addr, Account> = Map::new("acc");
+// Holds data related to individual accounts
+const BALANCES: Map<u32, Uint128> = Map::new("blc");
+const BOND_PLEDGES: Map<u32, PledgeData> = Map::new("bnd");
+const GATEWAY_PLEDGFES: Map<u32, PledgeData> = Map::new("gtw");
```
jstuczyn: `PLEDGFES` typo

contracts/vesting/src/storage.rs
```diff
@@ -1,16 +1,121 @@
...
+const DELEGATIONS: Map<(u32, &[u8], u64), Uint128> = Map::new("dlg");
```
jstuczyn: 1. Could we use some type aliases for the types used in the key? I had to go looking for what `&[u8]` and `u64` are supposed to represent here.
2. Isn't it possible to instead of `&[u8]` use the `IdentityKey` (`String` alias) type?
3. Is block `time` what we want for the third part of the key? Maybe `height` is more appropriate? (note:  I have absolutely no idea, it just felt weird to be that we are looking up values by time, which isn't super predictable). Should it even be part of the key to begin with? Perhaps it would be better as a value?
4. Do we even need a composite key given that the storage key, generated via `generate_storage_key` is bound to be unique for each delegation?

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -124,105 +110,88 @@ impl Account {
...
-    pub fn delegation_keys_for_mix(&self, mix: &str, storage: &dyn Storage) -> Vec<u64> {
-        let key = self.delegations_key();
-        let delegations: Map<(&[u8], u64), Uint128> = Map::new(&key);
-        delegations
-            .prefix(mix.as_bytes())
-            .keys(storage, None, None, Order::Ascending)
-            // Scan will blow up on first error
-            .scan((), |_, x| x.ok())
-            .collect::<Vec<u64>>()
+    pub fn delegation_block_times_for_mix(
```
jstuczyn: Do we care about block **times** or block heights instead?

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -124,105 +110,88 @@ impl Account {
...
+        &self,
+        mix: &str,
+        storage: &dyn Storage,
+    ) -> Result<Vec<u64>, ContractError> {
+        let delegations = load_delegations_for_mix(self.storage_key(), mix, storage)?;
```
jstuczyn: Also, how many values do you reckon could be loaded at the most? Because the of usual, it's an unbonded call and all of the associated "nice" stuff could happen...

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -231,18 +200,15 @@ impl Account {
         mix: IdentityKey,
         storage: &dyn Storage,
     ) -> Result<Uint128, ContractError> {
-        Ok(self
-            .delegations_for_mix(mix, storage)?
+        Ok(load_delegations_for_mix(self.storage_key(), &mix, storage)?
```
jstuczyn: Same as before. Another unbonded call : (
This thing is the reason there exists a memoised value for total delegation of particular mixnode. Perhaps it would also be appropriate here?

contracts/vesting/src/vesting/account/vesting_account.rs
```diff
@@ -115,21 +114,15 @@ impl VestingAccount for Account {
...
-        let mut amount = Uint128::zero();
-        for (mix, block_time) in delegations_keys {
-            amount += delegations.load(storage, (&mix, block_time))?
-        }
-        amount = Uint128::new(amount.u128().min(max_vested));
+
+        let coin = load_delegations_all(self.storage_key(), storage)?
+            .into_iter()
+            .filter(|((_mix, block_time), _amount)| *block_time < start_time)
+            .fold(Uint128::zero(), |acc, ((_mix, _block_time), amount)| {
```
jstuczyn, line 123: ibid.

contracts/vesting/src/contract.rs
```diff
@@ -277,9 +282,18 @@ fn try_create_periodic_vesting_account(
     env: Env,
     deps: DepsMut,
 ) -> Result<Response, ContractError> {
-    if info.sender != ADMIN_ADDRESS {
-        return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
+    cfg_if::cfg_if! {
+        if #[cfg(test)] {
```
neacsu: This doesn't look very nice, putting test code in the same place with production code. Isn't it possible to get the sender address in the test code? Why would we need this here?

contracts/vesting/src/errors.rs
```diff
@@ -4,7 +4,7 @@ use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq)]
 pub enum ContractError {
-    #[error("{0}")]
+    #[error("{} - {0}", line!())]
```
neacsu: This looks like something we could use on more than one error type, if it's really needed

jstuczyn (commented):
Just one sanity check question: does the vesting contract check whether user has the specified amount of tokens available to them?

contracts/vesting/src/storage.rs
durch: 1. Done,
2. Yup
3. `height` is probably a better choice
4. Its possible for them to delegate to the same node more then once, so we keep track of each individual delegation due to vesting accounting, ie we want to be able to say how many delegated coins are vested vs unvested, although the way that is implemented could be reworked to just store delegation total.

contracts/vesting/src/vesting/account/mod.rs
durch: I'm hoping that no one will delegate to a mix a 100 times, I can't imagine what the use case would be, then again, someone might do it just for fun.

durch: Does it make a difference that we're not materialising all the values, but iterating and folding?

contracts/vesting/src/vesting/account/mod.rs
durch: Does it make a difference that we're not materialising all the values, but iterating and folding? Vesting contract does to have access to that mix node, what we're getting here are total_delegations of this particular vesting contract for a given mix node

durch: This is annoying, I need a way to just generate the pagination code...

contracts/vesting/src/contract.rs
durch: Admin address is set at init, I can probably mock that, have it be nicer, good call

contracts/vesting/src/errors.rs
durch: I was chasing down an error, that was not even coming from the contract, pretty sure we don't really need it

durch:
> Just one sanity check question: does the vesting contract check whether user has the specified amount of tokens available to them?

~That's a great check, as we've moved to just passing in the amount, I've not introduced the check, great catch!~

I was totally wrong, looks like the paranoid me of the past has actually wrote code to check for that...

+ [gateway_bonding](https://github.com/nymtech/nym/blob/cc63f3f21b2474ae619a9a1c0939efe05bcd40a7/contracts/vesting/src/vesting/account/gateway_bonding_account.rs#L24)
+ [mixnode_bonding](https://github.com/nymtech/nym/blob/cc63f3f21b2474ae619a9a1c0939efe05bcd40a7/contracts/vesting/src/vesting/account/mixnode_bonding_account.rs#L24)
+ [mixnode_delegation](https://github.com/nymtech/nym/blob/cc63f3f21b2474ae619a9a1c0939efe05bcd40a7/contracts/vesting/src/vesting/account/delegating_account.rs#L24)
+ [withdraw](https://github.com/nymtech/nym/blob/cc63f3f21b2474ae619a9a1c0939efe05bcd40a7/contracts/vesting/src/contract.rs#L121)

contracts/vesting/src/traits/delegating_account.rs
```diff
@@ -18,12 +18,12 @@ pub trait DelegatingAccount {
...
 
     // track_delegation performs internal vesting accounting necessary when
-    // delegating from a vesting account. It accepts the current block time, the
+    // delegating from a vesting account. It accepts the current block height, the
     // delegation amount and balance of all coins whose denomination exists in
     // the account's original vesting balance.
     fn track_delegation(
         &self,
-        block_time: Timestamp,
+        block_time: u64,
```
neacsu:
```suggestion
        block_height: u64,
```

contracts/vesting/src/storage.rs
durch: I'll resolve this and merge, but lets continue the discussion if we need to, I just don't want this sitting in here for too long. Current storage is better then it was at least memory wise, we can continue to optimise it

PR #1065, opened 2022-01-24
De-'float'-ing `Interval` (`Display` impl + `serde`)

jstuczyn:


durch (approved):
Good stuff with date repr

PR #1066, opened 2022-01-24
Feature/wasm client

neacsu:
There is an `unused-unit` lint that will fail for now in the CI, but this is regarded as a false positive and should eventually get fixed:
https://github.com/rustwasm/wasm-bindgen/issues/2774


jstuczyn (approved):
Wow, that was way simpler (in terms of required changes) than I thought it would be.

By the way, perhaps you could temporarily allow the failing lint so that the CI would be happier? via `#![allow(clippy::...` on the top of the file

PR #1073, opened 2022-01-26
Add trait to mock client dependency in DelayForwarder

octol:
- Add trait for mocking the mixnode client dependency in the `DelayForwarder`.
- Create basic test instantiating the `DelayForwarder` and pass it a packet.

A alternative approach here could be to move the entire `Client` behind a trait instead of strictly what's needed for testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3904,6 +3904,8 @@ dependencies = [
  "mixnode-common",
  "nonexhaustive-delayqueue",
  "nymsphinx",
+ "nymsphinx-params",
+ "nymsphinx-types",
  "pemstore",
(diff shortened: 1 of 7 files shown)


neacsu (approved):

jstuczyn (approved):
Conceptually looks good to me. The only thing I'm maybe uncertain of is the name of the trait, i.e. `SendWithoutResponse`. I've been kinda liberally using `send`, `send_with_response` and `send_without_response` in multiple places, so I'm not sure if it might not be confusing later on. But I guess that's the problem for the future once other parts of the codebase are looked at.

PR #1075, opened 2022-01-26
Tidy nym-mixnode module visibility

octol:
- Some minor module visibility stuff I noticed when reading through the code. 
- Remove duplicate `load_identity_keys()` function.

### DIFF
--- a/mixnode/src/commands/describe.rs
+++ b/mixnode/src/commands/describe.rs
@@ -1,6 +1,7 @@
-use crate::commands::*;
 use crate::config::Config;
 use crate::node::node_description::NodeDescription;
+use clap::Args;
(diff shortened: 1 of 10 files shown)


neacsu (approved):
Those are all nice changes. Just as a note, this command refactor work could mostly apply to gateway & clients code too, so maybe we should have issues for doing this to the other components as well, so that we keep the codebase homogenous.

octol:
> Those are all nice changes. Just as a note, this command refactor work could mostly apply to gateway & clients code too, so maybe we should have issues for doing this to the other components as well, so that we keep the codebase homogenous.

Yep agree!

jstuczyn (approved):
lgtm

PR #1076, opened 2022-01-27
Make wallet_address mandatory for mixnode init

octol:


neacsu (commented):
Why does it have to be mandatory at `init`?

tommyv1987 (approved):
LGTM, tested, everything working as expected.

tommyv1987:
@neacsu - when the user `init`'s their node it currently provides the user with the owner signature which they will then use to bond their mixnode in the wallet or cli.

Otherwise running it without providing it upfront resulted in the following:
```
Error: wallet decoding failed: bad encoding
Exiting
```

jstuczyn:
@tommyv1987 with this change I wonder, would it be a good idea to limit number of things you can set in `run` command? I.e. for example once you set your wallet address in `init`, you'd have to stick with it (or manually modify config file, but that kinda implies you semi-know what you're doing)

jstuczyn (approved):

PR #1077, opened 2022-01-27
Setup basic test for mixnode stats reporting

octol:
Part of: https://github.com/nymtech/nym/issues/2059

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ panic = "abort"
 
 [workspace]
 
+resolver = "2"
 members = [
     "clients/client-core",
     "clients/native",
--- a/mixnode/Cargo.toml
+++ b/mixnode/Cargo.toml
@@ -49,6 +49,7 @@ version-checker = { path="../common/version-checker" }
 
 [dev-dependencies]
 serial_test = "0.5"
+tokio = { version="1.8", features = ["rt-multi-thread", "net", "signal", "test-util"] }
 
 [build-dependencies]
 vergen = { version = "5", default-features = false, features = ["build", "git", "rustc", "cargo"] }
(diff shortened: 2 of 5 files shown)


jstuczyn:
It looks like those new tests are not passing?

octol:
> It looks like those new tests are not passing?

Yeah that assert for the transient state is iffy, removed for now

jstuczyn:
Furthermore, I'm not really a fan of calling `sleep` inside the test code. Couldn't we just "advance" time without waiting for it? I.e. take a look at: https://docs.rs/tokio/latest/tokio/time/fn.advance.html

octol:
> Furthermore, I'm not really a fan of calling `sleep` inside the test code. Couldn't we just "advance" time without waiting for it? I.e. take a look at: https://docs.rs/tokio/latest/tokio/time/fn.advance.html

Updated to use this instead. However this depends on `test-util` in tokio (and `time`).

octol:
> > Furthermore, I'm not really a fan of calling `sleep` inside the test code. Couldn't we just "advance" time without waiting for it? I.e. take a look at: https://docs.rs/tokio/latest/tokio/time/fn.advance.html
> 
> Updated to use this instead. However this depends on `test-util` in tokio (and `time`).

and also `resolver = 2`

octol:
See https://github.com/nymtech/nym/pull/1086 for changing to rust edition 2021, with which we can drop `resolver = 2`

jstuczyn (approved):

octol:
> See #1086 for changing to rust edition 2021, with which we can drop `resolver = 2`

Seems like we still need `resolver = 2`, even after merging #1086. Looking at compiler output it still pulls in dev-dependency features. Maybe deps also need to opt-in to 2021 for it to work automatically

PR #1083, opened 2022-01-27
Feature/flexible vesting + utility queries

durch:
Change log:

+ [`CreateAccount`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/contracts/vesting/src/messages.rs#L47) message now has a [`VestingSpecification`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/contracts/vesting/src/messages.rs#L17) attribute that allows for flexible vesting schedules. Rust `nymd` [`create_periodic_vesting_account`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/common/client-libs/validator-client/src/nymd/traits/vesting_signing_client.rs#L276) function has also changed to take the `vesting_spec` parameter.

Three utility functions added to Rust `nymd` vesting client:

+ [`get_account`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/common/client-libs/validator-client/src/nymd/traits/vesting_query_client.rs#L183)
+ [`get_mixnode_pledge`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/common/client-libs/validator-client/src/nymd/traits/vesting_query_client.rs#L191)
+ [`get_gateway_pledge`](https://github.com/nymtech/nym/blob/90701f843fee7f2730d6a0a6db7478e870355250/common/client-libs/validator-client/src/nymd/traits/vesting_query_client.rs#L199)

### DIFF
--- a/contracts/vesting/src/contract.rs
+++ b/contracts/vesting/src/contract.rs
@@ -4,7 +4,7 @@ use crate::storage::{account_from_address, ADMIN, MIXNET_CONTRACT_ADDRESS};
 use crate::traits::{
     DelegatingAccount, GatewayBondingAccount, MixnodeBondingAccount, VestingAccount,
 };
-use crate::vesting::{populate_vesting_periods, Account};
+use crate::vesting::{populate_vesting_periods, Account, PledgeData};
 use config::defaults::DENOM;
 use cosmwasm_std::{
     coin, entry_point, to_binary, BankMsg, Coin, Deps, DepsMut, Env, MessageInfo, QueryResponse,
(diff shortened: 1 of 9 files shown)


contracts/vesting/src/contract.rs
```diff
@@ -389,14 +388,25 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> Result<QueryResponse, Contr
...
-    Ok(Some(account))
+    account.load_mixnode_pledge(deps.storage)
+}
+
+pub fn try_get_gateway(address: &str, deps: Deps) -> Result<Option<PledgeData>, ContractError> {
+    let account = account_from_address(address, deps.storage, deps.api)?;
+    account.load_gateway_pledge(deps.storage)
+}
+
+pub fn try_get_account(address: &str, deps: Deps) -> Result<Account, ContractError> {
```
jstuczyn, line 408: Shouldn't this return `Option<Account>`? What if given address is not vesting?

contracts/vesting/src/messages.rs
```diff
@@ -13,6 +13,27 @@ pub struct InitMsg {
 #[serde(rename_all = "snake_case")]
 pub struct MigrateMsg {}
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, Default)]
+pub struct VestingSpecification {
+    start_time: Option<u64>,
+    period_seconds: Option<u32>,
+    num_periods: Option<u32>,
```
jstuczyn: If you're already passing `Option<VestingSpecification>` when creating vesting accounts, does it make sense for all the fields here to be wrapped in `Option` as well?

contracts/vesting/src/vesting/mod.rs
```diff
@@ -1,19 +1,21 @@
...
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct VestingPeriod {
     pub start_time: u64,
+    pub period_seconds: u32,
 }
 
 impl VestingPeriod {
     pub fn end_time(&self) -> Timestamp {
-        Timestamp::from_seconds(self.start_time + VESTING_PERIOD)
+        Timestamp::from_seconds(self.start_time + self.period_seconds as u64)
```
jstuczyn, line 18: if you have to cast it here to u64, why not keep `period_seconds` as u64 to begin with?

contracts/vesting/src/contract.rs
durch: It used to, the thing is that validate will throw an error if the account does not exist, so I think we can just use that, although it was late yesterday, so I can just put the Option back in :)

contracts/vesting/src/messages.rs
durch: Each field has a default, so any combination of them can be passed in,, hence the multitude of Options

contracts/vesting/src/vesting/mod.rs
durch: Yeah, probably best, I always try to save on types, which usually turns out to be silly

PR #1084, opened 2022-01-28
Tag contract errors, and print out lines for easier QA

durch:
Since vesting <> mixnet interplay emits a few messages back and forth this should allow for easier tracing where exactly things break. We can remove this once we're happy with how things work


jstuczyn:
would `deps.debug()` help?

PR #1085, opened 2022-01-28
Fixes infinite loops in requests involving pagination

jstuczyn:
Turns out, pagination can be `Some` even if further requests are not required


neacsu (approved):
Maybe also add a comment for each, as this is not immediately obvious why the extra check is needed

PR #1090, opened 2022-02-01
Prevent contract overwriting

durch:
Prevents overwriting vesting accounts

### DIFF
--- a/contracts/vesting/src/contract.rs
+++ b/contracts/vesting/src/contract.rs
@@ -304,6 +304,12 @@ fn try_create_periodic_vesting_account(
     if info.sender != ADMIN.load(deps.storage)? {
         return Err(ContractError::NotAdmin(info.sender.as_str().to_string()));
     }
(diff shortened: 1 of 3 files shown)


tommyv1987:
- Steps to test:
Create a new account and a new vesting period, then try to overwrite the initial vesting period. The response back from the contract successfully returns an account already exists with the following:

```
called `Result::unwrap()` on an `Err` value: AbciError(18, Log("failed to execute message; message index: 0: VESTING (45): Account already exists: nymt12aq68fwzv7e38jhm77xdqx20ydywf0mxfmd8sf: execute wasm contract failed: invalid request"))
thread 'tests::create_periodic_vesting_for_new_user' panicked at 'called `Result::unwrap()` on an `Err` value: AbciError(18, Log("failed to execute message; message index: 0: VESTING (45): Account already exists: nymt12aq68fwzv7e38jhm77xdqx20ydywf0mxfmd8sf: execute wasm contract failed: invalid request"))
```

Happy with this.

jstuczyn (approved):

PR #1091, opened 2022-02-01
Add UpdateMixnodeConfigOnBehalf to vestng contract

durch:


tommyv1987:
Happy with this. 
Steps to test:
Migrate both vesting and mixnet contracts.
Have an existing account with a node bonded and perform the following:
```
 let mut rng = rand::thread_rng();
        let profit_percent: u8 = rng.gen_range(1..100);

        let response = client.vesting_update_mixnode_config(profit_percent)
            .await
            .unwrap();

        println!("{:?}", response);
 ```
 Then, get the account details and the mixnode profit percentage should be updated with the value set above: 
 
 ```
 Some(MixNodeBond { pledge_amount: Coin { denom: "unymt", amount: Uint128(100000000) }, total_delegation: Coin { denom: "unymt", amount: Uint128(0) }, owner: Addr("address"), layer: One, block_height: 442421, mix_node: MixNode { host: "1.1.1.1", mix_port: 1789, verloc_port: 1790, http_api_port: 8000, sphinx_key: "key", identity_key: "key", version: "0.12.1", profit_margin_percent: 99 }, proxy: Some(Addr("address")) })
 ```

**Comment by jstuczyn** [comment]
Should we perhaps futureproof ourselves a bit and make it take Option<ProfitMargin> and Option<OperatingCosts> to let you update both?

File: contracts/mixnet/src/mixnodes/transactions.rs [hunk 1]
```diff
@@ -248,7 +277,19 @@ pub(crate) fn try_update_mixnode_config(
...
+    let mut response = Response::new();
+
+    if let Some(proxy) = proxy {
+        // Returns one_unym proxy had to send in order to execute the contract to contract transaction, this is potentially leaky as anyone can say that they're a proxy,
+        // and they could potentially leak 1 unym per transaction, altough I'm pretty sure transaction fees make that silly.
+        let return_one_unymt = BankMsg::Send {
+            to_address: proxy.as_str().to_string(),
+            amount: vec![one_unym()],
+        };
+        response = response.add_message(return_one_unymt);
```
neacsu: The variable name and response message (and fwiw the comment too) aren't valid for all networks

contracts/vesting/src/vesting/account/mixnode_bonding_account.rs
```diff
@@ -6,12 +6,34 @@ use cosmwasm_std::{wasm_execute, Coin, Env, Response, Storage, Uint128};
 use mixnet_contract_common::{ExecuteMsg as MixnetExecuteMsg, MixNode};
 use vesting_contract_common::events::{
     new_vesting_mixnode_bonding_event, new_vesting_mixnode_unbonding_event,
+    new_vesting_update_mixnode_config_event,
 };
 use vesting_contract_common::one_unym;
```
neacsu: Not necessarily part of this PR, but this function name also shouldn't make reference to token name

contracts/mixnet/src/mixnodes/transactions.rs
```diff
@@ -248,7 +277,19 @@ pub(crate) fn try_update_mixnode_config(
             .ok_or(ContractError::NoBondFound)
     })?;
 
-    Ok(Response::new())
+    let mut response = Response::new();
+
+    if let Some(proxy) = proxy {
+        // Returns one_unym proxy had to send in order to execute the contract to contract transaction, this is potentially leaky as anyone can say that they're a proxy,
+        // and they could potentially leak 1 unym per transaction, altough I'm pretty sure transaction fees make that silly.
```
neacsu: Is the 1 unym story documented somewhere (at least as in-line comments)? I only got the brief idea that this is necessary, but I don't currently understand why that is.

durch: The only documentation is that comment, would be good to reify is else where

contracts/vesting/src/vesting/account/mixnode_bonding_account.rs
durch: These are both good calls, I'll fix it

contracts/mixnet/src/mixnodes/transactions.rs
jstuczyn: Have we actually mentioned this issue to cosmwasm people? Maybe there's a more sane workaround to this?

durch: I've not brought it to them, I'll open a thing in their discord

durch:
> Should we perhaps futureproof ourselves a bit and make it take Option and Option to let you update both?

I'll merge it like this for now, its been dangling a while

PR #1093, opened 2022-02-02
Wallet signing on MacOS

mmsinclair:
Adds two GitHub Actions:

1. Create a release when pushing to a tag `nym-wallet-*`
2. Builds, signs and uploads the wallet `dmg` to the release

Once merged into develop, the process of triggering these actions is as follows:

- write code, create PR, approve and merge PR to `develop`
- create a new release and name tag `nym-wallet-v1.0.0`
- this triggers the GH Action to mark the release as `draft` and in parallel starts building on MacOS

Soon, we can add GH Actions to do the same for Windows and Linux

### DIFF
--- a/nym-wallet/src-tauri/tauri.conf.json
+++ b/nym-wallet/src-tauri/tauri.conf.json
@@ -26,7 +26,7 @@
       "copyright": "Copyright © 2021-2022 Nym Technologies SA",
       "category": "Business",
       "shortDescription": "Nym desktop wallet allows you to manage your Nym tokens",
-      "longDescription": "The Nym wallet allows you to manage your Nym tokens and to stake on mixnodes, gateways and validators.",
+      "longDescription": "",
       "deb": {
         "depends": [],
         "useBootstrapper": false
--- a/nym-wallet/src-tauri/tauri.conf.json
+++ b/nym-wallet/src-tauri/tauri.conf.json
@@ -1,7 +1,7 @@
 {
   "package": {
     "productName": "nym-wallet",
-    "version": "0.12.1"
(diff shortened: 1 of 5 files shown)


nym-wallet/src-tauri/tauri.conf.json
```diff
@@ -1,7 +1,7 @@
 {
   "package": {
     "productName": "nym-wallet",
-    "version": "0.12.1"
+    "version": "1.0.0"
```
jstuczyn, line 4: Should we also update Cargo.toml file?

nym-wallet/src-tauri/tauri.conf.json
```diff
@@ -23,10 +23,10 @@
...
       "resources": [],
       "externalBin": [],
-      "copyright": "",
-      "category": "DeveloperTool",
-      "shortDescription": "",
-      "longDescription": "",
+      "copyright": "Copyright © 2021-2022 Nym Technologies SA",
+      "category": "Business",
+      "shortDescription": "Nym desktop wallet allows you to manage your Nym tokens",
+      "longDescription": "The Nym wallet allows you to manage your Nym tokens and to stake on mixnodes, gateways and validators.",
```
jstuczyn: Well, technically you can't stake on gateways (at all) and the wallet itself doesn't let you to stake on validators

nym-wallet/src-tauri/tauri.conf.json
mmsinclair: Yes will do that too and probably also the package.json

jstuczyn (approved):

PR #1099, opened 2022-02-04
Fixed underflow in rewarding all delegators

jstuczyn:
The issue was that `remaining_delegators -= MIXNODE_DELEGATORS_PAGE_LIMIT;` has been called in the last loop iteration, where `remaining_delegators` has already been smaller than `MIXNODE_DELEGATORS_PAGE_LIMIT` (and it has type `usize`)

### DIFF
--- a/validator-api/src/nymd_client.rs
+++ b/validator-api/src/nymd_client.rs
@@ -268,10 +268,6 @@ impl<C> Client<C> {
     where
         C: SigningCosmWasmClient + Sync,
     {
-        // determine how many times we are going to have to call the delegator rewarding,
-        // note that it doesn't include the "base" call to `RewardMixnode` that rewards one page
-        let further_calls = node.total_delegations / MIXNODE_DELEGATORS_PAGE_LIMIT;
-
         // start with the base call to reward operator and first page of delegators
         let msgs = vec![(node.to_reward_execute_msg(interval_id), vec![])];
         let memo = format!(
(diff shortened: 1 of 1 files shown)


validator-api/src/nymd_client.rs
```diff
@@ -281,20 +277,27 @@ impl<C> Client<C> {
...
-        );
-        for _ in 0..further_calls {
-            let delegators_in_call = remaining_delegators.min(MIXNODE_DELEGATORS_PAGE_LIMIT);
-            let msgs = vec![delegator_rewarding_msg.clone()];
-            let memo = format!("rewarding another {} delegators", delegators_in_call);
-            self.execute_multiple_with_retry(msgs, Default::default(), memo)
-                .await?;
-
-            remaining_delegators -= MIXNODE_DELEGATORS_PAGE_LIMIT;
+        // reward rest of delegators (if applicable)
```
durch, line 280: In think we can just change that sub to saturating sub, ie:
```
let mut remaining_delegators = node.total_delegations.saturating_sub(MIXNODE_DELEGATORS_PAGE_LIMIT);
```

jstuczyn: I was actually about to apply your suggestion, but then realised it wouldn't totally solve the problem, since `further_calls` variable is actually calculated incorrectly (in the case when `node.total_delegations % MIXNODE_DELEGATORS_PAGE_LIMIT == 0`). However, I used your `saturating_sub` idea to simplify the loop

PR #1100, opened 2022-02-04
nym-gateway: upgrade clap and use declarative cli argument parsing

octol:
Upgrade clap to latest and revamp command line parsing for `nym-gateway`

Changes in behaviour:

- Changed `--eth_endpoint` to `--eth-endpoint` for consistency
- Flags related to `eth` feature flag not available for `run`, consistent with `init` (`--testnet-mode`, `--eth-endpoint`, `--validators`, `--mnemonic`)

### DIFF
--- a/gateway/src/commands/mod.rs
+++ b/gateway/src/commands/mod.rs
@@ -3,8 +3,8 @@
 
 use std::process;
 
-use crate::config::Config;
-use clap::ArgMatches;
+use crate::{config::Config, Cli};
+use clap::Subcommand;
 use colored::Colorize;
 use crypto::bech32_address_validation;
 use url::Url;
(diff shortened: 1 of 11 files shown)


gateway/src/commands/mod.rs
```diff
@@ -15,30 +15,66 @@ pub(crate) mod run;
...
-#[cfg(not(feature = "coconut"))]
+#[cfg(all(feature = "eth", not(feature = "coconut")))]
 const DEFAULT_ETH_ENDPOINT: &str = "https://rinkeby.infura.io/v3/00000000000000000000000000000000";
-#[cfg(not(feature = "coconut"))]
+#[cfg(all(feature = "eth", not(feature = "coconut")))]
 const DEFAULT_VALIDATOR_ENDPOINT: &str = "http://localhost:26657";
 // A dummy mnemonic
-#[cfg(not(feature = "coconut"))]
+#[cfg(all(feature = "eth", not(feature = "coconut")))]
 const DEFAULT_MNEMONIC: &str = "typical regret aware used tennis noise resource crisp defy join donate orient army item immense clean emerge globe gift chronic loan flat enter egg";
```
jstuczyn, line 24: Uhm... I'm slightly confused by why that even exists

jstuczyn: (i.e. the `DEFAULT_MNEMONIC`)

gateway/src/commands/mod.rs
```diff
@@ -50,82 +86,72 @@ fn parse_validators(raw: &str) -> Vec<Url> {
...
     }
 
-    #[cfg(not(feature = "coconut"))]
-    if let Some(eth_endpoint) = matches.value_of(ETH_ENDPOINT) {
-        config = config.with_eth_endpoint(String::from(eth_endpoint));
-    } else {
-        config = config.with_eth_endpoint(String::from(DEFAULT_ETH_ENDPOINT));
-    }
+    // Set the test mode to always on, except if build with `feature = eth`, then we check the flag
+    config = config.with_testnet_mode(true);
```
jstuczyn, line 126: I think you forgot to remove that one, because just underneath it, it's properly locked behind the feature flag : )

mixnode/src/commands/init.rs
```diff
@@ -12,7 +12,7 @@ use super::OverrideConfig;
 
 #[derive(Args, Clone)]
 pub(crate) struct Init {
-    /// Initialise the mixnode
+    /// Id of the nym-mixnode we want to create config for
```
jstuczyn, line 15: I don't think we really need the `nym-` prefix here

jstuczyn (commented):
I think I mentioned it to you before, the config template was rather finicky. Just as a sanity check, can you make sure that if you compile the gateway with either `eth` or `coconut` feature, `init` will nicely create a config file and `run` will be able to read it from the disk?

gateway/src/commands/mod.rs
octol: Currently we do 
```rust
    if !cfg!(feature = "eth") || matches.is_present(TESTNET_MODE_ARG_NAME) {
        config.with_testnet_mode(true)
```
so the above line is to cover for when `not(feature = eth)`. That said, I just tried to match the old behaviour, perhaps we want to change this?

gateway/src/commands/mod.rs
octol: Seems like it was added very recently, so thought I'd leave it in

gateway/src/commands/mod.rs
octol: But yes you're right, this currently logic is fishy, I'm redoing it. And good tip about generating the config template to confirm it all works! (it didn't...)

octol:
These changes + fixes already merged in to develop

Issue #1096, opened 2022-02-03
Make mainnet coin transfers work


futurechimp:
Refactor keypair generation so that coin transfers work in mainnet. Currently the wallet is able to derive addresses for mainnet, but can't transfer coins (due to the way the login flow works).

PR #1128, opened 2022-02-23
Mainnet coin transfers in wallet

octol:
To enable mainnet coin transfers in the wallet we need to be able to pass the prefix and denomination and not rely on the default.

Part of: https://github.com/nymtech/nym/issues/1096


common/client-libs/validator-client/src/nymd/fee/gas_price.rs
```diff
@@ -71,11 +71,9 @@ impl FromStr for GasPrice {
     }
 }
 
-impl Default for GasPrice {
-    fn default() -> Self {
-        format!("{}{}", defaults::GAS_PRICE_AMOUNT, defaults::DENOM)
-            .parse()
-            .unwrap()
+impl GasPrice {
+    pub fn default(denom: String) -> Result<Self, NymdError> {
```
jstuczyn: Perhaps it shouldn't be called `default` anymore given that it requires some arguments for construction?

common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -88,7 +88,7 @@ impl NymdClient<SigningNymdClient> {
         vesting_contract_address: Option<AccountId>,
         erc20_bridge_contract_address: Option<AccountId>,
         signer: DirectSecp256k1HdWallet,
-        gas_price: Option<GasPrice>,
+        gas_price: GasPrice,
```
jstuczyn: So `connect_with_signer` makes `gas_price` mandatory, while in `connect_with_mnemonic` it's still optional -> any particular reason for that choice?

jstuczyn: Also with so many arguments, maybe at some point there should be a `Config` struct passed to the client? (Not unneccesarily now, but perhaps at some point) I guess it could have fields like all the contract addresses, gas price, etc.

common/client-libs/validator-client/src/nymd/wallet/mod.rs
```diff
@@ -107,6 +105,8 @@ impl DirectSecp256k1HdWallet {
         self.secret.to_string()
     }
 
+    //pub fn prefix(
```
jstuczyn: ? : )

nym-wallet/src-tauri/src/coin.rs
```diff
@@ -126,13 +110,15 @@ impl Sub for Coin {
 }
 
 impl Coin {
+  #[allow(unused)]
```
jstuczyn, line 113: If it's unused why not just remove it?

nym-wallet/src-tauri/src/coin.rs
```diff
@@ -171,34 +157,45 @@ impl Coin {
...
   pub fn denom(&self) -> Denom {
     self.denom.clone()
   }
-}
 
-impl TryFrom<Coin> for CosmWasmCoin {
-  type Error = BackendError;
+  // Helper function that returns the local denom in terms of the specified network denom.
+  fn denom_as_string(&self, network_denom: &CosmosDenom) -> Result<String, BackendError> {
```
jstuczyn, line 166: I wonder if having separate Minor/Major denoms even make sense as in cosmos there's only one name used for all coins (say `unym` and `nym` is only ever used for UI/UX purposes)

common/client-libs/validator-client/src/nymd/fee/gas_price.rs
octol: Yes, although the gas price in itself is still the default one. I'll see what better name I can come up with

common/client-libs/validator-client/src/nymd/mod.rs
octol: > So `connect_with_signer` makes `gas_price` mandatory, while in `connect_with_mnemonic` it's still optional -> any particular reason for that choice?

In `connect_with_mnemonic` we have access to `denom` so we can still use the default gas price in case it's not provided. Not so in `connect_with_signer`

common/client-libs/validator-client/src/nymd/wallet/mod.rs
octol: Let's get rid of that ;)

nym-wallet/src-tauri/src/coin.rs
octol: Yes down the line I think we should consider removing that capability, only store in Minor denom and just have the option to display in Major when needed

octol:
This was pushed to develop

PR #1133, opened 2022-02-28
Use network explorer map data with disputed areas

Baro1905:
See https://github.com/zcreativelabs/react-simple-maps/issues/251 and https://github.com/nvkelso/natural-earth-vector/issues?q=is%3Aissue+crimea+is%53Aclosed


mmsinclair:
@Baro1905 thanks for the PR 😀 

Sadly, as you may have noticed, the data from Natural Earth Vectors used by `react-simple-maps` is the problem. Sadly, the 110m disputed areas map still has Crimea's geometry in Russia. So unfortunately the `ne_110m_admin_0_countries.json` is not any better than the regular 110m map geometry 😔 

However, what I have done is to move the geometry manually inside the 110m geometry file as suggested in https://github.com/nvkelso/natural-earth-vector/issues/489#issuecomment-903066759. The result is as follows:

image

image

PR #1151, opened 2022-03-18
Feature/signature on deposit

### LINKED ISSUE #1138: Get coconut credential for deposited tokens
- [x] Write a design (https://docs.google.com/document/d/1Zrs2gju4sxAWeOOiiI4AotWmuFftc_VW6lcaVLX7414/edit)
- [x] Implement the signing part
- [x] Write tests for contract and other parts of the system

neacsu:
Implementing the plan described at https://docs.google.com/document/d/1Zrs2gju4sxAWeOOiiI4AotWmuFftc_VW6lcaVLX7414

Remaining work: 

- [x] Write tests

Closes https://github.com/nymtech/nym/issues/1138


clients/credential/src/client.rs
```diff
@@ -0,0 +1,69 @@
...
+    nymd_client: NymdClient<SigningNymdClient>,
+    denom: Denom,
+    contract_address: AccountId,
+}
+
+impl Client {
+    pub fn new() -> Self {
+        let nymd_url = Url::from_str(NYMD_URL).unwrap();
+        let mnemonic = Mnemonic::from_str(MNEMONIC).unwrap();
+        let nymd_client = NymdClient::connect_with_mnemonic(
```
jstuczyn, line 28: I guess it's not specifically for this PR, but in general looking at the constructor for `NymdClient` I think it's time to overhaul it maybe in favour of some builder.

clients/credential/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+async-trait = "0.1.52"
+bip39 = "1.0.1"
+clap = { version = "3.0.10", features = ["cargo", "derive"] }
+pickledb = "0.4.1"
```
jstuczyn, line 13: If this client is only for local tests, I guess that's fine, but `pickledb` hasn't been updated for 2 years at this point and we already have a rather unified database management in our code (via `sqlx`). Do we really want to split it again?

clients/credential/src/main.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn, line 1: What's the point of this binary? I presume it's just for local tests?

clients/native/src/commands/init.rs
```diff
@@ -107,15 +108,17 @@ async fn _prepare_temporary_credential(validators: &[Url], raw_identity: &[u8])
...
-        serial_number: params.random_scalar(),
-        binding_number: params.random_scalar(),
-        voucher_value: hash_to_scalar(BANDWIDTH_VALUE.to_be_bytes()),
-        voucher_info: hash_to_scalar(String::from("BandwidthVoucher").as_bytes()),
-    };
+    let bandwidth_credential_attributes = BandwidthVoucher::new(
+        &params,
+        &BANDWIDTH_VALUE.to_string(),
+        VOUCHER_INFO,
+        String::new(),
```
jstuczyn: What's with those empty strings everywhere? Perhaps it's a sign the constructor should get adjusted then?

common/client-libs/validator-client/src/client.rs
```diff
@@ -711,6 +711,13 @@ impl ApiClient {
         Ok(self.validator_api.blind_sign(request_body).await?)
     }
 
+    pub async fn signature(
```
jstuczyn: I think I'd call it "partial_bandwidth_credential"

common/client-libs/validator-client/src/validator_api/error.rs
```diff
@@ -7,4 +7,7 @@ pub enum ValidatorAPIError {
         #[from]
         source: reqwest::Error,
     },
+
+    #[error("Request failed with error message - {0}")]
+    GenericRequestFailure(String),
```
jstuczyn, line 12: Now, is it really a `GenericRequestFailure`? from the quick glance I had, it seems to be `reqwest`-based

common/client-libs/validator-client/src/validator_api/mod.rs
```diff
@@ -261,6 +261,18 @@ impl Client {
         .await
     }
 
+    pub async fn signature(
```
jstuczyn: As I was saying, I'd probably have renamed it

common/client-libs/validator-client/src/validator_api/routes.rs
```diff
@@ -11,6 +11,7 @@ pub const ACTIVE: &str = "active";
 pub const REWARDED: &str = "rewarded";
 
 pub const COCONUT_BLIND_SIGN: &str = "blind-sign";
+pub const COCONUT_SIGNATURE: &str = "signature";
```
jstuczyn: ibid. `partial-bandwidth-credential`, because imo `signature` is waay too generic. even `coconut-signature` would be too generic

common/coconut-interface/src/lib.rs
```diff
@@ -79,27 +79,39 @@ impl VerifyCredentialBody {
...
     total_params: u32,
 }
 
 impl BlindSignRequestBody {
     pub fn new(
         blind_sign_request: &BlindSignRequest,
+        tx_hash: String,
+        signature: String,
         public_attributes: &[Attribute],
+        public_attributes_plain: Vec<String>,
```
jstuczyn, line 107: Why are we passing *both* `public_attributes` and `public_attributes_plain`? It seems to be redundant

common/coconut-interface/src/lib.rs
```diff
@@ -79,27 +79,39 @@ impl VerifyCredentialBody {
...
     #[getset(get = "pub")]
+    public_attributes_plain: Vec<String>,
+    #[getset(get = "pub")]
     total_params: u32,
 }
 
 impl BlindSignRequestBody {
     pub fn new(
         blind_sign_request: &BlindSignRequest,
+        tx_hash: String,
```
jstuczyn, line 104: Can't we use a strong type for that? I think we got `tx::Hash` from the event, no?

common/coconut-interface/src/lib.rs
```diff
@@ -112,14 +124,34 @@ impl BlindSignRequestBody {
     }
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct BlindedSignatureResponse {
-    pub blinded_signature: BlindedSignature,
+    pub remote_key: [u8; 32],
```
jstuczyn, line 133: Why are we using an array rather than a stronger `x25519::PublicKey`? Remember, not every set of bytes is a valid elliptic curve point so we might potentially blow up in the future

common/coconut-interface/src/lib.rs
```diff
@@ -112,14 +124,34 @@ impl BlindSignRequestBody {
...
+        }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = self.remote_key.to_vec();
+        bytes.extend_from_slice(&self.encrypted_signature);
+        bytes
+    }
+
+    pub fn from_bytes(bytes: Vec<u8>) -> Self {
```
jstuczyn: Usually `from_bytes` method take byte slices, i.e. `&[u8]`. Also, you're not doing any bound checks. What will happen if you provide fewer than 32 bytes?

common/cosmwasm-smart-contracts/coconut-bandwidth-contract/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "coconut-bandwidth-contract"
```
jstuczyn: for consistency sake with other contract common crates, could you add a `-common` suffix to the name?

common/cosmwasm-smart-contracts/coconut-bandwidth-contract/src/deposit.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct DepositData {
+    deposit_info: String,
+    verification_key: String,
```
jstuczyn: I think we should keep the naming consistent since given this is the ed25519 public key (I presume), it should be called the `identity_key`

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
-    pub serial_number: PrivateAttribute,
+    serial_number: PrivateAttribute,
     // a random secret value generated by the client used to bind multiple credentials together
-    pub binding_number: PrivateAttribute,
+    binding_number: PrivateAttribute,
     // the value (e.g., bandwidth) encoded in this voucher
-    pub voucher_value: PublicAttribute,
+    voucher_value: PublicAttribute,
+    // the plain text value (e.g., bandwidth) encoded in this voucher
+    voucher_value_plain: String,
```
jstuczyn, line 33: why do we have to keep both of them? Isn't it the case that `H_{s} (plain) = attribute`?

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
-    pub voucher_info: PublicAttribute,
+    voucher_info: PublicAttribute,
+    // the plain text information
+    voucher_info_plain: String,
+    // the hash of the deposit transaction
+    tx_hash: String,
+    // base58 encoded private key ensuring the depositer requested these attributes
+    signing_key: String,
+    // base58 encoded private key ensuring only this client receives the signature share
+    encryption_key: String,
```
jstuczyn: Is there any reason for keeping those as strings? Why can't we use more strongly typed types?

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
+    pedersen_commitments_openings: Vec<Attribute>,
+    blind_sign_request: BlindSignRequest,
+    use_request: bool,
 }
 
-impl BandwidthVoucherAttributes {
+impl BandwidthVoucher {
+    pub fn new_with_blind_sign_req(
+        private_attributes: [PrivateAttribute; 2],
+        public_attributes_plain: [&str; 2],
```
jstuczyn: I think we should parameterise those with our constants, i.e. `private_attributes: [PrivateAttribute; PRIVATE_ATTRIBUTES]`

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
+
     pub fn get_public_attributes(&self) -> Vec<PublicAttribute> {
         vec![self.voucher_value, self.voucher_info]
     }
 
+    pub fn encryption_key(&self) -> String {
+        self.encryption_key.clone()
+    }
+
+    pub fn pedersen_commitments_openings(&self) -> &Vec<Attribute> {
```
jstuczyn, line 137: Is clippy not complaining about that one? It really dislikes returning `&Vec<T>` and prefers dereferenced `&[T]` insntead

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
+
+    pub fn get_public_attributes_plain(&self) -> Vec<String> {
+        vec![
+            self.voucher_value_plain.clone(),
+            self.voucher_info_plain.clone(),
+        ]
+    }
+
     pub fn get_private_attributes(&self) -> Vec<PrivateAttribute> {
         vec![self.serial_number, self.binding_number]
```
jstuczyn, line 157: I guess there's no much point in pointing it now since our entire codebase is polluted with this issue, but we should avoid cloning/copying our secret values (like private attributes, secret keys, etc) around.

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
-    params: &Parameters,
-    attributes: &BandwidthVoucherAttributes,
-    validators: &[Url],
-) -> Result<Signature, Error> {
-    let public_attributes = attributes.get_public_attributes();
-    let private_attributes = attributes.get_private_attributes();
-
-    obtain_aggregate_signature(params, &public_attributes, &private_attributes, validators).await
+    pub fn sign(&self, request: &BlindSignRequest) -> String {
+        let private_key = PrivateKey::from_base58_string(&self.signing_key).unwrap();
```
jstuczyn: Are we sure the unwrap is fine?

common/credentials/src/coconut/bandwidth.rs
```diff
@@ -7,56 +7,166 @@
...
+        let mut message = request.to_bytes();
+        message.extend_from_slice(self.tx_hash.as_bytes());
+        private_key.sign(&message).to_base58_string()
```
jstuczyn: Why not retain the stronger type? : )

common/credentials/src/coconut/utils.rs
```diff
@@ -63,45 +66,65 @@ pub async fn obtain_aggregate_verification_key(
...
+    } else {
+        client.signature(attributes.tx_hash()).await?
+    };
+    let encrypted_signature = response.encrypted_signature;
+    let remote_key = PublicKey::from_bytes(&response.remote_key)?;
+    let local_key = PrivateKey::from_base58_string(attributes.encryption_key())?;
+
+    let encryption_key =
+        recompute_shared_key::<ctr::Ctr64LE<Aes128>, blake3::Hasher>(&remote_key, &local_key);
+    let zero_iv = stream_cipher::zero_iv::<ctr::Ctr64LE<Aes128>>();
```
jstuczyn: Can you instead use the `Aes128Ctr` alias so that we would have it in a single place and it would be easier to adjust and/or fix things in the future? : )

jstuczyn: Also, I presume Ania approved the `zero_iv`? I don't see any reason why it would be the wrong choice, but I'm just making a sanity check here. Writing crypto is scary.

common/credentials/src/coconut/utils.rs
```diff
@@ -63,45 +66,65 @@ pub async fn obtain_aggregate_verification_key(
...
+        &encryption_key,
+        &zero_iv,
+        &encrypted_signature,
     );
 
-    let blinded_signature = client
-        .blind_sign(&blind_sign_request_body)
-        .await?
-        .blinded_signature;
+    let blinded_signature = BlindedSignature::from_bytes(&blinded_signature_bytes).unwrap();
```
jstuczyn: What if the `unwrap` fails?

jstuczyn: I.e. it very likely might since we just use `aes` in counter mode and not fancier variants like `gcm` so it has no checksum and thus we don't know if our decryption produced original plaintext.

common/credentials/src/coconut/utils.rs
```diff
@@ -110,42 +133,24 @@ pub async fn obtain_aggregate_signature(
...
-        &blind_sign_request,
-        &client,
-        &validator_partial_vk.key,
-    )
-    .await?;
+    let first =
+        obtain_partial_credential(params, attributes, &client, &validator_partial_vk.key).await?;
     shares.push(SignatureShare::new(first, 1));
 
     for (id, validator_url) in validators.iter().enumerate().skip(1) {
```
jstuczyn, line 145: Note: you dont have to obtain partial credentials from EVERY validator. A threshold number of them will suffice

common/credentials/src/coconut/utils.rs
```diff
@@ -110,42 +133,24 @@ pub async fn obtain_aggregate_signature(
...
     }
 
     let mut attributes = Vec::with_capacity(private_attributes.len() + public_attributes.len());
-    attributes.extend_from_slice(private_attributes);
-    attributes.extend_from_slice(public_attributes);
+    attributes.extend_from_slice(&private_attributes);
+    attributes.extend_from_slice(&public_attributes);
 
     let mut indices: Vec<u64> = Vec::with_capacity(validators_partial_vks.len());
     for i in 0..validators_partial_vks.len() {
```
jstuczyn, line 161: Spoiler, `indices.push((i + 1) as u64);` will break very soon. The signers are not guaranteed to have increasing indices from `1..n`. Just wanted to give you a heads up

contracts/coconut-bandwidth/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+mod error;
+// mod queries;
```
jstuczyn: any reason for leaving those commented out?

validator-api/src/coconut/client.rs
```diff
@@ -0,0 +1,35 @@
...
+use validator_client::nymd::{tx::Hash, NymdClient, QueryNymdClient, TxResponse};
+
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait Client {
+    async fn get_tx(&self, tx_hash: &str) -> Result<TxResponse>;
+}
+
+pub struct QueryClient {
```
jstuczyn, line 16: Why are we defining a completely separate client for this? Couldn't we just use our existing nymd client in validator API? Also are those traits really helping?

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
+use rand_07::rngs::OsRng;
 use rocket::fairing::AdHoc;
 use rocket::serde::json::Json;
-use rocket::State;
+use rocket::State as RocketState;
+use std::sync::Arc;
+use tokio::sync::{Mutex, RwLock};
+
+pub struct State {
+    client: Arc<RwLock<dyn LocalClient + Send + Sync>>,
```
jstuczyn, line 37: Why are we storing a some client by dynamic dispatch here? What's wrong with our `nymd_client::Client` that's already thread safe?

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
+    key_pair: KeyPair,
+    signed_deposits: sled::Db,
+    rng: Arc<Mutex<OsRng>>,
```
jstuczyn, line 40: Why do we have to keep the rng behind mutex here? It feels so weird. Do we have to keep track of its internal state or something?

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
 use rocket::serde::json::Json;
-use rocket::State;
+use rocket::State as RocketState;
+use std::sync::Arc;
+use tokio::sync::{Mutex, RwLock};
+
+pub struct State {
+    client: Arc<RwLock<dyn LocalClient + Send + Sync>>,
+    key_pair: KeyPair,
+    signed_deposits: sled::Db,
```
jstuczyn: I have some real mixed feelings about this one. I really prefer to keep all of our storage consistent in a single storage container, i.e. `sqlx`. It was a real mess in the past when it was all split into multiple solutions... But I might be biased. Perhaps @mmsinclair might have an opinion on this one?

validator-api/src/coconut/error.rs
```diff
@@ -0,0 +1,74 @@
...
+    DifferentPublicAttributes(String, String),
+
+    #[error("Sled error - {0}")]
+    SledError(#[from] sled::Error),
+
+    #[error("No signature found")]
+    NoSignature,
+}
+
+impl<'r, 'o: 'r> Responder<'r, 'o> for CoconutError {
```
jstuczyn, line 70: It seems like we had an identical thing defined before. Can't we just reuse it?

validator-api/src/coconut/deposit.rs
```diff
@@ -0,0 +1,430 @@
...
+        return Err(CoconutError::InconsistentPublicAttributes);
+    }
+
+    let tx_hash_str = blind_sign_request_body.tx_hash();
+    let mut message = blind_sign_request.to_bytes();
+    message.extend_from_slice(tx_hash_str.as_bytes());
+
+    let signature = Signature::from_base58_string(blind_sign_request_body.signature())?;
+
+    let attributes: &Vec<_> = tx
```
jstuczyn, line 34: I'll expand on this a bit more in the "main" comment, but wouldn't it be easier to just query for data?

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
+    ) -> Result<BlindedSignatureResponse> {
+        let (keypair, shared_key) = {
+            let mut rng = *self.rng.lock().await;
+            new_ephemeral_shared_key::<ctr::Ctr64LE<Aes128>, blake3::Hasher, _>(
+                &mut rng, remote_key,
+            )
+        };
+
+        let chunk_data = signature.to_bytes();
+        println!("Blinded signature bytes {:?}", chunk_data);
```
jstuczyn: println

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
+
+        let zero_iv = stream_cipher::zero_iv::<ctr::Ctr64LE<Aes128>>();
```
jstuczyn: the same thing for just using an alias for the `Aes128Ctr` or perhaps even do something more generic like we've done in `common/nymsphinxn/params/src/lib.rs`, i.e. 
`pub type GatewayEncryptionAlgorithm = Aes128Ctr`
`pub type PacketEncryptionAlgorithm = Aes128Ctr`

We could define something like `pub type ValidatorApiCredentialEncryptionAlgorithm = Aes128Ctr` in some crate shared between the api and clients

validator-api/src/coconut/mod.rs
```diff
@@ -1,15 +1,101 @@
...
+            stream_cipher::encrypt::<ctr::Ctr64LE<Aes128>>(&shared_key, &zero_iv, &chunk_data);
+
+        let response =
+            BlindedSignatureResponse::new(encrypted_data, keypair.public_key().to_bytes());
+
+        // Atomically insert data, only if there is no signature stored in the meantime
+        // This prevents race conditions on storing two signatures for the same deposit transaction
+        if self
+            .signed_deposits
+            .compare_and_swap(tx_hash, None as Option<&[u8]>, Some(response.to_bytes()))?
```
jstuczyn: Don't we need to store it alongside some timestamp so that we'd known when we could purge it?

jstuczyn (changes requested):
Alright. So a couple of extra general things that I haven't covered with my comments:

1. What's the purpose of the clients/credential client? It seem to have hardcoded mnemonic and addresses. It mostly looks as if  you just use it for local tests, but in that case, should it really get commited to the repo and in particular to the `clients` directory? For that reason I haven't really looked much into it, but if it's to stay, we should do something about that `pickledb`.

2. You probably meant to add the signed-deposits to `.gitignore` and not commit them : )

3. Regarding `get_tx` and getting events from the chain. Well. I had a brief and unpleasant experience with that function such that essentially ~ 5% of requests were failing to get deserialized properly. I still need to find some time and investigate it properly to file a bug report to `cosmrs`. But since we have a smart contract, why not go for a more battle-tested solution and just querying our smart contract directly rather than trying to find events by some atttributes? I think it might end up simplifying some things. I'm happy to discuss and brainstorm this point together in more detail.

4. I'm not sure what to feel about the return of `sled` to validator-api. I really prefer unified storage, but maybe in this context it makes sense? Again, not sure...

5. You seem to use `String` in a lot of places were a stronger type is available. I guess this isn't like a "YOU MUST CHANGE IT" thing, but perhaps something to keep in mind in the future.


Oh and of course feel free to ping me again any time to have a look at it again. I feel like towards the end I was just skimming at it rather than properly reviewing it so I might have missed some things.

common/client-libs/validator-client/src/validator_api/routes.rs
mmsinclair: It may also be worth adding an extra route level specifically for the bandwidth so:

```
/api/v1/coconut/bandwidth/blind-sign
/api/v1/coconut/bandwidth/partial-credential
```

Because the number of private and public parameters is pretty specific to bandwidth and isn't really useful in other cases, unless by chance, they use the same parameters and proofs. Bitter experience from the PCC.

validator-api/src/coconut/mod.rs
mmsinclair: `sqlite` and `sqlx` together would be best everywhere, another technology will be a bit confusing. Also `sqlite` is pretty easy to administrate at runtime.

clients/credential/src/client.rs
neacsu: Totally agree, it is a bit overloaded, and there's contract-specific stuff that needs to go in still.

clients/credential/Cargo.toml
neacsu: `pickledb` won't stick around, I've just used that as an easy solution for this client. The `credential` client is just a piece of binary that does what the tauri client did i.e. provide a place to gather the client-side code. The idea is to then move code around to where it should reside, but that is not yet decided. When we get a decision on what the buying coconut interface will look like, we can start moving the code and remove this binary. The storage solution will then depend on that, but it won't be `pickledb` for sure :D

clients/credential/src/main.rs
neacsu: Answered above. Testing + accumulating code until it's ready for move.

clients/native/src/commands/init.rs
neacsu: Instead of those default values, I think I'll put some actual generated keys in there, even though those clients are not yet integrated with the code in the placeholder credential client.

common/client-libs/validator-client/src/client.rs
neacsu: I'll rename it (and the others connected to it)

common/client-libs/validator-client/src/validator_api/error.rs
neacsu: It's not `request`-based, it's the solution I found for deserialising the error from the server to the client. Instead of the server sending an http error, that doesn't have the information we're looking for, it sends a valid response, with status `BadRequest` and the server-specific error message as text. This `GenericRequestFailure` type can at least parse that into a readable output to the clients.

common/client-libs/validator-client/src/validator_api/mod.rs
neacsu: 👍

common/client-libs/validator-client/src/validator_api/routes.rs
neacsu: Extra route levels sounds good to me. I presume the `/api/` part is in reference to what we use on our own validator-api machine, as there is no reference to it in the code afaik. It's just our way of redirecting to local 8080 port.

common/coconut-interface/src/lib.rs
neacsu: Because we need to check the public attributes value against what's being recorded on the chain, especially the deposit value. And unless I'm missing something, there's no way to go back from `Attribute` to plain value, as it's basically a hash.

common/coconut-interface/src/lib.rs
neacsu: This is the serialised structure, and `tx::Hash` doesn't implement `Serialize, Deserialize`. The current approach is to just use strings (see the existing `signature` and `public_attributes` fields)

common/coconut-interface/src/lib.rs
neacsu: The same reason as for the `BlindSignRequestBody`, `x25519::PublicKey` doesn't implement `Serialize, Deserialize`. There are checks on the other end though, and tests should cover the blow up part.

common/coconut-interface/src/lib.rs
neacsu: I'll add some error handling in here

common/cosmwasm-smart-contracts/coconut-bandwidth-contract/Cargo.toml
neacsu: Didn't see that convention, I'll rename it.

common/cosmwasm-smart-contracts/coconut-bandwidth-contract/src/deposit.rs
neacsu: 👍

common/credentials/src/coconut/bandwidth.rs
neacsu: The same reason as above i.e. to check what's encoded in the request.

common/credentials/src/coconut/bandwidth.rs
neacsu: I think this would be a good place to use strong types, yes. There would have to be some hard-coded values in the native and socks5 clients I think, but that will change when the credential client is integrated into them.

common/credentials/src/coconut/bandwidth.rs
neacsu: Will do

common/credentials/src/coconut/bandwidth.rs
neacsu: It doesn't seem to complain, no

common/credentials/src/coconut/bandwidth.rs
neacsu: At least those are pretty short-lived.

common/credentials/src/coconut/bandwidth.rs
neacsu: Switching to the strong type will remove the `unwrap` altogether.

common/credentials/src/coconut/bandwidth.rs
neacsu: Yep, will do that, and do the conversion when building the `BlindSignRequestBody`

common/credentials/src/coconut/utils.rs
neacsu: Yes, I'll put up a type alias.

neacsu: I'll double check with her on this.

common/credentials/src/coconut/utils.rs
neacsu: I think I've missed converting this one to error handling. I'll change it.

common/credentials/src/coconut/utils.rs
neacsu: Correct, but I don't think there is a threshold value anywhere right now. Once DGK is done, I think we can focus on that more.

common/credentials/src/coconut/utils.rs
neacsu: Can't wait for it then :D

contracts/coconut-bandwidth/src/lib.rs
neacsu: Yes, I forgot to remove it :D

validator-api/src/coconut/client.rs
neacsu: The idea is that I wrap the client in this so I can implement the trait specific to what I'm using it for. And the trait is useful so that I can create a mock client in unit tests and not depend on the nymd chain running. I really think we should expand this approach to the entire nymd client and enable more stuff to be tested in the codebase, but maybe we can have a talk about this with the others.

validator-api/src/coconut/mod.rs
neacsu: I think I've answered this in the comment above i.e. to enable more testing

validator-api/src/coconut/mod.rs
neacsu: Well, we had this talk a while ago about not re-initialising the rng every time it is used, as a general rule which might not necessarily apply to `OsRng`. As it is being used in `mut` calls, the `Mutex` does the job.

validator-api/src/coconut/mod.rs
neacsu: I'll look into using `sqlx` for this.

validator-api/src/coconut/error.rs
neacsu: If you mean the one used in `node_status_api/models.rs`, that had the issue of not passing the error to the clients when I tried it here. Not sure if it works better for their specific endpoints (I suspect not).

validator-api/src/coconut/deposit.rs
neacsu: I'll respond to this and the "main" comment here, as it's better threaded. The reasons for using an event approach instead of a storage approach are: at some point, there was an idea of watching the chain for events (which we've abandoned because of the websocket client complexity), and because it _should_ be more lightweight to keep those things forever in a low-level event, then in the more flexible, high-level cosmwasm storage. The query for data thing will be part of the double-spending storage anyway, but that at least needs a faster, overall view of the state at a certain point, whereas here, pin-pointing the query to a block is more then enough for our needs.

validator-api/src/coconut/mod.rs
neacsu: Sorry about that, I was debugging something, but removed it in a later commit.

validator-api/src/coconut/mod.rs
neacsu: Yes, a new crate in `common` sounds good.

validator-api/src/coconut/mod.rs
neacsu: I guess the plan is, if we decide later to actually introduce a non-infinite TTL, those values would get the timestamp of the time when we introduce the change. Until then, there is no need for extra complexity imo.

common/credentials/src/coconut/utils.rs
neacsu: I've double checked on this with Ania. Since this key is only used for encrypting the payload once, the IV can safely be zero.

PR #1158, opened 2022-03-23
Feature/vesting full

fmtabbara:


mmsinclair (approved):

mmsinclair:
@octol @fmtabbara looks like it is only clippy that is blocking the build, so worth getting that working and then we can merge this in

mmsinclair (changes requested):
Also, looks like there are some errors in the Rust tests that need fixing

tommyv1987:
@fmtabbara - looks good mate! let's get this in.

PR #1161, opened 2022-03-24
wallet: set nymd_url and api_url independently

octol:
Currently the connection test is done by walking through the list of available validators, and for each entry in serial
1) Create a client 
2) try a call to `get_mixnet_contract_version()`
3) Use the first one that succeeds

This implies that we pick the `nymd_url` and `api_url` from the same validator entry.

This PR propses to do two things:

1) Switch to the simpler connection health check that was implemented earlier for `Config` in the wallet. This is based on a simple query of the `nymd_url` appended with `/health`. A validator, that contains both a `nymd_url` and optionally a `api_url`, is marked as connectable if the call to `nymd_url/health` is successful. Unlike the previously used check that relied on creating a client and trying a nymd call, this one is done for all validators in the list.

2) From the list of validators that passed the health check, choose the first validator for the `api_url`, and pick one at random for `nymd_url`.


octol:
This draft is superseded by https://github.com/nymtech/nym/pull/1170

PR #1181, opened 2022-03-30
Fix cache bootstrap

durch:


durch:
This is no longer needed as the cache is no longer triggering rewarded set updates

PR #1182, opened 2022-03-31
Crypto part of the Groth's NIDKG

jstuczyn:
Alright, this [draft] pull request includes the first batch of work required for distributed key generation so that we could have our coconut credentials. This part focuses on the crypto-part of the problem. The implementation is based on [Non-interactive distributed key generation and key resharing by Jens Groth](https://eprint.iacr.org/2021/339.pdf) paper. It is also further supported by the extremely helpful description provided by Ania and Alfredo available on our overleaf: https://www.overleaf.com/project/61fc031513021c2738dee13c. 

Unfortunately in order to create this implementation, the [bls12-381 curve library](https://github.com/zkcrypto/bls12_381) had to be forked and we will need to operate on the fork until https://github.com/zkcrypto/bls12_381/issues/10 is resolved as we require being able to serialize curve elements inside `Gt` so that we could create the lookup table required for the baby-step giant-step algorithm used during chunk decryption. This also means that once DKG is incorporated into the coconut, the underlying curve library used there will need to be switched to the same fork. 

The PR also includes couple of benchmarks and those regarding dealing verification and share recovery should be of interest. I've run them on my local machine with ryzen 5900X and the relevant results are as follows:
- it takes 3.43s to verify a single dealing (and recover associated share) that has been created for 20 party setup. This means that every single party in the system will need to dedicate ~4.35 minutes of compute time for key creation. (each party has to recover shares from 19 other parties and it has to be performed 4 times since we intend to have 4 attributes in each credential and  `3.43s * 19 * 4 = 260.7s ~ 4.35 minutes`)
- it takes 16.83s to verify a single dealing (and recover associated share) that has been created for 100 party setup. This implies that every single party in the system will need to dedicate ~2 **hours** of compute time for key creation (`16.83 * 99 * 4 = 111 minutes`)

The remaining work before this PR can be moved out of the draft includes:
~~- serialization implementation of the remaining data structures that are required for end-to-end operation (such as `DecryptionKey` or `Dealing`),~~
~~- implementation of `VKCombine` and `VKVfy` algorithms~~

~~However, in its current state, I think all the existing crypto primitives are ready to get reviewed.~~


common/crypto/dkg/src/dealing.rs
```diff
@@ -0,0 +1,485 @@
...
+    pub public_coefficients: PublicCoefficients,
+    pub ciphertexts: Ciphertexts,
+    pub proof_of_chunking: ProofOfChunking,
+    pub proof_of_sharing: ProofOfSecretSharing,
+}
+
+impl Dealing {
+    // I'm not a big fan of this function signature, but I'm not clear on how to improve it while
+    // allowing the dealer to skip decryption of its own share if it was also one of the receivers
+    pub fn create(
```
aniampio, line 32: In the `create` function we should also add as an optional input argument the secret value (i.e., the constant value of the polynomial) known by the validator who is calling this function. This secret value will be useful if we want to do a re-sharing of the same secret and verification key.

common/crypto/dkg/src/dealing.rs
```diff
@@ -0,0 +1,485 @@
...
+        } else {
+            (dealing, None)
+        }
+    }
+}
+
+impl Dealing {
+    // rather than returning a bool for whether the dealing is valid or not, a Result is returned
+    // instead so that we would have more information regarding a possible failure cause
+    pub fn verify(
```
aniampio, line 108: Here we also add as an optional input argument the secret value known by the initiating validator (so we can verify a resharing if it happened)

common/crypto/dkg/src/bte/proof_chunking.rs
```diff
@@ -0,0 +1,1084 @@
...
+type FirstChallenge = Vec<Vec<Vec<u64>>>;
+
+// TODO: perhaps break it down into separate arguments after all
+#[cfg_attr(test, derive(Clone))]
+pub struct Instance<'a> {
+    /// y_1, ..., y_n
+    public_keys: &'a [PublicKey],
+
+    /// R_1, ..., R_m
+    randomizers_r: &'a [G1Projective; NUM_CHUNKS],
```
aniampio: This naming of the value is a bit misleading: the actual randomizer value is r, while R is the public value associated with it (you can think about it as a commitment to r or a public value associated with r). 
Note: In general, I would suggest we try in our whole codebase (for when we implement cryptographic primitives and schemes) to have a consistent notation for capital letters and lower-case letters. Sadly, somehow Rust doesn't seem to be happy about using capital letters

common/crypto/dkg/src/bte/proof_sharing.rs
```diff
@@ -0,0 +1,617 @@
...
+
+        true
+    }
+}
+
+#[derive(Debug)]
+#[cfg_attr(test, derive(Clone, PartialEq))]
+pub struct ProofOfSecretSharing {
+    // TODO: ask @AP for better names for those
+    f: G1Projective,
```
aniampio: Comment regarding notation: let's try to use a consistent notation for values that in the description of a cryptographic protocol are denoted with lower case letters vs upper case letters.

common/crypto/dkg/src/bte/proof_chunking.rs
```diff
@@ -0,0 +1,1084 @@
...
+            yy += instance.public_keys[i - 1].0 * delta_i;
+        }
+
+        let second_challenge =
+            Self::compute_second_challenge(&first_challenge, &responses_chunks, &ds, &yy);
+
+        // compute responses
+
+        // TODO: is this correct capacity?
+        let mut responses_r = Vec::with_capacity(PARALLEL_RUNS);
```
aniampio: Shouldn't there be n of those responses?

aniampio (commented):
The crypto looks good! Left a few minor comments.

common/crypto/dkg/src/bte/proof_chunking.rs
jstuczyn: Yup, you're totally right -> adjusted.

common/crypto/dkg/src/dealing.rs
aniampio: Correction: "optional input argument the secret value" -> "optional input argument the partial verification key"

PR #1208, opened 2022-04-13
Try to export typescript types for RewardEstimationResponse again

jstuczyn:


jstuczyn:
Turns out this is way trickier than I initially anticipated and thus I'm parking this since it's no longer a "5min fix"

PR #1219, opened 2022-04-22
Bugfix/delegation reconcile

jstuczyn:


jstuczyn:
Note: remaining work: add migration that clears all delegations and events with 0 value

PR #1223, opened 2022-04-25
Network Explorer: add economic dynamics

gala1234:
In the file `explorer/src/components/Socials.tsx` the discord icon is set to non be displayed.
3 weeks ago, was implemented to display this icon. But it couldn't be deployed to `prod`, as the server wasn't ready. And still is not ready today.
This is why in `develop` the discord icon is set to be displayed, but we don't want it yet.


.github/workflows/network-explorer.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - uses: rlespinasse/github-slug-action@v3.x
     - uses: actions/setup-node@v2
       with:
-        node-version: '16'
+        node-version: '16.13.2'
```
mmsinclair: it would be best if it runs without the specific version of node and just the latest LTS node 16 - I know Fouad change this to make things work

could you try it with just:
```suggestion
        node-version: '16'
```

explorer/src/components/MixNodes/Economics/Rows.ts
```diff
@@ -0,0 +1,53 @@
...
+      case 'Low':
+        return inclusionProbability;
+      case 'VeryHigh':
+        return 'Very High';
+      case 'VeryLow':
+        return 'Very Low';
+      default:
+        return '-';
+    }
+  };
```
mmsinclair: It is normally better to move a method like this out of the component and pass `inclusionProbability` as an argument, for two reasons:

1. you can unit test it
2. it isn't redeclared and instantiated every time the component renders (so better rendering performance - not that is matters in this case)

explorer/src/icons/InfoSVG.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { useTheme } from '@mui/material/styles';
+
+export const InfoSVG: React.FC = () => {
```
mmsinclair: Is this a special SVG? Normally Marcin and Jess use icons from here: https://mui.com/material-ui/material-icons/

image

Could you replace it with `import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';`?

ts-packages/mui-theme/src/theme/common.ts
```diff
@@ -46,6 +46,11 @@ export interface NymPaletteVariant {
       standby: string;
     };
   };
+  selectionChance: {
```
mmsinclair, line 49: Nice 😁

mmsinclair (changes requested):
It is looking good @gala1234, nicely done 😁 

Only a few minor things to change and then good to merge.

mmsinclair (approved):

PR #1230, opened 2022-04-28
Fix checkpointing

durch:


common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
```diff
@@ -247,24 +247,24 @@ impl DelegatorRewardParams {
     }
 }
 
-#[derive(Debug, Clone, JsonSchema, PartialEq, Serialize, Deserialize, Copy)]
+#[derive(Debug, Clone, JsonSchema, PartialEq, Serialize, Deserialize)]
 pub struct StoredNodeRewardResult {
     reward: Uint128,
-    lambda: Uint128,
-    sigma: Uint128,
+    lambda: String,
```
jstuczyn: just leaving a comment here to remember to [presumably] revert it back to Uint128 before merging

durch: No, no, leave it as string, Uint128 will truncate the float

durch: Never mind, I see what you did there, will test it out in QA

contracts/mixnet/src/constants.rs
```diff
@@ -1,9 +1,8 @@
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-// approximately 1 week (assuming 5s per block)
-// i.e. approximately quarter of the interval (there are 3600 * 60 * 7 = 604800 seconds in a week, i.e. ~604800 / 5 = 120960 blocks)
-pub const MINIMUM_BLOCK_AGE_FOR_REWARDING: u64 = 120960;
+// approximately 1 epoch (assuming 5s per block)
+pub const MINIMUM_BLOCK_AGE_FOR_REWARDING: u64 = 720;
```
jstuczyn, line 5: another note to remember to revert that before merging : )

durch: I don't think we need to revert this, if they get in the active set they should be rewarded

PR #1236, opened 2022-05-02
Mitigate PM change attack

durch:
Allow at most one PM update in a 30 day interval

### DIFF
--- a/contracts/mixnet/src/mixnodes/transactions.rs
+++ b/contracts/mixnet/src/mixnodes/transactions.rs
@@ -941,12 +941,14 @@ pub mod tests {
         )
         .is_ok());
 
+        env.block.time = env.block.time.plus_seconds(MIN_PM_UPDATE_INTERVAL - 1);
+
         // fails if too soon after bonding
         assert!(
             try_update_mixnode_config(deps.as_mut(), env.clone(), info_alice.clone(), 20).is_err()
         );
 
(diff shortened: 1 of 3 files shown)


contracts/mixnet/src/mixnodes/transactions.rs
```diff
@@ -882,4 +906,49 @@ pub mod tests {
...
+            sig1
+        )
+        .is_ok());
+
+        // fails if too soon after bonding
+        assert!(
+            try_update_mixnode_config(deps.as_mut(), env.clone(), info_alice.clone(), 20).is_err()
+        );
+
+        env.block.time = env.block.time.plus_seconds(MIN_PM_UPDATE_INTERVAL + 1);
```
neacsu: Maybe use something closer to this limit on the error path as well

contracts/mixnet/src/mixnodes/transactions.rs
```diff
@@ -152,6 +154,8 @@ fn _try_add_mixnode(
         storage::TOTAL_DELEGATION.save(deps.storage, identity, &Uint128::zero())?;
     }
 
+    storage::LAST_PM_UPDATE_TIME.save(deps.storage, identity, &env.block.time.seconds())?;
```
jstuczyn, line 157: We should probably also save into this map when bonding mixnode, so somebody wouldn't be able to bond with low profit margin and then, before 30 days are over update it to a way higher one.

PR #1238, opened 2022-05-03
Endpoints for average mixnode uptime

octol:


octol:
These changes will be merged in with https://github.com/nymtech/nym/pull/1223 instead

PR #1271, opened 2022-05-16
Optimize fetch and loop

durch:
# Description

Optimizes `calculate_delegator_reward` and `calculate_operator_reward` to fetch less and loop faster.

# Checklist:

### DIFF
--- a/contracts/mixnet/src/rewards/transactions.rs
+++ b/contracts/mixnet/src/rewards/transactions.rs
@@ -115,9 +115,8 @@ pub fn calculate_operator_reward(
     let accumulated_rewards = mixnodes()
         .changelog()
         .prefix(bond.identity())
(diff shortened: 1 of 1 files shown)


jstuczyn (commented):
There is only one question I have regarding changes in both functions. Since now we're calculating rewards from `last_claimed_height`, should the `fold` still be starting from `Uint128::zero()`, or should it rather use the value from `last_claimed_height`?

durch:
> There is only one question I have regarding changes in both functions. Since now we're calculating rewards from `last_claimed_height`, should the `fold` still be starting from `Uint128::zero()`, or should it rather use the value from `last_claimed_height`?

Rewards from the `last_claimed_height` have been sent to them already, so we'd be sending them multiples of already claimed rewards in this case every time they claim again. The intent is to only calculate unclaimed rewards.

PR #1275, opened 2022-05-17
Replace checked with saturating sub

durch:
# Description

Fixes underflow when there are more tokens delegated/pledge then there are tokens vesting

# Checklist:


jstuczyn (approved):
lgtm. just add the entry to changelog and let's merge it

PR #1284, opened 2022-05-24
Fix uptime integer division, change lambda and sigma from ticked to regular

### LINKED ISSUE #1286: Fix integer division problem for performance and operator cost parts of the rewards calculation

durch:
# Description

Fixes: [1286](https://github.com/nymtech/nym/issues/1286)

+ Adds `estimated_node_profit` and `estimated_operator_cost` to `reward-estimate` response.
+ Fixes` estimated_delegator_reward` to be calculated from `node_profit`
+ Switch to `lambda` and `sigma` for operator and delegator reward calculation, instead of `lambda'` and `sigma'`

# Checklist:

### DIFF
--- a/common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
+++ b/common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
@@ -318,6 +318,14 @@ impl NodeRewardResult {
     }
 }
 
+pub struct RewardEstimate {
+    pub total_node_reward: u64,
+    pub operator_reward: u64,
+    pub delegators_reward: u64,
+    pub node_profit: u64,
+    pub operator_cost: u64,
+}
+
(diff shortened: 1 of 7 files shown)


common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
```diff
@@ -427,20 +427,31 @@ impl MixNodeBond {
     pub fn estimate_reward(
         &self,
         params: &RewardParams,
-    ) -> Result<(u64, u64, u64), MixnetContractError> {
+    ) -> Result<(u64, u64, u64, u64, u64), MixnetContractError> {
```
jstuczyn: I think we reached a point where those should probably get extracted to some separate struct, returning a tuple with 5 values seem to be rather confusing : )

PR #1300, opened 2022-06-01
explorer: adding stake saturation o node list

gala1234:
# Description

Closes: #XXXX

# Checklist:


explorer/src/pages/Mixnodes/index.tsx
```diff
@@ -248,6 +248,27 @@ export const PageMixnodes: React.FC = () => {
         </MuiLink>
       ),
     },
+    {
+      field: 'stake_saturation',
+      headerName: 'Stake Saturation',
```
doums, line 253: I don't know if an internationalization system exists in the explorer (like `react-intl` or `i18next` etc). If yes, maybe this text should be replaced by the corresponding translation.

gala1234: we don't have it yet, all text is in EN and hardcoded. Maybe, if Strapi works well, we can use it on Nym wallet, explorer...

tommyv1987 (approved):
LGTM

PR #1302, opened 2022-06-01
Claim and compound reward wallet endpoints

durch:
# Description

In addition to the endpoints this introduces the `execute` procedural macro which generates `execute_*` and `simulate_*` `nymd` client methods. The bloat that we have there has been bugging me for some time, and the simulations only add to it, especially in the wallet. 

I've applied to macro to the claim and compound reward methods only, so that we can test how it works and decide if we're happy with it.

https://github.com/nymtech/nym/issues/2028

# Checklist:


nym-wallet/src-tauri/src/operations/mixnet/rewards.rs
```diff
@@ -0,0 +1,48 @@
...
+    Ok(())
+}
+
+#[tauri::command]
+pub async fn claim_delegator_reward(
+    mix_identity: IdentityKey,
+    state: tauri::State<'_, Arc<RwLock<State>>>,
+) -> Result<(), BackendError> {
+    nymd_client!(state)
+        .execute_claim_delegator_reward(mix_identity, None)
```
jstuczyn: As I was saying on keybase, we really don't want to be ignoring the `fee` argument here and passing a `None` all the time, since then we are not going to be able to use the simulated gas fees (keybase://chat/nymtech#dev-core/14687)

common/execute/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    };
+
+    let ts = proc_macro::TokenStream::from(simulate_block);
+    simulate_f.block = Box::new(parse_macro_input!(ts as Block));
+
+    let execute_block = quote! {
+        {
+            let (req, fee) = self.#name(#(#execute_args),*);
+            let fee = Fee::Auto(Some(self.simulated_gas_multiplier));
```
jstuczyn: If available we want to use the provided fee : )

durch: Thats a good catch, will fix that now

jstuczyn (approved):

PR #1306, opened 2022-06-06
Set probability to 1 if less nodes then the limit

### LINKED ISSUE #2448: NE - Correct how “Active Set Probability" is shown on the node page
This requires a back-end change to show the correct probability. But basically if we have less than 720 nodes (or whatever the max reward set is) in the network the probability for all the nodes should be 100%. However at the moment the set number is not being taken into account when calculating this probability.

durch:
# Description

Set probability of nodes to be included in the rewarded and active sets to 1. if there are less node then the rewarded_set limit.

Closes: [#226](https://github.com/nymtech/nym/issues/2448)

### DIFF
--- a/validator-api/src/node_status_api/routes.rs
+++ b/validator-api/src/node_status_api/routes.rs
@@ -211,6 +211,14 @@ pub(crate) async fn get_mixnode_inclusion_probability(
     identity: String,
 ) -> Json<Option<InclusionProbabilityResponse>> {
     let mixnodes = cache.mixnodes().await;
+    let rewarding_params = cache.epoch_reward_params().await.into_inner();
+
+    if mixnodes.len() <= rewarding_params.rewarded_set_size() as usize {
+        return Json(Some(InclusionProbabilityResponse {
+            in_active: 1.0.into(),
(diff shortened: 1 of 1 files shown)


validator-api/src/node_status_api/routes.rs
```diff
@@ -211,14 +211,21 @@ pub(crate) async fn get_mixnode_inclusion_probability(
     identity: String,
 ) -> Json<Option<InclusionProbabilityResponse>> {
     let mixnodes = cache.mixnodes().await;
+    let rewarding_params = cache.epoch_reward_params().await.into_inner();
+
+    if mixnodes.len() <= rewarding_params.rewarded_set_size() as usize {
+        return Json(Some(InclusionProbabilityResponse {
+            in_active: 1.0.into(),
```
jstuczyn: I'd probably do 2 separate if checks here for each of the sets, because it's possible we'll have 100% to be in rewarded set, but not in the active set. For example, assume the following:
- we have 200 nodes
- rewarded set size is set to 300
- active set size is set to 100
- we have 100% to be in rewarded set, but the same is not true for the active set

PR #1308, opened 2022-06-06
Feature/report node hardware info

### LINKED ISSUE #2030: Add mixnode hardware self-reporting to HTTP API
In order for us to speed the mixnet up in future, it would help us to know what kind of hardware people are typically running mixnodes on, and what hardware crypto helpers are available on their systems. 

This issue should add basic hardware reporting to the mixnodes, with hardware info available on the node at http://localhost:8000/hardware

futurechimp:
# Description

Closes: nymtech/nym#2030

Basic hardware reporting for mixnodes showing up in the HTTP API. 

# Checklist:

### DIFF
--- a/mixnode/src/node/http/hardware.rs
+++ b/mixnode/src/node/http/hardware.rs
@@ -1,3 +1,4 @@
+use cupid::ExtendedTopologyEnumeration;
 use rocket::serde::{json::Json, Serialize};
 use sysinfo::{System, SystemExt};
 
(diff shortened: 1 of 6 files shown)


mixnode/src/node/http/hardware.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// Sysinfo gives back basic stuff like number of CPU cores and available memory. If available, this includes the hardware encryption
+/// extensions report
+fn hardware_from_sysinfo(crypto_hardware: Option<CryptoHardware>) -> Option<Hardware> {
+    if System::IS_SUPPORTED {
+        let mut system = System::new_all();
+        let total_memory = system.free_memory();
+        system.refresh_all();
+        let ram = format!("{}KB", total_memory);
```
neacsu: The call is for `free_memory()` here though. We either change the call or the log message. Personally, I think the total memory is more relevant here (and also less revealing of what's running on the mixnode at a certain moment)

PR #1324, opened 2022-06-08
Introduce staking supply, contract and wallet

### LINKED ISSUE #2023: Make staking supply configurable.
As not all coins are currently distributed, we have a problem in the payment calculations: they assume a staking supply of 750 million tokens when it is in fact significantly lower. 

Let's make this a contract variable so that we can set it to a value which makes more sense, and then update as coins are released. 

This involves:

- [x] the contract
- [x] the admin screens

durch:
# Description

Introduces staking supply into wallet calculations

Closes: https://github.com/nymtech/nym/issues/2023

# Checklist:

### DIFF
--- a/common/cosmwasm-smart-contracts/mixnet-contract/src/reward_params.rs
+++ b/common/cosmwasm-smart-contracts/mixnet-contract/src/reward_params.rs
@@ -78,9 +78,9 @@ impl NodeEpochRewards {
     ) -> Result<Uint128, MixnetContractError> {
         // change all values into their fixed representations
         let delegation_amount = U128::from_num(delegation_amount.u128());
-        let circulating_supply = U128::from_num(epoch_reward_params.circulating_supply());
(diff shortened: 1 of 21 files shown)


common/cosmwasm-smart-contracts/mixnet-contract/src/reward_params.rs
```diff
@@ -105,7 +105,7 @@ pub struct EpochRewardParams {
     epoch_reward_pool: Uint128,
     rewarded_set_size: Uint128,
     active_set_size: Uint128,
-    circulating_supply: Uint128,
+    staking_supply: Uint128,
```
jstuczyn, line 109: You want to put some `serde(alias)` action here so that you would be able to deserialize old data from the `circulating_supply` field

durch: wow, thats a great idea!

PR #1329, opened 2022-06-08
Fix validator docker build on other platforms

elliot-u410:
# Description

The `Dockerfile` for validator failed to build when run on arm macOS hosts, even when targeting `amd64`. This is because the `setup.sh` script was not accounting for different naming of `libwasmvm` on various architectures and using `ldd` which routinely segfaults under emulation. The solution is simply to use `find` instead of `ldd` and allow for different naming.

This change also makes the build-arg variables required instead of silently building without them set.

### DIFF
--- a/docker/validator/setup.sh
+++ b/docker/validator/setup.sh
@@ -1,8 +1,11 @@
 #!/bin/sh
 
+set -ue
+
 git clone https://github.com/CosmWasm/wasmd.git
 cd wasmd
 git checkout "${WASMD_VERSION}"
+WASMD_COMMIT_HASH=$(git rev-parse HEAD)
 mkdir build
 go build \
     -o build/nymd -mod=readonly -tags "netgo,ledger" \
--- a/docker/validator/setup.sh
+++ b/docker/validator/setup.sh
@@ -14,5 +14,4 @@ go build \
     -X github.com/CosmWasm/wasmd/app.Bech32Prefix=${BECH32_PREFIX} \
     -X 'github.com/cosmos/cosmos-sdk/version.BuildTags=netgo,ledger'" \
     -trimpath ./cmd/wasmd
-WASMVM_SO=$(ldd build/nymd | grep libwasmvm.so | awk '{ print $3 }')
-cp "${WASMVM_SO}" build/
+find .. -type f -name 'libwasm*.so' -exec cp {} build \;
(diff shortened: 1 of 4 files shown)


docker/validator/setup.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/sh
 
+set -ue
```
elliot-u410, line 3: This change also makes the build-arg variables required instead of silently building without them set.

docker/validator/setup.sh
```diff
@@ -1,8 +1,11 @@
...
+
 git clone https://github.com/CosmWasm/wasmd.git
 cd wasmd
 git checkout "${WASMD_VERSION}"
+WASMD_COMMIT_HASH=$(git rev-parse HEAD)
```
neacsu, line 8: You're right, this one wasn't being declared. Could you instead declare this in `docker/validator/Dockerfile`, similarly to how we do it for `WASMD_VERSION`? i.e.:
```
ARG WASMD_COMMIT_HASH
```
That way, we keep those values in the `docker-compose.yml` file

neacsu (changes requested):
Nice catch on this one. We can enable validator build in docker on M1s now by the looks of it. I think this is worth a change in the CHANGELOG as well, under the `Fixed` section.

Also, I think in `docker/validator/Dockerfile` we should do
```
-FROM golang:buster as go_builder
+FROM amd64/golang:buster as go_builder
```
and
```
-FROM ubuntu:20.04
+FROM amd64/ubuntu:20.04
```

That way, this will only run on x86 and on M1s under emulation. Unfortunately native builds for M1 won't be possible until there's arm support for wasmvm (https://github.com/CosmWasm/wasmvm/issues/53)

Thank you for contributing to this project.

docker/validator/setup.sh
elliot-u410: We could but then it'd be possible to select a commit hash that doesn't correspond to the version. Is that really a desirable feature? If someone wants to build a specific commit right now, they can specify `WASMD_VERSION` as a commit hash instead and it will build that commit.

Since tags have a 1:1 correspondence to commit hashes, it seems like you'd want to take one or the other, but not both.

neacsu: Right. Then we could remove `WASMD_COMMIT_HASH` from docker-compose.yml altogether and only get the hash with `git` in the script, so there's no ambiguity there.

elliot-u410:
@neacsu I tested without setting the platform to amd64 and I'm able to see the two validators working together on my M1. Perhaps my test is insufficient but it seems to be working AFAICT.

Also:
  * I removed the unused `WASMD_COMMIT_HASH` env var.
  * I updated WASMD_VERSION to 0.27 because it fails to build on 0.26 on my M1.
  * I updated the changelog.

neacsu (approved):

PR #1331, opened 2022-06-09
Add locked token pledge limit

### LINKED ISSUE #2024: Cap staking in unvested accounts to 100K.

durch:
# Description

Closes: https://github.com/nymtech/nym/issues/2024

# Checklist:

### DIFF
--- a/common/client-libs/validator-client/src/nymd/mod.rs
+++ b/common/client-libs/validator-client/src/nymd/mod.rs
@@ -29,6 +29,7 @@ use network_defaults::DEFAULT_NETWORK;
 use serde::Serialize;
 use std::convert::TryInto;
 use vesting_contract_common::ExecuteMsg as VestingExecuteMsg;
+use vesting_contract_common::QueryMsg as VestingQueryMsg;
 
 pub use crate::nymd::cosmwasm_client::client::CosmWasmClient;
 pub use crate::nymd::cosmwasm_client::signing_client::SigningCosmWasmClient;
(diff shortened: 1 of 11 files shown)


common/client-libs/validator-client/src/nymd/mod.rs
```diff
@@ -939,6 +950,18 @@ impl<C> NymdClient<C> {
         )
     }
 
+    #[execute("vesting")]
+    fn _vesting_update_locked_pledge_cap(
```
jstuczyn, line 997: Do we even need to have the ability to dynamically update it? From what I understood we set it once and leave it like that

common/cosmwasm-smart-contracts/vesting-contract/src/messages.rs
```diff
@@ -113,6 +113,9 @@ pub enum ExecuteMsg {
     UpdateStakingAddress {
         to_address: Option<String>,
     },
+    UpdateLockedPledgeCap {
+        amount: Uint128,
```
jstuczyn, line 117: This is more of a general thought rather than comment to this specific line of code, but recently I've been thinking about how we're using `Uint128`/`u128` to represent coin values in various places. I think in general we should be using the `Coin` type directly instead wherever possible as to be explicit about the denomination and not get accidentally screwed if somebody decides to do something weird with coins of different denom of if we accidentally messed up a migration or something.

common/client-libs/validator-client/src/nymd/mod.rs
durch: I bet we’ll need to update it at some point :)

common/cosmwasm-smart-contracts/vesting-contract/src/messages.rs
durch: I agree, it’s just that denominations are a pain, if we had(maybe have) good constructors for Coin I’d be using them more.

jstuczyn: I'd say `pub fn new(amount: u128, denom: impl Into<String>)` is pretty good : )

common/client-libs/validator-client/src/nymd/mod.rs
jstuczyn: I've got a feeling you might be right : )

PR #1332, opened 2022-06-09
[wallet] style changes to match new design

doums:
# [wallet] style changes to match new design

This PR introduces new background colors to match the new UI presents in figma.

https://www.figma.com/file/KiYKZnWeefSALAe0j2dAPI/Nym-Wallet


nym-wallet/src/theme/theme.tsx
```diff
@@ -43,12 +43,16 @@ const darkMode: NymPaletteVariant = {
   topNav: {
     background: '#111826',
   },
+  nav: {
+    background: '#FFFFFF',
+    hover: '#',
```
mmsinclair: This is a strange value to set - does `hover` maybe need to be made optional?

mmsinclair (changes requested):

nym-wallet/src/theme/theme.tsx
doums: At this moment `hover` is not used at all, should be safe to remove it.

PR #1342, opened 2022-06-10
Bugfix: Prevent multiple send requests

### LINKED ISSUE #1340: [Issue] send button not protected from multiple clicks
**Describe the issue**
The send button of the wallet is not protected of multiple clicks. If the send button is clicked twice, it creates a condition where the wallet tries to send 2 transactions but with the same nonce. Luckily the second transaction fails, but the message received is showing first the failure of the second transaction, then when the first succeed, the screen is updated with the success message of the first transaction.

**Expected behaviour**
The send button should not trigger the wallet to try to send n transactions if pushed n times.

**Steps to Reproduce**
Steps to reproduce the behaviour:
- open wallet
- send nyms
- click frantically on the send button

**Screenshots**

Screenshot from 2022-06-10 10-18-05

 - Application: Wallet
 - OS: ubuntu 22.04
 - Version: 1.0.4

fmtabbara:
Closes: #1340

### DIFF
--- a/nym-wallet/src/pages/send/SendWizard.tsx
+++ b/nym-wallet/src/pages/send/SendWizard.tsx
@@ -62,6 +62,8 @@ export const SendWizard = () => {
   };
 
   const handleSend = async () => {
+    setIsLoading(true);
+
     const formState = methods.getValues();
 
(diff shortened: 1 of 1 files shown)


tommyv1987:
Tested and working. Merging.

PR #1345, opened 2022-06-10
Change epoch length to 10 minutes

### LINKED ISSUE #2033: Different epoch lengths for different networks

durch:
# Description

Testing how easy it is to change the epoch, tests will fail since the rewards calc depends on the epoch length

Closes https://github.com/nymtech/nym/issues/2033

# Checklist:

### DIFF
--- a/common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
+++ b/common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
@@ -431,6 +431,7 @@ impl MixNodeBond {
 
     pub fn estimate_reward(
         &self,
+        base_operator_cost: u64,
         params: &RewardParams,
     ) -> Result<RewardEstimate, MixnetContractError> {
         let total_node_reward = self
@@ -439,15 +440,15 @@ impl MixNodeBond {
             .checked_to_num::<u128>()
             .unwrap_or_default();
         let node_profit = self
-            .node_profit(params)
+            .node_profit(params, base_operator_cost)
             .checked_to_num::<u128>()
             .unwrap_or_default();
         let operator_cost = params
             .node
-            .operator_cost()
+            .operator_cost(base_operator_cost)
             .checked_to_num::<u128>()
             .unwrap_or_default();
-        let operator_reward = self.operator_reward(params);
+        let operator_reward = self.operator_reward(params, base_operator_cost);
         // Total reward has to be the sum of operator and delegator rewards
         let delegators_reward = node_profit.saturating_sub(operator_reward);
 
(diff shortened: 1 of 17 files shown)


common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
```diff
@@ -479,21 +480,25 @@ impl MixNodeBond {
...
-    pub fn node_profit(&self, params: &RewardParams) -> U128 {
+    pub fn node_profit(&self, params: &RewardParams, base_operator_cost: u64) -> U128 {
         self.reward(params)
             .reward()
-            .saturating_sub(params.node.operator_cost())
+            .saturating_sub(params.node.operator_cost(base_operator_cost))
     }
 
-    pub fn operator_reward(&self, params: &RewardParams) -> u128 {
+    pub fn operator_reward(&self, params: &RewardParams, base_operator_cost: u64) -> u128 {
```
jstuczyn, line 489: why is `base_operator_cost` a `u64`? It makes it very non-obvious to what it's supposed to be representing as all of our `coin` related things were kept in some variant of `U128`

contracts/mixnet/src/contract.rs
```diff
@@ -435,7 +440,15 @@ pub fn query(deps: Deps<'_>, env: Env, msg: QueryMsg) -> Result<QueryResponse, C
     Ok(query_res?)
 }
 
-fn migrate_contract_state_params(deps: DepsMut<'_>) -> Result<(), ContractError> {
+fn update_epoch_duration(deps: DepsMut<'_>) -> Result<(), ContractError> {
+    let mut epoch = crate::interval::storage::current_epoch(deps.storage)?;
+    epoch.update_duration(600);
```
jstuczyn, line 448: Should we be changing  length of the current, existing, epoch to 10min for this test -> is it not going to cause some havoc for current epoch rewarding?
Why not instead force it to transition to a new epoch of length 10min?

common/cosmwasm-smart-contracts/mixnet-contract/src/mixnode.rs
durch: It was always `u64`, it was just hidden away in `network_defaults`

contracts/mixnet/src/contract.rs
durch: Its a good idea, but I don't see how we can do without overcomplicating this. We'd need to update rewarded sets, distribute rewards and snapshot, not to mention mess with how epochs advance, as it can't advance while its in progress by design. With this change the validator-api will see that epoch needs advancing, and will just advance it, doing all the usual steps.

I'd suggest we test how this goes on QA, and if something horrible happens we'll think it through some more

PR #1353, opened 2022-06-13
client: allow rerun init for gateway config

### LINKED ISSUE #2017: Fix re-running init for client/socks5-client
Rerunning `init` for the socks5 client fails due to not preserving the gateway fields that it detects and sets on the first run.

When re-running init we should either reuse existing keys (like I think it's currently intended?), but also allow the user to re-register without having to manually go and delete files

This blocks: https://github.com/nymtech/nym/issues/2417

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/2017

Currently re-running init fails for the native and socks5 client. The gateway registration is skipped, but that also means the gateway related fields in the config is never set so when the config file is written the previously set gateway fields are erased.

This PR changes so that rerunning init will reuse the gateway config fields, and adds another flag `--force-register-gateway` that will trigger gateway registration regardless if the config file and associated keys already exists or not.

# Checklist:

- [X] SOCKS5 client
- [x] Native client


clients/client-core/src/config/mod.rs
```diff
@@ -122,6 +122,14 @@ impl<T: NymConfig> Config<T> {
         };
     }
 
+    pub fn with_gateway_endpoint_from_config(&mut self, other_config: &Config<T>) {
```
jstuczyn: It seems like it would be cleaner to just pass a `GatewayEndpoint` as an argument rather than the entire config

jstuczyn (approved):
lgtm

clients/client-core/src/config/mod.rs
octol: Had a stab at this, see the latest commits

PR #1355, opened 2022-06-13
feat(wallet): add a custom link component to match new design

doums:
# [wallet] link component style

This PR introduces a custom `Link` component to match the new design. Plus refactor MUI `Link` occurrences to the new one.

https://www.figma.com/file/KiYKZnWeefSALAe0j2dAPI/Nym-Wallet?node-id=2285%3A37345

image


mmsinclair (changes requested):
I like the idea of this component and it should be applied more widely to any Nym UI 😁 

So please can you move it to `ts-packages/react` and make it a component there?

It also needs a story in the `ts-packages/react` storybook so we can all see how it works. I suspect you'll have to move the theme changes to the Nym theme and not the wallet theme.

mmsinclair (approved):
Nice @doums, much cleaner 😁

PR #1365, opened 2022-06-14
Put circulating supply back as Option

durch:
# Description

# Checklist:


jstuczyn:
I think that given that `#[serde(alias)`] is not a problem, I think this PR is no longer required?

Issue #276, opened 2020-07-09
The messages pushed from gateway should be encrypted.


jstuczyn:
Right now when the client is sending any [sphinx] messages to the gateway they are encrypted with the shared key. However, any messages the gateway received that should be sent back to the client are not yet encrypted - I probably forgot about that part. Obviously that should get fixed!

PR #1386, opened 2022-06-20
Explorer: #276 change avg. uptime name to "Routing Score" and tooltip text 

### LINKED ISSUE #2395: NE - Change name and tooltip for Avg. Uptime:
Change name to "Routing Score"

Tooltip: Node's routing score is relative to that of the network. Each time a node is tested, the test packets have to go through the full path of the network (a gateway + 3 nodes). If a node in the path drop packets it will affect the score of other nodes in the test.

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2395

Before:
image

Now:
image


doums (approved):
lgfm

mmsinclair (approved):

PR #1428, opened 2022-07-04
Feature/wallet bonding with fees rebase

fmtabbara:
# Description

Display simulated gas fees for bonding/unbonding mixnodes and gateways with locked or unlocked coins

## Closes:
https://github.com/nymtech/nym/issues/2379
https://github.com/nymtech/nym/issues/2378
https://github.com/nymtech/nym/issues/2377
https://github.com/nymtech/nym/issues/2376


nym-wallet/src/components/Modals/LoadingModal.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Box, CircularProgress, Modal, Stack, Typography } from '@mui/material';
+
+const modalStyle = {
+  position: 'absolute' as 'absolute',
```
doums: are you sure about this cast `'absolute' as 'absolute'` ?

ts-packages/types/src/types/global.ts
```diff
@@ -1,13 +1,10 @@
 import { MixNode, MajorCurrencyAmount } from './rust';
 
-export enum EnumNodeType {
-  mixnode = 'mixnode',
-  gateway = 'gateway',
-}
+export type TNodeType = 'mixnode' | 'gateway';
```
doums, line 3: :godmode:

doums:
LGTM

nym-wallet/src/components/Modals/LoadingModal.tsx
fmtabbara: I think I borrowed this style object from some code Mark wrote. So I'm not sure why the cast is there but my guess is to satisfy something in the MUI sx prop. will see if we can remove the cast without any issues

fmtabbara: I managed to remove it :100:

PR #1432, opened 2022-07-04
feat(wallet): add generic dialog component

doums:
# Generic component for Dialogs

This PR introduces a new component to render Dialogs based on the new Figma design.

image


doums:
We decided to make the necessary changes in `SimpleModal` component and use it instead of create a bare new one.

PR #1434, opened 2022-07-05
feat(wallet): add confirmation modal component

doums:
# ConfirmationModal

Add a new component to render confirmation Modals.

image

### DIFF
--- a/nym-wallet/src/components/Modals/ConfirmationModal.tsx
+++ b/nym-wallet/src/components/Modals/ConfirmationModal.tsx
@@ -1,0 +1,79 @@
+import React from 'react';
+import {
+  Breakpoint,
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  SxProps,
+  Typography,
+} from '@mui/material';
+
+export interface Props {
+  open: boolean;
+  onConfirm: () => void;
+  onClose?: () => void;
+  children?: React.ReactNode;
+  title: React.ReactNode | string;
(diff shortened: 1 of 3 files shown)


nym-wallet/src/components/Modals/ConfirmationModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  title: React.ReactNode | string;
+  subTitle?: React.ReactNode | string;
+  confirmButton: React.ReactNode | string;
+  disabled?: boolean;
+  sx?: SxProps;
+  fullWidth?: boolean;
+  maxWidth?: Breakpoint;
+}
+
+const ConfirmationModal = ({
```
fmtabbara: Can we use export const here, instead of the default export?

nym-wallet/src/components/Modals/ConfirmationModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+        (typeof subTitle === 'string' ? (
+          <Typography fontWeight={400} variant="subtitle1" fontSize={12} color={(t) => t.palette.nym.text.muted}>
+            {subTitle}
+          </Typography>
+        ) : (
+          subTitle
+        ))}
+    </DialogTitle>
+  );
+  const confirmButtonComp =
```
fmtabbara: Can these two components start with uppercase `TitleComp` and `ConfirmButtonComp`. I think it's a little clear on a quick look that they are return components

fmtabbara (approved):

PR #1435, opened 2022-07-05
Feature/wallet modal back button

fmtabbara:
# Description

Add an optional back button to the SimpleModal component

image


doums:
In Back button, the icon seems not to be correctly centered, try using ` sx={{ px: 1 }}` in the parent `<Button />` component

PR #1442, opened 2022-07-06
using a tag instead of link component for external link

### LINKED ISSUE #2371: NE -  Fix Network Explorer validators link
Now at the Network Explorer, the validators links browse to https://explorer.nymtech.net/https://blocks.nymtech.net/validators, and it's a broke page. 
It should browse to: https://blocks.nymtech.net/validators

Branch: `300-fix-ne-validarors-link`

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2371

I've been trying to make the `Link` component, work well with external link, without success, for this reason, finally I use the `a` tag instead for them.

On the other hand, we have the `originalNavOptions` exported from `explorer/src/context/nav.tsx`, I think, we can removed the exported one from `explorer/src/components/Nav.tsx`


doums (approved):
LGTM

PR #1469, opened 2022-07-20
fix type update issues with actions (bonding, delegating etc.)

fmtabbara:
# Description

Refactor to fix some actions that were not working due to the new types updates to coin and denom.


mmsinclair (commented):
lgtm

PR #1472, opened 2022-07-21
Feature/rewarding revamp

### LINKED ISSUE #2018: Use native cosmwasm types (`Uint128`, `Decimal`) for reward calculations, instead of the Fixed type we use now to try and fix rounding errors.

### LINKED ISSUE #2089: Rust client code to calculate operator and delegator rewards
As a delegator or operator,
I want to calculate the rewards for an epoch
So that I know I'm being paid the correct rewards

Features:
- user provides `uptime` value
- estimate of rewards for an epoch is return

### LINKED ISSUE #2093: Deploy v2 mixnet payments contract

jstuczyn:
# Description

Closes https://github.com/nymtech/nym/issues/2093
Closes https://github.com/nymtech/nym/issues/2018
Closes: https://github.com/nymtech/nym/issues/2089

Alright. So this one is a massive one. Probably one of the biggest PRs I've ever done on this repo, but I think it could be worthy of bumping our mixnet contract version to `2.0.0`. It's still nowhere being fully done, however, I think the API on the mixnet contract is in 95% stabilised at this point and the only changes should be just bugfixes.

Therefore, I'm creating this draft so that if there is anyone with some down-time who wants to help me update the rest of the codebase to be compatible with the new changes they would be able to do so. (if you attempt to pick up some part of the codebase, please leave a comment so we wouldn't accidentally step on each other's toes)

Myself I'm finishing writing/fixing unit-tests in the mixnet contract itself. Then my plan is to move to `nymd-client` and then the rest of the codebase.

I tried to describe all the changes in [this doc](https://github.com/nymtech/nym/blob/feature/rewarding-revamp/contracts/mixnet/v2-changes.md), however I might have missed something due to the extent of the changes. I still recommend at least having a glance at the `##Overview` section.

But the two biggest changes are: mixnodes no longer being stored by their identity keys, instead we use a monotonically increasing id and changing our delegation rewarding to more closely resemble what cosmos is doing and thus simplifying the complexity. However, implementing those two changes had a cascading effect on the rest of the codebase. Essentially one changed type in mixnet contract implies broken validator-api, explorer-api, vesting-contract, etc.... And this is something we have to deal with now.

I have, however, one personal request to @durch: if you could take a look at the vesting contract that would be amazing since you understand it best. Also, I think, I left some questions to you in the mixnet contract regarding tracking vesting commands to make sure I used them correctly. (I tagged them with `@DU`). 

# Checklist:


durch:
@jstuczyn, I'll pick vesting contract, no problem :)

contracts/mixnet/src/interval/pending_events.rs
```diff
@@ -0,0 +1,403 @@
...
+        Some(details)
+            if details.rewarding_details.still_bonded()
+                && !details.bond_information.is_unbonding =>
+        {
+            details
+        }
+        _ => {
+            // if mixnode is no longer bonded or in the process of unbonding, return the tokens back to the
+            // delegator;
+            // TODO: @DU: do we need to do any vesting-specific tracking here?
```
durch: Yes, probably easiest is to send a `TrackUndelegate` message, as if it was a regular delegation

PR #1481, opened 2022-07-29
Feature/bonding refactor

fmtabbara:
# Description

Closes:
https://github.com/nymtech/nym/issues/2452
https://github.com/nymtech/nym/issues/2441
https://github.com/nymtech/nym/issues/2393
https://github.com/nymtech/nym/issues/2386
https://github.com/nymtech/nym/issues/2385
https://github.com/nymtech/nym/issues/2383
https://github.com/nymtech/nym/issues/2382


nym-wallet/src-tauri/src/operations/mixnet/bond.rs
```diff
@@ -198,3 +221,58 @@ pub async fn get_operator_rewards(
...
+        Ok(node_description) => node_description,
+        Err(_e) => NodeDescription::new(
+            "-".to_string(),
+            "-".to_string(),
+            "-".to_string(),
+            "-".to_string(),
+        ),
+    };
+    return response;
+}
```
fmtabbara: One of the Rustys should probably check this code :S

nym-wallet/src-tauri/src/operations/mixnet/bond.rs
```diff
@@ -198,3 +210,54 @@ pub async fn get_operator_rewards(
...
+    match response {
+        Ok(res) => {
+            let json = res.json::<NodeDescription>().await;
+            match json {
+                Ok(json) => Ok(json),
+                Err(e) => Err(e),
+            }
+        }
+        Err(e) => Err(e),
+    }
```
doums, line 249: I think you could write this match pattern as follows
```
    match response {
        Ok(res) => res.json::<NodeDescription>().await,
        Err(e) => Err(e),
    };

```

nym-wallet/src-tauri/src/operations/mixnet/bond.rs
```diff
@@ -198,3 +210,54 @@ pub async fn get_operator_rewards(
...
+pub async fn get_mix_node_description(
+    host: &str,
+    port: u16,
+) -> Result<NodeDescription, BackendError> {
+    let res = fetch_mix_node_description(host, port).await;
+
+    match res {
+        Ok(node_description) => Ok(node_description),
+        Err(e) => Err(BackendError::ReqwestError { source: e }),
+    }
```
doums: you could use `map_err` here
```
    return fetch_mix_node_description(host, port)
        .await
        .map_err(|e| BackendError::ReqwestError { source: e });
```

fmtabbara: :+1:

nym-wallet/src-tauri/src/operations/mixnet/bond.rs
fmtabbara: I did try this but it was causing me an error. I'll keep as it is for now.

nym-wallet/src-tauri/src/operations/mixnet/bond.rs
fmtabbara: Updated this as recommended here.

nym-wallet/src-tauri/src/operations/mixnet/bond.rs
doums: hum probably because in my snippet there should be no `;` at the end of the match `block`, my bad

PR #1493, opened 2022-08-04
Introduce Validator Client Binary

tommyv1987:
# Description
- enables end user to perform actions via a cli tool
- has the ability to be used across environment dependant of --config file passed
- contains the ability to perform vesting and normal operations

# Checklist:


tools/validator-client-scripts/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+log = "0.4"
+pretty_env_logger = "0.4"
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+tokio = { version = "1.11", features = [ "net", "rt-multi-thread", "macros", "signal"] }
+streaming-stats = "0.2.3"
+
+
+# we don't really need those directly, but including them somehow solves dependency issue
+url = "2.2"
```
jstuczyn: see if you can remove those now actually. What what I recall I had to include them because I was depending on `validator-client` through `git`

tools/validator-client-scripts/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+bs58 = "0.4"
+clap = { version = "3.2.8", features = ["derive"] }
+csv = "1.1"
+dotenv = "0.15.0"
+log = "0.4"
+pretty_env_logger = "0.4"
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+tokio = { version = "1.11", features = [ "net", "rt-multi-thread", "macros", "signal"] }
+streaming-stats = "0.2.3"
```
jstuczyn: This might also be redundant now due to commands you have already removed (same as `csv`)

tools/validator-client-scripts/Makefile
```diff
@@ -0,0 +1,8 @@
+install:
```
jstuczyn: Cant we put those `make` commands in our main `Makefile`?

tools/validator-client-scripts/README.md
```diff
@@ -0,0 +1,32 @@
+# NYM - CLI TOOL
+## Overview
+Set of simple CLI commands to perform basic actions, such as upload contract, bond mixnode, etc.
+Note: to work it requires rust 1.60+ 
+
+## The binary
+    - Build the binary 
+    - cargo build --release
```
jstuczyn: you can add you can use `cargo build --release --bin validator-client-scripts` so that it would only build this binary rather than everything in the repo (which should speed things up)

tools/validator-client-scripts/README.md
```diff
@@ -0,0 +1,32 @@
+# NYM - CLI TOOL
+## Overview
+Set of simple CLI commands to perform basic actions, such as upload contract, bond mixnode, etc.
+Note: to work it requires rust 1.60+ 
+
+## The binary
+    - Build the binary 
+    - cargo build --release 
+    - this will compile the binary to "your/main/nym/path/target/release"
+    - the binary is named ./validator-client-scripts-rs
```
jstuczyn: Didn't we drop the `-rs` suffix?

tools/validator-client-scripts/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Executing commands
+    -./validator-scripts-rs --help
```
jstuczyn: `-rs` suffix

tools/validator-client-scripts/README.md
```diff
@@ -0,0 +1,32 @@
...
+    - will produce a list of it's capabilites 
+
+The binary requires mandatory OPTIONS:        
+```
+--config-env-file <CONFIG_ENV_FILE>
+--mixnet-contract <MIXNET_CONTRACT>      
+--mnemonic <MNEMONIC>                    
+--nymd-url <NYMD_URL>                    
+--vesting-contract <VESTING_CONTRACT>
```
jstuczyn, line 22: I wonder, should contracts be mandatory if you don't need to interact with them? For example should you need to provide anything if you just want to send tokens between accounts (or just want to create a dummy account address)?

tools/validator-client-scripts/envs/mainnet.env
```diff
@@ -0,0 +1,20 @@
...
+
+RUST_LOG=info
+RUST_BACKTRACE=1
+
+BECH32_PREFIX=n
+MIX_DENOM=unym
+MIX_DENOM_DISPLAY=nym
+STAKE_DENOM=unyx
+STAKE_DENOM_DISPLAY=nyx
+DENOMS_EXPONENT=6
```
jstuczyn, line 11: This is just nitpicking, but in _theory_ the tokens could have different exponents (but it's such a niche case we can ignore it)

tools/validator-client-scripts/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+
+async fn wait_for_interrupt() {
+    if let Err(e) = tokio::signal::ctrl_c().await {
+        error!(
+            "There was an error while capturing SIGINT - {:?}. We will terminate regardless",
+            e
+        );
+    }
+    println!(
+        "Received SIGINT - the gateway will terminate now (threads are not yet nicely stopped, if you see stack traces that's alright)."
```
jstuczyn: gateway will terminate? : )
though it was most likely caused by me, but hey, `git blame` will show your name here hehe

tools/validator-client-scripts/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    #[clap(long)]
+    pub(crate) nymd_url: Option<String>,
+
+    #[clap(long)]
+    pub(crate) mnemonic: Option<bip39::Mnemonic>,
+
+    // it can only be `None` in the case of contract upload or init
+    #[clap(long)]
+    pub(crate) mixnet_contract: Option<AccountId>,
```
jstuczyn, line 94: aha, so the contracts are actually optional unlike what the readme says

tools/validator-client-scripts/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "validator-client-scripts"
+version = "0.1.0"
+edition = "2018"
```
doums: Humble question, is there a reason why we are not using the 2021 edition?

jstuczyn: The answer is that this binary was originally written before version 2021 was a thing. But an excellent catch, @tommyv1987 go and update that : ) (all you have to do is change `edition = "2018"` to `edition = "2021"`)

tools/validator-client-scripts/README.md
tommyv1987: fixed in pending inbound commit

tools/validator-client-scripts/README.md
tommyv1987: fixed in pending inbound commit

tools/validator-client-scripts/README.md
tommyv1987: These are opts as you've seen below :wink:

tools/validator-client-scripts/src/main.rs
tommyv1987: Updated

tools/validator-client-scripts/envs/mainnet.env
tommyv1987: nice to make note of it, we can amend where we see in the future on different configs.

tools/validator-client-scripts/README.md
tommyv1987: removed :)

tools/validator-client-scripts/Makefile
tommyv1987: sure thang :)

tools/validator-client-scripts/Cargo.toml
tommyv1987: resolved in inbound commit

tools/validator-client-scripts/envs/mainnet.env
```diff
@@ -0,0 +1,20 @@
...
+MIXNET_CONTRACT_ADDRESS=n14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sjyvg3g
+VESTING_CONTRACT_ADDRESS=n1nc5tatafv6eyq7llkr2gv50ff9e22mnf70qgjlv737ktmt4eswrq73f2nw
+BANDWIDTH_CLAIM_CONTRACT_ADDRESS=n19lc9u84cz0yz3fww5283nucc9yvr8gsjmgeul0
+COCONUT_BANDWIDTH_CONTRACT_ADDRESS=n19lc9u84cz0yz3fww5283nucc9yvr8gsjmgeul0
+MULTISIG_CONTRACT_ADDRESS=n19lc9u84cz0yz3fww5283nucc9yvr8gsjmgeul0
+REWARDING_VALIDATOR_ADDRESS=n10yyd98e2tuwu0f7ypz9dy3hhjw7v772q6287gy
+STATISTICS_SERVICE_DOMAIN_ADDRESS="http://127.0.0.1:8090"
+NYMD_VALIDATOR="https://rpc.nyx.nodes.guru/"
+API_VALIDATOR="https://validator.nymtech.net/api/"
```
neacsu, line 20: Given that there already is a `envs` directory in the repo, I don't see any reason to replicate that here, as any future change would need to be done in two places instead of one.

tools/validator-client-scripts/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+        .init();
+}
+
+fn create_client(args: &Args) -> Client {
+    let network_details = NymNetworkDetails::new_from_env();
+    let client_config = nymd::Config::try_from_nym_network_details(&network_details)
+        .expect("failed to construct valid validator client config with the provided network");
+    NymdClient::connect_with_mnemonic(
+        client_config,
+        args.nymd_url
```
neacsu, line 190: `network_details` from above already contains a `nymd_url` from the .env files. I would say, if the url is provided as cli option, use it, but if it's not, don't panic on it with the message `nymd url was not provided` and instead default to the value from .env file.

tools/validator-client-scripts/src/main.rs
neacsu: I would add that they are not used anywhere anymore now, so we could actually remove the contract address arguments altogether.

tools/validator-client-scripts/src/commands/migrate_contract.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    #[clap(long)]
+    pub gas: Option<u64>,
+}
+
+pub(crate) async fn migrate(client: Client, args: Args) {
+    println!("Starting contract migration!");
+
+    let memo = args.memo.unwrap_or_else(|| "contract migration".to_owned());
+    let contract_address = client.mixnet_contract_address();
```
neacsu, line 28: We should support (at least) two configurable contract migrations: mixnet and vesting. Hardcoding it to mixnet here would limit that for us. I would say, pass the migrating contract address as an migration argument instead.

tools/validator-client-scripts/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+    ClaimDelegatorReward(commands::claim_delegator_reward::Args),
+    CompoundOperatorReward(commands::compound_operator_reward::Args),
+    CompoundDelegatorReward(commands::compound_delegator_reward::Args),
+    Init(commands::init_contract::Args),
+    Migrate(commands::migrate_contract::Args),
+    Upload(commands::upload_contract::Args),
+    UnbondMixnode(commands::unbond_mixnode::Args),
+    UnbondGateway(commands::unbond_gateway::Args),
+    DelegateToMixnode(commands::delegate_to_mixnode::Args),
+    UnDelegateFomMixnode(commands::undelegate_from_mixnode::Args),
```
neacsu, line 61:
```suggestion
    UndelegateFromMixnode(commands::undelegate_from_mixnode::Args),
```
Small nit, preventing the command from looking like `un-delegate-from-mixnode`, which looks a bit strange to me. Also fixes a typo.

neacsu (changes requested):

mmsinclair:
Replaced with https://github.com/nymtech/nym/pull/1577

PR #1505, opened 2022-08-08
feat(wallet-balance): hide vesting ui when vesting balance is empty

### LINKED ISSUE #2463: Remove vesting section in balance screen once vesting period has finalised and all locked tokens are move to liquid balance
-  Given a user has finished their vesting period
 - And they have no more balance (LOCKED tokens) in their vesting contract (all transferred to their liquid balance)
 - Then we should not display the vesting period in the wallet anymore

doums:
# Description

Closes: [#210](https://github.com/nymtech/nym/issues/2463)

### DIFF
--- a/nym-wallet/src/pages/balance/index.tsx
+++ b/nym-wallet/src/pages/balance/index.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Box } from '@mui/material';
 import { AppContext } from '../../context/main';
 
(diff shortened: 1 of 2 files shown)


nym-wallet/src/pages/balance/index.tsx
```diff
@@ -9,9 +9,26 @@ import { TransferModal } from './components/TransferModal';
...
+    }
+    if (
+      originalVesting &&
+      currentVestingPeriod === 'After' &&
+      tokenAllocation?.locked === '0' &&
+      tokenAllocation?.vesting === '0' &&
+      tokenAllocation?.spendable === '0'
+    ) {
+      setShowVestingCard(false);
+    }
```
fmtabbara: What about 

```
if (
      originalVesting &&
      currentVestingPeriod === 'After' &&
      tokenAllocation?.locked === '0' &&
      tokenAllocation?.vesting === '0' &&
      tokenAllocation?.spendable === '0'
    ) {
      setShowVestingCard(false);
    } else if (originalVesting) {
       setShowVestingCard(true)
    }
```
What you have done is fine and works but I think we should set it up so that it is either setShowVestingCard(true) or set showVestingCard(false). At the moment both of your if statements can be triggered if `originalVesting` is true

fmtabbara (approved):

PR #1524, opened 2022-08-11
317 forms menus titles UI

### LINKED ISSUE #2354: Wallet: Global update forms, menus and titles of screens
- Left side menu: order of categories 1.balance, 2.send 3.receive 4.delegation, 5.bond + way smaller information about version(according to figma)

- Top bar menu with accounts and networks - different font size and thickness + to big avatar

- Title name to small - its like this everywhere in each category, please lets change this
image

- Forms: all forms and modules I'm bulding in figma are created like this:
-- spacing between main elements 24px
-- top and bottom margins 24px
-- smaller lists like these confirmation screens 12px between each element
image

image

- Form when we are showing user address it should he should not be able to mark it. Right now it seems like this is active imput

image

- When you are clicking copy button in wallet(balance screen) container in background is moving- width and hight should be always the same :)
image

- [x] design
- [x] development

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2354


doums (approved):
LGTM

nym-wallet/src/components/Delegation/DelegateModal.tsx
```diff
@@ -211,7 +211,7 @@ export const DelegateModal: React.FC<{
           initialValue={identityKey}
           readOnly={Boolean(initialIdentityKey)}
           textFieldProps={{
-            autoFocus: !initialIdentityKey,
+            focused: !initialIdentityKey,
```
fmtabbara: note sure why this changed? The autofocus doesnt seem to work now.

When the delegate modal opens the `Node identity key` field should focus`

nym-wallet/src/components/NymCard.tsx
```diff
@@ -21,7 +21,11 @@ export const NymCard: React.FC<{
 }> = ({ title, subheader, Action, Icon, noPadding, borderless, children, dataTestid }) => (
   <Card variant="outlined" sx={{ overflow: 'auto', ...(borderless && { border: 'none', dropShadow: 'none' }) }}>
     <CardHeader
-      sx={{ p: 3, color: (theme: Theme) => theme.palette.text.primary }}
+      sx={{
+        p: 3,
+        color: (theme: Theme) => theme.palette.text.primary,
```
fmtabbara: inside the `sx` prop you dont need to access the theme directly. You can use a shorthand `text.primary`. Try it. It should work.

nym-wallet/src/components/Accounts/AccountOverview.tsx
```diff
@@ -6,7 +6,7 @@ import { AccountAvatar } from './AccountAvatar';
 export const AccountOverview = ({ account, onClick }: { account: AccountEntry; onClick: () => void }) => (
   <Button
     startIcon={<AccountAvatar name={account.id} small />}
-    sx={{ color: 'text.primary' }}
+    sx={{ color: 'text.primary', fontSize: 14 }}
```
fmtabbara, line 9: This is ok for now but in the future we need to come back and fix this up. Ideally these values should come from somewhere in the theme which will make things more manageable and consistent.

nym-wallet/src/layouts/AppLayout.tsx
```diff
@@ -39,9 +39,9 @@ export const ApplicationLayout: React.FC = ({ children }) => {
             <Nav />
           </Box>
           {appVersion && (
-            <Box color="#888" ml={5} mt={8}>
+            <Typography sx={{ color: (t) => t.palette.grey[500], fontSize: 14, ml: 5, mt: 8 }}>
```
fmtabbara: You dont need to access the theme in this way when using `sx` prop. Try `color: `grey.500`

fmtabbara (changes requested):

nym-wallet/src/components/Accounts/AccountOverview.tsx
gala1234: I agree completely, doing these Figma diff tickets I'm having the feeling that we need some consistency, that we could be losing @fmtabbara

nym-wallet/src/components/Delegation/DelegateModal.tsx
gala1234: @fmtabbara I had some issues with the autofocus, but now I'm testing it, and seems ok. I'm going to revert this change, and keep it in mind if I see something weird in the future. Thanks

fmtabbara (approved):

PR #1537, opened 2022-08-16
Added keybase team variable

raphael-walther:


.github/workflows/support-files/notifications/send_message.js
```diff
@@ -38,6 +38,11 @@ function validateContext() {
       'Paperkey is not defined. Please set env var KEYBASE_NYMBOT_PAPERKEY',
     );
   }
+  if (!context.env.KEYBASE_NYMBOT_TEAM) {
+    throw new Error(
+      'Team is not defined. Please set env var KEYBASE_NYMBOT_TEAM',
+    );
+  }
```
mmsinclair: All the current GitHub Actions will fail, because they do not set this environment variable. So this check is a very bad idea.

.github/workflows/support-files/notifications/send_message.js
```diff
@@ -89,7 +94,7 @@ async function sendKeybaseMessage(messageBody) {
     });
 
     const channel = {
-      name: 'nymtech_bot',
```
mmsinclair, line 92: I suggest you rather try the env var and fallback to the default

```suggestion
      name: context.env.KEYBASE_NYMBOT_TEAM || 'nymtech_bot',
```

mmsinclair (changes requested):

.github/workflows/support-files/notifications/send_message.js
raphael-walther: That's a better solution.

mmsinclair (approved):

PR #1540, opened 2022-08-17
Nym-connect - store in local storage the service provider

### LINKED ISSUE #2311: Connect - load balance SP selection
### Why?

Users choose a service provider for a particular service. When there are lots of SPs, users are just going to choose the first one. This will put a large amount of load on just one SP.

We want to spread the load.

For testing stats collection, we also need to make the gateway sticky.

**NB:** after the testing phase is done, we want to load balance the gateways too and use any gateway for any service.

### What?

- [ ] Change the service selection dropdown to only contain the top level services, i.e. `keybase`, `electrum` (and `telegram` when it is added)
- [ ] When a user has not selected a service:
  - a random SP+GW pair should be selected
  - the selection should be stored using https://github.com/tauri-apps/tauri-forage
- [ ] If a user has selected a service previously
  - load the SP+GW pair from `local-forage`
  - check against the list of service providers, and if the selected one doesn't exist, fall back to selecting a new pair and storing (see above)

doums:
# Description

Closes: [#XXXX](https://github.com/nymtech/nym/issues/2311)

image
image

### DIFF
--- a/nym-connect/CHANGELOG.md
+++ b/nym-connect/CHANGELOG.md
@@ -1,3 +1,14 @@
+## [nym-connect-v1.0.2](https://github.com/nymtech/nym/tree/nym-connect-v1.0.2) (2022-08-18)
+
+### Changed
+
+- nym-connect: "load balance" the service providers by picking a random Service Provider for each Service and storing in local storage so it remains sticky for the user ([#1540])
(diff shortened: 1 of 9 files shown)


mmsinclair (approved):
Great work @doums 😀 

I added a few commits:
- the service description is shown for the current selection in the selector (rather than the SP description)
- bumped the version
- updated the CHANGELOG

PR #1543, opened 2022-08-18
fix(nym-wallet): bonding, enhance error handling

doums:
# Description

Add better error logging. Add user feedback (dialog UI) when an error occurs while loading bonded node data.

Note: The error dialog will only show up when the error prevents to load bonded node data. Not on "no blocking" error.


gala1234 (approved):
I didn't know we have this `Console` util, god to know.

PR #1548, opened 2022-08-22
fix(nym-wallet): typing

doums:
# Description

fix typing errors

### DIFF
--- a/nym-wallet/src/components/Delegation/DelegateModal.tsx
+++ b/nym-wallet/src/components/Delegation/DelegateModal.tsx
@@ -1,11 +1,12 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import { Box, Typography, SxProps } from '@mui/material';
 import { IdentityKeyFormField } from '@nymproject/react/mixnodes/IdentityKeyFormField';
 import { CurrencyFormField } from '@nymproject/react/currency/CurrencyFormField';
 import { CurrencyDenom, FeeDetails, DecCoin, decimalToFloatApproximation } from '@nymproject/types';
 import { Console } from 'src/utils/console';
 import { useGetFee } from 'src/hooks/useGetFee';
-import { simulateDelegateToMixnode, simulateVestingDelegateToMixnode } from 'src/requests';
+import { simulateDelegateToMixnode, simulateVestingDelegateToMixnode, tryConvertIdentityToMixId } from 'src/requests';
+import { debounce } from 'lodash';
 import { SimpleModal } from '../Modals/SimpleModal';
 import { ModalListItem } from '../Modals/ModalListItem';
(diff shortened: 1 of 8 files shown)


nym-wallet/src/components/Delegation/DelegateModal.tsx
```diff
@@ -134,7 +132,7 @@ export const DelegateModal: React.FC<{
     }
   };
 
-  const handleConfirm = async ({ mixId, value }: { mixId: number; value: DecCoin }) => {
+  const handleConfirm = async ({ id, value }: { id: number; value: DecCoin }) => {
```
jstuczyn: I think using `mixId` was more explicit here (I attempted to use it everywhere on rust side to make it clear what we're actually referring to)

nym-wallet/src/components/Delegation/DelegateModal.tsx
```diff
@@ -171,6 +169,29 @@ export const DelegateModal: React.FC<{
...
+  const resolveMixId = useCallback(
+    debounce(async (idKey) => {
+      if (!idKey || !validateKey(idKey, 32)) {
+        return;
+      }
+      let res;
+      try {
+        res = await tryConvertIdentityToMixId(idKey);
+      } catch (e) {
+        Console.warn(`failed to resolve mix_id for "${idKey}": ${e}`);
```
jstuczyn, line 181: Well, this can actually mean one of 2 things:
- we failed to resolve it (as you mention), implying some issue with the validator
- [more likely] mixnode with the specified identity key doesn't actually exist

jstuczyn (approved):

nym-wallet/src/components/Delegation/DelegateModal.tsx
doums: Yes I understand, but there is already a variable declared as `mixId` in the upper scope, so ESLint complained about that, and it's true that it's not a good practice to have overlaps of variable name from different scopes, in a TS code base.
So this change fix that lint error. In addition, reading the function body, it's pretty clear for what `id` actually stand for.

doums: But I can find a compromise like something:
```
async ({ mixId: id, value }: { mixId: number; value: DecCoin }) => {
```

nym-wallet/src/components/Delegation/DelegateModal.tsx
doums: Yes, in both cases "failed to resolve mix_id" is generic enough, isn't it?
If not, what do you suggest?

jstuczyn: I'm not 100% sure how the message will get displayed to the user, so maybe that's fine after all

PR #1565, opened 2022-08-25
fix(wallet): error UI feedback

doums:
Enhance the error UI when there is no matching mix_id for a given mixnode address, but that address has a valid format.

image


jstuczyn:
I'd perhaps change the messaging behind the error. Users are not going to care (nor understand) what `mix_id` is. How about, something along the lines of "Mixnode with this identity does not seem to be currently bonded"?

jstuczyn (approved):

PR #1570, opened 2022-08-26
Feature/credential mode

neacsu:
Activate the option to enable credential mode when building with `coconut` feature

### DIFF
--- a/clients/native/src/client/mod.rs
+++ b/clients/native/src/client/mod.rs
@@ -199,9 +199,9 @@ impl NymClient {
             Some(bandwidth_controller),
         );
 
-        if self.config.get_base().get_disabled_credentials_mode() {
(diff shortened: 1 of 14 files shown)


jstuczyn:
There's a lot of weird-feature locking happening here - can we just completely remove all `eth`-related things and simplify it all (it's not like anyone is nor even can use it)?

octol:
> There's a lot of weird-feature locking happening here - can we just completely remove all `eth`-related things and simplify it all (it's not like anyone is nor even can use it)?

+1 on removing `eth`, if we can do it :)

neacsu:
I am all for removing it, but I think we'll need to first have the coconut credential fully functional before going that road. From what I know, the regulatory reasons we got `eth` in the first place are still there

PR #1574, opened 2022-08-30
fix(nym-wallet): profit margin input

doums:


nym-wallet/src/context/bonding.tsx
```diff
@@ -161,6 +160,7 @@ export const BondingContextProvider = ({ children }: { children?: React.ReactNod
             data.bond_information.mix_node.host,
             data.bond_information.mix_node.http_api_port,
           );
+          const profitMargin = (Number(data.rewarding_details.cost_params.profit_margin_percent) * 100).toString();
```
jstuczyn: Isn't this literally the `toPercentFloatString` you introduced? : )

jstuczyn (approved):

nym-wallet/src/context/bonding.tsx
doums: nope, here this is the inverse operation :p

jstuczyn: Oh. Right. I wonder whether it would make sense to also move it to separate function then (in case we might need it elsewhere)

PR #1584, opened 2022-09-02
Migrate heights to timestamps

neacsu:
# Description

Introduce migration tx, as described in https://github.com/nymtech/nym/issues/1993


jstuczyn (approved):
Looks good to me, but given what we're changing here, some heavy QA-ing is definitely in order : )

PR #1587, opened 2022-09-02
Fix/wallet godzilla

doums:
# Desc

Add a confirmation modal when unbonding.

image


jstuczyn:
From my point of view it looks good, but I think somebody more well-versed in typescript should also take a look at it

mmsinclair (approved):

PR #1591, opened 2022-09-05
socks5-client: graceful shutdown

### LINKED ISSUE #2020: Graceful shutdown of socks5 client

### LINKED ISSUE #2419: Connect - Gracefully disconnect
Currently we run the socks5 client in its own thread and runtime, and the abruptly terminate it when the user clicks disconnect.

Instead we should send a message to the socks5 client to stop, and politely wait for it to finish.

This includes creating a channel for reporting state back to the UI from the client task.

Blocked by: https://github.com/nymtech/nym/issues/2020

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/2020
Closes: https://github.com/nymtech/nym/issues/2419

Shutdown sock5-client when signalled to do so. Hopefully this should also get rid of the stacktrace when clicking disconnect in nym-connect.

Some of these components are used elsewhere too, where we wire up the shutdown signalling, but don't actually fully utilize it.

# Checklist:


clients/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -159,9 +166,21 @@ impl LoopCoverTrafficStream<OsRng> {
...
-            self.on_new_message().await;
+        let mut shutdown = self.shutdown.clone();
+        while !shutdown.is_shutdown() {
+            tokio::select! {
+                biased;
+                _ = shutdown.recv() => {
+                    log::trace!("LoopCoverTrafficStream: Received shutdown");
+                }
+                next = self.next() => {
+                    if next.is_some() {
```
jstuczyn, line 177: if `next` is none, doesn't it mean that we received shutdown and we should just terminate?

clients/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -159,9 +166,21 @@ impl LoopCoverTrafficStream<OsRng> {
...
         )));
 
-        while self.next().await.is_some() {
-            self.on_new_message().await;
+        let mut shutdown = self.shutdown.clone();
+        while !shutdown.is_shutdown() {
+            tokio::select! {
+                biased;
+                _ = shutdown.recv() => {
+                    log::trace!("LoopCoverTrafficStream: Received shutdown");
```
jstuczyn, line 174: if we received shutdown, shouldn't we break the loop?

clients/client-core/src/client/mix_traffic.rs
```diff
@@ -66,9 +70,23 @@ impl MixTrafficController {
...
+                    Some(mix_packets) => {
+                        self.on_messages(mix_packets).await;
+                    },
+                    None => {
+                        log::trace!("MixTrafficController: Stopping since channel closed");
+                        break;
+                    }
+                },
+                _ = self.shutdown.recv() => {
+                    log::trace!("MixTrafficController: Received shutdown");
```
jstuczyn, line 85: again, should we perhaps break the loop here? (unless it will happen automatically since `self.shutdown.is_shutdown()` will return true now

clients/client-core/src/client/real_messages_control/acknowledgement_control/action_controller.rs
```diff
@@ -246,14 +252,18 @@ impl ActionController {
     }
 
     pub(super) async fn run(&mut self) {
-        loop {
-            // at some point there will be a global shutdown signal here as the third option
+        while !self.shutdown.is_shutdown() {
             tokio::select! {
                 // we NEVER expect for ANY sender to get dropped so unwrap here is fine
```
jstuczyn: hehe, so when I wrote those comments about senders that couldn't be dropped I **might** had written this in the context of binary never being shutdown gracefully - perhaps double check it? (or just to be safe, match on the result and never unwrap it)

clients/client-core/src/client/real_messages_control/acknowledgement_control/retransmission_request_listener.rs
```diff
@@ -121,9 +126,17 @@ where
 
     pub(super) async fn run(&mut self) {
         debug!("Started RetransmissionRequestListener");
-        while let Some(timed_out_ack) = self.request_receiver.next().await {
-            self.on_retransmission_request(timed_out_ack).await;
+
+        while !self.shutdown.is_shutdown() {
+            tokio::select! {
+                Some(timed_out_ack) = self.request_receiver.next() => {
```
jstuczyn: actually what would be the behaviour in this `select` if value returned was `None`? would it still get caught?

common/client-libs/gateway-client/src/socket_state.rs
```diff
@@ -98,9 +100,15 @@ impl PartiallyDelegated {
             let mut fused_receiver = notify_receiver.fuse();
```
jstuczyn, line 104: as a side note, I think it's now possible to remove this stupid fused streams since this method is no longer built under `wasm`

clients/client-core/src/client/cover_traffic_stream.rs
octol: When `.recv()` fires then it should toggle `shutdown.is_shutdown()`

clients/client-core/src/client/mix_traffic.rs
octol: It does happen automatically :)

PR #1619, opened 2022-09-13
NE: Display raw bond instead of self %

### LINKED ISSUE #2260: NE - change "Self" with raw Bond in below screen:
The title should be `Bond`, meaning the values need to be in NYM instead of a percentage.
And add this tooltip: "Node operator's share of stake"
Screenshot 2022-09-13 at 12 20 38

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2260

I added also `display none`  to the class `.MuiDataGrid-iconButtonContainer`, to avoid this grey circle we have hovering the right side of each column header: 

image

It's allows me to fit a bit better some columns width.


doums (approved):
LGTM

PR #1628, opened 2022-09-15
NE & Wallet: NE TableToolbar ui and Wallet new display when no delegations

### LINKED ISSUE #2264: Wallet: create delegation screen first step & NE: table toolbar ui update
This ticket require changes on the Network Explorer and the Wallet.

First step delegation screen - before any delegation user should see this screen with link to NYM explorer and explanation why they should use filters
image

_**Network Explorer:**_
Explorer filters currently are not visible enough

so let's change link to button in explorer and add copy to delegation screen that user for optimal search results should use advanced filters
image
mobile
image

_**Wallet:**_
Create this first step when users doesn't have delegations

image

Figma file location:
image

@benedettadavico I've been having a look with Marcin on the ui tickets done to fix Figma differences and he found one missing thing, that wasn't included on any of our tickets, and I'm going to use this one to include the change. Is the mnemonic on the signup, below screenshot is how it's looking currently on develop.

image

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2264

This ticket contains changes on NE:

New **Explorer** UI for the Table toolbar:

image
image

**Wallet**, when the user doesn't have delegations, the delegations page will display the following:

image


nym-wallet/src/pages/delegation/index.tsx
```diff
@@ -290,33 +290,71 @@ export const Delegation: FC<{ isStorybook?: boolean }> = ({ isStorybook }) => {
...
+              isLoading={isLoading}
+              items={delegations}
+              onItemActionClick={handleDelegationItemActionClick}
+            />
+          ) : isLoading ? (
+            <Box sx={{ width: '100%', display: 'flex', justifyContent: 'center' }}>
+              <CircularProgress />
+            </Box>
+          ) : (
+            <Box sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
```
doums: I think we should avoid nested ternary, they make the code harder to read (especially in JSX).
Here you could compute the rendering of this part outside the `return` call, using cleaner `if`, `else` flow etc.

fmtabbara (approved):
Approved but take a look at Pierre's comment

doums (approved):

PR #1637, opened 2022-09-19
feat(wallet): new buy nym feature

doums:
# Description

This PR introduces a first implementation of the "Buy NYM" feature (through Bity exchange interface).
https://www.figma.com/file/KiYKZnWeefSALAe0j2dAPI/Nym-Wallet?node-id=4124%3A26292&t=Ln3dSok0COUoxHYO-0

image


nym-wallet/src/components/Buy/Tutorial.tsx
```diff
@@ -0,0 +1,122 @@
...
+        </Typography>
+      </Stack>
+      <Typography fontWeight={600} variant="h6">
+        {title}
+      </Typography>
+      {typeof text === 'string' ? (
+        <Typography sx={{ color: (t) => t.palette.nym.text.muted }}>{text}</Typography>
+      ) : (
+        text
+      )}
```
fmtabbara: If `typeof text === 'string'` then you display it inside of `<Typography />`. 
Why not have text always be a `ReactNode` and pass in 
`<Typography sx={{ color: (t) => t.palette.nym.text.muted }}>{text}</Typography>` ?
Then here just use `{ text }` to display it here.

nym-wallet/src/components/CopyToClipboard.tsx
```diff
@@ -35,6 +35,7 @@ export const CopyToClipboard = ({ text = '', iconButton }: { text?: string; icon
           sx={{
             color: 'text.primary',
           }}
+          disabled={!text}
```
fmtabbara, line 38: Instead of disabling here if `!text` then you could just not render this copy component if text is undefined

fmtabbara (approved):

nym-wallet/src/components/CopyToClipboard.tsx
doums: Yes we could, but I followed what the design suggests. Let's keep it like it is for now.

gala1234 (approved):

PR #1647, opened 2022-09-22
Setup with 1 epoch and full test that skips key update

### LINKED ISSUE #1941: DKG: Simplify library to only have one epoch and not use forward secrecy

neacsu:
Closes: https://github.com/nymtech/nym/issues/1941


common/crypto/dkg/src/bte/mod.rs
```diff
@@ -243,7 +243,7 @@ impl Zeroize for Tau {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd)]
 pub struct Epoch(EpochStore);
```
jstuczyn: I think we might be actually possible to completely get rid of `Epoch` since we don't have to do any key updating

common/crypto/dkg/src/bte/mod.rs
```diff
@@ -326,6 +326,30 @@ pub fn setup() -> Params {
     }
 }
 
+pub fn setup_no_fwd_secrecy() -> Params {
```
jstuczyn: I'd just update the existing `setup`

jstuczyn (commented):
There's also bunch of things you could remove - everything that has to do with `Tau` I _think_ can go. 
The same should be true for `try_update_to_next_epoch`, `try_update_to`, etc in `DecryptionKey`  because we no longer have to update keys

neacsu:
@jstuczyn I see little benefit from removing code that might become useful to us (or even someone else) at some point, as long as it doesn't bother us with anything.

common/crypto/dkg/src/bte/mod.rs
neacsu: The reason I haven't update the existing function is because a number of unit tests rely on it to pe parametrised with `lambda_t` as 32, and those unit tests.

common/crypto/dkg/src/bte/mod.rs
neacsu: See answer in main thread i.e. I would leave the epoch implementation there, just in case

common/crypto/dkg/src/bte/encryption.rs
```diff
@@ -24,7 +24,8 @@ pub struct Ciphertexts {
 }
 
 impl Ciphertexts {
-    pub fn verify_integrity(&self, params: &Params, epoch: Epoch) -> bool {
+    pub fn verify_integrity(&self, params: &Params) -> bool {
+        let epoch = Epoch::default();
```
jstuczyn: I thought we wanted to get rid of `Epoch` completely? 
I might be misremembering (you'd have to try and run unit tests to see if it still passes), but you might be able to get rid of /simplify`Tau` conversion since you're just going to end up with 0 bits

common/crypto/dkg/src/bte/keys.rs
```diff
@@ -414,395 +292,31 @@ impl PublicKeyWithProof {
 #[zeroize(drop)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct DecryptionKey {
-    // note that the nodes are ordered from "right" to "left"
-    pub(crate) nodes: Vec<Node>,
+    pub(crate) inner: Node,
```
jstuczyn: I think we could further simplify it and remove `Node` struct (by moving everything from there into the `DecryptionKey`; I **think**, but I'm not 100% sure, you might be able to remove `Tau` from `Node` because it's always going to be constant in our case)

common/crypto/dkg/src/bte/mod.rs
```diff
@@ -305,7 +234,7 @@ pub struct Params {
 pub fn setup() -> Params {
     let f0 = hash_g2(b"f0", SETUP_DOMAIN);
 
-    let fs = (1..=MAX_EPOCHS_EXP)
+    let fs = (1..=1)
```
jstuczyn: I'm surprised clippy is not complaining here that you're iterating over a single element : )

common/crypto/dkg/src/bte/mod.rs
```diff
@@ -316,7 +245,7 @@ pub fn setup() -> Params {
     let h = hash_g2(b"h", SETUP_DOMAIN);
 
     Params {
-        lambda_t: MAX_EPOCHS_EXP,
+        lambda_t: 1,
```
jstuczyn: you might have to ask @aniampio here whether we might be able to get rid of `lambda_t` completely

common/crypto/dkg/src/bte/mod.rs
neacsu: I changed it, just in case future versions of clippy decide to kick us :D

neacsu:
@jstuczyn @aniampio As per our last discussion, I've removed the tau and epoch references, to diminish the number of unnecessary operations. Can you have another look?

PR #1660, opened 2022-10-05
fix: typescript generate types

doums:
# Description

Add and fix some TS types generation from Rust types.

### DIFF
--- a/common/cosmwasm-smart-contracts/mixnet-contract/src/interval.rs
+++ b/common/cosmwasm-smart-contracts/mixnet-contract/src/interval.rs
@@ -71,11 +71,12 @@ pub struct Interval {
     epochs_in_interval: u32,
 
     // TODO add a better TS type generation
-    #[cfg_attr(feature = "generate-ts", ts(type = "undefined"))]
+    #[cfg_attr(feature = "generate-ts", ts(type = "string"))]
     #[serde(with = "string_rfc3339_offset_date_time")]
     current_epoch_start: OffsetDateTime,
     current_epoch_id: EpochId,
-    #[cfg_attr(feature = "generate-ts", ts(type = "{ secs: number; nanos: number; }"))]
+    #[cfg_attr(feature = "generate-ts", ts(type = "string"))]
+    #[serde(with = "humantime_serde")]
     epoch_length: Duration,
     total_elapsed_epochs: EpochId,
 }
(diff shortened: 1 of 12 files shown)


fmtabbara (approved):

common/cosmwasm-smart-contracts/mixnet-contract/src/interval.rs
```diff
@@ -60,14 +60,22 @@ pub(crate) mod string_rfc3339_offset_date_time {
...
 pub struct Interval {
     id: IntervalId,
     epochs_in_interval: u32,
 
+    // TODO add a better TS type generation
+    #[cfg_attr(feature = "generate-ts", ts(type = "undefined"))]
     #[serde(with = "string_rfc3339_offset_date_time")]
     current_epoch_start: OffsetDateTime,
     current_epoch_id: EpochId,
+    #[cfg_attr(feature = "generate-ts", ts(type = "{ secs: number; nanos: number; }"))]
```
jstuczyn: if it's only for display purposes, would it perhaps make sense to use `#[serde(with = "humantime_serde")]` here instead?

common/cosmwasm-smart-contracts/mixnet-contract/src/interval.rs
```diff
@@ -60,14 +60,22 @@ pub(crate) mod string_rfc3339_offset_date_time {
...
+    feature = "generate-ts",
+    ts(export_to = "ts-packages/types/src/types/rust/Interval.ts")
+)]
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
 pub struct Interval {
     id: IntervalId,
     epochs_in_interval: u32,
 
+    // TODO add a better TS type generation
+    #[cfg_attr(feature = "generate-ts", ts(type = "undefined"))]
```
jstuczyn: is `undefined` the correct way of handling it? if we don't care about that field, how about explicitly ignoring it? (or maybe doing some string conversion?)

ts-packages/types/src/types/rust/RewardEstimate.ts
```diff
@@ -1,6 +1,6 @@
-export type RewardEstimate = {
+export interface RewardEstimate {
```
jstuczyn, line 1: Yeah, I dont think you want to be changing it here, because whenever somebody regenerates the types it will get overwritten

doums: yes the change here is auto generation overriding previous handwritten type

PR #1665, opened 2022-10-06
Wallet: Bonded node settings

### LINKED ISSUE #2322: Update of bonding flow part 2 - settings
**What is new**

we are adding full page container with node settings - currently we have only modal
image

**Settings structure**

General
   - Info
   - Parameters
  
Unbond
image

What is new here is that we are adding second left side menu to general category. and inside we have info and parameters.

**Unbond**
image

because we want to prevent situation when somebody is unbonding his node by mistake we have this safe mechanism to type in input UNBOND and confirm this with button.
image

after that we have our to standard modals with details and confirmation.

figma file: https://www.figma.com/file/KiYKZnWeefSALAe0j2dAPI/Nym-Wallet?node-id=3239%3A50783

- [x] design
- [ ] copy
- [x] development
- [x] qa

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2322


nym-wallet/src/pages/bonding/node-settings/NodeSettings.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const theme = useTheme();
+  const { network } = useContext(AppContext);
+  const { bondedNode, unbond, isLoading } = useBondingContext();
+  const navigate = useNavigate();
+  const query = useQuery();
+  const queryTab = query.get('tab');
+  const tabs = getTabs();
+
+  const [confirmationDetails, setConfirmationDetails] = useState<ConfirmationDetailProps>();
+  const [value, setValue] = React.useState(queryTab === 'unbond' ? 1 : 0);
```
fmtabbara: I think you might be able to lose a lot of this logic if you change your `nodeSettingsNav` enum to an array.
`['General',  'Unbond']`.

Then this becomes `const [value, setValue] = React.useState(nodeSettingsNav[0])`
and you pass the array of tabs as the value of the `tabs` prop on the Tabs component

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+  const theme = useTheme();
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting, isDirty, isValid },
+  } = useForm({
+    resolver: yupResolver(bondedInfoParametersValidationSchema),
+    mode: 'onChange',
```
fmtabbara, line 26: I'm not sure if you need this. See if it works without.

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/ParametersSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting, isDirty, isValid },
+  } = useForm({
+    resolver: yupResolver(bondedNodeParametersValidationSchema),
+    mode: 'onChange',
+    defaultValues:
+      bondedNode.type === 'mixnode'
+        ? {
+            operatorCost: bondedNode.bond.amount,
```
fmtabbara: Not sure if this is right??

fmtabbara (changes requested):

nym-wallet/src/pages/bonding/node-settings/NodeSettings.tsx
```diff
@@ -0,0 +1,147 @@
...
+        }
+        Action={
+          <Button
+            size="small"
+            sx={{
+              color: 'text.primary',
+            }}
+            onClick={() => navigate('/bonding')}
+            startIcon={<Close />}
+          ></Button>
```
fmtabbara: Change this button to an IconButton component and add <CloseIcon /> from mui icons as a child

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx
gala1234: @fmtabbara I need it because if the values don't change the button should be disable, if not, and the user send the same parameters already set-up, and is received an error from the api

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/ParametersSettings.tsx
gala1234: @fmtabbara still we are waiting for the gateways settings designs, probably they will have less params. Now, I'm not sure neither, if we are going to deploy this changes before having the gateways setting and we can have bonded gateways, then, I need to display settings only for mixnodes.

nym-wallet/src/pages/bonding/node-settings/NodeSettings.tsx
gala1234: @fmtabbara I was afraid of leaving the logic depending on the order of the items on this array, this is why I used the `enum`.

nym-wallet/src/context/bonding.tsx
```diff
@@ -36,6 +36,7 @@ import { attachDefaultOperatingCost, toPercentFloatString, toPercentIntegerStrin
 
 // TODO add relevant data
 export type TBondedMixnode = {
+  type: 'mixnode';
```
doums: Adding this is a bit of an antipattern because adding a property of value `mixnode` to a type actually describing a mixnode is redundant.

With TS we should rely on typing instead. Using [type-predicates ](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

For instance, there are already convenient functions to check the type of bonded node (using type predicates), check for `isMixnode` and `isGateway` :)

So instead of writing this kind of check
```
bondedNode.type === 'mixnode'
```
you would use

```
isMixnode(bondedNode)
```

No need to add a `type` property :)

nym-wallet/src/context/bonding.tsx
```diff
@@ -46,17 +47,27 @@ export type TBondedMixnode = {
...
   host: string;
+  httpApiPort: number;
+  mixPort: number;
+  verlocPort: number;
+  version: string;
+  operatorCost?: string;
 };
 
 export interface TBondedGateway {
+  type: 'gateway';
```
doums: Same as mixnode type

nym-wallet/src/context/bonding.tsx
fmtabbara: Glad we can make use of those type guards :smiley:

gala1234: @doums I like it, indeed I felt as I was doing something redundant

nym-wallet/src/pages/bonding/node-settings/node-settings.constant.tsx
```diff
@@ -0,0 +1,7 @@
+enum Options {
+  General,
+  Unbond,
+}
+type NavItem = keyof typeof Options;
+
+export const nodeSettingsNav: NavItem[] = ['General', 'Unbond'];
```
doums: Instead of using enum (which are JS thing), I always recommend to use TS types instead.

So here a more "TS" way to do this is using its typing system again :p
```
export const navItems = ['General', 'Unbond'] as const;

export type NodeSettingsNav = typeof navItems[number]; // type NodeSettingsNav = 'General' | 'Unbond';
```
With that, instead of having some runtime values, you have a real type you can rely on
```type NodeSettingsNav = 'General' | 'Unbond';```

Okay now we need to do a bit of refactoring to embrace it

First you need to refactor a bit `nym-wallet/src/components/Tabs.tsx`, change the signature of the component to accept a `readonly` array of string (this is what we want as we expect tabs to stay the same). Plus, we need to change the type of the current selected tab to the value of the tab itself and not an index (which can be unreliable):
```
type Props = {
  tabs: readonly string[];
  selectedTab: string;
  disabled?: boolean;
  onChange?: (event: React.SyntheticEvent, tab: string) => void;
  disableActiveTabHighlight?: boolean;
  tabSx?: SxProps;
  tabIndicatorStyles?: {};
};

export const Tabs = ({
  tabs,
  selectedTab,
  disabled,
  disableActiveTabHighlight,
  onChange,
  tabSx,
  tabIndicatorStyles,
}: Props) => (
```
Note that I voluntarily dropped the use of `React.FC` because it's bad to use it (why? check [this](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/))

Dont forget to map the corresponding value for each Tab:
```
      <Tab key={tabName} label={tabName} sx={{ textTransform: 'capitalize' }} value={tabName} disabled={disabled} />
```

Now, you can simplify the logic in `nym-wallet/src/pages/bonding/node-settings/NodeSettings.tsx`
```
//...
import { navItems, NodeSettingsNav } from './node-settings.constant';
//...

  const [value, setValue] = React.useState<NodeSettingsNav>('General');

  const handleChange = (event: React.SyntheticEvent, tab: string) => {
    setValue(tab as NodeSettingsNav);
  };

  useEffect(() => {
    if (location.state === 'unbond') {
      setValue('Unbond');
    }
  }, [location]);

// ...
              <Tabs
                tabs={navItems}
// ...
        <Divider />
        {value === 'General' && bondedNode && <NodeGeneralSettings bondedNode={bondedNode} />}
        {value === 'Unbond' && bondedNode && (

```
As you can see, the code is a bit more clean, but most importantly, our state is typed to be a union type of `General` or `Unbond`. So you can't mess up with wrong mapping between value and index, for example.

Finally, don't forget to refactor as well `nym-wallet/src/components/Bonding/modals/NodeSettingsModal.tsx`
```
      <Tabs tabs={['System variables']} selectedTab="System variables" disableActiveTabHighlight />
```

That's it :)

Please all my above note are only suggestion, the code you wrote works well and is perfectly functional. I just try to share a more "idiomatic" way to work with TS :) Which lead to cleaner code base in the long run.

File: nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx [hunk 8]
```diff
@@ -0,0 +1,246 @@
...
+        mix_port: mixPort,
+        verloc_port: verlocPort,
+        http_api_port: httpApiPort,
+        version,
+      };
+      try {
+        await updateMixnodeConfig(MixNodeConfigParams);
+        setOpenConfirmationModal(true);
+      } catch (error) {
+        console.error(error);
```
doums: you should use `Console.error` instead (to have a trace of the error in tauri logs as well)

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/ParametersSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+          sx={{
+            width: 1,
+            px: 2,
+            borderRadius: 0,
+            bgcolor: 'background.default',
+            color: (theme) => theme.palette.nym.nymWallet.text.blue,
+            '& .MuiAlert-icon': { color: (theme) => theme.palette.nym.nymWallet.text.blue, mr: 1 },
+          }}
+        >
+          <strong>Profit margin can be changed once a month, your changes will be applied in the next interval</strong>
```
doums: We should avoid to use "raw" HTML tags like `strong` `span` `div` etc. You'd might use the MUI equivalent ones to keep consistency.

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/ParametersSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+                  {...register('profitMargin')}
+                  name="profitMargin"
+                  label="Profit margin"
+                  fullWidth
+                  error={!!errors.profitMargin}
+                  helperText={errors.profitMargin?.message}
+                  InputProps={{
+                    endAdornment: (
+                      <InputAdornment position="end">
+                        <span>%</span>
```
doums: same as above

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/ParametersSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+        interval_operating_cost: {
+          denom: bondedNode.bond.denom,
+          amount: data.operatorCost.toString(),
+        },
+      };
+      try {
+        await updateMixnodeCostParams(MixNodeCostParams);
+        setOpenConfirmationModal(true);
+      } catch (error) {
+        console.error(error);
```
doums: You'd might use `Console` instead

doums (approved):

PR #1670, opened 2022-10-09
Node family management

### LINKED ISSUE #1964: "Family" feature for node families + layers

durch:
# Description

Closes: https://github.com/nymtech/nym/issues/1964

I've also removed Cargo.lock from the contracts workspace, it only creates merge conflicts, don't think it adds any value.

# Checklist:


contracts/mixnet/src/families/storage.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<Family>> + '_> {
+        let v: Vec<&dyn Index<Family>> = vec![&self.label];
+        Box::new(v.into_iter())
+    }
+}
+
+// storage access function.
+pub fn families<'a>() -> IndexedMap<'a, String, Family, FamilyIndex<'a>> {
+    let indexes = FamilyIndex {
+        label: UniqueIndex::new(|d| d.label.clone(), "faml"),
```
jstuczyn: could you move storage keys to `constants.rs` to keep it all consistent?

common/cosmwasm-smart-contracts/mixnet-contract/src/msg.rs
```diff
@@ -273,6 +334,14 @@ impl ExecuteMsg {
...
 pub enum QueryMsg {
+    // families
+    GetFamilyByHead {
+        head: String,
+        proxy: Option<String>,
+    },
+    GetFamilyByLabel {
+        label: String,
+    },
     // state/sys-params-related
```
jstuczyn, line 379: I think it might also be useful to be able to query for all families

contracts/mixnet/src/families/storage.rs
```diff
@@ -0,0 +1,174 @@
...
+    };
+    IndexedMap::new("fam", indexes)
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Family {
```
jstuczyn: since this guy is being returned by our queries, it should go to the common types (plus it'd probably need some `ts_rs` love for our frontenders)

contracts/mixnet/src/families/storage.rs
```diff
@@ -0,0 +1,174 @@
...
+    head: FamilyHead,
+    proxy: Option<Addr>,
+    label: String,
+    members: HashSet<String>,
```
jstuczyn: is cosmwasm fine with `HashSet`? (also, my PTSD kicking in here: what if a family has a shit-ton of members, don't we want to potentially store them in a separate `Map`)?

contracts/mixnet/src/families/storage.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn storage_key(&self) -> String {
+        storage_key(self.head_identity(), self.proxy.as_ref())
+    }
+
+    pub fn create(&self, store: &mut dyn Storage) -> Result<(), MixnetContractError> {
+        match families().save(store, self.storage_key(), self) {
+            Ok(()) => Ok(()),
+            Err(e) => match &e {
+                StdError::GenericErr { msg } => {
+                    if msg.starts_with("Violates unique constraint") {
```
jstuczyn: can't we just lookup for whether family with given label already exists and get rid of this error message content comparison?

contracts/mixnet/src/families/transactions.rs
```diff
@@ -0,0 +1,391 @@
...
+    validate_node_identity_signature(
+        deps.as_ref(),
+        owner,
+        &owner_signature,
+        existing_bond.identity(),
+    )?;
+
+    let family_head = FamilyHead::new(existing_bond.identity());
+    Family::new(family_head, proxy, label, deps.storage)?;
+    Ok(Response::default())
```
jstuczyn, line 74: perhaps some basic events would be useful here?

contracts/mixnet/src/families/transactions.rs
```diff
@@ -0,0 +1,391 @@
...
+        .owner
+        .item(deps.storage, owner.clone())?
+        .ok_or(MixnetContractError::NoAssociatedMixNodeBond {
+            owner: owner.clone(),
+        })?
+        .1;
+
+    ensure_bonded(&existing_bond)?;
+
+    validate_family_signature(
```
jstuczyn, line 197: isn't signature verification a bit redundant here? we know the sender owns the mixnode in question and he must had provided a signature to join the family

contracts/mixnet/src/families/transactions.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+fn _try_head_kick_member(
+    deps: DepsMut,
+    owner: &Addr,
+    owner_signature: String,
+    member: IdentityKeyRef<'_>,
+    proxy: Option<Addr>,
+) -> Result<Response, MixnetContractError> {
+    let existing_bond = crate::mixnodes::storage::mixnode_bonds()
```
jstuczyn: thinking about it, don't we need to check whether the existing bond actually has the same identity as the family head? (i.e. what would have happened if you created family under identity `X`, unbonded and then rebonded with identity `Y`?)

contracts/mixnet/src/families/transactions.rs
```diff
@@ -0,0 +1,391 @@
...
+        .owner
+        .item(deps.storage, owner.clone())?
+        .ok_or(MixnetContractError::NoAssociatedMixNodeBond {
+            owner: owner.clone(),
+        })?
+        .1;
+
+    ensure_bonded(&existing_bond)?;
+
+    validate_node_identity_signature(
```
jstuczyn: again, signature might be possibly redundant

contracts/mixnet/src/families/storage.rs
durch: Seems to be fine, I was thinking about a separate Map, not sure

contracts/mixnet/src/families/storage.rs
durch: We could, but we get the check for free here, but its only an implementation detail

contracts/mixnet/src/families/transactions.rs
durch: Yeah, probably true, but I think we lose nothing by verifying it, trust but verify and all that, and who knows maybe we change something down the lines, and this proves useful, sort of a better safe then sorry thing

contracts/mixnet/src/families/transactions.rs
durch: Great call, we should totally check

durch: Although, looking at the implementation, if the identity changed, the head is no longer head, so it can't kick anyone out

PR #1671, opened 2022-10-10
Introduced a client-configurable option to force it to use extended packet size

jstuczyn:
# Checklist:


octol (commented):
There is a warning when receiving an extended packet saying it was unexpected, should we keep that?

jstuczyn:
> There is a warning when receiving an extended packet saying it was unexpected, should we keep that?

good point - we probably should remove it

PR #1687, opened 2022-10-17
Add per account pledge caps

### LINKED ISSUE #1989: Vesting contract - change 100k cap to 10%

durch:
# Description

Closes: https://github.com/nymtech/nym/issues/1989

# Checklist:

### DIFF
--- a/common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
+++ b/common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 use cosmwasm_std::{Addr, Coin, Timestamp, Uint128};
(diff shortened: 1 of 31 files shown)


common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
```diff
@@ -42,6 +44,48 @@ impl PledgeData {
     }
 }
 
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub enum PledgeCap {
+    Percent(u8),
```
jstuczyn: I'd just internally keep it as `Decimal` so if somebody has an insane idea of "change the pledge cap to 7.5%" we wouldn't need to do another rework; plus I'd assume all of maths would be easier

common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
```diff
@@ -42,6 +44,48 @@ impl PledgeData {
...
+    Absolute(Uint128),
+}
+
+impl FromStr for PledgeCap {
+    type Err = String;
+
+    fn from_str(cap: &str) -> Result<Self, Self::Err> {
+        match cap.parse::<f32>() {
```
jstuczyn: wouldn't the validator complain about that float operation here?

common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
```diff
@@ -42,6 +44,48 @@ impl PledgeData {
...
+impl Default for PledgeCap {
+    fn default() -> Self {
+        PledgeCap::Absolute(Uint128::from(100_000_000_000u128))
+    }
+}
+
+impl PledgeCap {
+    pub fn valid(&self) -> bool {
+        match &self {
+            PledgeCap::Percent(p) => *p > 0 && *p <= 100,
```
jstuczyn: If you're locking it between 0-100%, we already have a `Percent` type with that constraint (https://github.com/nymtech/nym/blob/develop/common/cosmwasm-smart-contracts/mixnet-contract/src/types.rs#L31) though it'd probably have to be moved outside mixnet-contract specific types then

contracts/vesting/src/vesting/account/mod.rs
```diff
@@ -32,6 +32,7 @@ pub struct Account {
     pub periods: Vec<VestingPeriod>,
     pub coin: Coin,
     storage_key: u32,
+    pub pledge_cap: Option<PledgeCap>,
```
jstuczyn, line 36: You need to put `#[serde(default)]` here so that it wouldn't blow up on trying to deserialize old accounts without defined pledge_cap

common/cosmwasm-smart-contracts/vesting-contract/src/lib.rs
durch: Crap, good call, yeah it will probably, altough that can never be executed in wasm, but yeah, moving to Decimal will help that as well

PR #1690, opened 2022-10-18
Display interval timings

### LINKED ISSUE #2235: Wallet - add notion of where in the epoch and interval we are in the node setting.
image

Additional info about when will be the Interval.

After change additional info below inputs about when change will be applied

fmtabbara:
# Description

Display the next epoch time and the next interval date and time.

If the user has updated operating cost and/or profit margin percentage, display the latest requested update.

Closes: https://github.com/nymtech/nym/issues/2235


nym-wallet/src/pages/bonding/types.ts
```diff
@@ -68,3 +68,8 @@ export interface BondState {
   bondStatus: BondStatus;
   error?: string | null;
 }
+
+export interface ChangeMixCostParams {
+  mix_id: number;
+  new_costs: MixNodeCostParams;
```
doums, line 75: Not possible to use camelCase for field naming?

PR #1693, opened 2022-10-20
Display routing scores for bonded gateway

fmtabbara:
# Description

Dsiplay current and average routing scored for bonded gateways

Closes: https://github.com/nymtech/nym/issues/2228

### DIFF
--- a/nym-wallet/src/context/bonding.tsx
+++ b/nym-wallet/src/context/bonding.tsx
@@ -79,7 +79,7 @@ export interface TBondedGateway {
   mixPort: number;
   verlocPort: number;
   version: string;
-  routingScore: {
+  routingScore?: {
     current: number;
     average: number;
   };
@@ -226,6 +226,7 @@ export const BondingContextProvider = ({ children }: { children?: React.ReactNod
       return { current: report.most_recent, average: report.last_day };
     } catch (e) {
       Console.error(e);
+      return undefined;
     }
   };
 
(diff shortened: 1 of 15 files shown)


nym-wallet/src/context/bonding.tsx
```diff
@@ -282,14 +296,15 @@ export const BondingContextProvider = ({ children }: { children?: React.ReactNod
...
-
+          const routingScore = await getGatewayReportDetails(data.gateway.identity_key);
           setBondedNode({
             name: nodeDescription?.name,
             identityKey: data.gateway.identity_key,
             ip: data.gateway.host,
             location: data.gateway.location,
             bond: data.pledge_amount,
             proxy: data.proxy,
+            routingScore,
```
doums, line 308: `routingScore` is not typed to be nullable (aka `| undefined`) in `TBondedGateway`. But if  `getGatewayReportDetails` fails for some reason, `routingScore` will be `undefined`, and potentially cause some client crash since it's an object with properties, and accessing a property on `undefined` value will cause a runtime crash.

nym-wallet/src/context/bonding.tsx
```diff
@@ -78,6 +79,10 @@ export interface TBondedGateway {
   mixPort: number;
   verlocPort: number;
   version: string;
+  routingScore: {
+    current: number;
+    average: number;
+  };
```
doums, line 85: As the request to fetch routing data could potentially crash/or returns an error for some reasons, a better typing would be to make it nullable :  `routingScore?: ...`

nym-wallet/src/context/bonding.tsx
fmtabbara: Good point. Will update this.

PR #1694, opened 2022-10-20
Add two more sphinx extended packet sizes (8kb and 16kb)

### LINKED ISSUE #1952: Add additional packet sizes (8kb and 16kb) to easier explore parameter space

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/1952

Add two more extended sphinx packet sizes, 8kb and 16kb, in addition to the already existing 32kb one. 

These are strictly for performance _testing only_, and should **not be used** by end users as privacy guarantees would be lower than the default packet sizes.

Update configuration file option `use_extended_packet_size` to take a string (`extended8`, `extended16`, `extended32`)
```
use_extended_packet_size = 'extended8'
```

# Checklist:

### DIFF
--- a/clients/client-core/src/client/cover_traffic_stream.rs
+++ b/clients/client-core/src/client/cover_traffic_stream.rs
@@ -14,6 +14,7 @@ use nymsphinx::params::PacketSize;
 use nymsphinx::utils::sample_poisson_duration;
 use rand::{rngs::OsRng, CryptoRng, Rng};
 use std::pin::Pin;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -140,6 +141,19 @@ impl LoopCoverTrafficStream<OsRng> {
 
     pub fn set_custom_packet_size(&mut self, packet_size: PacketSize) {
         self.packet_size = packet_size;
+    }
+
+    pub fn try_set_extended_packet_size(&mut self, size: &str) -> bool {
+        if let Some(packet_size) = PacketSize::from_str(size)
+            .ok()
+            .and_then(PacketSize::as_extended_size)
+        {
+            log::debug!("Setting extended packet size: {:?}", packet_size);
+            self.set_custom_packet_size(packet_size);
+            true
+        } else {
+            false
+        }
     }
 
     async fn on_new_message(&mut self) {
--- a/clients/client-core/src/client/cover_traffic_stream.rs
+++ b/clients/client-core/src/client/cover_traffic_stream.rs
@@ -14,7 +14,6 @@ use nymsphinx::params::PacketSize;
 use nymsphinx::utils::sample_poisson_duration;
 use rand::{rngs::OsRng, CryptoRng, Rng};
 use std::pin::Pin;
-use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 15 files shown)


clients/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -142,6 +143,19 @@ impl LoopCoverTrafficStream<OsRng> {
         self.packet_size = packet_size;
     }
 
+    pub fn try_set_extended_packet_size(&mut self, size: &str) -> bool {
```
jstuczyn: would it be perhaps cleaner to pass an explicit enum here so that it couldn't fail?

clients/client-core/src/client/real_messages_control/mod.rs
```diff
@@ -90,6 +91,19 @@ impl Config {
     pub fn set_custom_packet_size(&mut self, packet_size: PacketSize) {
         self.packet_size = packet_size;
     }
+
+    pub fn try_set_extended_packet_size(&mut self, size: &str) -> bool {
```
jstuczyn: same note as with loop cover traffic

clients/client-core/src/config/mod.rs
```diff
@@ -247,8 +247,8 @@ impl<T: NymConfig> Config<T> {
         self.debug.disable_main_poisson_packet_distribution
     }
 
-    pub fn get_use_extended_packet_size(&self) -> bool {
-        self.debug.use_extended_packet_size
+    pub fn get_use_extended_packet_size(&self) -> Option<String> {
```
jstuczyn: I'd suggest it should then return an enum : )
You should take a look at one of the enums in our wallet types (unfortunately I don't remember the name); it has some derived trait that allows for serde from/to string

clients/native/src/client/mod.rs
```diff
@@ -103,8 +102,13 @@ impl NymClient {
             topology_accessor,
         );
 
-        if self.config.get_base().get_use_extended_packet_size() {
-            stream.set_custom_packet_size(PacketSize::ExtendedPacket)
+        if let Some(size) = self.config.get_base().get_use_extended_packet_size() {
```
jstuczyn, line 105: I'd argue that if user provided any invalid value in their config we should blow up

clients/webassembly/src/client/config.rs
```diff
@@ -107,6 +107,11 @@ pub struct Debug {
 
 impl From<Debug> for ConfigDebug {
     fn from(debug: Debug) -> Self {
+        // For now we just always use the (older) 32kb extended size
+        let use_extended_packet_size = debug
+            .use_extended_packet_size
+            .then(|| String::from("extended32"));
```
jstuczyn: IF we're going to go with strings as opposed to enums, `"extended32"` should be exported via some constant in case it was ever changed

common/nymsphinx/params/src/packet_sizes.rs
```diff
@@ -30,7 +36,28 @@ pub enum PacketSize {
     AckPacket = 2,
 
     // for example for streaming fast and furious in uncompressed 10bit 4K HDR quality
-    ExtendedPacket = 3,
+    ExtendedPacket32 = 3,
+
+    // for example for streaming fast and furious in heavily compressed lossy RealPlayer quality
+    ExtendedPacket8 = 4,
+
+    // for example for streaming fast and furious in compressed XviD quality
```
jstuczyn, line 44: hehe, I like those comments : )

octol:
Reworked using enums, this is much better :)

PR #1695, opened 2022-10-20
NE: Adding Gateways details page

### LINKED ISSUE #2236: NE - gateway profile page to show liveness, uptime history and other stats ie available storage
Here is the [API endpoint](https://validator.nymtech.net/api/v1/status/gateway/14E6nscnRag4Ch1xiZjkRE77Bs8Nu5jhK52uTzD8wh3B/report) for the data. Use `most_recent` from the API for "Routing score" and `last_day` for "Avg. score". If `last_day` was not available use `last_hour` instead
And here is the [API with measurement history](https://validator.nymtech.net/api/v1/status/gateway/14E6nscnRag4Ch1xiZjkRE77Bs8Nu5jhK52uTzD8wh3B/history) for Routing Score timeline chart

image

Figma file: Nym explorer, designs under yellow stripe(gateways)

gala1234:
# Description

Closes: https://github.com/nymtech/nym/issues/2236

A quick glance on the new Gateways details page:
image

On this PR is also fixed a currently bug we have on the gateways list, where we can't go to the following pages even when we have more items than the pagination limit, now will appear the following (if it's needed): 
image


explorer/src/components/DetailTable.tsx
```diff
@@ -43,42 +46,60 @@ function formatCellValues(val: string | number, field: string) {
...
-    </Table>
-  </TableContainer>
-);
+        </TableHead>
+        <TableBody>
+          {rows.map((eachRow) => (
+            <TableRow key={eachRow.id} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
+              {columnsData?.map((_, index) => (
+                <TableCell
+                  key={_.title}
```
doums: We should give a name to that `_` variable as we are using it

explorer/src/context/hooks.ts
```diff
@@ -1,30 +1,46 @@
...
-  fn: Function,
+  id: string,
+  fn: (argId: string) => Promise<T>,
   errorMessage: string,
-): [ApiState<T>, WrappedApiFn<T>] => {
-  const [value, setValue] = React.useState<ApiState<T>>();
-  const wrappedFn = React.useCallback(async () => {
+): [ApiState<T> | undefined, () => Promise<ApiState<T>>, () => void] => {
+  // stores the state
+  const [value, setValue] = React.useState<ApiState<T> | undefined>();
```
fmtabbara: I'm not sure if you need the `| undefined` here. If the state begins as undefined I think TS can infer it.

explorer/src/context/hooks.ts
```diff
@@ -1,30 +1,46 @@
...
+
+      // delegate to user function to get data and set if successful
+      const data = await fn(id);
+      const newValue: ApiState<T> = {
         isLoading: false,
         data,
-      });
-      return data;
+      };
+      setValue(newValue);
```
fmtabbara, line 34: I think this could just be
```
setValue({
  isLoading: false,
  data
})
```

explorer/src/components/Gateways.ts
```diff
@@ -21,3 +28,21 @@ export function gatewayToGridRow(arrayOfGateways: GatewayResponse): GatewayRowTy
...
+    owner: gateway.owner,
+    identity_key: gateway.gateway.identity_key || '',
+    location: gateway?.gateway?.location || '',
+    bond: gateway.pledge_amount.amount || 0,
+    host: gateway.gateway.host || '',
+    clients_port: gateway.gateway.clients_port || 0,
+    mix_port: gateway.gateway.mix_port || 0,
+    routing_score: `${report.most_recent}%`,
+    avg_uptime: `${report.last_day || report.last_hour}%`,
+  };
```
doums, line 47: We should conform to the JS/TS naming convention whenever we can (ofc in JS/TS side), so when we are creating our types, use `camelCase` for field/variable declaration. Even if it maps to external data that uses its own different convention (`snake_case` in this case because it comes from Rust).

explorer/src/context/hooks.ts
```diff
@@ -1,30 +1,46 @@
...
         isLoading: false,
-      });
-      return undefined;
+      };
+      setValue(newValue);
+      return newValue;
     }
   }, [setValue, fn]);
-  return [value || { isLoading: true }, wrappedFn];
+  return [value || { isLoading: true }, wrappedFetchFn, clearValueFn];
```
fmtabbara: I'm not sure about this `value || {isLoading: true}`
If the value is always undefined the the loading state will always be true. I think the loading state should be determined by the value held in the state?

explorer/src/pages/GatewayDetail/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  React.useEffect(() => {
+    if (uptimeReport?.data && selectedGateway) {
+      setEnrichGateway(gatewayEnrichedToGridRow(selectedGateway, uptimeReport.data));
+    }
+  }, [uptimeReport, selectedGateway]);
+
+  React.useEffect(() => {
+    if (enrichGateway) {
+      setStatus([enrichGateway?.mix_port, enrichGateway?.clients_port]);
```
fmtabbara: shouldnt need the optional chaining operator (`?`) here as we have determined that `enrichGateway` exists through if statement

explorer/src/context/gateway.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface GatewayContextProviderProps {
+  gatewayIdentityKey: string;
+}
+
+/**
+ * Provides a state context for a gateway by identity
+ * @param gatewayIdentityKey   The identity key of the gateway
+ */
+export const GatewayContextProvider: React.FC<GatewayContextProviderProps> = ({ gatewayIdentityKey, children }) => {
```
doums: Please, in general, avoid using  `React.FC` -> https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components

explorer/src/context/gateway.tsx
```diff
@@ -0,0 +1,62 @@
...
+    () => ({
+      uptimeReport,
+      uptimeStory,
+    }),
+    [
+      {
+        uptimeReport,
+        uptimeStory,
+      },
+    ],
```
doums: You don't need to wrap the `memo` dependencies inside an object. You should list them directly instead.
```
    [uptimeReport,  uptimeStory],
```

explorer/src/context/hooks.ts
doums: to clean a bit the code you could initiate `value` during `useState` declaration
```ts
  // stores the state
  const [value, setValue] = React.useState<ApiState<T> | undefined>({ isLoading: true });
  ```
and
```ts
  return [value, wrappedFetchFn, clearValueFn];
```

explorer/src/components/Gateways.ts
gala1234: @doums I was in doubt, as all fields on the explorer has this convention instead the camelCase. I'd say that here is an "explorer convention". What do you think, should i make here the exception using camelCase?

I was checking and already started using the camelCase for the economic stats, I'm changing these parameters to camelCase too

explorer/src/context/hooks.ts
gala1234: Confirmed, I don't need it : )

explorer/src/context/hooks.ts
gala1234: you are right

gala1234: I moved the isLoading set, before we start fetch the data, here wasn't the correct place to do it

explorer/src/context/hooks.ts
gala1234: @fmtabbara, do you mean by not returning the newValue object after set it? the function is expected to have a return value and the logic doesn't work if it doesn't.

explorer/src/context/hooks.ts
gala1234: Completely agree, after reading Fouad's comment I realised that it was the correct form, changed. Thank you

explorer/src/context/hooks.ts
fmtabbara: I just thought instead of saving the object into a variable and setting that variable in state, just set the object directly to state?

fmtabbara: I see what you mean! Ignore me.

explorer/src/context/hooks.ts
```diff
@@ -1,30 +1,47 @@
...
+      const newValue: ApiState<T> = {
         error: error instanceof Error ? error : new Error(errorMessage),
         isLoading: false,
-      });
-      return undefined;
+      };
+      setValue(newValue);
+      return newValue;
     }
   }, [setValue, fn]);
```
doums: Not having `id` and `errorMessage` as dependencies to that `useCallback` hook is probably an error. You should add them too
```
[setValue, fn, id, errorMessage]
```

explorer/src/pages/GatewayDetail/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+    title: 'Owner',
+    headerAlign: 'left',
+    flex: 1,
+  },
+];
+
+/**
+ * Shows gateway details
+ */
+const PageGatewayDetailsWithState: React.FC<{ selectedGateway: GatewayResponseItem | undefined }> = ({
```
doums: `React.FC`

explorer/src/pages/GatewayDetail/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+      <Grid container spacing={2} mt={0}>
+        <Grid item xs={12} md={4}>
+          {status && (
+            <ContentCard title="Gateway Status">
+              <TwoColSmallTable
+                loading={false}
+                keys={['Mix port', 'Client WS API Port']}
+                values={status.map((each) => each)}
+                icons={status.map((elem) => !!elem) || [false, false]}
```
doums: Im not sure about that `|| [false, false]` ?
It will never be the case right? Cause you provide the function `(elem) => !!elem)` anyway.
Not sure if I understand it to be honest.

explorer/src/pages/GatewayDetail/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+const PageGatewayDetailGuard: React.FC = () => {
+  const [selectedGateway, setSelectedGateway] = React.useState<GatewayResponseItem | undefined>();
+  const { gateways } = useMainContext();
+  const { id } = useParams<{ id: string | undefined }>();
+
+  React.useEffect(() => {
+    if (gateways?.data) {
+      setSelectedGateway(gateways.data.find((gateway) => gateway.gateway.identity_key === id));
+    }
+  }, [gateways]);
```
doums: You probably want to add `id` in the `useEffect` deps.

explorer/src/pages/GatewayDetail/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  if (gateways?.isLoading) {
+    return <CircularProgress />;
+  }
+
+  if (gateways?.error) {
+    // eslint-disable-next-line no-console
+    console.error(gateways?.error);
```
doums, line 143: Maybe use `Console` module instead.

explorer/src/pages/GatewayDetail/index.tsx
gala1234: Yep, you're right

explorer/src/pages/GatewayDetail/index.tsx
doums: VERY SILLY FEEDBACK xD I tought during a sec it was a wallet PR..

doums (approved):
LGTM

doums:
I'll recreate the PR to merge it into release v1.1 branch, great work Gala ;)

PR #1698, opened 2022-10-20
fix up  operating cost

fmtabbara:
# Description

Operating cost is now available directly through the the "getMixnodeDetails" response. I've updated the type on the frontend and removed a unnecessary extra request. I've also harmonized the operating cost fields and field validations between bonding a new node and updating an existing node.


doums (approved):
lgtm

PR #1702, opened 2022-10-21
socks5: wait to close buffer

### LINKED ISSUE #1701: [Issue] SOCKS5 proxy can stop before transferring all received data
**Issue**
Due to some reordering in the packets, a SOCKS5 proxy connection might close and stop transferring data to the user before actually sending everything.

**Expected behavior**
The SOCKS5 proxy connection closes after all data was sent.

**Details**
At the end of a communication between a socks5 proxy and a network requester, they exchange a message saying no more bytes will be sent, the connection should close. 
Under heavy load, the proxy cannot keep up, and the network requester sends the closing message 30 seconds after the last bits of data. Due to reordering, the proxy can receive this packet before the rest, and will mark the connection as closed while adding this packet to the `OrderedBuffer`. This behavior causes the next bytes batch from this buffer to close the connection, even though some data are still in the buffer, or in transit.

This issue was encountered while running multiple file downloads at the same time.
I was using a small testent with 2 * 3 mix nodes, one service provider and 20 proxy. 
  
**Proposed fix**
Upon receiving the closing packet, mark the connection as closed at the index of this message. Effectively close the connection only upon returning this packet from the buffer. 
The joined `diff` file should fix this behavior.
[sock5_reordering.txt](https://github.com/nymtech/nym/files/9839054/sock5_reordering.txt)

octol:
# Description

Part of: https://github.com/nymtech/nym/issues/1960
Closes: https://github.com/nymtech/nym/issues/1701

This is the fix proposed by @simonwicky in https://github.com/nymtech/nym/issues/1701

# Checklist:

### DIFF
--- a/common/socks5/proxy-helpers/src/connection_controller.rs
+++ b/common/socks5/proxy-helpers/src/connection_controller.rs
@@ -58,7 +58,7 @@ impl ActiveConnection {
         self.ordered_buffer.write(ordered_message);
     }
 
-    fn fn read_from_buf(&mut self) -> (Option<Vec<u8>>, u64) {
+    fn read_from_buf(&mut self) -> (Option<Vec<u8>>, u64) {
         self.ordered_buffer.read()
     }
 }
--- a/common/socks5/proxy-helpers/src/connection_controller.rs
+++ b/common/socks5/proxy-helpers/src/connection_controller.rs
@@ -38,7 +38,7 @@ pub enum ControllerCommand {
 
 struct ActiveConnection {
     is_closed: bool,
-    closed_at_index: u64,
+    closed_at_index: Option<u64>,
     connection_sender: Option<ConnectionSender>,
     ordered_buffer: OrderedMessageBuffer,
 }
@@ -53,7 +53,7 @@ impl ActiveConnection {
             }
         };
         if is_closed {
-            self.closed_at_index = ordered_message.index;
+            self.closed_at_index = Some(ordered_message.index);
         }
         self.ordered_buffer.write(ordered_message);
     }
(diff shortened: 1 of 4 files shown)


common/socks5/proxy-helpers/src/connection_controller.rs
```diff
@@ -127,21 +132,23 @@ impl Controller {
...
             } else if !is_closed {
                 error!("Tried to write an empty message to a not-closing connection. Please let us know if you see this message");
             }
-            // if messages get unordered, make sure we don't lose information about
-            // remote socket getting closed!
-            active_connection.is_closed |= is_closed;
 
             if let Some(payload) = active_connection.read_from_buf() {
+                if let Some(closed_at_index) = active_connection.closed_at_index {
+                    if payload.last_index > closed_at_index {
```
jstuczyn, line 142: is it possible for payload's `last_index` to be greater than `closed_at_index`? isn't it the case we expect the last received packet to be the closing packet?

simonwicky: Yes. When reading the `OrderedBuffer`, the index returned actually indicates the number of messages returned so far (or the index of the one yet to come), kinda like TCP SEQ and ACK numbers. 
So if for example the closing message has index 30, the `OrderedBuffer` will deliver that message along with the index 31, effectively making `last_index` bigger than `closed_at_index` when we want to close the connection.  
I've proposed it that way to avoid messing with index's offset in the `OrderedBuffer::read()` method.

octol:
I was playing around with this patch for some casual keybase use and I didn't spot any situations where the new logic closed the connection earlier than the old logic, but I did spot some cases (very few, but indeed happening) where the new logic closes the connection slightly later then before

PR #1719, opened 2022-10-28
Feature/explorer operating cost

doums:
# Description

Closes: nymtech/nym#2318


explorer-api/src/mix_node/models.rs
```diff
@@ -36,6 +36,7 @@ pub(crate) struct PrettyDetailedMixNodeBond {
     pub avg_uptime: u8,
     pub estimated_operator_apy: f64,
     pub estimated_delegators_apy: f64,
+    pub operating_cost: Coin,
```
jstuczyn, line 39: Isn't the value returned here supposed to be "our" `Coin` as opposed to cosmwasm's?

explorer-api/src/mix_nodes/models.rs
```diff
@@ -153,6 +154,11 @@ impl ThreadsafeMixNodesCache {
         let denom = &node.mixnode_details.original_pledge().denom;
         let rewarding_info = &node.mixnode_details.rewarding_details;
 
+        let mut operating_cost = rewarding_info.cost_params.interval_operating_cost.clone();
+        if operating_cost.denom == "unym" {
```
jstuczyn: isn't the if redundant? it should always be returned in either `unym` or `nym`

explorer-api/src/utils.rs
```diff
@@ -0,0 +1,19 @@
+use cosmwasm_std::Uint128;
+use mixnet_contract_common::Coin;
+
+pub fn unym_to_nym(unym: Coin, fallback_amount: Option<u128>) -> Coin {
```
jstuczyn: if we **really** need it, I think we have this functionality elsewhere, but I'd argue its redundant

explorer-api/src/mix_node/models.rs
doums: image
it's the case right?

explorer-api/src/mix_nodes/models.rs
doums: No? `operating_cost` is of type Coin, which can be either "nym" or "unym" values (as far I understand it). So the logic is to convert only when it's "unym", right?

explorer-api/src/utils.rs
doums: If there is a utility function that exists already, i'd be glad to use it, but I didnt find one, so if you can give me that I'll appreciate.

> but I'd argue its redundant

not sure to understand why

PR #1720, opened 2022-10-28
Client: multiplex connection data streams

### LINKED ISSUE #1939: Client and SOCKS5 connection multiplexing

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/1939
Part of: https://github.com/nymtech/nym/issues/1960

## Transmission lanes

Introduces the concept of `TransmissionLanes` in `client-core`.

```rust
pub enum TransmissionLane {
    General,
    Reply,
    Retransmission,
    Control,
    ConnectionId(ConnectionId),
}

```

The idea is that the `OutQueueControl` will multiplex sphinx packets according to these categories. Large data transfers will not block other connections, and smaller messages will be given priority.

## Tagging messages

The SOCKS5 client will tag sent messages accordingly, and is relatively straightforward as it owns the whole pipeline from incoming TCP connection to transmitting sphinx packets to the gateway via the `gateway-client`.

The `network-requester` however, will need to modify `ClientRequest` to append a `connection_id` to `ClientRequest::Send`, as well as add another message type, `ClientRequest::CloseConnection(..)`.

## Connection handling in the `OutQueueControl`

Because there might be a not insignificant delay between accepting TCP connections and transmitting the sphinx packets, depending on load, the `OutQueueControl` needs to be able to discard packets destined for a stale and closed connection.

## Message priority in `OutQueueControl`

Initially a very rudimentary heuristic is used to prioritize packets in the `OutQueueControl`, where small packets will be sent first.

## Testing

Confirmed in keybase that:
1. _sending_ a large video in keybase at the same time as
2. doing a ping with `curl --socks5-hostname localhost:1080 https://nymtech.net/.wellknown/connect/healthcheck.json`
does not affect each other (too much)

## TODO

- [x] Multiplex outgoing connections in `nym-socks5-client`
- [x] Multiplex outgoing connections originating from `nym-network-requester`
- [x] We need a way to signal that closed or stale connections should be removed from the out queue controller
- [x] Better packet prioritization. Let small packets through.

# Checklist:


clients/client-core/src/client/real_messages_control/acknowledgement_control/retransmission_request_listener.rs
```diff
@@ -113,10 +117,10 @@ where
 
         // send to `OutQueueControl` to eventually send to the mix network
         self.real_message_sender
-            .unbounded_send(vec![RealMessage::new(
-                prepared_fragment.mix_packet,
-                frag_id,
-            )])
+            .unbounded_send((
+                vec![RealMessage::new(prepared_fragment.mix_packet, frag_id)],
```
jstuczyn, line 121: Looking at how those are being sent in channels in multiple places, would it perhaps make sense to make `TransmissionLane` part of the `RealMessage` struct instead of sending tuples everywhere?

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -44,6 +48,15 @@ const MAX_DELAY_MULTIPLIER: u32 = 6;
 // The minium multiplier we apply to the base average Poisson delay.
 const MIN_DELAY_MULTIPLIER: u32 = 1;
 
+// When prioritizing traffic it can be used to divide lanes into new and old.
+const MSG_CONSIDERED_OLD_AFTER_SECS: u64 = 5;
```
jstuczyn: should those perhaps go to our config file so that we could mess with them more easily?

clients/client-core/src/client/real_messages_control/mod.rs
```diff
@@ -30,6 +30,14 @@ use crate::client::reply_key_storage::ReplyKeyStorage;
 mod acknowledgement_control;
 mod real_traffic_stream;
 
+/// The `RealMessagesController` through the `OutQueueControl` listens to closed connections.
+// WIP(JON) This guy is used across 1) native client, 2) socks5 client, 3) network-requester 4)
+// client-core. Currently the definition is duplicated, and the type sent is `ConnectionId` in the
+// socks5 common code, but u64 elsewhere.
+// Need to figure out a better way to share this.
+pub type ClosedConnectionSender = mpsc::UnboundedSender<u64>;
```
jstuczyn: That's a tricky one, isn't it. Maybe we need another common crate for client connections?

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -222,8 +241,199 @@ where
...
+    /// Buffer containing all real messages keyed by transmission lane.
+    received_buffer: ReceivedBuffer,
+
+    /// Incoming channel for being notified of closed connections, to avoid sending traffic
+    /// unnecessary
+    closed_connection_rx: ClosedConnectionReceiver,
+}
+
+#[derive(Default)]
+struct ReceivedBuffer {
```
jstuczyn: While the name works in that context, I'd perhaps rename it because we already have `ReceivedBuffer` and all of that for messages received from mixnet, Uhm... maybe something like `IncomingBuffer` or `PendingBuffer` or `TransmissionBuffer`? I'm not sure honestly

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -222,8 +241,199 @@ where
...
+        buffer
+            .iter()
+            .rev()
+            .map(|(k, _)| *k)
+            .take(OLDEST_LANE_SET_SIZE)
+            .copied()
+            .collect()
+    }
+
+    fn store(&mut self, lane: &TransmissionLane, real_messages: Vec<RealMessage>) {
```
jstuczyn: nitpicking here, but `self.buffer.entry(...)` would probably be cleaner

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -409,28 +633,47 @@ where
                 next_delay.as_mut().reset(next_poisson_delay);
             }
 
-            // check if we have anything immediately available
-            if let Some(real_available) = self.received_buffer.pop_front() {
-                return Poll::Ready(Some(StreamMessage::Real(Box::new(real_available))));
+            // In addition to closing connections on receiving such messages, also close
+            // connections when sufficiently stale.
+            self.received_buffer.prune_stale_connections();
```
jstuczyn: while I don't think it's a huge problem in this case, we generally want to avoid doing a lot of work in `poll` method

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -452,32 +695,55 @@ where
...
+                log::trace!("handling real_messages: size: {}", real_messages.len());
+
+                // First store what we got for the given connection id
+                self.received_buffer.store(&conn_id, real_messages);
+                let real_next = self
+                    .received_buffer
+                    .pop_next_message_at_random()
+                    .expect("we just added one");
+
+                // if there are more messages immediately available, notify the runtime
```
jstuczyn: what about that? : )

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -492,24 +758,46 @@ where
         }
     }
 
+    fn log_status(&self) {
```
jstuczyn, line 444: that's a good idea!

clients/client-core/src/client/real_messages_control/acknowledgement_control/retransmission_request_listener.rs
octol: Well I just tried making this change, yes and no :)
At first it seems to make things more tidy. But the lane is per chunk of messages, so keeping it separate makes it easy to add a single burst of traffic to the relevant buffer in the `OutQueueControl`

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: Agree! I'll add this to the todo for the next PR

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: `TransmissionBuffer` sounds like a good one

clients/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: Unfortunately in this case with `real_messages` being moved and the invariant of no empty entries, the correponding code using `entry` with `.and_modify(..).or_insert(..)` doesn't work :(

clients/client-core/src/client/real_messages_control/mod.rs
octol: Yep, created `client-connections` crate

PR #1733, opened 2022-11-02
Wallet: make input's label always shrink

gala1234:
# Description Make input label state be always 'reduced'

Closes: 
https://github.com/nymtech/nym/issues/2256 

We are having problems with the shrink input property, and reading the mui doc, I found this solution. On mui doc they tell that this issue is common on number input, datetime input, Stripe input. But we are having it in more places. 

Also, to be consistent, I added it to all inputs to behave the same always.

Maybe we could consider to have a custom textField, to avoid adding this property on each textField component.

Input's new look:
image

image

image


doums (approved):
lgtm

fmtabbara (approved):
I wonder if we create a new component which wraps the Textfield component like this
```
export const NymInput = (props: TextFieldProps) => 
    <TextField {...props} InputLabelProps={{ shrink: true }} />;
```

and then replace all `<TextField ... />` with `<NymInput ... />

PR #1743, opened 2022-11-08
Nym SDK + webassembly client changes

mmsinclair:
Adds a new SDK Typescript package that provides the Nym WASM client as a Web Worker

There are two examples of a chat app to show how to use it

# Checklist:


clients/webassembly/src/client/mod.rs
```diff
@@ -281,18 +287,21 @@ impl NymClient {
             let this = JsValue::null();
 
             while let Some(reconstructed) = reconstructed_receiver.next().await {
-                if let Some(ref callback) = on_message {
-                    for msg in reconstructed {
+                for msg in reconstructed {
+                    if let Some(ref callback_binary) = on_binary_message {
```
jstuczyn, line 291: isn't it the case that it will attempt to use binary callback even on text messages?

clients/webassembly/src/binary_message_helper.rs
```diff
@@ -0,0 +1,217 @@
...
+            .unchecked_into::<IStringMessage>(),
+    )
+}
+pub(crate) fn parse_binary_payload(message: &[u8]) -> (u8, String, &[u8]) {
+    // 1st byte is the kind
+    let kind = message[0];
+
+    // then the size as u64 big endian
+    let mut size = [0u8; 8];
+    size.clone_from_slice(&message[1..9]);
```
jstuczyn, line 131: you're calling this function from `parse_binary_message`, where you only check for whether the `message` is shorter  than 2 bytes. But here you attempt to read at least 9 so it could blow up by reading out of range

clients/webassembly/src/binary_message_helper.rs
```diff
@@ -0,0 +1,217 @@
...
+    let size = u64::from_be_bytes(size) as usize;
+
+    // then the headers
+    let headers = String::from_utf8_lossy(&message[9..9 + size]).into_owned();
```
jstuczyn, line 135: and also here if `size` was malformed (or somebody was lying) and set it to some stupidly high value

sdk/typescript/examples/plain-html/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    return;
+  }
+
+  // // mixnet v1
+  // const validatorApiUrl = 'https://validator.nymtech.net/api';
+  // const preferredGatewayIdentityKey = 'E3mvZTHQCdBvhfr178Swx9g4QG3kkRUun7YnToLMcMbM';
+
+  // mixnet v2
+  const validatorApiUrl = 'https://qwerty-validator-api.qa.nymte.ch/api'; // "http://localhost:8081";
+  const preferredGatewayIdentityKey = undefined; // '36vfvEyBzo5cWEFbnP7fqgY39kFw9PQhvwzbispeNaxL';
```
jstuczyn, line 23: if we're not even using it, why not remove it

clients/webassembly/src/client/mod.rs
mmsinclair: yes, I left the two handlers in so that the old `js-example` in the `webassembly` directory would work still. I could take it out?

clients/webassembly/src/binary_message_helper.rs
mmsinclair: yes, good spot, I'll add some extra checks and tests

clients/webassembly/src/binary_message_helper.rs
mmsinclair: I'll add some checking for suspicious behaviour 😉

sdk/typescript/examples/plain-html/src/index.ts
mmsinclair: mmmm, I really want to remove all of that and make it just work with `gateway1.nymtech.net` on mainnet

PR #1753, opened 2022-11-10
Update/last minute release updates

### LINKED ISSUE #2204: Wallet - snags and tweaks
1. In the Node Settings, change "Your changes will be ONLY saved on the display. Remember to change the values on your node’s config file too" to -> **"Changing these values will ONLY change the data about your node on the blockchain. Remember to change your node’s config file with the same values too."**

2.  In the Node Settings, change "Unbonding is irreversible and it won’t be possible to restore the current state of your node again" to -> **"Remember you should only unbond if you want to remove your node from the network for good. Unbonding is irreversible and once done it won't be possible to restore the current state of your node again."**

3. In below image: 
- The button does not operate when node is bonded using coins from vesting account. 
- A cost modal should be added for tx fee.
- The text in the confirmation modal should change to -> **"Your changes are submitted to the blockchain. Remember to change your node’s config file with the same values too"**  
- Remove descriptions under all sections (Port, Host and Version) as they are wrong 
- Change the button's text to read **"Submit changes to the blockchain"**
- The button should be aligned with the input fields 
Screenshot 2022-11-09 at 21 59 49

6. In below image:
- The input field for Profit Margin is filling the entire container, it should be the same size and style as the bottom field
- Change the button's text to read **"Save changes"** 
- Button behaviour same as above; does not work when vesting account coins are used for bonding
- A cost modal needs to be added
- The confirmation modal's text should change to lower case
- The button should be aligned to the other side
Screenshot 2022-11-09 at 21 54 13

5. In below image:
- Remove the tooltip icon next to "Vesting Schedule"
- The font used for "Transfer unlocked tokens" looks bigger than the other headers. Let's make it same size as "Vesting Schedule"
- Change "Transferable tokens" to -> **"Unlocked transferable tokens"**
- The font on 400 NYM looks bold, it should be same size and style as the balance amount. 
Screenshot 2022-11-09 at 22 33 34

6. Tooltip for `Operator rewards` talks about manual compounding which is wrong, it should change to -> 
**"This is your (operator) rewards 
including the PM and cost. Rewards are 
automatically compounded every epoch. 
You can redeem your rewards at any time."**

7. the version is still showing as v1.0.9
Screenshot 2022-11-10 at 12 25 03

Changes are applied on light version of wallet:
image

image

fmtabbara:
# Description

- Fix bug preventing node and cost updates being made on bonds using vesting tokens

- Updates to working and styling

Closes: https://github.com/nymtech/nym/issues/2204

### DIFF
--- a/nym-wallet/src/components/Bonding/BondedMixnode.tsx
+++ b/nym-wallet/src/components/Bonding/BondedMixnode.tsx
@@ -42,7 +42,7 @@ const headers: Header[] = [
     header: 'Operator rewards',
     id: 'operator-rewards',
     tooltipText:
-      'This is your (operator) new rewards including the PM and cost. You can compound your rewards manually every epoch or unbond your node to redeem them.',
+      'This is your (operator) rewards including the PM and cost. Rewards are automatically compounded every epoch. You can redeem your rewards at any time.',
   },
   {
     header: 'No. delegators',
--- a/nym-wallet/src/pages/balance/vesting.tsx
+++ b/nym-wallet/src/pages/balance/vesting.tsx
@@ -130,16 +130,10 @@ const TokenTransfer = () => {
       </Grid>
       <Grid item>
         <Typography variant="subtitle2" sx={{ color: (t) => t.palette.nym.text.muted, mt: 2 }}>
-          Transferable tokens
+          Unlocked transferable tokens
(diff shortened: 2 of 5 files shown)


doums (approved):
lgtm

PR #1759, opened 2022-11-15
Added migration code for updating vesting contract address

jstuczyn:
# Description

This PR introduces default migration to the mixnet contract so that it would be possible to update the stored vesting contract address. This is required if we ever wanted to deploy any new environment since there exists a circular dependency on contract addresses between vesting and mixnet contracts during instantiation and you can't know both of them ahead of time (well, not without changes from wasmd 0.29 that we are NOT on)

# Checklist:


tommyv1987:
All good.

PR #1795, opened 2022-11-23
Feature/gateway client protocol version

### LINKED ISSUE #1917: Client: rework client <=> gateway compatibility check
The purpose of this work is:
- to allow operators to upgrade asynchronously

Consequences:
- the mixnet splits as operators upgrade
- some clients may break, if no gateways can work with them

Currently the client checks that the major and minor semver matches.

https://github.com/nymtech/nym/blob/122f5d9f2e5c1ced96e3b9ba0c74ef8b7dbde2c7/gateway/gateway-requests/src/registration/handshake/mod.rs

Add a method of:
- exchanging protocol versions during handshake
- needs to be backwards compatible (operators will not need to redeploy immediately, and the upgrade can be rolling)

The following work needs this:
https://github.com/nymtech/nym/issues/1954

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/1917

Introduces simple `protocol_version` field exchanged for each `authenticate` and `register` message between client and the gateway. Currently it's fully optional and if one party doesn't specify it, everything proceeds without a problem. However, once we release a breaking change (like a 1.2.0) this behaviour should change.

# Checklist:

### DIFF
--- a/common/client-libs/gateway-client/src/client.rs
+++ b/common/client-libs/gateway-client/src/client.rs
@@ -443,7 +443,7 @@ impl GatewayClient {
         // right now there are no failure cases here, but this might change in the future
         match gateway_protocol {
             None => {
-                warn!("the gateway we're connected to has not specified it's protocol version. It's probably running version < 1.1.X, but that's still fine for now. It will become a hard error in 1.2.0");
+                warn!("the gateway we're connected to has not specified its protocol version. It's probably running version < 1.1.X, but that's still fine for now. It will become a hard error in 1.2.0");
                 // note: in 1.2.0 we will have to return a hard error here
                 Ok(())
             }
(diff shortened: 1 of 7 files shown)


common/client-libs/gateway-client/src/client.rs
```diff
@@ -436,6 +436,33 @@ impl GatewayClient {
...
     }
 
+    fn check_gateway_protocol(
+        &self,
+        gateway_protocol: Option<u8>,
+    ) -> Result<(), GatewayClientError> {
+        // right now there are no failure cases here, but this might change in the future
+        match gateway_protocol {
+            None => {
+                warn!("the gateway we're connected to has not specified it's protocol version. It's probably running version < 1.1.X, but that's still fine for now. It will become a hard error in 1.2.0");
```
octol: Nitpick: its

octol (approved):

common/client-libs/gateway-client/src/client.rs
jstuczyn: good catch

PR #1820, opened 2022-11-30
fix compile issue with v1.1.4 of the mixnode

timkuijsten:
This fixes the following two errors when compiling the mixnode:
error[E0412]: cannot find type `ApiClient` in crate `validator_client`
error[E0433]: failed to resolve: could not find `ApiClient` in `validator_client`

The gateway has similar code but somehow just compiles, same for the develop branch. Happens with the mixnode since [v1.1.1](e4f3483).


octol:
Thanks and good spotting this! The issue is that the `validator-client` dependency should have the `nymd-client` feature enabled

octol (changes requested):
I suggest adding `nymd-client` to the `validator-client` dependency in `common/mixnode-common/Cargo.toml` instead

timkuijsten:
> I suggest adding `nymd-client` to the `validator-client` dependency in `common/mixnode-common/Cargo.toml` instead

Done, tested and rebased on [release/v1.1.4](eff1f383c3c8b9be5e67d2afaf441c211d2d939d).

timkuijsten:
Fixed in nym-mixnode v1.1.19 on OpenBSD 7.3.

PR #1821, opened 2022-11-30
fix compiling with LibreSSL 3.6.x

timkuijsten:
Update the openssl-sys and openssl crate to v0.9.78 and v0.10.43, respectively. These version are compatible with LibreSSL 3.6.x (see https://github.com/sfackler/rust-openssl/pull/1705).

Tested on OpenBSD 7.2.

Fixes the following compiler error:
--- stderr  
thread 'main' panicked at '                       

This crate is only compatible with OpenSSL (version 1.0.1 through 1.1.1, or 3.0.0), or LibreSSL 2.5  
through 3.4.1, but a different version of OpenSSL was found. The build is now aborting      
due to this version mismatch.  

', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.72/build/main.rs:295:5

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3680,16 +3680,28 @@ checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
 [[package]]
 name = "openssl"
-version = "0.10.38"
+version = "0.10.43"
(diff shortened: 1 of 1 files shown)


timkuijsten:
Fixed in nym-mixnode v1.1.19 on OpenBSD 7.3.

PR #1827, opened 2022-12-01
nym-cli generate instantiate message for mixnet and vesting contracts

### LINKED ISSUE #1906: Add command to nym-cli for generating instantiate messages
When calling the `nym-cli cosmwasm init` command for initializing contracts we need to pass that command an init message in JSON form.

There are 5 different messages and each of them has complex messages.

The idea is to add a simple command that will help the user generate those messages in JSON format so it can be used with the `init` command.

farbanas:
# Description

Closes: nymtech/nym#1906

# Checklist:


farbanas:
Should I add other contracts straight away (multisig, coconut-dkg, coconut-bandwidth) ?

common/commands/src/validator/cosmwasm/generators/mixnet.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, Parser)]
+pub struct Args {
+    #[clap(long, empty_values = false)]
+    pub rewarding_validator_address: String,
+
+    #[clap(long, empty_values = false)]
+    pub vesting_contract_address: String,
+
+    #[clap(long, default_value = "unymt")]
+    pub rewarding_denom: String,
```
neacsu: Given that those 3 fields already have default values, as given by the env setup, I would say mark them as optional (`Option<String>`) without any `clap` default values. And in the `generate` function bellow, use the values from `Args` or, if not specified, the values from environment (e.g. `std::env::var(network_defaults::var_names::REWARDING_VALIDATOR_ADDRESS)`)

common/commands/src/validator/cosmwasm/generators/vesting.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use vesting_contract_common::InitMsg;
+
+#[derive(Debug, Parser)]
+pub struct Args {
+    #[clap(long, empty_values = false)]
+    pub mixnet_contract_address: String,
+
+    #[clap(long, default_value = "unymt", empty_values = false)]
+    pub mix_denom: String,
```
neacsu: Same comment as in `mixnet.rs`

neacsu (commented):
> Should I add other contracts straight away (multisig, coconut-dkg, coconut-bandwidth) ?

Yes, I think it would be useful to have those added as well, as we already deployed them in QA.

As for the `Args` values for those contracts, I would propose, for simplicity, the following:
```
coconut-bandwidth: pool_addr -> mandatory, no default value
coconut-dkg: group_addr -> mandatory, no default value
coconut-dkg: admin -> mandatory, no default value
multisig: group_addr -> mandatory, no default value
multisig: threshold -> hard-coded value of AbsolutePercentage { percentage: 0.67 } as we're not going to play with this value
multisig: max_voting_period -> default value to something like 1 hour or 1 day
```

PR #1831, opened 2022-12-01
client: add --no-cover and update --fastmode

octol:
# Description

- Bring the `--fastmode` debug flag up to date
- Add `--no-cover` debug mode

# Checklist:

### DIFF
--- a/clients/native/src/commands/mod.rs
+++ b/clients/native/src/commands/mod.rs
@@ -80,6 +80,7 @@ pub(crate) struct OverrideConfig {
     disable_socket: bool,
     port: Option<u16>,
     fastmode: bool,
+    no_cover: bool,
 
     #[cfg(feature = "coconut")]
     enabled_credentials_mode: bool,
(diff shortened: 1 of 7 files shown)


clients/native/src/commands/mod.rs
```diff
@@ -141,6 +142,10 @@ pub(crate) fn override_config(mut config: Config, args: OverrideConfig) -> Confi
         config.get_base_mut().set_high_default_traffic_volume();
     }
 
+    if args.no_cover {
+        config.get_base_mut().set_no_cover_traffic();
```
jstuczyn, line 146: shouldn't you also disable poisson there given the flag description?

clients/socks5/src/commands/mod.rs
```diff
@@ -136,6 +137,10 @@ pub(crate) fn override_config(mut config: Config, args: OverrideConfig) -> Confi
         config.get_base_mut().set_high_default_traffic_volume();
     }
 
+    if args.no_cover {
+        config.get_base_mut().set_no_cover_traffic();
```
jstuczyn, line 141: same comment regarding poisson

clients/native/src/commands/mod.rs
octol: `set_no_cover_traffic` disables that too

PR #1863, opened 2022-12-07
Renaming validator-api to nym-api

futurechimp:
# Description

Closes: team-core#214

Renames validator-api to nym-api.

# Checklist:

- [*] added a changelog entry to `CHANGELOG.md`


common/client-libs/validator-client/Cargo.toml
```diff
@@ -29,7 +29,7 @@ futures = "0.3"
 
 coconut-interface = { path = "../../coconut-interface" }
 network-defaults = { path = "../../network-defaults" }
-validator-api-requests = { path = "../../../validator-api/validator-api-requests", features = ["coconut"] }
+validator-api-requests = { path = "../../../nym-api/nym-api-requests", features = ["coconut"] }
```
jstuczyn: you'd probably also want to rename the crate name itself too (in cargo.toml)

futurechimp: Thanks, I hadn't caught that, good to have the Stuczynskinator on the case.

futurechimp:
Ok I think this develop merge should do the trick - can you have a look @jstuczyn and @octol ?

common/client-libs/validator-client/src/client.rs
```diff
@@ -827,17 +827,17 @@ impl CoconutApiClient {
 
 #[derive(Clone)]
 pub struct ApiClient {
-    pub validator_api: validator_api::Client,
+    pub validator_api: nym_api::Client,
```
jstuczyn: some leftover `validator_api`

common/client-libs/validator-client/src/client.rs
```diff
@@ -827,17 +827,17 @@ impl CoconutApiClient {
...
 }
 
 impl ApiClient {
     pub fn new(api_url: Url) -> Self {
-        let validator_api_client = validator_api::Client::new(api_url);
+        let nym_api_client = nym_api::Client::new(api_url);
 
         ApiClient {
-            validator_api: validator_api_client,
+            validator_api: nym_api_client,
```
jstuczyn: ibid.

common/commands/src/validator/mixnet/query/query_all_gateways.rs
```diff
@@ -15,7 +15,7 @@ pub struct Args {
 }
 
 pub async fn query(args: Args, client: &QueryClientWithValidatorAPI) {
```
jstuczyn: client is called `QueryClientWithValidatorAPI` -> I guess it should also be renamed

futurechimp:
Lots more changes done. Everything compiles nicely now, and except in TypeScript client I think most things have renamed. 

Please don't merge this to develop as we need to talk about whether updating config file keys is going to break old clients (I suspect it will). We will need to figure out a good upgrade process.

jstuczyn:
> Lots more changes done. Everything compiles nicely now, and except in TypeScript client I think most things have renamed.
> 
> Please don't merge this to develop as we need to talk about whether updating config file keys is going to break old clients (I suspect it will). We will need to figure out a good upgrade process.

you can use `#[serde(alias = "old-name")]` to preserve compatibility

futurechimp:
Ok I've tested out 1.1.x config files with `validator_api_urls` in them, as well as the new key `nym_api_urls`. Deserialization works well in both cases. 

I'll leave it here for the afternoon in case anyone can think of other stuff which could break (currently I can't think of any)

PR #1886, opened 2022-12-12
Make layers always balanced

durch:
Assures that all families are in the same layer, and that all layers have equal (if possible) number of nodes

# Checklist:

### DIFF
--- a/nym-api/src/epoch_operations/mod.rs
+++ b/nym-api/src/epoch_operations/mod.rs
@@ -59,16 +59,6 @@ pub struct RewardedSetUpdater {
     storage: NymApiStorage,
 }
 
-// Weight of a layer being chose is reciprocal to current count in layer
-fn layer_weight(l: &Layer, layer_assignments: &HashMap<Layer, f32>) -> f32 {
-    let total = layer_assignments.values().fold(0., |acc, i| acc + i);
-    if total == 0. {
-        1.
-    } else {
-        1. - (layer_assignments.get(l).unwrap_or(&0.) / total)
-    }
-}
-
 impl RewardedSetUpdater {
     pub(crate) async fn current_interval_details(
         &self,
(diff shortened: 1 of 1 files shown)


validator-api/src/epoch_operations/mod.rs
```diff
@@ -90,44 +99,113 @@ impl RewardedSetUpdater {
...
-            let family = mix_to_family.get(&mix.bond_information.identity().to_owned());
-            // Get layer already assigned to nodes family, if any
-            let family_layer = family.and_then(|h| families_in_layer.get(h.identity()));
+        let mut nodes_in_families = vec![];
+        let mut regular_nodes = Vec::with_capacity(rewarded_set.len());
 
-            // Same node families are always assigned to the same layer, otherwise layer selected by a random weighted choice
-            let layer = if let Some(layer) = family_layer {
-                layer.to_owned()
+        for (idx, node) in rewarded_set.drain(..).enumerate() {
```
jstuczyn:
```suggestion
        for (idx, node) in rewarded_set.into_iter().enumerate() {
```
so that we could get rid of `&mut` in the argument?

validator-api/src/epoch_operations/mod.rs
```diff
@@ -88,47 +78,79 @@ impl RewardedSetUpdater {
...
+        for (_head, members) in families.iter_mut() {
+            let smallest_layer = layers
+                .iter()
+                .min_by_key(|(_layer, members)| members.len())
+                .map(|(layer, _members)| *layer)
+                .unwrap_or(Layer::One);
+
+            // What if a single family is bigger then the layer
+            if members.len() > target_layer_count {
+                members.truncate(target_layer_count)
```
jstuczyn: but isn't it the case that after we truncate this vec, we will "lose" those members?

durch:
Relevant parts merged into audit fixes

PR #2685, opened 2022-12-13
use experimental text

fmtabbara:
# Description

Use following text in the default view of NymConnect "Connect to the Nym mixnet" & "This is experimental software. Do not rely on it for strong anonymity (yet)."


tommyv1987:
Looking good here too, ready to be in 1.1.4

Issue #2130, opened 2022-12-09
Wallet - bonding and delegation tables updates


MarcinSN:
Bonding table:
- [x] When node has set name, show the name instead of "Mix node". If node has not set name show only title "Mix node"
- [x] Tooltip with IP after hovering on node address
![image](https://user-images.githubusercontent.com/95295201/207008423-bbae59de-ac3b-4d3a-8931-a880bdf806ad.png)

- [x] Adding next epoch element in table
![image](https://user-images.githubusercontent.com/95295201/206721110-a6e40719-cdfd-4111-a019-43c816cc8d8f.png)

Delegation:
- [x] Adding next epoch element in table
- [x] Rearange elements(screen, button, network explorer positions)
- [x] If delegation has been done with tokens from vesting account, the tooltip should appear when hovering anywhere on the line
![image](https://user-images.githubusercontent.com/95295201/206721383-80dfe525-aadd-4dfc-a016-82cc69f04862.png)

Issue #2128, opened 2022-12-09
Wallet - node settings - delete epoch info


MarcinSN:
Interval info stays

![image](https://user-images.githubusercontent.com/95295201/207008034-e2707230-5710-49fb-ba9f-b2a12644973d.png)

PR #2709, opened 2022-12-15
Feat/2130 tables update

gala1234:
# Description

Closes:
[2130](https://github.com/nymtech/nym/issues/2130)
[2128](https://github.com/nymtech/nym/issues/2128)

- Node Settings, the divider is better positioned and was removed the epoch info, this is how it looks now:
image

- Bond table, if the node does not have a name, we show nothing instead of the node name, the tooltip with the host will be in the walletId instead of the node name, the next epoch info has also been added:
image

- Delegation table, re-arrange elements(screen, button, network explorer positions) and adding epoch info. Adding tooltip if delegation has been done with tokens from vesting account, the tooltip should appear when hovering anywhere on the line:
image


nym-wallet/src/components/Bonding/BondedMixnode.tsx
```diff
@@ -63,6 +66,7 @@ export const BondedMixnode = ({
   network?: Network;
   onActionSelect: (action: TBondedMixnodeActions) => void;
 }) => {
+  const [nextEpoch, setNextEpoch] = useState<string | undefined | Error>();
```
fmtabbara: I don't think you need to specify undefined as a type if the state begins as undefined.

nym-wallet/src/components/IdentityKey.tsx
```diff
@@ -1,13 +1,15 @@
...
+export const IdentityKey = ({ identityKey, tooltipTitle }: { identityKey: string; tooltipTitle?: string }) => (
+  <Tooltip title={tooltipTitle || ''} placement="top" arrow>
+    <Stack direction="row" width="fit-content">
+      <Typography variant="body2" component="span" fontWeight={400} sx={{ mr: 1, color: 'text.primary' }}>
+        {splice(6, identityKey)}
+      </Typography>
+      <CopyToClipboard value={identityKey} sx={{ fontSize: 18 }} />
+    </Stack>
+  </Tooltip>
 );
```
fmtabbara: Is it possible to wrap <IdentityKey /> in a tooltip instead of passing in this extra prop and wrapping inside the component?

gala1234: @fmtabbara It doesn't work. I've been trying fist to wrap directly with the tooltip, and I didn't find the way to make it work.  But wrapping directly the component it does

nym-wallet/src/components/Bonding/BondedMixnode.tsx
gala1234: @fmtabbara you are right, changed!

fmtabbara (approved):

Issue #2208, opened 2022-11-04
NE - add correct decimals to the gateway details page


benedettadavico:
Make it consistent with the mixnode page, ie 4 decimal places
<img width="1048" alt="image" src="https://user-images.githubusercontent.com/46782255/199948508-119c066c-cf01-4c6a-84f5-c4f1ad46fec2.png">

PR #2741, opened 2022-12-21
fix(explorer): set gateway bond 6 decimals

doums:
# Description

set 6 decimal places for gateway bond number (we use 6 decimals for mixnode pages)

Closes: #2208


gala1234 (approved):
looks good!

PR #2764, opened 2023-01-04
Initial Rust client SDK

### LINKED ISSUE #2669: Rust SDK - Initial version

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/2669

Initial version of the Rust client SDK

# Checklist:


clients/client-core/src/client/base_client/non_wasm_helpers.rs
```diff
@@ -85,3 +85,10 @@ pub async fn setup_fs_reply_surb_backend<P: AsRef<Path>>(
         setup_fresh_backend(db_path, debug_config).await
     }
 }
+
+pub fn setup_empty_reply_surb_backend(debug_config: &DebugConfig) -> browser_backend::Backend {
+    browser_backend::Backend::new(
```
jstuczyn: if you want to create an empty backend, I'd perhaps use `backend::Empty` since `browser_backend` **will** change once it's implemented

clients/client-core/src/client/key_manager.rs
```diff
@@ -119,7 +161,21 @@ impl KeyManager {
...
+
+        Ok(())
+    }
+
+    pub fn store_key_gateway_only(
+        &self,
+        client_pathfinder: &ClientKeyPathfinder,
+    ) -> io::Result<()> {
+        match self.gateway_shared_key.as_ref() {
+            None => panic!("WIP(JON)"),
```
jstuczyn: I wonder if perhaps we should change the startup procedure to unify all this gateway key  issue. maybe they should be kept separate or something?

clients/client-core/src/client/replies/reply_storage/backend/mod.rs
```diff
@@ -6,10 +6,12 @@ use async_trait::async_trait;
 use std::error::Error;
 use thiserror::Error;
 
-#[cfg(target_arch = "wasm32")]
+//#[cfg(target_arch = "wasm32")]
```
jstuczyn: and if you use `Empty`, i'd restore this target lock : )

clients/client-core/src/config/persistence/key_pathfinder.rs
```diff
@@ -7,12 +7,12 @@ use std::path::{Path, PathBuf};
 
 #[derive(Debug)]
 pub struct ClientKeyPathfinder {
```
jstuczyn, line 9: I wonder if this should be part of sdk. In principle sdk could be used by wasm, no? Meaning there's no filesystem

clients/client-core/src/error.rs
```diff
@@ -55,6 +55,9 @@ pub enum ClientCoreError {
     #[error("The address of the gateway is unknown - did you run init?")]
     GatwayAddressUnknown,
 
+    #[error("failed to register receiver for reconstructed mixnet messages")]
+    FailedToRegisterReceiver,
```
jstuczyn, line 59: Is it even possible for this one to be thrown? If we fail to register receiver it means some task has crashed so we'd have already got another error

common/task/src/manager.rs
```diff
@@ -208,7 +208,7 @@ pub struct TaskClient {
 
 impl TaskClient {
     #[cfg(not(target_arch = "wasm32"))]
-    const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
+    const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(1);
```
jstuczyn: Is 1s enough though? If you have bunch of reply surbs in memory, it might take more than 1s to flush them to storage (during shutdown)

sdk/rust/nym-sdk/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+client-core = { path = "../../../clients/client-core", features = ["fs-surb-storage"]}
+crypto = { path = "../../../common/crypto" }
+gateway-client = { path = "../../../common/client-libs/gateway-client" }
+gateway-requests = { path = "../../../gateway/gateway-requests" }
+network-defaults = { path = "../../../common/network-defaults" }
+nymsphinx = { path = "../../../common/nymsphinx" }
+task = { path = "../../../common/task" }
+
+futures = "0.3"
+log = "0.4"
```
jstuczyn, line 19: *nudge nudge*, `tracing` : )

sdk/rust/nym-sdk/examples-common/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+pub fn setup_logging() {
```
jstuczyn: haven't we already written identical function in `common/logging`?

sdk/rust/nym-sdk/src/mixnet/config.rs
```diff
@@ -0,0 +1,36 @@
+use client_core::config::DebugConfig;
+use network_defaults::mainnet;
+use url::Url;
+
+pub struct Config {
```
jstuczyn, line 5: I wonder if with this config we would be somehow able to refactor the "main" `Config`s for the clients and get rid of that `PhantomData`

clients/client-core/src/client/key_manager.rs
octol: It certainly feels like the shared gateway key is conceptually separate from the other client keys

clients/client-core/src/error.rs
octol: Possibly, but I'm not certain enough that I want to put an `unwrap` there instead

sdk/rust/nym-sdk/Cargo.toml
octol: yep yep, it's on the todo list :)

clients/client-core/src/config/persistence/key_pathfinder.rs
octol: Yep true that. Well I think we need it for the path case though, so how to handle it for wasm I'll let future me/us figure out. YAGNI ;)

PR #2834, opened 2023-01-12
Feature/gateway graceful shutdown

### LINKED ISSUE #2019: Graceful shutdown of gateway

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/2019

# Checklist:

### DIFF
--- a/common/client-libs/mixnet-client/src/forwarder.rs
+++ b/common/client-libs/mixnet-client/src/forwarder.rs
@@ -16,6 +16,7 @@ type MixForwardingReceiver = mpsc::UnboundedReceiver<MixPacket>;
 pub struct PacketForwarder {
     mixnet_client: Client,
     packet_receiver: MixForwardingReceiver,
+    shutdown: task::TaskClient,
 }
 
 impl PacketForwarder {
@@ -25,6 +26,7 @@ impl PacketForwarder {
         initial_connection_timeout: Duration,
         maximum_connection_buffer_size: usize,
         use_legacy_version: bool,
+        shutdown: task::TaskClient,
     ) -> (PacketForwarder, MixForwardingSender) {
         let client_config = Config::new(
             initial_reconnection_backoff,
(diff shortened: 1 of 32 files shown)


common/client-libs/mixnet-client/src/forwarder.rs
```diff
@@ -40,12 +42,38 @@ impl PacketForwarder {
...
     }
 
     pub async fn run(&mut self) {
+        while !self.shutdown.is_shutdown() {
+            tokio::select! {
+                biased;
+                _ = self.shutdown.recv() => {
+                    log::trace!("PacketForwarder: Received shutdown");
+                }
+                Some(mix_packet) = self.packet_receiver.next() => {
```
octol, line 58: what about making it stop on channel close? In many cases I suspect that tasks might not even need a `TaskClient` if they stop on channel close

gateway/src/commands/mod.rs
```diff
@@ -136,46 +138,61 @@ pub(crate) fn override_config(mut config: Config, args: OverrideConfig) -> Confi
...
+    // }
 }
 
 // this only checks compatibility between config the binary. It does not take into consideration
 // network version. It might do so in the future.
-pub(crate) fn version_check(cfg: &Config) -> bool {
+pub(crate) fn ensure_config_version_compatibility(cfg: &Config) -> Result<(), GatewayError> {
     let binary_version = env!("CARGO_PKG_VERSION");
     let config_version = cfg.get_version();
     if binary_version != config_version {
```
octol: fyi clippy suggest to change to `==` and swap the `if` `else`

gateway/src/error.rs
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io;
+use std::path::PathBuf;
+use thiserror::Error;
+use validator_client::nyxd::AccountId;
+use validator_client::ValidatorClientError;
+
+#[derive(Debug, Error)]
+pub(crate) enum GatewayError {
```
octol, line 11: Nice error enum, I should do more like this

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -412,15 +413,19 @@ where
...
     {
         trace!("Started listening for ALL incoming requests...");
 
-        loop {
+        while !shutdown.is_shutdown() {
             tokio::select! {
+                biased;
+                _ = shutdown.recv() => {
+                    log::trace!("client_handling::AuthenticatedHandler: received shutdown");
+                }
```
octol, line 427: Shutdown also for None = handle_request below?

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -412,15 +413,19 @@ where
...
     where
         S: AsyncRead + AsyncWrite + Unpin,
         St: Storage,
     {
         trace!("Started listening for ALL incoming requests...");
 
-        loop {
+        while !shutdown.is_shutdown() {
             tokio::select! {
+                biased;
```
octol: Note that this will mean read_websocket_message always will take priority over mix_receiver.next

common/client-libs/mixnet-client/src/forwarder.rs
jstuczyn: good idea

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
jstuczyn: no, because some requests might not produce a response that has to be sent back to the client.

common/client-libs/mixnet-client/src/forwarder.rs
jstuczyn: while that could work, I think having explicit shutdown check here would be better as if the request queue has built up on `packet_receiver` channel, it will take quite a while for the channel to close on the receiver side.

PR #2835, opened 2023-01-12
Feature/centralized allow list

### LINKED ISSUE #2794: Centralize the allow.list
At the moment the fact that each network requester maintains its own allowed list makes the system too unreliable. We should load a standard allowed.list from a remote server instead.

futurechimp:
# Description

Closes: #2794 

When the network requester starts up, it downloads a standard list of allowed domains from a wellknown location, so that all network requesters will work reliably for all supported services.


service-providers/network-requester/src/main.rs
```diff
@@ -26,7 +26,7 @@ struct Run {
     #[clap(long)]
     open_proxy: bool,
 
-    /// Websocket port to bind to
+    /// Websocket port to bind to. Defaults to `network_defaults::DEFAULT_WEBSOCKET_LISTENING_PORT` (currently 1977)
```
octol: This will print that constant name verbatim in `--help`, not the value, just checking that it's intentional?

futurechimp:
Aha, good catch, thanks @octol! I didn't realize that was going to be printed.

PR #2837, opened 2023-01-13
chore: added --all-targets --all-features to clippy

farbanas:
# Description

Follow up to the discussion on the daily yesterday, I've run a couple of benchmarks on my machine (ignore the bad formatting of the output).

`WITH --all-targets` and `WITHOUT --all-targets`actually has `--all-features` flag as well enabled.

So the outputs are:
- `--workspace --all-features --all-targets`
- `--workspace --all-features`
- `--workspace --all-targets`
- `--workspace`

paste
paste (1)

We currently use the last case, without `--all-targets` and without `--all-features`. The increase in run time is ~30 seconds if we enable both. That is a 60% increase, but it's still not a lot while we may get the info of what's failing faster.

In CI the differences are even smaller. Look at the `Run clippy` step.

paste (3)

paste (2)

It's 37s vs 18s, a 19s increase.

The question is however, would this be useful to you guys? If you wouldn't have any benefits from this then it doesn't make sense to increase the time it takes to run the CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace -- -D warnings
+          args: --workspace --all-targets --all-features -- -D warnings
 
       # COCONUT stuff
 


octol:
IMHO it's worth running with everything enabled. YMMV

octol (approved):

PR #2847, opened 2023-01-16
Fix typescript examples

solarex1:
Both `plain-html` and `react-webpack-with-theme-example` are broken on the develop branch.

- `plain-html` is calling removed methods.
- `react-webpack-with-theme-example` is referencing an undefined type.


mmsinclair:
Thanks for your PR @solarex1 - we have been fixing these issues in other branches (see https://github.com/nymtech/nym/pull/2884)

The `develop` branch should be working with the new WASM client interface (similar to the fixes you supplied, so hopefully your code should just work).

We haven't published a new package to NPM yet and are hoping to do this next week.

Feel free to raise an issue or submit another PR if things still don't work for you.

PR #2853, opened 2023-01-17
cargo: try out inheriting workspace metadata

octol:
# Description

As part of publishing our crates, we should try to simplify common metadata and dependencies

# Checklist:


durch (approved):
This is pretty great!

Issue #2733, opened 2022-12-20
Use our hosted runners for building and releasing


farbanas:
We're currently using shared GH action runners which have bad performance and they are also overbooked. We have our hosted machines which should be used for those instead.

PR #2864, opened 2023-01-17
[#2733] Releasing on our self-hosted runners and less failing actions

### LINKED ISSUE #2857: Add explorer-api to releases

farbanas:
# Description

Closes: #2857 and #2733 

image

instead of 

image


.github/workflows/nym-release-publish.yml
```diff
@@ -71,6 +66,7 @@ jobs:
         if: github.event_name == 'release'
         with:
           files: |
+            target/release/explorer-api
```
farbanas, line 70: Not sure regarding `explorer-api` if it's fine to release it here since I'm not bumping its version every week at the moment. Maybe we should start bumping its version?

PR #2904, opened 2023-01-24
a bunch of systemd service files

gyrusdentatus:
This should make life much easier to automate stuff. I used a chatgpt with some manual edits, so please check it out also yourself but it all works for me.

The nym-client there is twice just if there is a need to run two instances on one box. I kept the README port 1978 for that service.

Cheers.

# Description

Closes: #XXXX

# Checklist:

### DIFF
--- a/scripts/systemd/nym-client-sp.service
+++ b/scripts/systemd/nym-client-sp.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=NYM Client
+
+[Service]
+User=nym
+ExecStart=/usr/local/bin/nym-client run --id service-provider -p 1978
+LimitNOFILE=65535
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+
--- a/scripts/systemd/nym-client.service
+++ b/scripts/systemd/nym-client.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=NYM Client
+
+[Service]
+User=nym
+ExecStart=/usr/local/bin/nym-client run --id client
+LimitNOFILE=65535
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+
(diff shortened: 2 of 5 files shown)


gyrusdentatus:
Come on guys these are just benign systemd.service file templates for all major services. @futurechimp

gyrusdentatus:
This is a totally benign merge request, which will only save time for a lot of people.

 @mfahampshire Please merge this, unless you do not think the templates are OK.

ovhpse:
Hello,

You should at least add ```Group=nym``` as the default group will be `root`:
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#User=

gyrusdentatus:
It is assumed you have a user nym, iirc useradd or adduser create the group as well for the user on Debian based systems.

Therefore - no. My
Nym user has default group nym.

From the link above:

If no group is set, the default group of the user is used. This setting does not affect commands whose command line is prefixed with "+".

gyrusdentatus:
> Hello,
> 
> 
> 
> You should at least add ```Group=nym``` as the default group will be `root`:
> 
> https://www.freedesktop.org/software/systemd/man/systemd.exec.html#User=

... @ovhpse

tommyv1987:
Thank you for putting this PR up, however, we deem it not required in the repo. All the nymtech.net/docs pages contain this information if you're looking to script the process, here's something:

```
#!/bin/bash

urls=(
    "https://nymtech.net/docs/nodes/gateway-setup.html"
    "https://nymtech.net/docs/nodes/mix-node-setup.html"
    "https://nymtech.net/docs/nodes/network-requester-setup.html"
)

for url in "${urls[@]}"; do
    filename=$(basename "$url" ".html" | sed 's/-setup$//')

    curl -s "$url" | pup 'code[class="language-ini"] text{}' > "nym-$filename.service"
done

PR #2907, opened 2023-01-24
[poc] nym-connect android initial version

doums:
# Description

Initial proof-of-concept version of nym-connect for android. Rough edges included.


clients/socks5/src/client/config/mod.rs
```diff
@@ -35,13 +35,21 @@ impl NymConfig for Config {
...
+    #[cfg(not(feature = "mobile"))]
     fn default_root_directory() -> PathBuf {
         dirs::home_dir()
             .expect("Failed to evaluate $HOME value")
             .join(".nym")
             .join("socks5-clients")
     }
 
+    #[cfg(feature = "mobile")]
+    fn default_root_directory() -> PathBuf {
```
octol: Would you mind checking where this call is originating from in nym-connect-android? I suspect we can avoid calling it altogether and then we can get rid of this

PR #2920, opened 2023-01-26
nym-api tests

benedettadavico:
Adding nym-api tests


tommyv1987 (approved):
Nice work @benedettadavico 🚀

PR #2922, opened 2023-01-26
Audit fixes

durch:
# Description

Various fixes

# Checklist:


jstuczyn:
for oak-13 we should put `set_contract_version` in `migrate` function so that it would get populated and updated on every migration. Furthermore we should probably import `cw2` as dependency that has that functionality rather than trying to reimplement it.

jstuczyn:
~oak-2 might not be yet fully mitigated due to similar issue as described in: https://gitlab.nymte.ch/nym/security/-/issues/34. Since the `node_identity_signature` is made only on the node's owner, malicious proxy could just replay the signature it has seen when the mixnode bonded (since it would be identical)~

edit: nah, should be fine after all since only the vesting contract is allowed to be the proxy now and given our threat model, the signature replay is not going to happen.

PR #2929, opened 2023-01-27
fix(nc-android): panic on socks5 connect

doums:
# Description

Core panic on socks5 connection

### DIFF
--- a/clients/socks5/src/client/mod.rs
+++ b/clients/socks5/src/client/mod.rs
@@ -209,21 +209,23 @@ impl NymClient {
 
     pub async fn start(self) -> Result<TaskManager, Socks5ClientError> {
         #[cfg(not(feature = "mobile"))]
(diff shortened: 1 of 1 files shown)


clients/socks5/src/client/mod.rs
```diff
@@ -209,23 +210,26 @@ impl NymClient {
...
-            fs_reply_surb_backend,
+            non_wasm_helpers::setup_fs_reply_surb_backend(
+                Some(self.config.get_base().get_reply_surb_database_path()),
+                self.config.get_debug_settings(),
+            )
+            .await?,
         );
 
+        #[cfg(feature = "mobile")]
+        let base_builder: BaseClientBuilder<Backend, QueryNyxdClient> =
```
octol: I'd annotate the generic types on the rhs so you don't have to write he type twice. Also, is `Backend` really needed?

octol (approved):

PR #2939, opened 2023-01-30
Added an option to set custom 'host' for the native client

### LINKED ISSUE #1674: [Feature Request] Give the possibility to set the listening address
**Is your feature request related to a problem? Please describe.**
Could it be possible to set the listening address in the configuration

Refering to this line https://github.com/nymtech/nym/blob/d95df4b2869257383e005ec29dc81161a88ac7d2/clients/native/src/websocket/listener.rs#L29

**Where does the feature fit in the Nym real estate?**
- nym-client

**What is this solving?**
Flexibility for application using docker or k8s for deploying and using the nym-client

**Is this an update to packages or libraries?**
https://github.com/nymtech/nym/blob/d95df4b2869257383e005ec29dc81161a88ac7d2/clients/native/src/websocket/listener.rs#L29

### LINKED ISSUE #2912: Add `host` option to client init
At the moment, the Nym native client is hardcoded to listen to host 127.0.0.1. We set it up this way to make a conservative security assumption, essentially "don't open up a listening port to the world". 

We're getting developer feedback now that this is a problem in a Dockerized environment, where it can be convenenient to listen on 0.0.0.0. 

I think it makes sense to keep the default behaviour (start the client bound to 127.0.0.1) but add a new option in `init` so that if needed, the user can specify the listening address according to their needs (0.0.0.0, 1.2.3.4 etc).

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/2912
Closes: https://github.com/nymtech/nym/issues/1674

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ Post 1.0.0 release, the changelog format is based on [Keep a Changelog](https://
 ### Added
 
 - remove coconut feature and unify builds ([#2890])
+- native-client: is now capable of listening for requests on sockets different than `127.0.0.1` ([#2939]). This can be specified via `--host` flag during `init` or `run`. Alternatively a custom `host` can be set in `config.toml` file under `socket` section.
 
 [#2890]: https://github.com/nymtech/nym/pull/2890
+[#2939]: https://github.com/nymtech/nym/pull/2939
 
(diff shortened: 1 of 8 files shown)


tommyv1987:
@jstuczyn all good with this one to get merged into develop and into the release branch 👍

PR #3032, opened 2023-02-14
NC Android - UI changes

### LINKED ISSUE #3019: NC Android - Mobile screens
New things

- Size of fonts, elements
- White button with Start
- Green button with Stop
- No red/disconnet button in mobile version
- Status info is underlined and clickable - we are showing connection info in modals
- If there is a status change we should also use Android notifications to inform user about this change

Image

doums:
# Description

Closes: #3019


nym-connect-android/src/components/PowerButton/PowerButton.tsx
```diff
@@ -74,10 +74,17 @@ export const PowerButton: FCWithChildren<{
     return undefined;
   };
 
+  let statusText: string;
+  if (status === 'connected') {
+    statusText = 'stop';
+  } else {
+    statusText = status === 'disconnected' ? 'start' : '';
+  }
+
```
fmtabbara: Maybe makes this into a switch statement?
default can be empty string

fmtabbara (approved):

PR #3042, opened 2023-02-16
Generalise MessageReceiver

durch:
# Description

Closes: #XXXX

# Checklist:


clients/client-core/src/client/received_buffer.rs
```diff
@@ -438,13 +442,16 @@ impl FragmentedMessageReceiver {
...
+        &mut self,
+        mut shutdown: task::TaskClient,
+    ) -> Result<(), MessageRecoveryError> {
         debug!("Started FragmentedMessageReceiver with graceful shutdown support");
         while !shutdown.is_shutdown() {
             tokio::select! {
                 new_messages = self.mixnet_packet_receiver.next() => {
                     if let Some(new_messages) = new_messages {
-                        self.received_buffer.handle_new_received(new_messages).await;
+                        self.received_buffer.handle_new_received(new_messages).await?;
```
jstuczyn, line 454: should we be returning an error and triggering shutdown if we received a malformed message?

common/nymsphinx/src/receiver.rs
```diff
@@ -7,6 +7,8 @@ use crypto::asymmetric::encryption;
 use crypto::shared_key::recompute_shared_key;
 use crypto::symmetric::stream_cipher;
 use crypto::symmetric::stream_cipher::CipherKey;
+use nym_outfox::error::OutfoxError;
+use nym_outfox::lion::lion_transform_decrypt;
```
jstuczyn, line 11: I'd argue `nym_outfox` stuff shouldn't be inside `nymsphinx` crates. Instead there should be some common crate that those two depend on it. Idk, maybe `common/packets` or something?

jstuczyn (commented):
few extra notes:
- it seems some extra crate is missing that both `nymsphinx` and `outfox` would depend on. I'd imagine it would contain things like message reconstruction since it'd be the same for both,
- acks will have to change as from what I remember they're smaller than `outfox::MIN_MESSAGE_LEN`,
- with the current generic approach I'm not sure whether it's going to be possible to select your packet type via config, but maybe that's fine? 
- when you start incorporating dual packet support on the mixnode side, you'll probably want to set `config.debug.use_legacy_framed_packet_version` to `false` (in mixnode/gateway configs) and bump up the current version number, so that they'd known what to expect

durch:
>     * it seems some extra crate is missing that both `nymsphinx` and `outfox` would depend on. I'd imagine it would contain things like message reconstruction since it'd be the same for both,
Yeah, thats my general plan

>     * acks will have to change as from what I remember they're smaller than `outfox::MIN_MESSAGE_LEN`,
Noted
 
>     * with the current generic approach I'm not sure whether it's going to be possible to select your packet type via config, but maybe that's fine?
My idea was to eventually have some sort of branching at the top level, where both variants of the Receivers/Senders are constructed and we can dispatch based on the config or have an Enum where these receivers are variants and then dispatch based on the variant, still not clear on what to do, I'm leaning towards the enum path

>     * when you start incorporating dual packet support on the mixnode side, you'll probably want to set `config.debug.use_legacy_framed_packet_version` to `false` (in mixnode/gateway configs) and bump up the current version number, so that they'd known what to expect
Check

clients/client-core/src/client/received_buffer.rs
durch: I'd rather return an error, but we should probably ticket that in a different PR, this will take a while...

common/nymsphinx/src/receiver.rs
durch: Yeah, we'll evolve to that for sure

Issue #2978, opened 2023-02-07
NE - fix Routing Score values shown for the gateways


jessgess:
Currently we are showing 4 different values for GW routing score on the explorer: one on the GW list page and 3 on GW details page. 

What we want to show is Routing Score computed as `routingScore: ${report.most_recent}%` on the list and the details page. And the Avg. Score computed like `avgUptime: ${report.last_day || report.last_hour}%` on the details page.

octol:
Backend changes are complete in https://github.com/nymtech/nym/pull/3053

PR #3053, opened 2023-02-17
nym-api: uptime rework

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/2978

- Cache the uptime data and add it to the detailes mixnode and gateway endpoints (as performance).
- Uptime report endpoints now just fetch cached data.
- Add gateway avg uptime endpoint for symmetry with the corresponding mixnode endpoint.

# Checklist:

### DIFF
--- a/explorer/src/api/index.ts
+++ b/explorer/src/api/index.ts
@@ -94,9 +94,9 @@ export class Api {
   static fetchGateways = async (): Promise<GatewayBond[]> => {
     const res = await fetch(GATEWAYS_API);
     const gatewaysAnnotated: GatewayBondAnnotated[] = await res.json();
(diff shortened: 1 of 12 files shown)


octol:
Things to check:

Changes are in `nym-api`

In the two detailed (`/status/gateways/detailed` etc)  endpoints for mixnodes and gateways, there should be new a `node_performance` field with one sub-entry overlapping the old deprecated `performance` field (which we keep for backwards compatibility).

The report endpoints for gateways and mixnodes should also provide this data (3 uptime values), these should be the same as `node_performance` but scaled as a percentage value.

The new `avg_uptime` endpoints for mixnodes and gateways should give the same data as the `report` and `node_performance` ones. Here the `avg_uptime` field should be the same as `node_performance.last_24h`, but scaled to percent.

PR #3135, opened 2023-03-02
Wallet - custom fees and memo features

### LINKED ISSUE #2222: Wallet: optional gas and memo fields
All transactions should have an advanced area where the user can specify the gas to use and/or to set a memo field

Comes from `HAL-06` of vesting contract audit.

The suggestion is to give the user can option of specifying the gas, so they can choose to provide a number to avoid overspend (we use 1.3x factor by default on the simulation to avoid failures).

The memo field is a nice feature to have.

---

**Let's check if we can do this only in Send**
Affects:

- send
- bond (liquid + locked)
- delegate (liquid + locked)
- undelegate (liquid + locked)
- withdraw vested tokens

Image

- more options checkbox(same as advanced options in bonding flow)
- Description and two fields
- Gas field - whatever user put there will be displayed on next screen in summary - we are showing error if proposed fee is to low
- Memo optional field
- If user is using "Prefered Gas Fee" for this transaction we are displaying this in transaction summary
- If user typed something in memo input we are also showing this in summary(send details) screen

Image

doums:
# Description

introduces custom fees and `memo` fields for send operation

Closes: #2222

image

### DIFF
--- a/nym-wallet/src/components/Send/SendInputModal.tsx
+++ b/nym-wallet/src/components/Send/SendInputModal.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { Stack, TextField, Typography, SxProps } from '@mui/material';
+import { Stack, TextField, Typography, SxProps, FormControlLabel, Checkbox } from '@mui/material';
 import { CurrencyFormField } from '@nymproject/react/currency/CurrencyFormField';
 import { CurrencyDenom, DecCoin } from '@nymproject/types';
 import { validateAmount } from 'src/utils';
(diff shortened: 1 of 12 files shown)


fmtabbara:
When "more options" is deselected I think we should clear the state for memo and fee?

fmtabbara:
When you move to the "send details" screen, and then back to the "send" screen, the state for "more options" is lost.

nym-wallet/src/components/Send/SendInputModal.tsx
```diff
@@ -32,8 +36,14 @@ export const SendInputModal = ({
   onAddressChange: (value: string) => void;
   sx?: SxProps;
   backdropProps?: object;
+  userFees?: DecCoin;
+  memo?: string;
+  onUserFeesChange: (value: DecCoin) => void;
+  onMemoChange: (value: string) => void;
 }) => {
   const [isValid, setIsValid] = useState(false);
+  const [showMoreOptions, setShowMoreOptions] = useState(false);
```
fmtabbara: "more options" state is lost when you move to the next page and then back

fmtabbara (changes requested):

fmtabbara (approved):

PR #3157, opened 2023-03-07
Feature/bonding signature UI

### LINKED ISSUE #3109: Wallet - change the Bonding flow to include an additional step for the user to provide a unique signature when they bond their node
Bonding flow update

- 3 steps instead of 2

1 step
we are deleting signature field from form

Image

2 step - new one

we are providing message - user is signing in this message outside the wallet and after that pasting signature in input

image

3 step same as it was(previously second step)

doums:
# Description

Update bonding flows with new signatures system.

Closes: #3109


nym-wallet/src/components/Bonding/modals/BondMixnodeModal.tsx
```diff
@@ -72,28 +77,20 @@ export const BondMixnodeModal = ({
...
-        profit_margin_percent: toPercentFloatString(data.profitMargin),
-        interval_operating_cost: {
-          amount: data.operatorCost.amount.toString(),
-          denom: data.operatorCost.denom,
-        },
-      },
+      pledge: amountData.amount,
+      msgSignature: data.signature,
+      mixnode: mixnodeToTauri(mixnodeData),
+      costParams: costParamsToTauri(amountData),
```
fmtabbara, line 90: :ok_hand:

fmtabbara (approved):

nym-wallet/src/components/Bonding/modals/BondMixnodeModal.tsx
doums: Yeah, in the future we can probably refactor these types to avoid that kind of `convertToTheOtherType` logic

PR #3182, opened 2023-03-14
Features/wallet UI micro

### LINKED ISSUE #2183: Wallet - discrepancies figma vs wallet
image

Talk with Marcin, before you will start doing this, please :)

**Top navigation wallet**
- [ ] Change behaviour and design of dropdown(network selector) same dropdown design we will use in inputs for locked and liquid tokens

**Send**
- [ ] Back button - should be square (global change for back buttons)

**Receive**
- [ ] Alignments + delete text below title

**Accounts**
- [ ] create account - font size in yellow border

**Delegate more**
- [ ] amount input description - fix
- [ ] dropdown states - active, highlighted, disabled (same as network selector)
- [ ] In case of error, not enough funds - highlight of whole input like in bond mixnode form
- [ ] square back button

**Node settings**
- [ ] Alignments
- [ ] Underline of categories - align to words
- [ ] General/Info - Button changes to active after clicking in input - without any changes
- [ ]  General/parameters - lack of tooltips for PM and OC

doums:
# Description

Closes: nymtech/nym/issues/2183 and nymtech/nym/issues/3179

### DIFF
--- a/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
+++ b/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
@@ -27,7 +27,7 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
         </Typography>
         <Typography fontSize={12}>
           In order to import or create account(s) you need to log in with your password or create a new one. Creating a
-          new password will overwrite any old one. Make sure your menonics are all wirtten down before creating a new
+          new password will overwrite any old one. Make sure your mnemonics are all written down before creating a new
           password.
         </Typography>
       </Warning>
--- a/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
+++ b/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
@@ -22,10 +22,10 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
   >
     <Stack spacing={2}>
       <Warning sx={{ textAlign: 'center' }}>
-        <Typography fontWeight={600} sx={{ mb: 1 }}>
+        <Typography fontWeight={600} fontSize={14} sx={{ mb: 1 }}>
           This machine already has a password set on it
         </Typography>
-        <Typography>
+        <Typography fontSize={12}>
           In order to import or create account(s) you need to log in with your password or create a new one. Creating a
           new password will overwrite any old one. Make sure your menonics are all wirtten down before creating a new
           password.
@@ -35,7 +35,7 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
       {passwordCreationSteps.map((step, index) => (
         <Stack key={step} direction="row" spacing={1}>
           <Typography fontWeight={600}>{`${index + 1}.`}</Typography>
-          <Typography>{`${step}`}</Typography>
+          <Typography fontSize={14}>{`${step}`}</Typography>
         </Stack>
       ))}
       <Link
@@ -43,6 +43,7 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
         target="_blank"
         text="Open Nym docs for this guide in a browser window"
         fontWeight={600}
+        fontSize={14}
       />
     </Stack>
   </SimpleModal>
--- a/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
+++ b/nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
@@ -22,10 +22,10 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
   >
     <Stack spacing={2}>
       <Warning sx={{ textAlign: 'center' }}>
-        <Typography fontWeight={600} fontSize={14} sx={{ mb: 1 }}>
+        <Typography variant="body2" fontWeight={600} sx={{ mb: 1 }}>
           This machine already has a password set on it
         </Typography>
-        <Typography fontSize={12}>
+        <Typography variant="caption">
           In order to import or create account(s) you need to log in with your password or create a new one. Creating a
           new password will overwrite any old one. Make sure your mnemonics are all written down before creating a new
           password.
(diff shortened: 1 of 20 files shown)


nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
```diff
@@ -22,10 +22,10 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
   >
     <Stack spacing={2}>
       <Warning sx={{ textAlign: 'center' }}>
-        <Typography fontWeight={600} sx={{ mb: 1 }}>
+        <Typography fontWeight={600} fontSize={14} sx={{ mb: 1 }}>
```
fmtabbara: can we use `variant` here to achieve the correct fontsize? I believe the value `body1` might work

nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
```diff
@@ -22,10 +22,10 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
...
           This machine already has a password set on it
         </Typography>
-        <Typography>
+        <Typography fontSize={12}>
```
fmtabbara: can we use variant here to achieve the correct fontsize? I believe the value body2 might work

nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
```diff
@@ -35,14 +35,15 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
       {passwordCreationSteps.map((step, index) => (
         <Stack key={step} direction="row" spacing={1}>
           <Typography fontWeight={600}>{`${index + 1}.`}</Typography>
-          <Typography>{`${step}`}</Typography>
+          <Typography fontSize={14}>{`${step}`}</Typography>
```
fmtabbara: same as above

nym-wallet/src/components/Accounts/MultiAccountWithPwdHowTo.tsx
```diff
@@ -35,14 +35,15 @@ export const MultiAccountWithPwdHowTo = ({ show, handleClose }: { show: boolean;
...
         </Stack>
       ))}
       <Link
         href="https://nymtech.net/docs/stable/wallet#importing-or-creating-accounts-when-you-have-signed-in-with-mnemonic-but-a-password-already-exists-on-your-machine"
         target="_blank"
         text="Open Nym docs for this guide in a browser window"
         fontWeight={600}
+        fontSize={14}
```
fmtabbara: same as above

nym-wallet/src/components/TokenPoolSelector.tsx
```diff
@@ -40,20 +50,29 @@ export const TokenPoolSelector: FCWithChildren<{ disabled: boolean; onSelect: (p
...
-              } ${clientDetails?.display_mix_denom.toUpperCase()}`}
-            />
+            <Stack direction="row" alignItems="center" gap={2} width="100%">
+              <ListItemText
+                primary="Locked"
+                secondary={`${
+                  +tokenAllocation.locked + +tokenAllocation.spendable
+                } ${clientDetails?.display_mix_denom.toUpperCase()}`}
+                secondaryTypographyProps={{
+                  sx: { textTransform: 'uppercase', color: (t) => t.palette.nym.text.muted },
```
fmtabbara: for colour I think you can do `color: `nym.text.muted`

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx
```diff
@@ -126,7 +137,7 @@ export const InfoSettings = ({ bondedNode }: { bondedNode: TBondedMixnode | TBon
             </Grid>
           </Grid>
         </Grid>
-        <Divider flexItem />
+        <Divider sx={{ width: '100%' }} />
```
fmtabbara, line 132: can you use `variant=fullWidth` instead of setting the width this way?

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx
```diff
@@ -62,7 +62,15 @@ export const InfoSettings = ({ bondedNode }: { bondedNode: TBondedMixnode | TBon
...
 
   return (
-    <Grid container xs item>
+    <Grid
+      container
+      xs
+      item
+      sx={{
+        paddingTop: '0 !important',
+        paddingLeft: '0 !important',
```
fmtabbara: can you remove the sx prop and try removing the `item` prop to achieve the 0 padding? you may also need to add `spacing={0}` but possibly not.

nym-wallet/src/theme/theme.tsx
```diff
@@ -303,6 +303,24 @@ export const getDesignTokens = (mode: PaletteMode): ThemeOptions => {
...
+                    t.palette.mode === 'dark' ? `${t.palette.background.default} !important` : '#FFFFFF !important',
+                },
+                '&& .Mui-selected:hover': {
+                  backgroundColor: 'rgba(251, 110, 78, 0.08) !important',
+                },
+              },
+            },
+          },
+        },
+      },
```
fmtabbara, line 323: I'm not sure how I feel about adding these global styles. I think the best solution is to create our own `StyledSelect` component with this styling BUT perhaps  leave it like this for now if it works, and if we need to change it later we will

gala1234 (approved):
LGTM but Fouad suggestions are interesting to try

nym-wallet/src/pages/bonding/node-settings/settings-pages/general-settings/InfoSettings.tsx
doums: It does not work for some reason

nym-wallet/src/theme/theme.tsx
doums: As discussed with Marcin we want the same design for select like input/menu everywhere. And I double checked there is only one usage of `Select` mui comp, so the style changes does not hurt. Placing the style override in `theme` is the right place for this case.

PR #3199, opened 2023-03-20
chore: tidy up client `Debug` config section

jstuczyn:
# Description

While working on dual packets in clients, I had to add yet another entry into their configs. And I've realised how messy they are so I decided to tidy the `Debug` section a bit and get it here as a separate PR.

### DIFF
--- a/clients/native/src/client/config/template.rs
+++ b/clients/native/src/client/config/template.rs
@@ -110,10 +110,15 @@ host = '{{ socket.host }}'
 
 [debug]
 
+[[debug.traffic]]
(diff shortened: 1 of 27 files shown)


clients/native/src/client/config/template.rs
```diff
@@ -110,10 +110,15 @@ host = '{{ socket.host }}'
 
 [debug]
 
+[[debug.traffic]]
```
octol: Is this a backwards incompatible change that will require re-init?

jstuczyn: old configs should now get updated during `run`

PR #3217, opened 2023-03-24
Feature/dual packet sizes

### LINKED ISSUE #3189: Introduce dual packet sizes to our clients (as in use two packet sizes at the same time depending on message size)

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/3189

This PR allows clients to optionally specify a `secondary_packet_size` for its messages. Right now deciding which packet size should be used is rather naive. We check how many packets would be required with `primary` packet size vs `secondary` packet size. We then choose the smaller option.

This should be of interest to @simonwicky .

This PR depends on https://github.com/nymtech/nym/pull/3199 so it can't be merged until that one is dealt with first.

# Checklist:

### DIFF
--- a/sdk/rust/nym-sdk/examples/simple.rs
+++ b/sdk/rust/nym-sdk/examples/simple.rs
@@ -18,4 +18,6 @@ async fn main() {
     client
         .on_messages(|msg| println!("Received: {}", String::from_utf8_lossy(&msg.message)))
         .await;
+
+    client.disconnect().await;
 }
--- a/sdk/rust/nym-sdk/examples/simple.rs
+++ b/sdk/rust/nym-sdk/examples/simple.rs
@@ -18,6 +18,4 @@ async fn main() {
     client
         .on_messages(|msg| println!("Received: {}", String::from_utf8_lossy(&msg.message)))
         .await;
-
-    client.disconnect().await;
 }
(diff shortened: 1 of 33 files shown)


clients/client-core/src/client/real_messages_control/mod.rs
```diff
@@ -45,61 +43,29 @@ pub struct Config {
...
-    /// poisson distribution.
-    disable_main_poisson_packet_distribution: bool,
-
-    /// Predefined packet size used for the encapsulated messages.
-    packet_size: PacketSize,
-
-    /// Defines the minimum number of reply surbs the client would request.
-    minimum_reply_surb_request_size: u32,
+    /// Specifies all traffic related configuration options.
+    traffic: config::Traffic,
```
octol: I like this tidying up

sdk/rust/nym-sdk/examples/simple.rs
```diff
@@ -18,4 +18,6 @@ async fn main() {
     client
         .on_messages(|msg| println!("Received: {}", String::from_utf8_lossy(&msg.message)))
         .await;
+
+    client.disconnect().await;
```
octol: Will this be called?

octol (approved):

sdk/rust/nym-sdk/examples/simple.rs
jstuczyn: thinking about it, probably not. I just added it following builder.rs example, but I didn't realise it's using different call for waiting for messages

PR #3218, opened 2023-03-24
Don't clone reconstructor

durch:
Fix a problem where cloning of reconstructor resulted in unreconstructable messages


octol (approved):
Works on my machine!

PR #3222, opened 2023-03-27
Feature/available reader changes

### LINKED ISSUE #3213: Make the buffer size in `AvailableReader` depend on packet sizes the client is using + introduce read timeouts
When the client transmission rate is small, measured at the Poisson process, then a large buffer size causes large latencies for the connected application when sending large chunks of data, make the connection unreliable.

Reducing the buffer size, ideally while making it a multiple of the capacity of the select sphinx packet size, should solve this

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/3213

# Checklist:

### DIFF
--- a/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
+++ b/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
@@ -167,6 +167,7 @@ pub(super) async fn run_inbound<F, S>(
     remote_source_address: String,
     connection_id: ConnectionId,
     mix_sender: MixProxySender<S>,
+    available_plaintext_per_mix_packet: usize,
     adapter_fn: F,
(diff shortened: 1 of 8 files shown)


common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
```diff
@@ -61,6 +61,8 @@ where
         None => (Default::default(), true),
     };
 
+    error!("{} bytes", read_data.len());
```
octol: Leftover?

octol (approved):

common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
jstuczyn: indeed : )

PR #3247, opened 2023-03-29
Sphinx packet processing CPU cycles measurement

durch:
Adds `measure` macro that calculates cpu cycles for a given block and publishing it to a tracing subscriber

Sample output, both to file and stdout, only stdout has the tree view: 
```
pr 04 20:28:19 localhost nym-mixnode[3962675]: nym_mixnode::node::listener::connection_handler::handle_received_packet{}
Apr 04 20:28:19 localhost nym-mixnode[3962675]:     mixnode_common::packet_processor::processor::process_received{}
Apr 04 20:28:19 localhost nym-mixnode[3962675]:         mixnode_common::packet_processor::processor::perform_initial_unwrapping{}
Apr 04 20:28:19 localhost nym-mixnode[3962675]:             mixnode_common::packet_processor::processor::perform_initial_sphinx_packet_processing{}
Apr 04 20:28:19 localhost nym-mixnode[3962675]:                 2ms INFO mixnode_common::packet_processor::processor perform_initial_sphinx_packet_processing cpucycles: 621038
Apr 04 20:28:19 localhost nym-mixnode[3962675]:
Apr 04 20:28:19 localhost nym-mixnode[3962675]:             3ms INFO mixnode_common::packet_processor::processor perform_initial_unwrapping cpucycles: 5882448
Apr 04 20:28:19 localhost nym-mixnode[3962675]:
Apr 04 20:28:19 localhost nym-mixnode[3962675]:         3ms INFO mixnode_common::packet_processor::processor process_received cpucycles: 7628368
Apr 04 20:28:19 localhost nym-mixnode[3962675]:
Apr 04 20:28:19 localhost nym-mixnode[3962675]:     4ms INFO nym_mixnode::node::listener::connection_handler handle_received_packet cpucycles: 8786712
```


common/bin-common/src/logging/mod.rs
```diff
@@ -24,6 +28,18 @@ pub fn setup_logging() {
         .init();
 }
 
+pub fn setup_tracing(file_name: &str) {
```
jstuczyn: if we're using a global `setup_tracing`, should it always log to a file? It sounds like something we'd want to make configurable

common/mixnode-common/src/lib.rs
```diff
@@ -3,3 +3,24 @@
 
 pub mod packet_processor;
 pub mod verloc;
+
+#[cfg(feature = "cpu-cycles")]
+pub fn cpu_cycles() -> i64 {
+    cpu_cycles::cpucycles().unwrap_or(0)
```
jstuczyn: idk, it's kinda nitpicky, but it seems like that feature lock would want to live inside the function rather than having to define it twice. also, maybe let's make it an `Option` to explicitly state whether it's available or not? (rather than being possibly confused by getting a `0` if something blows up)

jstuczyn (approved):

common/bin-common/src/logging/mod.rs
durch: We're always logging to file, or rather stdout, once we actually have some sort of a tracing collector, we can upgrade

common/mixnode-common/src/lib.rs
durch: Its actually a very good point!

PR #3255, opened 2023-03-30
Feature/sdk socks5

### LINKED ISSUE #3237: Expose the socks5 functionality in a nicer interface in the Rust SDK.

neacsu:
# Description

Closes: #3237

# Checklist:

### DIFF
--- a/sdk/rust/nym-sdk/examples/socks5.rs
+++ b/sdk/rust/nym-sdk/examples/socks5.rs
@@ -1,4 +1,5 @@
 use nym_sdk::mixnet;
+use nym_socks5_client_core::config::Socks5;
 
 #[tokio::main]
 async fn main() {
@@ -7,21 +8,16 @@ async fn main() {
     println!("Connecting receiver");
     let mut receiving_client = mixnet::MixnetClient::connect_new().await.unwrap();
 
-    let config = mixnet::Config {
-        socks5_service_provider: Some(receiving_client.nym_address().to_string()),
-        ..Default::default()
-    };
+    let socks5_config = Socks5::new(receiving_client.nym_address().to_string());
+    let client_config = mixnet::Config::new(None, Some(socks5_config), None);
     let sending_client = mixnet::MixnetClientBuilder::new()
-        .config(config)
+        .config(client_config)
         .build::<mixnet::EmptyReplyStorage>()
         .await
         .unwrap();
 
     println!("Connecting sender");
     let mut sending_client = sending_client.connect_to_mixnet().await.unwrap();
-    // wait until socks5 server is started
-    println!("Wait 5 seconds for the socks5 setup to be done");
-    std::thread::sleep(std::time::Duration::from_secs(5));
 
     let proxy = reqwest::Proxy::all(format!(
         "socks5h://127.0.0.1:{}",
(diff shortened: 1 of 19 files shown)


sdk/rust/nym-sdk/examples/socks5.rs
```diff
@@ -7,27 +7,17 @@ async fn main() {
...
         .build::<mixnet::EmptyReplyStorage>()
         .await
         .unwrap();
 
     println!("Connecting sender");
-    let mut sending_client = sending_client.connect_to_mixnet().await.unwrap();
-    // wait until socks5 server is started
-    println!("Wait 5 seconds for the socks5 setup to be done");
-    std::thread::sleep(std::time::Duration::from_secs(5));
+    let mut sending_client = sending_client.connect_to_mixnet_via_socks5().await.unwrap();
```
octol, line 18: The name `connect_to_mixnet_via_socks5` kinda makes the user think that the client is connecting to the mixnet via socks5 when it's connecting to the mixnet and exposing a socks5 endpoint. But I can't come up with a better name .. :man_shrugging:

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -415,193 +399,80 @@ where
...
+            client_output,
+            client_state.clone(),
+            nym_address,
+            started_client.task_manager.subscribe(),
+        );
+        log::debug!(
+            "Waiting for {} seconds for the socks5 channel to boot",
+            socks5_listener_boot_time_secs
+        );
+        std::thread::sleep(std::time::Duration::from_secs(
```
octol: No need to sleep, the socks5 client should send a status message once the connection is established. I think nym-connect listens to this to know when to show connected in the UI

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -610,80 +481,57 @@ impl MixnetClient {
...
-        let lane = TransmissionLane::General;
-        let input_msg = match surbs {
-            IncludedSurbs::Amount(surbs) => {
-                InputMessage::new_anonymous(address, message, surbs, lane)
-            }
-            IncludedSurbs::ExposeSelfAddress => InputMessage::new_regular(address, message, lane),
-        };
-        self.send(input_msg).await
-    }
+    pub async fn connect_to_mixnet(self) -> Result<MixnetClient>
```
octol, line 505: Should this function return an error if the socks5 config is set (which strongly indicates a user mistake)?

sdk/rust/nym-sdk/examples/socks5.rs
neacsu: While that is true, and there is technically a way for the user to talk directly to the mixnet, I specifically removed the communication channels with the mixnet from the user interface for this type of connection, so that the socks5 port is the only way for it to use the mixnet. But I agree, this is a bit of a head spinner and will need to be well documented. The `socks5` example is one way to start that, but more docs around that will probably also be needed.

sdk/rust/nym-sdk/src/mixnet/client.rs
neacsu: I was thinking that maybe there is a scenario where you would build the client with the socks5 config but later decide not to use it when connecting. But I guess that is just overreaching, making the two flows specifically distinct is probably a better idea. I'll make this throw out an error for that case.

Issue #3256, opened 2023-03-30
Bootstrap Browser Extension Application


fmtabbara:
Bootstrap application with 

- React
- Webpack
- Typescript
- Eslint
- Material UI
- Theme wrapper

PR #3257, opened 2023-03-30
Chore/browser extension bootstrap

fmtabbara:
# Description

Initialise the Browser Extension application

Closes: #3256

# Checklist:


nym-browser-extension/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "jest": "^27.1.0",
+    "mini-css-extract-plugin": "^2.2.2",
+    "prettier": "2.3.2",
+    "react-refresh": "^0.10.0",
+    "react-refresh-typescript": "^2.0.2",
+    "style-loader": "^3.3.1",
+    "ts-jest": "^27.0.5",
+    "ts-loader": "^9.4.2",
+    "tsconfig-paths-webpack-plugin": "^3.5.2",
+    "typescript": "^4.6.2",
```
doums, line 65: Not possible to use TS version 5.0* ?

nym-browser-extension/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "eslint-plugin-jsx-a11y": "^6.5.1",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react": "^7.29.2",
+    "eslint-plugin-react-hooks": "^4.3.0",
+    "eslint-plugin-storybook": "^0.5.12",
+    "favicons-webpack-plugin": "^5.0.2",
+    "html-webpack-plugin": "^5.3.2",
+    "jest": "^27.1.0",
+    "mini-css-extract-plugin": "^2.2.2",
+    "prettier": "2.3.2",
```
doums: not sure if there is a reason to pin this specific version, latest is `2.8.7`

nym-browser-extension/package.json
mmsinclair: Best to keep it the same as the rest of the monorepo, of there might be problems.

I want to leave TS 5 to settle a bit before upgrading everything, because tools like eslint, webpack TS plugins, ts-jest etc need to catch up

nym-browser-extension/package.json
mmsinclair: Again, @fmtabbara use the same version as elsewhere in the monorepo. You could also upgrade everything else to the latest version.

doums (approved):

PR #3273, opened 2023-04-04
Feature/sdk coconut

### LINKED ISSUE #2755: Rust SDK - Support coconut, credential storage etc

neacsu:
# Description

Closes: #2755

# Checklist:

- [x] introduce ephemeral coconut credential database
- [x] separate credential acquirement methods into separate crate
- [x] expose acquirement methods in the SDK


sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -45,8 +52,26 @@ impl MixnetClientBuilder {
     }
 
     #[must_use]
-    pub fn config(mut self, config: Config) -> Self {
-        self.config = Some(config);
+    pub fn gateway(mut self, user_chosen_gateway: String) -> Self {
```
octol: Looking good! Probably want some docs above this method just to clarify the difference between this and `fn gateway_config()` below

Now that I'm thinking about it, maybe it should be called `fn request_gateway()`? Or something similar?

And/or, `fn gateway_config()` should be renamed to `fn registered_gateway()`?

neacsu: I've applied your proposed changes and also added some more (simple) documentation on the other builder methods.

PR #3278, opened 2023-04-04
Replace mobile flag with target_os = android

### LINKED ISSUE #2975: Replace `mobile` flag with `target_os = android`

octol:
# Description

Closes: #2975

Replace the `mobile` feature flag with `target_os = "android"`. The `mobile` feature flag is problematic since it's not additive and breaks things like `--all-features`.

# Checklist:


nym-connect/mobile/src-tauri/src/lib.rs
doums, line 1: I think it's a bad idea to remove those, as `#[cfg(mobile)]`/`#[cfg(desktop)]` are provided by tauri itself.
Even it seems the way to go according to their [guide](https://next--tauri.netlify.app/next/mobile/development/integrate#conditional-compilation).
So, specifically for the ` nym-connect/mobile/src-tauri` code I think we should stick with tauri guidelines.

octol: oh I see, yeah didn't know those two came from tauri

PR #3282, opened 2023-04-05
gateway: cap the number of configs to keep

octol:
# Description

Cap the number of gateway configs we keep around by deleting old ones at startup.

# Checklist:


octol:
No this is not what we want at all

PR #3294, opened 2023-04-11
A branch with all clippy warnings dealt with in contracts

futurechimp:
# Description

We've got a lot of Clippy warnings in the monorepo (nothing major from a functionality or security point of view). Everything is clean on first glance, but if you open up e.g. a subproject like `contracts/mixnet` in vscode, Clippy lights up like a yellow Christmas tree. 

Part of this is because the `clippy.toml` from the root of the monorepo doesn't get evaluated - copying it into place gets rid of all the `.unwrap()`s in test code, so that's simple enough. 

This PR shows all the rest of Clippy fixes - some iterator-related things, lots of unnecessary `clone`s, and some unneeded mutability mostly.

### DIFF
--- a/contracts/coconut-bandwidth/src/storage.rs
+++ b/contracts/coconut-bandwidth/src/storage.rs
@@ -86,9 +86,9 @@ mod tests {
         assert!(res.is_none());
 
         let mut spend_credential = SpendCredential::new(
-            funds.clone(),
+            funds,
             blind_serial_number.to_string(),
-            gateway_cosmos_address.clone(),
+            gateway_cosmos_address,
         );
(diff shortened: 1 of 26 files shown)


octol:
I suspect a reason why these were overlooked is that in CI we run `cargo clippy` with  `--all-targets` for the root workspace, but not for the contracts workspace

octol:
Unless we're on a CI resources budget, imho we should add `--all-targets` here https://github.com/nymtech/nym/blob/1ad97adc7c3f0e623a3e21eff110823ddd2cbdba/.github/workflows/contracts.yml#L66

PR #3332, opened 2023-04-20
Service provider directory support in nym-api, nym-cli, validator-client

### LINKED ISSUE #3241: Cache service provider contract in nym-api
Internal changes to support #3242

### LINKED ISSUE #3242: Expose service-provider-directory contract data in nym-api endpoints
Add a new `services/` endpoint that lists all registered service providers.

### LINKED ISSUE #3296: Add service-provider-directory support to validator-client
See the note here on how we should go about testing this: https://github.com/nymtech/nym/pull/3332#issuecomment-1516155075

octol:
# Description

Closes: #3296
Closes: #3241
Closes: #3242
Closes: #3334 

- Add sp directory contract traits and methods to nyxd client
- Make the contract optional
- nym-api: add service provider endpoint and caching
- nym-cli: add query, announce and delete service provider


octol:
The way we should test this IMHO is the following:
1. Deploy `nym-api` from this branch, and make sure it works with qwerty where the service provider contract is already deployed, but crucially also that it works with mainnet where there is no contract yet. The service provider contract is supposed to be optional so we don't want anything breaking for networks where it's missing.
    1. For qwerty I'd expect the endpoints for the nym-api to be able to query announced services
    2. For mainnet, the endpoint should be there but it should return an empty list
    3. Announcing and deleting services is done via the `nym-cli`
1. Once the above is cleared, I think we should merge it into the release branch
1. When it's merged into the release branch we can start doing further more extensive testing of the service-provider contract in itself (which can be a follow up task)

common/client-libs/validator-client/src/nyxd/error.rs
```diff
@@ -22,7 +22,7 @@ use std::{io, time::Duration};
 #[derive(Debug, Error)]
 pub enum NyxdError {
     #[error("No contract address is available to perform the call")]
-    NoContractAddressAvailable,
+    NoContractAddressAvailable(String),
```
jstuczyn, line 25: If you're storing contract address in the error enum, you might as well return it in the error message

common/commands/src/validator/mixnet/operators/service/announce.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::context::SigningClient;
+
+#[derive(Debug, Parser)]
+pub struct Args {
+    #[clap(long)]
+    pub nym_address: String,
+
+    /// Deposit to be made to the service provider directory, in curent DENOMINATION (e.g. 'unym')
+    #[clap(long)]
+    pub deposit: u128,
```
jstuczyn, line 15: I wonder if it would make sense to change the `deposit` to be an explicit `Coin` so people wouldnt be confused by unym vs nym. But it's more of a note for the future

jstuczyn (approved):
lgtm

PR #3338, opened 2023-04-21
feat(wallet): update bond amount

### LINKED ISSUE #2134: Wallet - change "Bond more" action to "Change bond amount" to allow for increasing and decreasing the amount in the same modal
Andrew is working on the contract changes. Once done we can add "Change bond amount" to the action menu, where user should be able to increase or decrease the amount as long as it is not lower than minimum bond

Image

Image

Image

doums:
# Description

Closes: #2134


nym-wallet/src-tauri/src/state.rs
```diff
@@ -449,6 +449,30 @@ impl WalletStateInner {
...
+            std::cmp::Ordering::Less => {
+                let delta = DecCoin {
+                    amount: coin2.amount - coin1.amount,
+                    denom: coin1.denom.clone(),
+                };
+                Ok(delta)
+            }
+            std::cmp::Ordering::Equal => Ok(coin1.to_owned()),
+        }
+    }
```
doums, line 479: This could be potentially simplified:
Instead of coding a dedicated function to calculate the delta, I could simply calculate the sub then use the absolute value (one-liner). But `DecCoin` type does not implement any kind of `abs()`. But maybe we can implement that easily?

doums:
nym-wallet/src-tauri/src/operations/mixnet/bond.rs
nym-wallet/src-tauri/src/operations/simulate/mixnet.rs
nym-wallet/src-tauri/src/operations/simulate/vesting.rs
nym-wallet/src-tauri/src/operations/vesting/bond.rs

Yes the functions added in these 4 files `*update_pledge` are quite duplicated.
This could be improved by refactoring upper in the stack. For example at `nyxd` level? 

By implementing one "update_bond" function instead of providing 2functions pairs (one for decrease, one for increase).
Plus they get doubled for vesting support.
But I dont know if it's doable, and I found this refractor is quite out of the scope here

gala1234 (approved):

nym-wallet/src-tauri/src/error.rs
```diff
@@ -94,6 +94,8 @@ pub enum BackendError {
     WalletFileAlreadyExists,
     #[error("The wallet file is not found")]
     WalletFileNotFound,
+    #[error("Invalid update pledge request, no-op")]
```
octol: Consider replacing "no-op" with something that explains to the user why the pledge failed

nym-wallet/src-tauri/src/state.rs
```diff
@@ -449,6 +449,30 @@ impl WalletStateInner {
     pub fn remove_validator_url(&mut self, url: config::ValidatorConfigEntry, network: Network) {
         self.config.remove_validator_url(url, network)
     }
+
+    pub fn calculate_coin_delta(
+        &self,
+        coin1: &DecCoin,
+        coin2: &DecCoin,
+    ) -> Result<DecCoin, BackendError> {
+        match coin1.amount.cmp(&coin2.amount) {
```
octol, line 462: Maybe add a check that denoms are same on both coins

octol (approved):
LGTM

nym-wallet/src/context/main.tsx
```diff
@@ -82,6 +85,8 @@ export const AppProvider: FCWithChildren = ({ children }) => {
   const [isAdminAddress, setIsAdminAddress] = useState<boolean>(false);
   const [showSendModal, setShowSendModal] = useState(false);
   const [showReceiveModal, setShowReceiveModal] = useState(false);
+  const [printBalance, setPrintBalance] = useState<string>('-');
+  const [printVestedBalance, setPrintVestedBalance] = useState<string | undefined>();
```
fmtabbara, line 89: would this state be better off inside the useGetBalance hook?

nym-wallet/src/context/main.tsx
```diff
@@ -163,6 +168,18 @@ export const AppProvider: FCWithChildren = ({ children }) => {
...
+    if (userBalance.originalVesting) {
+      setPrintVestedBalance(`${toDisplay(userBalance.tokenAllocation?.spendableVestedCoins || 0)} ${currency}`);
+    }
+    if (userBalance?.balance?.amount) {
+      setPrintBalance(`${toDisplay(userBalance.balance.amount.amount)} ${currency}`);
+    } else {
+      setPrintBalance(`${toDisplay(0)} ${currency}`);
+    }
+  }, [userBalance, clientDetails]);
+
```
fmtabbara, line 182: would this state be better off inside the useGetBalance hook?

PR #3344, opened 2023-04-22
fix: add tokio in dependencies

sven-hash:


mfahampshire:
@sven-hash please target to `release/1.1.16` and please make sure to add the `documentation` tag to PRs in the future

mfahampshire (commented):
I dont really get why you're adding this dependency in there - if you're running the examples from the monorepo then this is already there. 

If we're talking about building from these examples then its a different question but I want to understand first

sven-hash:
I understand, how I did the "tutorial" was by copying the code snippet and the run them not by cloning the monorepo (mostly because of its size). So it was easier to just copy paste and then run them with `cargo run`

Maybe if you want people to run the example from the monorepo explain to clone it and the go to `sdk` folder where you can run them

mfahampshire:
I'll add a note: I thought that was obvious but obviously not! Nice catch.

mfahampshire:
[todo] add note + close

mfahampshire:
Closing in place of comment in #3370

PR #3362, opened 2023-04-26
refactor(wallet-bonding): fetch node data concurrently

doums:
# Description

Refactor the way requests are fired in the _Bonding_ page. Currently, optional backend queries (allowed to fails without blocking the whole page render) are fired sequentially. That means all requests are executed one by one, the next being executed only after the previous one has resolved/rejected (waiting for the server response). Slowing down the page and data loading.
The changes here implement concurrency mode (thx to [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)), meaning the requests are fired and resolve (or reject) concurrently in “parallel” without blocking between each others.

This improves overall load time of the _Bonding_ page ⚡

### DIFF
--- a/nym-wallet/src/context/bonding.tsx
+++ b/nym-wallet/src/context/bonding.tsx
@@ -186,18 +186,6 @@ export const BondingContextProvider: FCWithChildren = ({ children }): JSX.Elemen
     setBondedNode(undefined);
   };
 
-  // try {
-  //   const stakeSaturationResponse = await getMixnodeStakeSaturation(mixId);
-  //   additionalDetails.stakeSaturation = decimalToPercentage(stakeSaturationResponse.saturation);
-  //
-  //   const rawUncappedSaturation = decimalToFloatApproximation(stakeSaturationResponse.uncapped_saturation);
-  //   if (rawUncappedSaturation && rawUncappedSaturation > 1) {
-  //     additionalDetails.uncappedSaturation = Math.round(rawUncappedSaturation * 100);
-  //   }
-  // } catch (e) {
-  //   Console.log('getMixnodeStakeSaturation fails', e);
-  // }
-
   /**
    * Fetch mixnode **optional** data.
    * ⚠ The underlying queries are allowed to fail.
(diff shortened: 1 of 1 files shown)


nym-wallet/src/context/bonding.tsx
```diff
@@ -185,82 +186,145 @@ export const BondingContextProvider: FCWithChildren = ({ children }): JSX.Elemen
...
+    const promises = await Promise.allSettled([
+      requests.getMixnodeStatus(),
+      requests.getMixnodeUptime(),
+      requests.getMixnodeStakeSaturation(),
+      requests.getMixnodeRewardEstimation(),
+      requests.getInclusionProbability(),
+      requests.getMixnodeAvgUptime(),
+      requests.getNodeDescription(),
+      requests.getPendingOperatorRewards(),
+    ]);
```
fmtabbara:
```suggestion
    const requests = [
      () => getMixnodeStatus(mixId),
      () => getMixnodeUptime(mixId),
      () => getMixnodeStakeSaturation(mixId),
      () => getMixnodeRewardEstimation(mixId),
      () => getInclusionProbability(mixId),
      () => getMixnodeAvgUptime(),
      () => getNodeDescriptionRequest(host, port),
      () => getPendingOperatorRewards(clientDetails?.client_address || ''),
    ];

    const promises = await Promise.allSettled(requests);
```

nym-wallet/src/context/bonding.tsx
```diff
@@ -185,82 +186,145 @@ export const BondingContextProvider: FCWithChildren = ({ children }): JSX.Elemen
...
+
+    promises.forEach((res, index) => {
+      const reqNames = Object.keys(requests);
+      if (res.status === 'rejected') {
+        Console.warn(`${reqNames[index]} request fails`, res.reason);
+      }
+    });
```
fmtabbara:
```suggestion
    promises.forEach((res, index) => {
      if (res.status === 'rejected') {
        Console.warn('request fails', res.reason);
      }
    });
```

fmtabbara: I think the failure reason should give us some details about the request that failed?

nym-wallet/src/context/bonding.tsx
```diff
@@ -185,82 +186,145 @@ export const BondingContextProvider: FCWithChildren = ({ children }): JSX.Elemen
...
+
+    const promises = await Promise.allSettled([requests.getGatewayReport(), requests.getNodeDescription()]);
+
+    promises.forEach((res, index) => {
+      const reqNames = Object.keys(requests);
+      if (res.status === 'rejected') {
+        Console.warn(`${reqNames[index]} request fails`, res.reason);
+      }
+    });
+
```
fmtabbara: As the above suggestion

PR #3363, opened 2023-04-26
socks5: send empty keepalive msg to avoid triggering MIX_TTL during long downloads

### LINKED ISSUE #3215: Implement heartbeat messages between socks5 proxy and network requester
The Outbound Proxy has a timeout after which it closes. If it closes the InboundProxy follows 30 seconds later, even if data is still being transmitted. During long download, the socks5 client has nothing to send to the requester, which eventually causes this timeout to fire and the download is cut short. Adding small heartbeat message from the proxy to the network requester would prevent that.

octol:
# Description

Closes: https://github.com/nymtech/nym/issues/3215

During long downloads when the traffic can in some cases be entirely unidirectional, we risk trigger `MIX_TTL` while the connection is still alive.

This PR adds a time on which emoty empty keepalive messages are sent to keep a regular heartbeat going.

Co-author: @simonwicky, who investigated this issue

### DIFF
--- a/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
+++ b/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
@@ -36,7 +36,7 @@ async fn send_empty_close<F, S>(
         .expect("BatchRealMessageReceiver has stopped receiving!");
 }
 
-async fn send_empty_keepalive<F,S> (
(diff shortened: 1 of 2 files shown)


octol:
Closed due to wrong branch name used, correct PR is #3364

PR #3365, opened 2023-04-26
socks5: abort sending data if the connection is closed

### LINKED ISSUE #3366: socks5: sender waits for lanes to clear even though the connection is closed

octol:
# Description

Closes: #3366 

Make sure to stop sending as soon as the connection is closed. 

- abort if connection is closed while waiting for lane to clear
- prioritize closing connection before reading more data

Co-author: @simonwicky, who investiged this and came up with the solution

### DIFF
--- a/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
+++ b/common/socks5/proxy-helpers/src/proxy_runner/inbound.rs
@@ -53,7 +53,7 @@ async fn send_empty_keepalive<F, S>(
         .expect("BatchRealMessageReceiver has stopped receiving!");
 }
 
-#[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


octol:
Status: implementation is done, but the PR requires some systematic real-world testing before we can merge it in. I'll remove the milestone from the issue for now.

simonwicky (approved):
Looks good to me. 
Lanes related to download interruptions are cleared as expected, 30 sec after the download was interrupted.  
Normal downloads do not seem to be impacted in any way.  
This 30 sec can probably be reduced to 0 straight away (or with other proxy adjustments) which would further reduce the unavailability of a provider given someone is starting and cancelling lots of downloads. (scope of https://github.com/nymtech/nym/issues/3438)

PR #3373, opened 2023-05-02
Feature/nym browser extension login and send

fmtabbara:
# Description

Set up
App context
Connection to validator client
App connection config
Send page
Balance page
Required components

Closes: #3268 #3372

# Checklist:


nym-browser-extension/package.json
```diff
@@ -32,14 +35,18 @@
...
+    "@types/big.js": "^6.1.6",
+    "@types/crypto-js": "4.1.1",
     "@types/jest": "^27.0.1",
-    "@types/node": "^16.7.13",
+    "@types/node": "^18.16.1",
     "@types/react": "^18.0.26",
     "@types/react-dom": "^18.0.10",
     "@typescript-eslint/eslint-plugin": "^5.13.0",
     "@typescript-eslint/parser": "^5.13.0",
+    "big.js": "^6.2.1",
```
doums: is already included as regular dep

nym-browser-extension/src/routes/login.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { ForgotPassword, Login } from 'src/pages/auth';
+
+export const LoginRoutes = () => {
+  const { client } = useAppContext();
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    if (client) navigate('/user/balance');
+    else navigate('/login');
+  }, [client]);
```
doums, line 18: please put those curlybraces xD (without it, could be a pain for future addition in the if/else branches etc..)

nym-browser-extension/src/urls/index.ts
```diff
@@ -0,0 +1,4 @@
+export const blockExplorer = {
+  mainnet: 'https://blocks.nymtech.net',
+  qa: 'https://qwerty-blocks-explorer.qa.nymte.ch',
+};
```
doums: should we provide these from env var instead of hardcoding values?

nym-browser-extension/src/utils/coin.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const nymToUnym = (nym: number) => {
+  let unym;
+  try {
+    unym = Big(nym).mul(1_000_000);
+  } catch (e: any) {
+    console.warn(`unable to convert nym to unym: ${e}`);
+  }
+  return Number(unym);
+};
```
doums, line 21: ⚠ in case of error, `Number(undefined)` will returns  `NaN`

nym-browser-extension/yarn-error.log
doums, line 1: I think we can untrack and ignore it

nym-browser-extension/src/utils/coin.ts
fmtabbara: Should be OK here as the function can only be passed a number type?

doums: yes you right I feel bad now xD

doums (approved):

PR #3377, opened 2023-05-02
Use Loading Modal component when loading Delegations data

### LINKED ISSUE #3376: Wallet - keep the delegations page state while new data is being loaded

fmtabbara:
# Description

Show loading modal when updating delegations list.

Closes: #3376 

# Checklist:

### DIFF
--- a/nym-wallet/src/context/delegations.tsx
+++ b/nym-wallet/src/context/delegations.tsx
@@ -90,15 +90,7 @@ export const DelegationContextProvider: FC<{
     }
   };
 
-  const resetState = () => {
-    setError(undefined);
-    setTotalDelegations(undefined);
-    setTotalRewards(undefined);
-    setDelegations([]);
-  };
-
   const refresh = useCallback(async () => {
-    resetState();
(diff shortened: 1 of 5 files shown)


nym-wallet/src/context/delegations.tsx
```diff
@@ -90,15 +90,7 @@ export const DelegationContextProvider: FC<{
...
 
-  const resetState = () => {
-    setError(undefined);
-    setTotalDelegations(undefined);
-    setTotalRewards(undefined);
-    setDelegations([]);
-  };
-
   const refresh = useCallback(async () => {
-    resetState();
```
doums, line 101: I'm a bit concerned here. What happens if an error occurs for some reason (and so `setError("something")` is fired)? I think the error state will remain set even if the `refresh` is called.

fmtabbara: It's a good point. I just checked and the error is not being used anywhere. I've removed the error state as it's unused and now logging any errors that occur.

doums (approved):

PR #3384, opened 2023-05-04
Add name-service support to validator-client

octol:
# Description

Part of: #3355

Add support for nym-name-service contract to the validator-client

# Checklist:


nym-wallet/nym-wallet-types/src/network.rs
```diff
@@ -115,6 +115,7 @@ mod sandbox {
     pub(crate) const COCONUT_DKG_CONTRACT_ADDRESS: &str =
         "nymt1k8re7jwz6rnnwrktnejdwkwnncte7ek7kk6fvg";
     pub(crate) const SERVICE_PROVIDER_DIRECTORY_CONTRACT_ADDRESS: &str = "";
+    pub(crate) const NAME_SERVICE_CONTRACT_ADDRESS: &str = "";
```
jstuczyn: if it doesn't exist, does the entry even need to be here? won't it just be parsed into a `None` instead?

octol: Yep we can explicitly set it to `None`. I'll do that

PR #3404, opened 2023-05-10
nym-cli: support name-service contract

### LINKED ISSUE #3355: Support functionality for nym-name-service (nym-api, nym-cli, etc)

octol:
# Description

Closes: #3355

Add support to nym-cli for querying, registering, and deleting entries in the name service contract.

### DIFF
--- a/common/commands/src/validator/mixnet/query/query_all_names.rs
+++ b/common/commands/src/validator/mixnet/query/query_all_names.rs
@@ -1,0 +1,50 @@
+// Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::Parser;
+use comfy_table::Table;
+use nym_validator_client::nym_api::error::NymAPIError;
+
+use crate::context::QueryClientWithNyxd;
+use crate::utils::show_error;
+
+#[derive(Debug, Parser)]
+pub struct Args {
+    #[clap(value_parser)]
+    #[clap(help = "Optionally, the registered name to display")]
+    pub name: Option<String>,
+}
+
+pub async fn query(args: Args, client: &QueryClientWithNyxd) {
+    match client.nym_api.get_registered_names().await {
+        Ok(res) => {
+            if let Some(name) = args.name {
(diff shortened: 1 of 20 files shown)


common/commands/src/validator/mixnet/query/query_all_names.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[clap(value_parser)]
+    #[clap(help = "Optionally, the registered name to display")]
+    pub name: Option<String>,
+}
+
+pub async fn query(args: Args, client: &QueryClientWithNyxd) {
+    match client.nym_api.get_registered_names().await {
+        Ok(res) => {
+            if let Some(name) = args.name {
+                let name = res.iter().find(|name_entry| {
```
jstuczyn: I wonder if perhaps the contract should let you query by name natively given it (I think) has indexes on those

jstuczyn (approved):

common/commands/src/validator/mixnet/query/query_all_names.rs
octol: yep the contract allows you to do that, but this is the nym-api and currently it's only returning a list of all (cached) names

common/commands/src/validator/mixnet/query/query_all_names.rs
```diff
@@ -0,0 +1,53 @@
...
+                        name_entry.name.address.to_string(),
+                        name_entry.name.name.to_string(),
+                    ]);
+                }
+
+                println!("The registered names in the directory are:");
+                println!("{table}");
+            }
+        }
+        Err(NymAPIError::NotFound) => {
```
jstuczyn, line 48: should this error be even possible?

jstuczyn (approved):

common/commands/src/validator/mixnet/query/query_all_names.rs
octol: Currently on mainnet it does, for example

octol: Since that endpoint isn't (yet) deployed

Issue #3293, opened 2023-04-11
eWallet - create an account and add a password


MarcinSN:
![Image](https://user-images.githubusercontent.com/95295201/232067115-6a2e8fd2-df72-4d14-9d62-519569e36492.png)

figma file: https://www.figma.com/file/rZIsG8APtLQOVBsvpPnZXZ/Nym-Extension-Wallet?node-id=419%3A12948&t=6STwMBWJAR9Trh72-1

Ask Marcin for password to figma file

PR #3442, opened 2023-05-23
Feature/nym browser extension password encryption (single account)

fmtabbara:
# Description

Sets up use of wasm to access indexeddb 
Indexeddb used for storing encrypted mnemonics

Closes: #3293 

# Checklist:


nym-browser-extension/src/context/register.tsx
```diff
@@ -0,0 +1,44 @@
...
+  importExistingAccount: (password: string) => Promise<void>;
+};
+
+export const RegisterContextProvider = ({ children }: { children: React.ReactNode }) => {
+  const [userPassword, setUserPassword] = useState('');
+  const [userMnemonic, setUserMnemonic] = useState('');
+
+  const createAccount = async (mnemonic: string, password: string) => {
+    const storage = await new ExtensionStorage(password);
+    await storage.store_mnemonic('Default account', mnemonic);
```
jstuczyn, line 21: When we're creating new account, shouldn't we also provide the account name? Because with this you might end up accidentally overwriting existing  data

nym-browser-extension/storage/Makefile
```diff
@@ -1,2 +1,2 @@
 wasm-pack:
-	wasm-pack build --scope nymproject #--target no-modules
\ No newline at end of file
+	wasm-pack build
```
jstuczyn: why did we remove the `nymproject` scope?

nym-browser-extension/src/context/register.tsx
fmtabbara: The designs don't request an account name on initial set up. After the user has installed the extension they can rename this initial account + any further accounts that are added will require a name to be added.

Multiple accounts will be the next PR.

nym-browser-extension/storage/Makefile
fmtabbara: I wasn't sure if it was necessary? I can put it back if it's required, but everything seems to be working without it?

nym-browser-extension/src/context/register.tsx
jstuczyn: So I guess (_makes a personal note_) you might need additional method on the storage like `already_contains`?

nym-browser-extension/storage/Makefile
jstuczyn: I think this would add the `@nymproject` at the beginning of import paths so it's explicit its ours. Whether we need it or not, I'm not sure. Perhaps @mmsinclair would have an opinion.

fmtabbara: Ah OK, I see. I'll run it past Mark. I think it makes sense to have it if it's likely to get consumed in other packages.

nym-browser-extension/src/context/register.tsx
fmtabbara: Yes! That would be useful!

nym-browser-extension/src/hooks/useCreatePassword.tsx
doums, line 1: Will there be additional logics in this hook in the future?

doums (approved):

nym-browser-extension/src/context/register.tsx
jstuczyn: done in https://github.com/nymtech/nym/pull/3462

nym-browser-extension/src/hooks/useCreatePassword.tsx
fmtabbara: Possibly. The reason I created it is because there are two pages that use the exact same logic

PR #3459, opened 2023-05-26
Sign when announcing service providers to the directory contract

### LINKED ISSUE #3360: service-provider-directory: add signature check when announcing
Add signature check when announcing a service provider the same way we do with gateways and mixnodes, to ensure only the service provider operator is able to announce a service

This will likely change the invariants on allowed entries etc

octol:
# Description

Closes: #3360

Require signing when announcing service providers to the service provider directory contract.

# Checklist:

- [x] Create signable payload in nym-cli
- [x] Sign payload in nym-network-requester
- [x] Contract changes to verify signature


common/commands/src/validator/mixnet/mod.rs
```diff
@@ -14,6 +14,7 @@ pub struct Mixnet {
     pub command: MixnetCommands,
 }
 
+#[allow(clippy::large_enum_variant)]
```
jstuczyn, line 17: should we be allowing this? or would it perhaps make sense to just Box the problematic variants?

common/commands/src/validator/mixnet/operators/mod.rs
```diff
@@ -15,6 +15,7 @@ pub struct MixnetOperators {
     pub command: MixnetOperatorsCommands,
 }
 
+#[allow(clippy::large_enum_variant)]
```
jstuczyn, line 18: ibid.

common/cosmwasm-smart-contracts/service-provider-directory/Cargo.toml
```diff
@@ -7,5 +7,9 @@ edition = "2021"
 
 [dependencies]
 cosmwasm-std = { workspace = true }
+nym-contracts-common = { path = "../contracts-common", version = "0.4.0" }
 schemars = "0.8"
 serde = { workspace = true, features = ["derive"] }
+thiserror = { version = "1.0.23" }
```
jstuczyn: isn't `thiserror` in our workspace?

common/cosmwasm-smart-contracts/service-provider-directory/src/types.rs
```diff
@@ -7,28 +7,41 @@ use serde::{Deserialize, Serialize};
 /// The directory of services are indexed by [`ServiceId`].
 pub type ServiceId = u32;
 
+pub type IdentityKey = String;
```
jstuczyn: If you have to redefine our existing types, perhaps they should be moved to some contract-common crate instead?

contracts/name-service/src/state/names.rs
```diff
@@ -691,7 +692,6 @@ mod tests {
     #[test]
     #[ignore]
     fn max_page_limit_is_applied() {
-        // WIP(JON)
         todo!();
```
jstuczyn, line 695: : )

contracts/service-provider-directory/Cargo.toml
```diff
@@ -22,5 +23,8 @@ thiserror = "1.0.39"
 vergen = { version = "=7.4.3", default-features = false, features = ["build", "git", "rustc"] }
 
 [dev-dependencies]
-cw-multi-test = { workspace = true }
 anyhow = "1.0.40"
+cw-multi-test = { workspace = true }
+nym-crypto = { path = "../../common/crypto", features = ["asymmetric", "rand"] }
+rand_chacha = "0.2"
+rstest = "0.17.0"
```
jstuczyn, line 30: I've just looked up this crate and it looks really cool! I need to remember to use it myself

contracts/service-provider-directory/src/contract/execute.rs
```diff
@@ -74,47 +75,98 @@ fn return_deposit(service_to_delete: &Service) -> BankMsg {
...
+
+    // reconstruct the payload
+    let nonce = state::get_signing_nonce(deps.storage, sender.clone())?;
+
+    let msg = construct_service_provider_announce_sign_payload(nonce, sender, deposit, service);
+
+    if deps.api.verify_message(msg, signature, &public_key)? {
+        Ok(())
+    } else {
+        Err(SpContractError::InvalidEd25519Signature)
```
jstuczyn, line 99: Another thing we had to redefine, maybe at some point (not this PR!) we should move signing helpers and associated errors to separate crate?

contracts/service-provider-directory/src/contract/execute.rs
```diff
@@ -128,15 +180,12 @@ pub(crate) fn delete_nym_address(
     let services_to_delete = query::query_nym_address(deps.as_ref(), nym_address)?.services;
```
jstuczyn, line 183: Just making sure I understood it correctly, the number of  services that could be returned by this is bound by `MAX_NUMBER_OF_ALIASES_FOR_NYM_ADDRESS`, right?

contracts/service-provider-directory/src/state/nonce.rs
```diff
@@ -0,0 +1,71 @@
...
+use nym_contracts_common::signing::Nonce;
+
+pub const NONCES: Map<'_, Addr, Nonce> = Map::new(SIGNING_NONCES_NAMESPACE);
+
+pub fn get_signing_nonce(storage: &dyn Storage, address: Addr) -> Result<Nonce> {
+    let nonce = NONCES.may_load(storage, address)?.unwrap_or(0);
+    Ok(nonce)
+}
+
+pub fn update_signing_nonce(storage: &mut dyn Storage, address: Addr, value: Nonce) -> Result<()> {
```
jstuczyn: under what conditions would we want to update the nonce to an arbitrary value as opposed to only incrementing it?

contracts/service-provider-directory/src/state/services.rs
```diff
@@ -55,39 +55,38 @@ pub fn has_service(store: &dyn Storage, service_id: ServiceId) -> bool {
...
-    store: &dyn Storage,
-    nym_address: NymAddress,
-) -> Result<Vec<(ServiceId, Service)>> {
+pub fn load_nym_address(store: &dyn Storage, nym_address: NymAddress) -> Result<Vec<Service>> {
     let services = services()
         .idx
         .nym_address
         .prefix(nym_address.to_string())
         .range(store, None, None, Order::Ascending)
         .take(MAX_NUMBER_OF_ALIASES_FOR_NYM_ADDRESS as usize)
```
jstuczyn, line 81: given that clients shouldn't be allowed to store more than `MAX_NUMBER_OF_ALIASES_FOR_NYM_ADDRESS` values, do we even need that `take` here?

contracts/service-provider-directory/src/test_helpers/signing.rs
```diff
@@ -0,0 +1,41 @@
...
+    owner: &str,
+    service: ServiceDetails,
+    deposit: Coin,
+) -> SignableServiceProviderAnnounceMsg {
+    let owner = Addr::unchecked(owner);
+    let nonce = state::get_signing_nonce(deps.storage, owner.clone()).unwrap();
+    construct_service_provider_announce_sign_payload(nonce, owner, deposit, service)
+}
+
+pub fn ed25519_sign_message<T: Serialize + SigningPurpose>(
```
jstuczyn, line 27: yeah, I think we should definitely have some common crate...

jstuczyn: actually, don't we already have `common/cosmwasm-smart-contracts/contracts-common/src/signing`? i.e. where we define `SigningPurpose`  trait.  Maybe we could put those things there?

service-providers/network-requester/src/cli/mod.rs
```diff
@@ -111,6 +117,36 @@ fn try_upgrade_v1_1_13_config(id: &str) -> std::io::Result<()> {
     updated.save_to_file(None)
 }
 
+// this only checks compatibility between config the binary. It does not take into consideration
+// network version. It might do so in the future.
+fn version_check(cfg: &Config) -> bool {
```
jstuczyn, line 122: I think those in-binary version checks are rather useless at this point, but I guess that's a topic for another discussion

common/commands/src/validator/mixnet/mod.rs
octol: Yes, possibly. Though I this case I think the performance hit isn't an issue

common/cosmwasm-smart-contracts/service-provider-directory/Cargo.toml
octol: yes good catch

contracts/service-provider-directory/src/contract/execute.rs
octol: Correct

contracts/service-provider-directory/src/state/nonce.rs
octol: Good question, I just copied how it's done in the mixnet contract. It's in a private mod though

contracts/service-provider-directory/src/state/services.rs
octol: In theory not, but defensive coding

contracts/service-provider-directory/src/state/nonce.rs
jstuczyn: then I think  it's only used in tests and maybe I forgot to mark it as such

common/cosmwasm-smart-contracts/service-provider-directory/src/types.rs
octol: Yes, done!

contracts/service-provider-directory/src/state/nonce.rs
octol: Remove the `pub` on the function anyway

jstuczyn (approved):

PR #3496, opened 2023-06-01
fix(ns5-android): make lib calling callbacks

### LINKED ISSUE #3495: NC native android - fix callback calls

doums:
# Description

Closes: #3495

### DIFF
--- a/sdk/lib/socks5-listener/src/android.rs
+++ b/sdk/lib/socks5-listener/src/android.rs
@@ -1,17 +1,11 @@
+use crate::ClientState;
 use ::safer_ffi::prelude::*;
 use jni::{
     objects::{JClass, JObject, JString},
+    sys::jint,
     JNIEnv,
 };
-use safer_ffi::char_p::char_p_boxed;
-use safer_ffi::closure::{RefDynFnMut0, RefDynFnMut1};
-use std::{thread, time};
-
-extern "C" fn placeholder_startup_cb(address: char_p::Box) {
-    crate::rust_free_string(address)
-}
-
-extern "C" fn placeholder_shutdown_cb() {}
+use std::sync::{Arc, Mutex};
 
 fn init_jni_logger() {
     use android_logger::{Config, FilterBuilder};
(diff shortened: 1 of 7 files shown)


sdk/lib/socks5-listener/src/android.rs
```diff
@@ -47,53 +42,50 @@ pub unsafe extern "C" fn Java_net_nymtech_nyms5_NymProxy_run(
 
     log::debug!("using sp {}", sp_input);
 
-    // TODO pass this callback to blocking_run_client
-    env.call_method(cb_object, "onStart", "()V", &[])
-        .expect("failed to call Java callbacks");
-
     let service_provider = char_p::new(sp_input.as_str());
+
+    let arced = Arc::new(Mutex::new(env));
```
jstuczyn, line 47: You'd probably want to use tokio's async aware mutexes because the std ones could theoretically block the runtime

sdk/lib/socks5-listener/src/lib.rs
```diff
@@ -107,13 +103,15 @@ pub fn get_client_state() -> ClientState {
...
     storage_directory: Option<char_p::Ref<'_>>,
     service_provider: Option<char_p::Ref<'_>>,
-    on_start_callback: StartupCallback<'static>,
-    on_shutdown_callback: ShutdownCallback<'static>,
-) {
+    on_start_callback: F,
+    on_shutdown_callback: S,
+) where
+    F: FnMut(String) + Send + 'static,
+    S: FnMut() + Send + 'static,
```
jstuczyn, line 113: nitpicky: does it have to be an `FnMut` or could we get away with a normal `Fn` ( or maybe even something like we had initially, i.e. `extern "C" fn`)?

doums: `FnMut` is needed as for android we are using JNI to call java methods and the signature takes a `&mut self`
https://docs.rs/jni/0.21.1/jni/struct.JNIEnv.html#method.call_method

> or maybe even something like we had initially, i.e. extern "C" fn

Im not sure, can you share a snippet using `extern "C" fn` please?

sdk/lib/socks5-listener/src/android.rs
doums: Im not so sure, reading the ref https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use I think using async mutex here does not worth it?
what do you think?

sdk/lib/socks5-listener/src/lib.rs
jstuczyn: Ah, if we need it for android, then the other way will definitely not work

PR #3503, opened 2023-06-04
Replace openssl with rustls

octol:
# Description

CT-24

Explore replacing openssl by rustls across the main workspace, mostly to avoid the shared library dependency.

```
❯ ldd target/x86_64-unknown-linux-gnu/debug/nym-socks5-client
	linux-vdso.so.1 (0x00007ffe12363000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f7156f90000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f7154d21000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f7154b40000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7156fcf000)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6450,7 +6450,6 @@ dependencies = [
  "nym-credential-storage",
  "nym-crypto",
  "nym-socks5-client-core",
- "openssl",
  "rand 0.7.3",
  "safer-ffi",
  "serde",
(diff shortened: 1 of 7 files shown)


octol:
This was mostly a test, we should investigate making this switch (#3511), but for now lets start by linking in openssl statically #3504

octol:
Reviving this to explore switching to rustls to increase portability across platforms

PR #3507, opened 2023-06-05
Reduce SHUTDOWN_TIMEOUT to 3 sec

### LINKED ISSUE #3438: Investigate if we need to lower `SHUTDOWN_TIMEOUT` in  socks5 to zero (or almost zero)

octol:
# Description

Closes: #3438

Can't think of a scenario where we don't want to close one at the same
time as the other, but let's be conservative and keep it a very low
number for now.


tommyv1987 (approved):

jstuczyn (approved):
makes sense to me.  🚢  it

PR #3530, opened 2023-06-08
fix: pop handler to remove listener

sven-hash:
# Description

`unshift()` will not remove listener, it will just add them in it. `pop()` is a quick workaround to solve this

### DIFF
--- a/sdk/typescript/packages/sdk/src/mixnet/wasm/index.ts
+++ b/sdk/typescript/packages/sdk/src/mixnet/wasm/index.ts
@@ -69,31 +69,31 @@ export const createNymMixnetClient = async (options?: {
     subscribeToConnected: (handler) => {
       getSubscriptions<ConnectedEvent>(EventKinds.Connected).push(handler);
       return () => {
-        getSubscriptions<ConnectedEvent>(EventKinds.Connected).unshift(handler);
+        getSubscriptions<ConnectedEvent>(EventKinds.Connected).pop();
       };
     },
     subscribeToLoaded: (handler) => {
       getSubscriptions<LoadedEvent>(EventKinds.Loaded).push(handler);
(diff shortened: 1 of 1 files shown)


mmsinclair:
Thanks for the PR @sven-hash and apologies for the slow response.

mmsinclair:
I don't think this fix will work in the case of multiple handlers, have you tried using the "unregister function" you get from subscribing?

I want to add some unit tests for this before anything gets changed, so if you wouldn't mind waiting until we do that? It will definitely be in the next few days as work is ongoing in the SDK at the moment.

sven-hash:
> I don't this fix will work in the case of multiple handlers, have you tried using the "unregister function" you get from subscribing?
> 
> I want to add some unit tests for this before anything gets changed, so if you wouldn't mind waiting until we do that? It will definitely be in the next few days as work is ongoing in the SDK at the moment.

Thanks for your answer, it was a quick fix because we need it on our side and we weren't able to deploy a custom version of the wasm nym-client (even by following the DEVELOPPERS.md) meaning we weren't able to test a better but more "complex" solution

And no the unregister is not working because it use `unshift()` function who's not removing a listener but adding an element at the end of the list (see [firefox doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift))

Will wait for the new version of the SDK to be available then, and let us know if you need anything on our side

mmsinclair:
@sven-hash this should be fixed since #3659 and there is a unit test for it here: https://github.com/nymtech/nym/blob/develop/sdk/typescript/packages/sdk/src/mixnet/wasm/subscriptions.test.ts#L20

If you're still having trouble with unsubscribing let us know by opening a new issue.

PR #3531, opened 2023-06-08
Allow query network-requestor for open_proxy

### LINKED ISSUE #3461: network-requester: support report if they run an open proxy using `ControlRequest` API
There is an existing interface in the service-providers that can be queried. Implement the ability to query if a given network-requester is an open-proxy.

octol:
# Description

Closes: #3461

Add requests for being able to query a running network-requester across the
mixnet if it is an open proxy.

### DIFF
--- a/common/socks5/requests/src/lib.rs
+++ b/common/socks5/requests/src/lib.rs
@@ -32,6 +32,11 @@ pub enum Socks5RequestError {
 
     #[error(transparent)]
     ProviderInterfaceError(#[from] ServiceProviderMessagingError),
+
+    #[error("received unsupported request protocol version: {protocol_version:?}")]
+    UnsupportedProtocolVersion {
+        protocol_version: <Socks5Request as interface::ServiceProviderRequest>::ProtocolVersion,
+    },
 }
 
 #[cfg(test)]
--- a/common/socks5/requests/src/lib.rs
+++ b/common/socks5/requests/src/lib.rs
@@ -33,10 +33,17 @@ pub enum Socks5RequestError {
     #[error(transparent)]
     ProviderInterfaceError(#[from] ServiceProviderMessagingError),
 
-    #[error("received unsupported request protocol version: {protocol_version:?}")]
+    #[error("received unsupported request protocol version: {protocol_version}")]
     UnsupportedProtocolVersion {
         protocol_version: <Socks5Request as interface::ServiceProviderRequest>::ProtocolVersion,
     },
+}
+
+fn make_bincode_serializer() -> impl bincode::Options {
+    use bincode::Options;
+    bincode::DefaultOptions::new()
+        .with_big_endian()
+        .with_varint_encoding()
 }
 
 #[cfg(test)]
--- a/common/socks5/requests/src/request.rs
+++ b/common/socks5/requests/src/request.rs
@@ -15,6 +15,7 @@ pub type RemoteAddress = String;
 pub enum RequestFlag {
     Connect = 0,
     Send = 1,
+    OpenProxy = 2,
 }
 
 impl TryFrom<u8> for RequestFlag {
@@ -24,6 +25,7 @@ impl TryFrom<u8> for RequestFlag {
         match value {
             _ if value == (RequestFlag::Connect as u8) => Ok(Self::Connect),
             _ if value == (RequestFlag::Send as u8) => Ok(Self::Send),
+            _ if value == (RequestFlag::OpenProxy as u8) => Ok(Self::OpenProxy),
             value => Err(RequestDeserializationError::UnknownRequestFlag { value }),
         }
     }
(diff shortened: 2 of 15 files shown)


common/socks5/requests/src/lib.rs
```diff
@@ -32,6 +32,11 @@ pub enum Socks5RequestError {
 
     #[error(transparent)]
     ProviderInterfaceError(#[from] ServiceProviderMessagingError),
+
+    #[error("received unsupported request protocol version: {protocol_version:?}")]
```
jstuczyn: I think `ProtocolVersion` has a `Display`, no? (unless I'm misremembering)

common/socks5/requests/src/request.rs
```diff
@@ -288,6 +322,16 @@ impl Socks5RequestContent {
...
                 }))
             }
+            RequestFlag::Query => {
+                //let query = bincode::deserialize(&b[1..]).expect("WIP(JON)");
+                use bincode::Options;
+                let query = bincode::DefaultOptions::new()
+                    .with_big_endian()
+                    .with_varint_encoding()
+                    .deserialize(&b[1..])
+                    .expect("WIP(JON)");
```
jstuczyn: don't forget to handle that : )

common/socks5/requests/src/request.rs
```diff
@@ -317,6 +361,19 @@ impl Socks5RequestContent {
...
                 .collect(),
+
+            Socks5RequestContent::Query(query) => {
+                //let query_bytes: Vec<u8> = bincode::serialize(&query).expect("WIP(JON)");
+                use bincode::Options;
+                let query_bytes: Vec<u8> = bincode::DefaultOptions::new()
+                    .with_big_endian()
+                    .with_varint_encoding()
+                    .serialize(&query)
+                    .expect("WIP(JON");
```
jstuczyn: ibid.

common/socks5/requests/src/response.rs
```diff
@@ -175,6 +189,17 @@ impl Socks5ResponseContent {
                     .chain(res.into_bytes().into_iter())
                     .collect()
             }
+            Socks5ResponseContent::Query(query) => {
+                use bincode::Options;
+                let query_bytes: Vec<u8> = bincode::DefaultOptions::new()
+                    .with_big_endian()
+                    .with_varint_encoding()
+                    .serialize(&query)
+                    .expect("WIP(JON");
```
jstuczyn: ibid.

common/socks5/requests/src/response.rs
```diff
@@ -175,6 +189,17 @@ impl Socks5ResponseContent {
                     .chain(res.into_bytes().into_iter())
                     .collect()
             }
+            Socks5ResponseContent::Query(query) => {
+                use bincode::Options;
+                let query_bytes: Vec<u8> = bincode::DefaultOptions::new()
```
jstuczyn: it seems we're creating that `bincode::DefaultOptions::new().with_big_endian().with_varint_encoding()` in multiple places. maybe we should extract it to a separate `make_bincode_serializer` or `make_queries_serializer` function so it'd be easier to update it (if we ever decided to)

common/socks5/requests/src/response.rs
```diff
@@ -318,6 +352,12 @@ impl ConnectionError {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum QueryResponse {
+    OpenProxy(bool),
```
jstuczyn, line 361: it's just a personal bias, but I'm not the biggest fan of tuples in this context, imo named fields would be better so once/if we add more fields, we'd still know what corresponds to what

service-providers/network-requester/examples/query.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    //nym_bin_common::logging::setup_logging();
+    let mut client = MixnetClient::connect_new().await.unwrap();
+    //let provider: Recipient = "AN8eLxYWFitCkMn92zim3PrPszxJZDYyFFKP7qnnAAew.8UAxL3LwQBis6WpM3GGXaqKGaVdnLCpGJWumHT6KNdTH@77TSuVU8d1oXKbPzjec2xh4i3Wj5WwUyy9Lr36sm8gZm".parse().unwrap();
+    let provider: Recipient = "85xZUGsMi8koFGXnoKQpn3wkcajSghu99uW3P4aQqP3E.D5PdTNtPZnek28BsctU1pYGFzMSTx9z7g9EL9ryjDB5Z@9xJM74FwwHhEKKJHihD21QSZnHM2QBRMoFx9Wst6qNBS".parse().unwrap();
```
jstuczyn: I wonder if examples should have placeholder provider addresses for users to enter it manually because there's no guarantee what you entered here is going to be valid in a week/month, etc.

jstuczyn (approved):
It was indeed quite straightforward as you mentioned

common/socks5/requests/src/lib.rs
octol: ah yes, it has

service-providers/network-requester/examples/query.rs
octol: Yep idk what's best, it's kinda assumed that the example runner will need to lookup a network-requester manually

common/socks5/requests/src/request.rs
```diff
@@ -108,6 +124,13 @@ impl Serializable for Socks5Request {
         }
 
         let protocol_version = Socks5ProtocolVersion::from(b[0]);
+        if protocol_version > Self::max_supported_version() {
+            return Err(Socks5RequestError::UnsupportedProtocolVersion { protocol_version });
+        }
```
octol, line 129: @jstuczyn I added this after your review, what you think of it?

PR #3599, opened 2023-06-27
Remove unmaintained upgrade commands

### LINKED ISSUE #3600: Remove (or start maintaining) `upgrade` commands from all binaries

octol:
# Description

Closes: #3600 

Remove `upgrade` commands for `nym-mixnode`, `nym-gateway`, `nym-client`,
`nym-socks5-client`, since they are not maintained and keep confusing users.
Old configuration files are automatically converted when encountered.


gateway/src/config/mod.rs
```diff
@@ -159,6 +159,7 @@ impl Config {
         self
     }
 
+    #[allow(unused)]
```
jstuczyn: I guess since it was only used for upgrades, lets remove it? We can always add it back in if we need it

jstuczyn (approved):

PR #3616, opened 2023-06-28
Feature/gh action strapi

doums:
# Description

introduces a new GH workflow (reusable) to publish release data on strapi

### DIFF
--- a/.github/workflows/nym-connect-publish-macos.yml
+++ b/.github/workflows/nym-connect-publish-macos.yml
@@ -15,8 +15,15 @@ jobs:
       fail-fast: false
       matrix:
         platform: [macos-latest]
-
     runs-on: ${{ matrix.platform }}
+
+    outputs:
+      release_id: ${{ steps.create-release.outputs.id }}
+      release_date: ${{ fromJSON(steps.create-release.outputs.assets)[0].published_at }}
+      version: ${{ steps.release-info.outputs.version }}
+      filename: ${{ steps.release-info.outputs.filename }}
+      file_hash: ${{ steps.release-info.outputs.file_hash }}
+
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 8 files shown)


.github/workflows/nym-connect-publish-macos.yml
```diff
@@ -81,10 +88,38 @@ jobs:
...
+        name: Prepare release info
+        run: |
+          semver="${${{ github.ref_name }}##nym-connect-}" && semver="${semver##v}"
+          echo "version=$semver" >> "$GITHUB_OUTPUT"
+          echo "filename=nym-connect_$version_x64.dmg" >> "$GITHUB_OUTPUT"
+          echo "file_hash=${{ hashFiles('nym-connect/desktop/target/release/bundle/dmg/nym-connect_*_x64.dmg') }}" >> "$GITHUB_OUTPUT"
+
+  push-release-data:
+    if: ${{ (startsWith(github.ref, 'refs/tags/nym-connect-') && github.event_name == 'release') || github.event_name == 'workflow_dispatch' }}
+    uses: nymtech/nym/.github/workflows/push-release-data.yml@develop
```
farbanas: since it's in the same repo I think you can use `uses: ./.github/workflows/push-release-data.yml` instead. That way it will also use the version of that workflow for the tag that it's running for instead of develop (where something new and breaking might be)

.github/workflows/nym-release-publish.yml
```diff
@@ -76,3 +89,139 @@ jobs:
             target/release/nym-network-requester
             target/release/nym-network-statistics
             target/release/nym-cli
+
+      - id: release-info
+        name: Prepare release info
+        run: |
+          semver="${${{ github.ref_name }}##nym-wallet-}" && semver="${semver##v}"
```
farbanas: it should be `nym-binaries-`

.github/workflows/nym-release-publish.yml
```diff
@@ -76,3 +89,139 @@ jobs:
...
+    if: ${{ (startsWith(github.ref, 'refs/tags/nym-binaries-') && github.event_name == 'release') || github.event_name == 'workflow_dispatch' }}
+    uses: nymtech/nym/.github/workflows/push-release-data.yml@develop
+    needs: publish-nym
+    with:
+      release_tag: ${{ github.ref_name }}
+      release_id: ${{ needs.publish-nym.outputs.release_id }}
+      release_date: ${{ needs.publish-nym.outputs.release_date }}
+      download_base_url: https://github.com/nymtech/nym/releases/download/${{ github.ref_name }}
+      changelog_url: https://github.com/nymtech/nym/blob/${{ github.ref_name }}/CHANGELOG.md
+      version: ${{ needs.publish-nym.outputs.version }}
```
farbanas: The problem with versions here is that they are not the same for each of the binaries, they might diverge more in the future but currently mixnode is `v1.1.23` while others are `v1.1.22`. To get this completely correct you'll have to fetch the version for each of them from their respective `Cargo.toml` files.

.github/workflows/nym-connect-publish-ubuntu.yml
```diff
@@ -56,10 +63,38 @@ jobs:
...
+  push-release-data:
+    if: ${{ (startsWith(github.ref, 'refs/tags/nym-connect-') && github.event_name == 'release') || github.event_name == 'workflow_dispatch' }}
+    uses: nymtech/nym/.github/workflows/push-release-data.yml@develop
+    needs: publish-tauri
+    with:
+      release_tag: ${{ github.ref_name }}
+      release_id: ${{ needs.publish-tauri.outputs.release_id }}
+      release_date: ${{ needs.publish-tauri.outputs.release_date }}
+      download_base_url: https://github.com/nymtech/nym/releases/download/${{ github.ref_name }}
+      changelog_url: https://github.com/nymtech/nym/blob/${{ github.ref_name }}/nym-connect/CHANGELOG.md
```
farbanas: I think changelog is in `nym-connect/desktop`

farbanas (changes requested):

farbanas (approved):

PR #3635, opened 2023-07-03
Feature/v1 1 23 docs

mfahampshire:
* fixes broken links from `mdbook last-changed` plugin 
* includes commits from https://github.com/nymtech/nym/pull/3602 
* fixes broken links in readme to platform and wallet build instructions 
* adds tutorial on using element with the mixnet


documentation/dev-portal/src/tutorials/matrix.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Command-line**
+
+To start Matrix's Element client via Socks5 proxy connected to NymConnect open terminal and run:
+
+```sh
+element-desktop --proxy-server=socks5://127.0.0.1:1080
+```
+
+**Keybinding**
```
mfahampshire: I'd split this into 2: the current one is the GUI option, and then we can add a CLI option with making an `alias` ?

documentation/dev-portal/src/tutorials/monero.md
```diff
@@ -0,0 +1,43 @@
+# Monero NymConnect Integration
```
mfahampshire, line 1: why is this file in here again, its already deployed to prod

documentation/dev-portal/src/tutorials/matrix.md
serinko: Maybe I explain wrong. Both of these are GUI Element options, just one starts with a cmd and one initiate the same cmd with pressing keys. It's a nice way around.

serinko: Feel free to change it though, I know it's not a good time to discuss back and forth.

PR #3640, opened 2023-07-04
Verify address with identity key in name-service contract

octol:
# Description

Verify identity_key with nym-address in name-service contract


contracts/name-service/src/contract/execute.rs
```diff
@@ -135,6 +145,7 @@ pub fn register(
     name: NameDetails,
     owner_signature: MessageSignature,
 ) -> Result<Response> {
+    ensure_identity_key_is_part_of_nym_address(&name.address, &name.identity_key)?;
```
jstuczyn, line 148: if we have to introduce this step, wouldn't it make sense to just remove `name.identity_key` field and make it implicit from the address?

PR #3664, opened 2023-07-11
refactor(nc-desktop): add privacy level user settings

### LINKED ISSUE #3663: NC S desktop -  add privacy mode user setting
"High" -> default and normal privacy mode, with all protection enabled, slow speed
"Medium" -> alternative privacy mode with some protection disabled, speed is better

Use sentry to record privacy mode switches

doums:
# Description

Closes: #3663

For the need of the feature this PR introduces support to save and read user data, like permanent settings value, on disk.
This allows the user to change settings values and automatically write them on disk, so that when he quit and restart the app, these settings values are restored as expected.

The data file is `user-data.toml` and file paths are:
- Linux: `$XDG_DATA_HOME/nym-connect` or `$HOME/.local/share/nym-connect`
- macOS: `$HOME/Library/Application Support/nym-connect`
- Windows: `{FOLDERID_RoamingAppData}\nym-connect`

https://docs.rs/tauri/1.4.1/tauri/api/path/fn.data_dir.html is used to pick the right path and follows platform specific standards.

For now 2 values are stored:
```toml
monitoring = true
privacy_level = "High"
```

### screenshots

image
image
image


jstuczyn:
It feels like this thing should have been either added to the normal config:  https://github.com/nymtech/nym/blob/develop/nym-connect/desktop/src-tauri/src/config/mod.rs#L62
(In a separate section, similarly to what we do say in network requester: https://github.com/nymtech/nym/blob/develop/service-providers/network-requester/src/config/mod.rs#L66)
or if you want to keep it in separate file, the path should be kept in the paths struct (and persisted in the config: https://github.com/nymtech/nym/blob/develop/nym-connect/desktop/src-tauri/src/config/persistence.rs#L9)

doums:
@jstuczyn we discussed that point with @octol , and basically
"on the rust side, there is config per gateway"
"the main reason for those config is to keep track of keys it will select the config depending on which gateway is selected"
"so if you store it there, it will be only available for that specific gateway, and it's loaded only when you click connect"

So based on this we decided to make an implementation from scratch, using tauri api standard for user data (which in that case is exactly what we want) -> https://docs.rs/tauri/1.4.1/tauri/api/path/fn.data_dir.html 

To be more specific, here the purpose is to save user settings value, not "config" like app internal config stuff or even user config stuff. As data saved here are not intended to be edited  directly in the file by the user (there is the UI for that).
So `data_dir` is exactly the API to use for that. More globally tauri follows OS standard so that's great:

image

octol:
Yep I talked @doums into it :) 

I think the approach makes sense, we just need to make sure it's integrated into the config related work that @jstuczyn has been doing recently

jstuczyn:
> To be more specific, here the purpose is to save user settings value, not "config" like app internal config stuff or even user config stuff.

I feel like those should still be referenced (at the very least by filepath) in the actual config. Then it would give us an option to change those values on per-client basis. And most importantly we'd keep all settings consistent

doums:
> Then it would give us an option to change those values on per-client basis

@jstuczyn why would we need that in the case of "user data"?

PR #3678, opened 2023-07-17
Upgrade cosmwasm

neacsu:


common/client-libs/validator-client/src/nyxd/fee/mod.rs
```diff
@@ -144,29 +144,11 @@ mod sealed {
...
-
-    impl From<Decimal> for CosmosDecimal {
-        fn from(val: Decimal) -> Self {
-            val.0.into()
-        }
-    }
-
     #[derive(Serialize, Deserialize, Clone)]
     struct Coin {
         #[serde(with = "Denom")]
```
jstuczyn: iirc we could also remove that special serde annotation like you've done with Decimal

contracts/Cargo.toml
```diff
@@ -32,16 +32,17 @@ incremental = false
...
+cosmwasm-storage = "=1.2.5"
+cw-controllers = "=1.0.1"
+cw-multi-test = "=0.16.4"
+cw-storage-plus = "=1.0.1"
+cw-utils = "=1.0.1"
+cw2 = "=1.0.1"
+cw3 = "=1.0.1"
+cw3-fixed-multisig = "=1.0.1"
+cw4 = "=1.0.1"
+cw20 = "=1.0.1"
```
jstuczyn, line 48: huh. why do we depend on `cw20` now?

jstuczyn (approved):

contracts/Cargo.toml
jstuczyn: ah, right, because of forked multisig.

PR #3691, opened 2023-07-19
Feature/issue credentials

### LINKED ISSUE #3690: Rust SDK: Coconut API to issue a credential
This code uses the credential:
https://github.com/nymtech/nym/blob/develop/sdk/rust/nym-sdk/examples/bandwidth.rs

This code issues the credential:
https://github.com/nymtech/nym/tree/develop/clients/credential

- [x] Add commands to the https://github.com/nymtech/nym/tree/develop/common/commands/src/coconut package
- [x] Add to the Nym CLI
- [x] Add an example to the https://github.com/nymtech/nym/blob/develop/sdk/rust/nym-sdk/examples called `get_bandwidth_credential.rs`

Image

farbanas:
# Description

Closes: #3690


common/client-core/src/client/base_client/storage/mod.rs
```diff
@@ -94,7 +94,7 @@ impl MixnetClientStorage for Ephemeral {
 pub struct OnDiskPersistent {
     pub(crate) key_store: OnDiskKeys,
     pub(crate) reply_store: fs_backend::Backend,
-    pub(crate) credential_store: PersistentCredentialStorage,
+    pub credential_store: PersistentCredentialStorage,
```
farbanas: How can I get access to this field in SDK without removing the crate part?

jstuczyn: Since `OnDiskPersistent` implements `MixnetClientStorage` you can just call `credential_store()` on it

common/commands/src/coconut/issue_credentials.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Home directory of the client that is supposed to use the credential.
+    #[clap(long)]
+    pub(crate) client_home_directory: std::path::PathBuf,
+
+    /// A mnemonic for the account that buys the credential
+    #[clap(long)]
+    pub(crate) mnemonic: Option<bip39::Mnemonic>,
+
+    /// The amount of utokens the credential will hold. If recovery mode is enabled, this value
+    /// is not needed
```
jstuczyn: if the value is not needed, set it to an `Option<u64>` so that clap would not try to enforce it (and you'd have a nicer type to work with)

common/commands/src/coconut/issue_credentials.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub(crate) amount: u64,
+
+    /// Path to a directory used to store recovery files for unconsumed deposits
+    #[clap(long)]
+    pub(crate) recovery_dir: std::path::PathBuf,
+}
+
+pub async fn execute(args: Args, client: SigningClientWithNyxd) {
+    let network_details = NymNetworkDetails::new_from_env();
+    let coin = Coin::new(
```
jstuczyn: kinda relating to the above. if you're going to be using the recovery mode, do you need to be constructing the coin?

common/issue-credential/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "nym-issue-credential"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+log = "0.4"
+thiserror = "1.0"
```
jstuczyn:
```suggestion
log = { workspace = true }
thiserror = { workspace = true }
```

common/issue-credential/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::recovery_storage::RecoveryStorage;
+
+const SAFETY_BUFFER_SECS: u64 = 60; // 1 minute
+
+pub async fn issue_credential(
+    client: nym_validator_client::Client<DirectSigningNyxdClient>,
+    amount: Coin,
+    persistent_storage: &PersistentStorage,
+    recovery_storage_path: PathBuf,
+) {
```
jstuczyn: Since you have defined a `CredentialClientError`, perhaps return a `Result<(), CredentialClientError>` instead? .....

common/issue-credential/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+    let recovery_storage = setup_recovery_storage(recovery_storage_path).await;
+
+    block_until_coconut_is_available(&client).await.expect("");
```
jstuczyn: .... and then you'll be able to get rid of this `expect`

common/issue-credential/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+
+            break;
+        } else {
+            // Use 1 additional second to not start the next iteration immediately and spam get_current_epoch queries
+            let secs_until_final = epoch
+                .final_timestamp_secs()
+                .saturating_sub(current_timestamp_secs)
+                + 1;
+            info!("Approximately {} seconds until coconut is available. Sleeping until then. You can safely kill the process at any moment.", secs_until_final);
+            std::thread::sleep(Duration::from_secs(secs_until_final));
```
jstuczyn: that's a rather subtle thing, but you don't want to be using `thread::sleep` in the async context. what you probably meant to use was `tokio::time::sleep` instead

common/issue-credential/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+                "Recovered credentials in the amount of {}",
+                recovered_amount
+            );
+            return;
+        }
+    };
+
+    let state = nym_bandwidth_controller::acquire::deposit(&client.nyxd, amount.clone())
+        .await
+        .expect("");
```
jstuczyn: ibid.

common/issue-credential/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+                error!("Could not dump recovery data to file system due to {:?}, the deposit will be lost!", e)
+            }
+        }
+        exit(1);
+    }
+
+    info!(
+        "Succeeded adding a credential with amount {}{}",
+        &amount.amount, &amount.denom,
+    );
```
jstuczyn:
```suggestion
    info!("Succeeded adding a credential with amount {amount}");
```

common/issue-credential/src/recovery_storage.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl RecoveryStorage {
+    pub fn new(recovery_dir: PathBuf) -> std::io::Result<Self> {
+        create_dir_all(&recovery_dir)?;
+        Ok(Self { recovery_dir })
+    }
+
+    pub fn unconsumed_vouchers(&self) -> std::io::Result<impl Iterator<Item = BandwidthVoucher>> {
+        Ok(read_dir(&self.recovery_dir)?
+            .filter_map(|entry| entry.ok())
```
jstuczyn: I guess technically it works and does exactly what we need, but the amount of `filter_map` here looks a bit iffy. Perhaps it might make sense to try to rewrite it to use normal loop and propagate the errors?

jstuczyn (commented):
lgtm, but in general try to avoid using so many `unwrap` and `expect` and instead either propagate the errors or handle them directly.

common/commands/src/coconut/issue_credentials.rs
farbanas: Actually this comment is incorrect, it is left over from the credential binary that has the recovery mode. Instead of the recovery mode I am first trying to recover the credentials and then create new ones if nothing is recovered

jstuczyn (approved):

PR #3695, opened 2023-07-20
build(nc-desktop): sentry dsn as env vars

### LINKED ISSUE #3694: NC Desktop - remove sentry DSN from code
Currently, sentry DSNs are hardcoded

refactor: pass them at compile time

doums:
# Description

Closes: #3694

This PR remove hard-coded Sentry DSN (server URLs). Now they must be provided through env vars `SENTRY_DSN_RUST` and `SENTRY_DSN_JS`:

```dotenv
SENTRY_DSN_RUST=https://...
SENTRY_DSN_JS=https://...
```
Additionally, they can be set at build time (`yarn build`, CI releases etc) and they will be used automatically by the bundled binary.
Both vars have been added to repo secrets (https://github.com/nymtech/nym/settings/secrets/actions) and gh workflows have been updated accordingly.

For development (`yarn dev`) and for convenience, you can provide them in a `.env` file placed in `nym-conenct/desktop/`

### DIFF
--- a/nym-connect/desktop/src-tauri/Cargo.toml
+++ b/nym-connect/desktop/src-tauri/Cargo.toml
@@ -46,7 +46,7 @@ yaml-rust = "0.4"
 toml = "0.7"
 sentry = { version = "0.31.5", features = [ "anyhow" ] }
 sentry-log = "0.31.5"
-dotenv = "0.15.0"
+dotenvy = "0.15.7"
 
 nym-client-core = { path = "../../../common/client-core" }
 nym-api-requests = { path = "../../../nym-api/nym-api-requests" }
--- a/nym-connect/desktop/src-tauri/Cargo.toml
+++ b/nym-connect/desktop/src-tauri/Cargo.toml
@@ -46,6 +46,7 @@ yaml-rust = "0.4"
 toml = "0.7"
 sentry = { version = "0.31.5", features = [ "anyhow" ] }
 sentry-log = "0.31.5"
+dotenv = "0.15.0"
 
 nym-client-core = { path = "../../../common/client-core" }
 nym-api-requests = { path = "../../../nym-api/nym-api-requests" }
--- a/nym-connect/desktop/src-tauri/src/main.rs
+++ b/nym-connect/desktop/src-tauri/src/main.rs
@@ -3,7 +3,6 @@
     windows_subsystem = "windows"
 )]
 
-use dotenv::dotenv;
 use std::env;
 use std::sync::Arc;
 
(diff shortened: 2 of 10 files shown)


nym-connect/desktop/src-tauri/Cargo.toml
```diff
@@ -46,6 +46,7 @@ yaml-rust = "0.4"
 toml = "0.7"
 sentry = { version = "0.31.5", features = [ "anyhow" ] }
 sentry-log = "0.31.5"
+dotenv = "0.15.0"
```
octol: Consider `dotenvy` instead

nym-connect/desktop/src-tauri/src/main.rs
```diff
@@ -46,37 +46,21 @@ fn main() {
...
 
     if monitoring {
-        println!("Sentry reporting and monitoring is enabled");
-    } else {
-        println!("Monitoring is disabled, dropping sentry guard");
-        drop(_guard)
-    }
+        match monitoring::init() {
+            Ok(guard) => {
+                println!("Monitoring and error reporting enabled");
```
octol, line 54: General comment for the future

Something to ponder: log vs println?

I don't not yet what my opinion is, I just know that currently we're doing it a bit haphazardly. Possibly we should think it through and try to formulate a good guideline

doums: log is the way to go, the thing is for some reason (our custom log setup?) any `log::info/warn/error` calls in `main.rs` print nothing
That's why Im forced to use raw println

octol: Ah yes you're right, logging is setup further down in this function

nym-connect/desktop/src-tauri/src/monitoring.rs
```diff
@@ -0,0 +1,40 @@
+use std::env;
+
+use anyhow::{Context, Result};
+use sentry::ClientInitGuard;
+
+// env var keys
+const SENTRY_DSN_RUST: &str = "SENTRY_DSN_RUST";
+const SENTRY_DSN_JS: &str = "SENTRY_DSN_JS";
```
octol: These two are defined twice by the looks of it? Can we deduplicate?

nym-connect/desktop/src-tauri/src/main.rs
```diff
@@ -46,37 +45,21 @@ fn main() {
...
-        },
-    ));
-
-    sentry::configure_scope(|scope| {
-        scope.set_user(Some(sentry::User {
-            id: Some("nym".into()),
-            ..Default::default()
-        }));
-    });
+    let mut _sentry_guard;
```
octol, line 49: Suggestion: what about

```
let _sentry_guard = if monitoring {
    ...
```

octol (approved):

doums:
merged 34a47a9

PR #3698, opened 2023-07-24
Apply fix from feature/ephemera to develop too

neacsu:


common/nymsphinx/types/Cargo.toml
```diff
@@ -9,5 +9,9 @@ repository = { workspace = true }
 
 [dependencies]
 sphinx-packet = { version = "0.1.0" }
-nym-outfox = { path = "../../../nym-outfox" }
+nym-outfox = { path = "../../../nym-outfox", optional = true }
 thiserror = "1"
+
+[features]
```
jstuczyn, line 15: wouldnt it make more sense to have 2 features:
- sphinx
- outfox

and lock stuff accordingly? Then the library consumers could decide what they want and the features wouldn't be mutually exclusive.

durch: In general agree with the sentiment, if its not too much of a hassle (and it could be) lets do it now, otherwise we can iterate to it, to unblock this now...

durch (approved):

Issue #3581, opened 2023-06-20
Speedy mixnet- make a new mixnet client with below properties:


jessgess:
For this track we will be using the existing Socks5 based NymConnect, but with below changes on the client side:

- [x] remove cover traffic
- [x] remove delays
- [x] remove poisson sending
- [x] smaller packet size (lower send rate, more private) and larger packet size (higher send rate, less private)
- [x] use geographically close mixnodes for creating the path

PR #3713, opened 2023-07-27
Add geo-aware mixnet topology provider

octol:
# Description

Part of: #3581

Add a geoaware mixnet topology provider that uses ip lookup to group nodes.

Enable it in nym-connect when in medium / speedy mode (defaults to Europe)

Add flag `--geo-routing` flag to `nym-socks5-client`. Also, enable when `--medium-toggle` is used.


common/client-core/src/client/topology_control/geo_aware_provider.rs
```diff
@@ -0,0 +1,263 @@
...
+            Asia => write!(f, "AS"),
+            Africa => write!(f, "AF"),
+            Unknown => write!(f, "Unknown"),
+        }
+    }
+}
+
+// We map contry codes into group, which initially are continent codes to a first approximation,
+// but we do it manually to reserve the right to tweak this distribution for our purposes.
+impl From<&str> for CountryGroup {
```
jstuczyn: how about a `FromStr` instead?

common/client-core/src/client/topology_control/geo_aware_provider.rs
```diff
@@ -0,0 +1,263 @@
...
+use nym_topology::{
+    mix::Layer,
+    nym_topology_from_detailed,
+    provider_trait::{async_trait, TopologyProvider},
+    NymTopology,
+};
+use nym_validator_client::client::{MixId, MixNodeDetails};
+use url::Url;
+
+const EXPLORER_API_MIXNODES_URL: &str = "https://explorer.nymtech.net/api/v1/mix-nodes";
```
jstuczyn, line 16: that's gonna be a bit awkward for anything that's not mainnet. also I dont think we should be relying on the explorer...

common/client-core/src/client/topology_control/geo_aware_provider.rs
```diff
@@ -0,0 +1,263 @@
...
+
+async fn fetch_mixnodes_from_explorer_api() -> Option<Vec<PrettyDetailedMixNodeBond>> {
```
jstuczyn, line 19: i actually dont remember, but dont we have an explorer api client?

common/client-core/src/client/topology_control/geo_aware_provider.rs
```diff
@@ -0,0 +1,263 @@
...
+            acc
+        })
+}
+
+fn check_layer_integrity(mixnodes: Vec<MixNodeDetails>) -> Result<(), ()> {
+    let mut layer_counts = count_mixnodes_per_layer(mixnodes);
+    for layer in &[Layer::One, Layer::Two, Layer::Three] {
+        layer_counts.entry(*layer).or_insert(0);
+        let count = layer_counts[layer];
+        if count < 2 {
```
jstuczyn: why 2 : )

common/client-core/src/client/topology_control/geo_aware_provider.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+fn group_mixnodes_by_country_code(
+    mixnodes: Vec<PrettyDetailedMixNodeBond>,
+) -> HashMap<CountryGroup, Vec<MixId>> {
+    mixnodes
+        .into_iter()
+        .fold(HashMap::<CountryGroup, Vec<MixId>>::new(), |mut acc, m| {
+            if let Some(ref location) = m.location {
```
jstuczyn, line 184: hm. should we be ignoring nodes without location? maybe lets put them under unknown?

common/client-core/src/client/topology_control/geo_aware_provider.rs
octol: I didn't find one :(

common/client-core/src/client/topology_control/geo_aware_provider.rs
octol: yes, and yes. agree...

common/client-core/src/client/topology_control/geo_aware_provider.rs
octol: FromStr returns Result

common/client-core/src/client/topology_control/geo_aware_provider.rs
octol: I pulled it out of my behind. But let me at least extract out to a const

PR #3723, opened 2023-08-01
refactor(nc-desktop): use userdata storage to save user gateway&sp

### LINKED ISSUE #3707: NC - use user data to manage selected gateway and sp
Use new user data (tauri side) to save selected gateway and sp instead of `@tauri-apps/tauri-forage` (frontend)

doums:
# Description

Closes: #3707


nym-connect/desktop/src-tauri/src/config/user_data.rs
```diff
@@ -14,6 +14,18 @@ pub enum PrivacyLevel {
     Medium,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct SelectedGateway {
+    address: Option<String>,
```
octol, line 19: Can probably remove the `Option<>` here, since one level higher up `SelectedGateway` is already stored as an `Option<>`?

nym-connect/desktop/src-tauri/src/state.rs
```diff
@@ -104,6 +107,13 @@ impl State {
         &self.user_data
     }
 
+    pub fn clear_user_data(&mut self) -> Result<()> {
```
octol, line 110: do I understand it right in that this is basically `clear_user_data_on_disk`? (not clearing the in-mem `UserData`?)

doums: yes

nym-connect/desktop/src-tauri/src/config/user_data.rs
doums: When user toggle on and off "select your gateway blabla" without selecting any address, the user data would be saved on disk like that

```
[selected_gateway]
is_active = false
```

So we have to keep `address` field as an Option

octol (approved):
Looks good to me!
I think we should change the name of the clear function though, it's good to be explicit whenever we delete things from disk.

PR #3731, opened 2023-08-02
Feature/ephemera

neacsu:
# Description

Include Ephemera in the nym-api and start it as a service, if enabled by the flag.
In order for a nym-api to be able to sign up to be part of the Ephemera group, it will need to be part of the group contract that the ephemera contract points to.

After merging this PR, the next step would be to modify the access control of mixnet contract from verifying one address being the rewarding account to verifying the rewarder is part of a bigger group, controlled by the ephemera contract.

# Checklist:

- [x] QAed to check that, when not enabled, it doesn't mess with the current way of metrics and rewarding (@tommyv1987 )


durch (approved):
My commentary is Ephemera specific... does not really concern integration work, but wow...

It is giant, bunch of modules seem overly bloated, `id`, and `crypto` for example. I'd be much happier without having to maintain this...

PR #3759, opened 2023-08-10
Theme

gala1234:
# Description

Adding Nym theme to the dev-portal docs


mfahampshire:
looks good to me - I think we should these changes to the `book.toml` files in `docs/` and `operators/` as well

PR #3783, opened 2023-08-17
network-requester: disable poisson process by default

### LINKED ISSUE #3774: NR - reduce response time by removing poisson delay
We want to have just one set of NRs with option to have two different packet sizes. 
The default packet size should be the smaller one.

octol:
# Description

Closes: #3774

In the network-requester, by default disable the Poisson sending rate process.
We keep the cover traffic process running, but at a slow trickle to act as keepalive messages.

### DIFF
--- a/service-providers/network-requester/src/cli/run.rs
+++ b/service-providers/network-requester/src/cli/run.rs
@@ -83,6 +83,7 @@ pub(crate) async fn execute(args: &Run) -> Result<(), NetworkRequesterError> {
     }
 
     let mut config = try_load_current_config(&args.id)?;
+    dbg!(&config);
     config = override_config(config, OverrideConfig::from(args.clone()));
     log::debug!("Using config: {:#?}", config);
 
@@ -109,6 +110,5 @@ pub(crate) async fn execute(args: &Run) -> Result<(), NetworkRequesterError> {
         stats_provider_addr,
     )
     .await;
-    panic!();
     server.run_service_provider().await
 }
--- a/service-providers/network-requester/src/cli/run.rs
+++ b/service-providers/network-requester/src/cli/run.rs
@@ -83,7 +83,6 @@ pub(crate) async fn execute(args: &Run) -> Result<(), NetworkRequesterError> {
     }
 
     let mut config = try_load_current_config(&args.id)?;
-    dbg!(&config);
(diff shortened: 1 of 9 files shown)


octol:
QA notes: I've added printing the whole `Config` struct at the `debug` log-level, for both `init` and `run` commands

service-providers/network-requester/src/cli/run.rs
```diff
@@ -108,5 +109,6 @@ pub(crate) async fn execute(args: &Run) -> Result<(), NetworkRequesterError> {
         stats_provider_addr,
     )
     .await;
+    panic!();
```
jstuczyn: what's with the panic here?

service-providers/network-requester/src/config/mod.rs
```diff
@@ -81,8 +81,13 @@ impl NymConfigTemplate for Config {
 
 impl Config {
     pub fn new<S: AsRef<str>>(id: S) -> Self {
+        let mut base = BaseClientConfig::new(id.as_ref(), env!("CARGO_PKG_VERSION"));
+        // By default we have no cover traffic, but we have a slow trickle of cover traffic
+        // packages acting as keepalive messages.
+        base.set_no_cover_traffic_with_keepalive();
```
jstuczyn: couldn't we just set all of those as you know, the defaults?

octol: But we don't want them to be default for BaseClientConfig, only for when BaseClientConfig is used in the NR

service-providers/network-requester/src/config/mod.rs
```diff
@@ -159,9 +164,21 @@ impl Config {
     }
 }
 
-#[derive(Debug, Default, Clone, Deserialize, PartialEq, Serialize)]
+#[derive(Debug, Clone, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
-pub struct NetworkRequester {}
+pub struct NetworkRequester {
+    pub description: String,
```
jstuczyn: should description be really part of the config? imo (personal bias) it should be treated like the mixnodes'

octol: it's useful to stay consistent, so I will change that

jstuczyn (approved):

PR #3785, opened 2023-08-17
feat(explorer-api): add gateway geoip location

### LINKED ISSUE #3776: Explorer - look up gateways based on geo-location

doums:
# Description

Closes: #3776 

This PR update the route `/gateways` to add a `location` field (based on geoip) for each gateway in the response

image


octol (approved):
Looks good!

I would prefer change the existing endpoint, that is, also update the frontend side to use the new endpoint in the same PR, if it's practical. Then the mixnode and gateway endpoints would be symmetrical.

If it's not practical I guess it's ok to just create a new endpoint for now and migrate the frontend later on

PR #3796, opened 2023-08-22
gateway: disconnect inactive duplicate clients

### LINKED ISSUE #2845: [Issue] There is already an open connection to this client
**Describe the issue**
If the nym-client crash or stop without waiting "cleaning process", it's impossible to reconnect directly, the gateway still see an open connection
After a while (more than 12 hours) it's possible to reconnect again

**Expected behaviour**
Open a new connection with the gateway works directly

**Stack Traces**

```
 2023-01-11T16:51:42.270Z ERROR client_core::client::base_client                                          > Could not authenticate and start up the gateway connection - Gateway returned an error response - There is already an open connection to this client
Error: ClientCoreError(GatewayClientError(GatewayError("There is already an open connection to this client")))
```

**Steps to Reproduce**
Steps to reproduce the behaviour, if you're familiar with BDD syntax, please write it in this style:
- `nym-client run --id test`
- application stop abruptly
- When trying to reconnect this error appears ` Gateway returned an error response - There is already an open connection to this client`

**Screenshots**
If applicable, add screenshots to help explain your problem...

 - Application: [nym-client(1.1.5 and 1.1.4)]

**Additional context**
Please provide any other information

octol:
# Description

Closes: #2845

In `nym-gateway` when a duplicate client connection is detected, query the existing connection if it's still active. If it's still active, deny the new connection (as before). If it's NOT active, drop the previous connection and allow the new client to connect.

### DIFF
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -37,7 +37,7 @@ serde_json = { workspace = true }
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "sqlite", "macros", "migrate", ] }
 subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
 thiserror = "1"
-tokio = { version = "1.24.1", features = [ "rt-multi-thread", "net", "signal", "fs", "time" ] }
+tokio = { workspace = true, features = [ "rt-multi-thread", "net", "signal", "fs", "time" ] }
 tokio-stream = { version = "0.1.11", features = ["fs"] }
 tokio-tungstenite = "0.14"
 tokio-util = { version = "0.7.4", features = ["codec"] }
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -37,7 +37,7 @@ serde_json = { workspace = true }
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "sqlite", "macros", "migrate", ] }
 subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
 thiserror = "1"
-tokio = { version = "1.24.1", features = [ "rt-multi-thread", "net", "signal", "fs", ] }
+tokio = { version = "1.24.1", features = [ "rt-multi-thread", "net", "signal", "fs", "time" ] }
 tokio-stream = { version = "0.1.11", features = ["fs"] }
 tokio-tungstenite = "0.14"
 tokio-util = { version = "0.7.4", features = ["codec"] }
--- a/gateway/src/node/client_handling/active_clients.rs
+++ b/gateway/src/node/client_handling/active_clients.rs
@@ -1,13 +1,17 @@
 // Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::node::client_handling::websocket::message_receiver::MixMessageSender;
 use dashmap::DashMap;
+use futures::channel::{mpsc, oneshot};
 use nym_sphinx::DestinationAddressBytes;
 use std::sync::Arc;
 
+use super::websocket::message_receiver::{IsActiveRequestSender, MixMessageSender};
+
 #[derive(Clone)]
-pub(crate) struct ActiveClientsStore(Arc<DashMap<DestinationAddressBytes, MixMessageSender>>);
+pub(crate) struct ActiveClientsStore(
+    Arc<DashMap<DestinationAddressBytes, (MixMessageSender, IsActiveRequestSender)>>,
+);
 
 impl ActiveClientsStore {
     /// Creates new instance of `ActiveClientsStore` to store in-memory handles to all currently connected clients.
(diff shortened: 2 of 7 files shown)


gateway/src/node/client_handling/active_clients.rs
```diff
@@ -1,13 +1,17 @@
...
+use futures::channel::{mpsc, oneshot};
 use nym_sphinx::DestinationAddressBytes;
 use std::sync::Arc;
 
+use super::websocket::message_receiver::{IsActiveRequestSender, MixMessageSender};
+
 #[derive(Clone)]
-pub(crate) struct ActiveClientsStore(Arc<DashMap<DestinationAddressBytes, MixMessageSender>>);
+pub(crate) struct ActiveClientsStore(
+    Arc<DashMap<DestinationAddressBytes, (MixMessageSender, IsActiveRequestSender)>>,
```
jstuczyn: I think that unit tuple is getting a bit messy, should we perhaps move `(MixMessageSender, IsActiveRequestSender)` into separate struct? like idk 
```rust
struct Foo {
  foo: MixMessageSender,
  bar: IsActiveRequestSender,
}
```

or maybe (apart from it) we could use the inner pattern

gateway/src/node/client_handling/websocket/connection_handler/fresh.rs
```diff
@@ -75,6 +81,10 @@ pub(crate) struct FreshHandler<R, S, St> {
     pub(crate) socket_connection: SocketStream<S>,
     pub(crate) storage: St,
     pub(crate) coconut_verifier: Arc<CoconutVerifier>,
+    // Occasionally the handler is requested to ping the connected client for confirm that it's
+    // active, such as when a duplicate connection is detected. This hashmap stores the oneshot
+    // senders that are used to return the result of the ping to the handler requesting the ping.
+    pub(crate) is_active_ping_pending_replies: HashMap<u64, IsActiveResultSender>,
```
jstuczyn: wait, I'm confused. why do we need an entire map for those? isn't it the case that we'd only be handling a single ping/pong at a time?  if we receive a message with wrong tag, we just ignore it.

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -351,6 +356,20 @@ where
...
     }
 
+    /// Handles pong message received from the client.
+    /// If the client is still active, the handler that requested the ping will receive a reply.
+    async fn handle_pong(&mut self, msg: Vec<u8>) {
+        if let Ok(msg) = msg.try_into() {
+            let msg = u64::from_be_bytes(msg);
+            debug!("Received pong from client: {}", msg);
+            if let Some(tx) = self.inner.is_active_ping_pending_replies.remove(&msg) {
+                if let Err(err) = tx.send(true) {
```
jstuczyn: if we can never send a `false`, why not just use the `Notify`?

jstuczyn (approved):

gateway/Cargo.toml
```diff
@@ -37,7 +37,7 @@ serde_json = { workspace = true }
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "sqlite", "macros", "migrate", ] }
 subtle-encoding = { version = "0.5", features = ["bech32-preview"] }
 thiserror = "1"
-tokio = { version = "1.24.1", features = [ "rt-multi-thread", "net", "signal", "fs", ] }
+tokio = { version = "1.24.1", features = [ "rt-multi-thread", "net", "signal", "fs", "time" ] }
```
jstuczyn: while at it. change it to use `workspace = true`

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -363,10 +405,64 @@ where
...
+        let tag: u64 = rand::thread_rng().gen();
+        debug!("Got request to ping our connection: {}", tag);
+        self.inner
+            .send_websocket_message(Message::Ping(tag.to_be_bytes().to_vec()))
+            .await?;
+        Ok(tag)
+    }
+
+    /// Handles the ping timeout by responding back to the handler that requested the ping.
+    async fn handle_ping_timout(&mut self) {
```
jstuczyn: typo

gateway/src/node/client_handling/websocket/message_receiver.rs
```diff
@@ -1,7 +1,20 @@
...
 
-use futures::channel::mpsc;
+use futures::channel::{mpsc, oneshot};
 
 pub(crate) type MixMessageSender = mpsc::UnboundedSender<Vec<Vec<u8>>>;
 pub(crate) type MixMessageReceiver = mpsc::UnboundedReceiver<Vec<Vec<u8>>>;
+
+// Channels used for one handler to requester another handler to check that the client is still
+// active. The result is then passed back to the requesting handler in the oneshot channel.
+pub(crate) type IsActiveRequestSender = mpsc::UnboundedSender<IsActiveResultSender>;
```
jstuczyn, line 11: does it have to be an unbounded channel though?

jstuczyn: hm. or maybe it does

PR #3802, opened 2023-08-23
Bugfix/use correct tendermint dialect

jstuczyn:
# Description

Fully delegate rpc calls to inner fields so that correct tendermint dialect would be used (i.e. 0.34 as opposed to default 0.37). 

I've briefly looked at https://github.com/hobofan/ambassador to remove all the boilerplate code, but couldn't get it to work with async trait that has to be conditionally derived based on build target. And writing our own delegate macro would have been a huge overkill.


tommyv1987 (approved):
LGTM -  tested all working a, ok.

PR #3806, opened 2023-08-23
fix(wallet): gateway bond form wrong helper text

doums:
# Description

On location field the error message text displayed when wrong location values are entered is wrong.


tommyv1987 (approved):
LGTM

PR #3809, opened 2023-08-24
fix(nc-android): state sync cross ffi

### LINKED ISSUE #3771: NC - fix socks5 connection state
## Issue

The socks5 client process and the application are running it 2
different and independent **layers**.

The socks5 client connection is managed at the “native” layer,
written in Rust, consumed as a
[native library](https://github.com/nymtech/nym/tree/develop/sdk/lib/socks5-listener). That means, the socks5 process is running in the “native” layer when connected, entirely detached and independently of the Android layer (Java/Kotlin).

In the other hand, the application is running (obviously) in the
Android layer, and it lives according to the
[Activity lifecycles](https://developer.android.com/guide/components/activities/activity-lifecycle). Those lifecycles directly manage its state.

That means it's quite challenging to keep the real state of the
socks5 process perfectly **in sync** with the App one. That's even
more challenging since to comply with Android OS, we had to
wrap the socks5 connection as a long-running worker using
[WorkManager](https://developer.android.com/guide/background/persistent)

see https://developer.android.com/guide/background/persistent/how-to/long-running

Issue example:

If the user launch the socks5 proxy on the application:
1. in android layer the app launches a long-running work, call the native lib and the state is set to `connected`
2. in the native layer socks5proxy process is spawned, connects and starts running

Then if the user kill the app from the notification drawer

Image

- the android Task Manager kills the app entirely, including the
long running worker, without any kind of mechanism to catch that 
event in order to kill the underlying socks5proxy process running 
in the native layer
(for details see https://developer.android.com/guide/components/foreground-services#handle-user-initiated-stop)
- the app is closed, state is reset, that means `disconnected`
- but the socks5proxy process is still operating and running,
socks5 connection is still active, like a zombie process consuming network traffic
It can potentially keep consuming mobile data in the background and the user
has no idea
- the app state is unsync with the real socks5proxy state

## potential solution

Make the app and native layer communicating with each other using a client-server model (`gRPC` protocol could be a nice candidate)

doums:
# Description

Closes: #3771

### DIFF
--- a/sdk/lib/socks5-listener/src/lib.rs
+++ b/sdk/lib/socks5-listener/src/lib.rs
@@ -28,12 +28,13 @@ mod persistence;
 
 static SOCKS5_CONFIG_ID: &str = "mobile-socks5-test";
 const ANDROID_HEALTHCHECK_INTERVAL: Duration = Duration::from_secs(5);
+const HEALTHCHECK_TIMEOUT: Duration = Duration::from_secs(10);
 
 // hehe, this is so disgusting : )
 lazy_static! {
     static ref CLIENT_SHUTDOWN_HANDLE: Mutex<Option<Arc<Notify>>> = Mutex::new(None);
     static ref RUNTIME: Runtime = Runtime::new().unwrap();
-    static ref LAST_PING: Mutex<Option<Instant>> = Mutex::new(None);
+    static ref LAST_HEALTHCHECK_PING: Mutex<Option<Instant>> = Mutex::new(None);
 }
 static ENV_SET: AtomicBool = AtomicBool::new(false);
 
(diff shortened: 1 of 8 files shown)


sdk/lib/socks5-listener/src/lib.rs
```diff
@@ -150,6 +162,37 @@ pub fn blocking_run_client<'cb, F, S>(
         return;
     }
 
+    RUNTIME.spawn(async move {
```
octol: Can probably extract out this closure to a function

sdk/lib/socks5-listener/src/lib.rs
```diff
@@ -150,6 +162,37 @@ pub fn blocking_run_client<'cb, F, S>(
...
+                break;
+            }
+            let mut guard = LAST_PING.lock().await;
+            let Some(last_ping) = *guard else {
+                warn!("client has not been pinged yet - shutting down");
+                *guard = None;
+                stop_and_reset_shutdown_handle().await;
+                break;
+            };
+            if last_ping.elapsed() > Duration::from_secs(10) {
```
octol: Extract out constant, something like `HEALTHCHECK_TIMEOUT`

sdk/lib/socks5-listener/src/lib.rs
```diff
@@ -16,20 +16,24 @@ use std::marker::Send;
...
 mod persistence;
 
 static SOCKS5_CONFIG_ID: &str = "mobile-socks5-test";
+const ANDROID_HEALTHCHECK_INTERVAL: Duration = Duration::from_secs(5);
 
 // hehe, this is so disgusting : )
 lazy_static! {
     static ref CLIENT_SHUTDOWN_HANDLE: Mutex<Option<Arc<Notify>>> = Mutex::new(None);
     static ref RUNTIME: Runtime = Runtime::new().unwrap();
+    static ref LAST_PING: Mutex<Option<Instant>> = Mutex::new(None);
```
octol: Maybe `LAST_HEALTHCHECK_PING` to not confuse it with a network ping?

octol (approved):

PR #3812, opened 2023-08-25
nym-connect: filter gateways on compatible version

### LINKED ISSUE #3788: NymConnect sometimes fails to connect because the gateway it fetches from the validator-api to use is running an old version (of the gateway binary)
NymConnect queries the validator-api for gateways to select from. This appears to include gateways that are running old versions. Connecting then fails with

```
Failed to initialize: No gateway with id: 9HqEJiqjPUv75DoJdaM8sX8htCEAU9C6BMPhuA32obRf
```

Make nym-connect (in the Rust tauri layer), when quering nym-api for gateways also filter on compatible versions.

octol:
# Description

Closes: #3788

In nym-connect, filter out incompatible gateways based on the self-reported version.


tommyv1987 (approved):
LGTM

PR #3818, opened 2023-08-28
nym-connect: select gateway based on latency in medium mode

### LINKED ISSUE #3770: Speedy mode - selects gateway based on latency  in medium / speedy mode
The normal client supports selecting the gateway based on ping times instead of uniformly random across all gateways.

In `nym-connect` however, gateway selection is handled in the Tauri rust layer. We should extend this code to be able to do latency based selection in the same way as the normal client.

octol:
# Description

Closes: #3770

In nym-connect, select gateway based on latency in medium mode

image


doums (approved):
lgtm

PR #3819, opened 2023-08-28
feat(nc): disable privacy switch when active connection

### LINKED ISSUE #3816: NC - it should not be possible to toggle speedy mode while the connection is active
If the connection is active, it should not be possible to toggle speedy mode on/off

doums:
# Description

Closes: #3816


tommyv1987 (approved):
lgtm

Issue #3817, opened 2023-08-28
NC - clicking connect does not select a new network-requester


octol:
Clicking connect does not select a new network-requester.
Meaning that if there is an issue with the network-requester, the user is stuck

See also: #3834 for a similar issue

The fix is to call `get_services` more frequently on the TS frontend.

tommyv1987:
Closing

PR #3843, opened 2023-09-05
fix(nc): refresh nr list on privacy switch

### LINKED ISSUE #3834: NC - toggling speedy mode on/off does not refresh the set of network-requesters
Toggling speedy/medium mode on/off does not refresh the list of network-requesters, meaning you are still stuck with the incorrect NR until you restart. 

This has possibly large impact on download speeds if in speedy mode you are stuck with an older vanilla NR still restricted to the Poisson rate limited.

You can check this by trying to manually select NR where it will display the current list of NR.

doums:
# Description

Closes: #3834 #3817


octol (changes requested):
Possibly it's me not being knowledgeable enough about the frontend part, but I don't think this solves https://github.com/nymtech/nym/issues/3817 yet? 

To clarity a little, clicking disconnect/connect should choose a new NR each time, otherwise if the got assigned a bad NR on first connect then they are stuck with a non-functioning setup

octol:
> Possibly it's me not being knowledgeable enough about the frontend part, but I don't think this solves https://github.com/nymtech/nym/issues/3817 yet? 
> 
> 
> 
> To clarity a little, clicking disconnect/connect should choose a new NR each time, otherwise if the got assigned a bad NR on first connect then they are stuck with a non-functioning setup

Ok might be my bad, if it re-selects a new one each time, we're all good

octol (approved):

PR #3864, opened 2023-09-07
Remove unused import

mmsinclair:
Fixes linting errors


jstuczyn (approved):
after thorough and very time consuming review, I can conclude that: lgtm

PR #3866, opened 2023-09-08
[feat] Socks5 and Native client: run with hardcoded topology

jstuczyn:
# Description

Another thing to have come out of https://github.com/nymtech/nym/issues/3838. This PR introduces a `--custom-mixnet` flag to socks5 and native clients. It lets you provide a json file containing list of mixnodes and gateways that will be used by the client instead of querying the nym-api. Those nodes don't need to be bonded or anything. I can imagine it might be helpful for your tests, @tommyv1987 and @benedettadavico 

A sample json file can look as follows (some fields like `mixPort` or `version` are optional; furthermore both `snake_case` and `camelCase` should be valid for deserialization):
```json
{
    "mixnodes": {
      "1": [
        {
          "mixId": 14,
          "owner": "foomp",
          "host": "109.74.197.180",
          "mixPort": 1789,
          "identityKey": "AnPotiqc7NE8VPBRQcaJQR3nM3KMS3JferiMhPwQkrwi",
          "sphinxKey": "AdmUgaeNuDD7uWAWwyXqt2qocirNdonusm9SN2wsSwma",
          "layer": 1,
          "version": "1.1.26"
        }
      ],
      "2": [
        {
          "mixId": 7,
          "owner": "n19wln95zj5r3wnepgk6nf7lqx0zgufvgtlvyawf",
          "host": "3.121.86.221",
          "mixPort": 1789,
          "identityKey": "2wEzyW6Pihk19xbCK9kB81vYuDFUvTzTAr77arvxdfSg",
          "sphinxKey": "CRaQojHncye6fhCF9qDHrVoHb1ESnPhyhoXeYQjFbZpp",
          "layer": 2,
          "version": "1.1.24"
        }
      ],
      "3": [
        {
          "mixId": 8,
          "owner": "n1ay7t9wwp88w5rg2amzgryylsykf26r8wmk4h7m",
          "host": "3.64.10.148",
          "mixPort": 1789,
          "identityKey": "A2todXdDJrRj3h7a54QCC6GxbPhYrg31uHt2S7U1shFa",
          "sphinxKey": "5MXsxarEPmd3GcqYDVqVpzouv6Gr1pVKraWFzGU1oMd1",
          "layer": 3,
          "version": "1.1.24"
        }
      ]
    },
    "gateways": [
      {
        "owner": "n1k5uwsuewkqhf3s6n5zx9ps2wkdgs9p9fhcdujn",
        "host": "85.159.212.96",
        "mixPort": 1789,
[...]

PR #3877, opened 2023-09-12
Feature/gateway inbuilt nr

### LINKED ISSUE #3838: Introduce a local network requester directly inside a gateway

jstuczyn:
# Description

~~Depends on https://github.com/nymtech/nym/pull/3866 getting merged first~~

Closes: https://github.com/nymtech/nym/issues/3838

# Checklist:


clients/native/src/client/config/mod.rs
```diff
@@ -67,7 +67,7 @@ pub struct Config {
 }
 
 impl NymConfigTemplate for Config {
-    fn template() -> &'static str {
+    fn template(&self) -> &'static str {
```
octol, line 70: Stylistic choice to add self here? Doesn't look like it's needed?

common/client-core/src/init/types.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    // pub fn is_custom_new(&self) -> bool {
+    //     if let GatewaySetup::New { specification } = self {
+    //         specification.is_custom()
+    //     } else {
+    //         false
+    //     }
+    // }
+
+    // pub async fn choose_gateway(
```
octol: Commented out code?

gateway/src/commands/helpers.rs
```diff
@@ -0,0 +1,271 @@
...
+    let Some(opts) = opts else { return cfg };
+
+    // as of 12.09.23 the below is true (not sure how this comment will rot in the future)
+    // medium_toggle:
+    // - sets secondary packet size to 16kb
+    // - disables poisson distribution of the main traffic stream
+    // - sets the cover traffic stream to 1 packet / 5s (on average)
+    // - disables per hop delay
+    //
+    // fastmode:
```
octol: side note for the future: `fastmode` should be renamed to `fast-poisson` or something

gateway/src/config/template.rs
```diff
@@ -73,6 +77,9 @@ keys.public_sphinx_key_file = '{{ storage_paths.keys.public_sphinx_key_file }}'
 # derived shared keys and available client bandwidths.
 clients_storage = '{{ storage_paths.clients_storage }}'
 
+# Path to the configuration of the locally running network requester.
```
octol: imho we should probably call it embedded or something, locally to me just means running on the same machine

gateway/src/node/mod.rs
```diff
@@ -194,46 +194,114 @@ impl<St> Gateway<St> {
         packet_sender
     }
 
-    async fn wait_for_interrupt(
+    // TODO: rethink the logic in this function...
+    async fn maybe_start_network_requester(
```
octol: style: I prefer to have the check in the caller scope and not call it maybe ...

octol (approved):

clients/native/src/client/config/mod.rs
jstuczyn: I had an idea of providing different templates based on whether embedded NR is enabled or not. I think this still might be a useful feature in the further stages of project smoosh

PR #3896, opened 2023-09-15
replaced '--disable-sign-ext' with '--signext-lowering' when running wasm-opt

jstuczyn:
# Description

This should allow us to build all the contracts with the most recent version of rust once more. 
Suggestion came from: https://github.com/CosmWasm/cosmwasm/issues/1727#issuecomment-1646513138

for QA:

make sure we're still able to upload, init and migrate contracts on our current verison of nyxd. because as we found out in the past, compilation is only one part of the story


octol (approved):

benedettadavico:
lgtm :)

PR #3902, opened 2023-09-18
Use TUN device for forwarding wireguard traffic

### LINKED ISSUE #3903: Create TUN device as a mechanism for forwarding wireguard traffic

octol:
# Description

Closes: #3903 

Add TUN device as a mechanism of relaying traffic coming out of the wireguard tunnel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2957,6 +2957,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "etherparse"
+version = "0.13.0"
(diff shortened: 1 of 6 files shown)


gyrusdentatus:
hmmm… was there some tech deep-dive article on the reason why are you implementing wg into nym? Turning the whole thing into tun was always kinda my way of thinking - although I know you’ve been dabbling with wg for some months now but still not sure for what purpose; therefore  -  any written update/notes on this would be nice ! <3

PR #3905, opened 2023-09-19
Docs: make shell scripts exit on errors so that CI jobs fail on build errors

mmsinclair:
It is always best to add the following to all CI build scripts, so that errors bubble up to the runner:

```
set -o errexit
set -o nounset
set -o pipefail
```


serinko (approved):
I don't master bash, but with the argument you made this looks 100% solid.

PR #3935, opened 2023-09-29
Fix clippy for latest stable toolchain

octol:
# Description

- Unpin Rust toolchain for wallet
- cargo clippy --fix
- clippy::redundant-guards
- clippy::incorrect_partial_ord_impl_on_ord_type
- clippy::unwrap-or-default
- noop_method_call

### DIFF
--- a/common/nymsphinx/params/src/packet_sizes.rs
+++ b/common/nymsphinx/params/src/packet_sizes.rs
@@ -89,7 +89,7 @@ pub enum PacketSize {
 impl PartialOrd for PacketSize {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         // order them by actual packet size
-        self.size().partial_cmp(&other.size())
+        Some(self.cmp(other))
     }
 }
 
--- a/ephemera/src/core/ephemera.rs
+++ b/ephemera/src/core/ephemera.rs
@@ -120,7 +120,7 @@ impl<A: Application> Ephemera<A> {
         while !shutdown.is_shutdown() {
             tokio::select! {
                 biased;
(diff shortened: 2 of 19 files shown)


common/nymsphinx/params/src/packet_sizes.rs
```diff
@@ -89,7 +89,7 @@ pub enum PacketSize {
 impl PartialOrd for PacketSize {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         // order them by actual packet size
-        self.size().partial_cmp(&other.size())
+        Some(self.cmp(other))
```
jstuczyn, line 92: Are we sure this is doing the same thing as before? We seem to no longer be comparing on `size()`

jstuczyn: Ah. `Ord` already does it.

ephemera/src/core/ephemera.rs
```diff
@@ -120,7 +120,7 @@ impl<A: Application> Ephemera<A> {
         while !shutdown.is_shutdown() {
             tokio::select! {
                 biased;
-                _ = shutdown.recv() => {
+                () = shutdown.recv() => {
```
jstuczyn, line 123: I'm surprised clippy cares about that one. I guess you'll have plenty of those to fix now

jstuczyn (approved):

ephemera/src/core/ephemera.rs
octol: We're seeing it only in ephemera as it has pedantic enabled

PR #3941, opened 2023-09-30
Add sdk example for geo-aware topology provider

octol:
# Description

Add example for how to use the geoaware topology provider in the Rust SDK.

### DIFF
--- a/sdk/rust/nym-sdk/examples/geo_topology_provider.rs
+++ b/sdk/rust/nym-sdk/examples/geo_topology_provider.rs
@@ -1,10 +1,6 @@
 // Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-use nym_client_core::client::topology_control::geo_aware_provider::{
-    CountryGroup, GeoAwareTopologyProvider,
-};
-use nym_client_core::config::GroupBy;
 use nym_sdk::mixnet;
 use nym_sdk::mixnet::MixnetMessageSender;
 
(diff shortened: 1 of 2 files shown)


sdk/rust/nym-sdk/examples/geo_topology_provider.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use nym_client_core::client::topology_control::geo_aware_provider::{
```
jstuczyn: We'd probably want to re-export those types as part of the sdk

octol: Yes good idea

sdk/rust/nym-sdk/examples/geo_topology_provider.rs
```diff
@@ -0,0 +1,51 @@
...
+    // ... or on a nym-address. This means we use the geo location of the gateway that the
+    // nym-address is connected to.
+    //let group_by = GroupBy::NymAddress("id.enc@gateway".parse().unwrap());
+
+    let geo_topology_provider = mixnet::GeoAwareTopologyProvider::new(
+        vec![nym_api],
+        // We filter on the version of the mixnodes. Be prepared to manually update
+        // this to keep this example working, as we can't (currently) fetch to current
+        // latest version.
+        "1.1.31".to_string(),
```
jstuczyn, line 25: do we need to have that explicit version filtering? can we maybe make it optional and/or deprecate it?

PR #3951, opened 2023-10-03
Feature/localnet

jstuczyn:
# Description

A helper script to speed up local development of mixnodes/gateways. It startups a local network without relying on any validators, contracts, etc.

Just run `./localnet_start.sh` from `scripts` directory and watch the magic happen!
note, you might need the following:
- openssl
- python3
- jq
- tmux
- rust (duh)
installed


octol (approved):
This is super useful

tommyv1987:
Looking good:
image

PR #3966, opened 2023-10-05
Github Actions: Add cargo clean and workflow trigger

raphael-walther:


raphael-walther:
The target directory isn't present on an idle runner. So cargo clean isn't needed.

PR #3976, opened 2023-10-10
[DOCS] Update admonish version && Fix formatting

serinko:
# Description

CD/CI returns error on admonish plugin, This shall fix it.


tommyv1987 (approved):
changes look good, this should also fix - https://github.com/nymtech/nym/pull/3993

PR #3977, opened 2023-10-11
wireguard: parse the public key up front in the UDP handler

### LINKED ISSUE #3987: Parse peer public key up front in UDP handler

octol:
# Description

Closes: #3987, NC-49

Try to have a flow in the UDP listener where we parse the public key up front instead of relying on knowing the socket address.


durch (approved):
Great, sets the stage for integrating with the gateway client registry, love the use of DashMap.

Unibitrum:
unsubscribe

PR #3992, opened 2023-10-13
Feature/configurable socks5 bind address

### LINKED ISSUE #3477: [FEATURE REQUEST] add option to can bind nym-socks5-client to 0.0.0.0 instead of 127.0.0.1 with a "--host" flag
**Where does the feature fit in the Nym real estate?**
nym-socks5-client / Nym Connect 

**What is this solving?**
Add option to can bind nym-socks5-client to 0.0.0.0 instead of 127.0.0.1 with a "--host" flag without using docker
This could be useful to use socks5 client from other different machines in the same local network

**Is this an update to packages or libraries?**
I don't know

**Additional context**
There is a method to get this, but only using docker: https://nymtech.net/docs/clients/socks5-client.html#configuring-your-client-for-docker
I'm not using docker

When I try to use the flag in command: `./nym-socks5-client init --id docs-example --host 0.0.0.0`
output:
`error: unexpected argument '--host' found`

### LINKED ISSUE #3536: Allow the socks5 client (and all of its derivatives like `nym-connect`) to set custom listening `host`.
It seems to be a rather requested feature by the community: https://github.com/nymtech/nym/issues/3477

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/3536
Closes: https://github.com/nymtech/nym/issues/3477

### DIFF
--- a/clients/native/src/client/config/old_config_v1_1_20_2.rs
+++ b/clients/native/src/client/config/old_config_v1_1_20_2.rs
@@ -11,6 +11,7 @@ use crate::{
 use nym_bin_common::logging::LoggingSettings;
 use nym_client_core::config::disk_persistence::old_v1_1_20_2::CommonClientPathsV1_1_20_2;
 use nym_client_core::config::old_config_v1_1_20_2::ConfigV1_1_20_2 as BaseConfigV1_1_20_2;
+use nym_client_core::config::old_config_v1_1_30::ConfigV1_1_30 as BaseConfigV1_1_30;
 use nym_client_core::config::GatewayEndpointConfig;
 use nym_config::read_config_from_toml_file;
 use nym_network_defaults::DEFAULT_WEBSOCKET_LISTENING_PORT;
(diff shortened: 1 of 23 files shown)


durch (approved):

tommyv1987:
all good now, merging in... 🚀

twofaktor:
Where will be set this option, using the "init" or "run" command? with both?

Will be used "run" using flags `--host 0.0.0.0 --port 1080` to set either of the two parameters?

Thanks

jstuczyn:
> Where will be set this option, using the "init" or "run" command? with both?
> 
> Will be used "run" using flags `--host 0.0.0.0 --port 1080` to set either of the two parameters?
> 
> Thanks

you can use it with either, but there's a caveat:
- if you use it with `init`, it will be written to the `config.toml` file and it will be used in all subsequent `run`s
- if you use it with `run`, it will simply override whatever is already in `config.toml` and you will need to pass the `--host` flag every single time you want to use it; alternatively, if your client has already been initialised, you can modify `config.toml`. you will be able to find the setting under `[core.socks5].bind_adddress`. note that for that option to appear in the config.toml, you will need to run the client at least once after the update [i.e. when this feature goes live] so that the config gets properly migrated.

PR #3995, opened 2023-10-13
Replace HashMap with DashMap

durch:
Prepare for integrating with `nym-wireguard`, and `DashMap` looks pretty cool


octol (approved):
Looking good!

PR #4006, opened 2023-10-17
Add wrapper for wasm client and mix-fetch in nodejs

sebastinez:
# Description

Adds a NodeJS version for mix-fetch.
Adds a NodeJS wrapper for running a nym WASM client.

- Uses the existing SDK worker and a wasm packed for NodeJS.

This PR doesn't provide a full-fat version yet, since wasm-pack has some quirks on how the wasm bytes are loaded for `--target node`, eventually I'll be able to add it to this PR.

# Checklist:


mmsinclair:
Nice one @sebastinez 😀

If you could also add the output `dist/ts/node-client/esm` etc directories to:
https://github.com/nymtech/nym/blob/develop/sdk/typescript/scripts/publish.sh#L34-L37
and https://github.com/nymtech/nym/blob/develop/sdk/typescript/scripts/unpublish.sh#L21-L24

PR #4010, opened 2023-10-17
Integrate gateway registry with nym-wireguard

durch:
# Description

Integrate gateway registry with nym-wireguard, extract common types.

Related: NC-54


durch:
Rebased on the latest develop

common/types/src/lib.rs
```diff
@@ -9,8 +9,10 @@ pub mod error;
 pub mod fees;
 pub mod gas;
 pub mod gateway;
+pub mod gateway_client_registration;
 pub mod helpers;
 pub mod mixnode;
 pub mod pending_events;
 pub mod transaction;
 pub mod vesting;
+pub mod wireguard;
```
octol, line 20: I wonder if we should place all the wireguard stuff in this crate behind the `wireguard` feature toggle? Then we don't have to worry about nym-connect and the wallet pulling in wireguard

durch: This is the `PeerPublicKey` type, so its neater this way, I'll put it behind a feature flag tough

durch: Practical problem was that there was a circular dependency between gateway and wireguard, so the types got extracted out

octol (approved):

PR #3970, opened 2023-10-06
Feature/gateway api

jstuczyn:
# Description

Closes: https://github.com/nymtech/nym/issues/3893
Closes: https://github.com/nymtech/nym/issues/4017

This PR does slightly more than what its ticket might suggest. This is a next, albeit small, step towards the completed project smoosh. It introduces initial http API that will be used by nym-nodes, but currently only enabled on gateways. This enables gateways to self-describe themselves. In particular, what is relevant in this issue, to specify whether they support `wss` and on which port. Nym-api then aggregates this information and serves it to the clients.

# Checklist:

- [ ] added a changelog entry to `CHANGELOG.md`

PR #4016, opened 2023-10-18
Add configurable http api port to gateway

neacsu:


jstuczyn:
i imagine this will most likely get superseded by https://github.com/nymtech/nym/pull/3970 once merged. 
we'll want to have gateways/nym-nodes expose their API on one of predefined ports: 80/443 if properly setup with SSL (which will be obligatory down the line), or 8080 otherwise; so there's going to be no need to put anything in the contract

neacsu:
> i imagine this will most likely get superseded by #3970 once merged. we'll want to have gateways/nym-nodes expose their API on one of predefined ports: 80/443 if properly setup with SSL (which will be obligatory down the line), or 8080 otherwise; so there's going to be no need to put anything in the contract

Given that, from what I last heard, it's not clear if we will have these ports on the gateway, or in a reverse proxy, having it configurable should still be ok.

durch (approved):

jstuczyn:
> > i imagine this will most likely get superseded by #3970 once merged. we'll want to have gateways/nym-nodes expose their API on one of predefined ports: 80/443 if properly setup with SSL (which will be obligatory down the line), or 8080 otherwise; so there's going to be no need to put anything in the contract
> 
> Given that, from what I last heard, it's not clear if we will have these ports on the gateway, or in a reverse proxy, having it configurable should still be ok.

I think we should sync up with @mmsinclair on his current thoughts about that. because afaik we didn't want to put yet more data in the contract and nodes should have been using predictable ports. obviously, if operators wanted, they could set custom ports locally and expose the correct ones via proxy (but that's already covered by my feature)

neacsu:
Ok, I'll close this PR then, as the customisable port will already be covered by another PR.

PR #4032, opened 2023-10-24
wireguard: add packet relayer

### LINKED ISSUE #3938: Tear apart wg UDP listener and TUN device so they can be placed them on the opposite side of the mixnet

octol:
# Description

Closes: #3938
NC-48

Add PacketRelayer. The is the last step to enable splitting the UDP listener and the TUN device to exist on separate nodes.

### DIFF
--- a/common/wireguard/src/wg_tunnel.rs
+++ b/common/wireguard/src/wg_tunnel.rs
@@ -14,10 +14,7 @@ use tokio::{
     time::timeout,
 };
 
-use crate::{
-    error::WgError, event::Event, network_table::NetworkTable, registered_peers::PeerIdx,
-    tun_task_channel::TunTaskTx,
-};
+use crate::{error::WgError, event::Event, network_table::NetworkTable, registered_peers::PeerIdx};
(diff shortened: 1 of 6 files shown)


common/wireguard/src/wg_tunnel.rs
```diff
@@ -117,13 +113,17 @@ impl WireGuardTunnel {
...
-            tun_task_tx: tunnel_tx,
+            packet_tx,
             tag,
         };
 
         (tunnel, peer_tx, tag)
     }
 
+    fn new_tag() -> u64 {
+        rand::thread_rng().next_u64()
```
neacsu, line 125: Although small, there's still a possibility for collisions here, right? Maybe at least have some TODO comment for now reminding to check for tag collisions after generating a random one

neacsu (approved):

common/wireguard/src/wg_tunnel.rs
octol: Yep good point

PR #4056, opened 2023-10-25
suppress error output

tommyv1987:
- we can run the api and set the logging level differently on launch
- same offending characters are spamming the logs
- write to tmp file if the same offending characters appear

### DIFF
--- a/explorer-api/src/geo_ip/location.rs
+++ b/explorer-api/src/geo_ip/location.rs
@@ -118,12 +118,12 @@ impl GeoIp {
     }
 
     pub fn query(&self, address: &str, port: Option<u16>) -> Result<Option<Location>, GeoIpError> {
-        let ip: IpAddr = FromStr::from_str(address).or_else(|_| {
+        let p = port.unwrap_or(FAKE_PORT);
+        let ip_result: Result<IpAddr, GeoIpError> = FromStr::from_str(address).or_else(|_| {
             debug!(
                 "Fail to create IpAddr from {}. Trying using internal lookup...",
                 &address
             );
-            let p = port.unwrap_or(FAKE_PORT);
             match (address, p).to_socket_addrs() {
                 Ok(mut addrs) => {
                     if let Some(socket_addr) = addrs.next() {
@@ -142,7 +142,10 @@ impl GeoIp {
                     Err(GeoIpError::NoValidIP)
                 }
             }
-        })?;
+        });
+
+        let ip = ip_result?;
+
         let result = self
             .db
             .as_ref()
--- a/explorer-api/src/geo_ip/location.rs
+++ b/explorer-api/src/geo_ip/location.rs
@@ -107,18 +107,24 @@ impl GeoIp {
                         Ok(ip)
                     } else {
                         debug!("Fail to resolve IP address from {}:{}", &address, p);
-                        let mut failed_ips_guard = self.failed_addresses.failed_ips.lock().unwrap();
-                        if failed_ips_guard.insert(address.to_string()) {
-                            append_ip_to_file(address);
+                        if let Ok(mut failed_ips_guard) = self.failed_addresses.failed_ips.lock() {
+                            if failed_ips_guard.insert(address.to_string()) {
+                                append_ip_to_file(address);
+                            }
+                        } else {
+                            error!("Failed to acquire lock on failed_ips");
                         }
                         Err(GeoIpError::NoValidIP)
                     }
                 }
                 Err(_) => {
                     debug!("Fail to resolve IP address from {}:{}.", &address, p);
-                    let mut failed_ips_guard = self.failed_addresses.failed_ips.lock().unwrap();
-                    if failed_ips_guard.insert(address.to_string()) {
-                        append_ip_to_file(address);
+                    if let Ok(mut failed_ips_guard) = self.failed_addresses.failed_ips.lock() {
+                        if failed_ips_guard.insert(address.to_string()) {
+                            append_ip_to_file(address);
+                        }
+                    } else {
+                        error!("Failed to acquire lock on failed_ips");
                     }
                     Err(GeoIpError::NoValidIP)
                 }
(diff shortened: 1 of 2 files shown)


explorer-api/src/geo_ip/location.rs
```diff
@@ -78,20 +99,30 @@ impl GeoIp {
...
-            Ok(ip)
+            match (address, p).to_socket_addrs() {
+                Ok(mut addrs) => {
+                    if let Some(socket_addr) = addrs.next() {
+                        let ip = socket_addr.ip();
+                        debug!("Internal lookup succeeded, resolved ip: {}", ip);
+                        Ok(ip)
+                    } else {
+                        debug!("Fail to resolve IP address from {}:{}", &address, p);
+                        let mut failed_ips_guard = self.failed_addresses.failed_ips.lock().unwrap();
```
octol: imho don't unwrap if this fails, just skip appending and print warning

explorer-api/src/geo_ip/location.rs
```diff
@@ -78,20 +99,30 @@ impl GeoIp {
...
+                        if failed_ips_guard.insert(address.to_string()) {
+                            append_ip_to_file(address);
+                        }
+                        Err(GeoIpError::NoValidIP)
+                    }
+                }
+                Err(_) => {
+                    debug!("Fail to resolve IP address from {}:{}.", &address, p);
+                    let mut failed_ips_guard = self.failed_addresses.failed_ips.lock().unwrap();
```
octol: Same as other comment

explorer-api/src/helpers.rs
```diff
@@ -2,9 +2,27 @@
...
+    format!("{}/failed_ips.txt", home_dir)
+}
+
+pub fn append_ip_to_file(address: &str) {
+    if let Ok(mut file) = OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(failed_ips_filepath())
+    {
+        writeln!(file, "{}", address).expect("Failed to write to file");
```
octol: this one too, skip the expect :)

explorer-api/src/geo_ip/location.rs
jstuczyn: in this case I could argue the unwrap is fine given it implies the mutex got poisoned, so there isn't much you could do anyway - but perhaps put an `expect` there instead for easier debugging

explorer-api/src/helpers.rs
jstuczyn: but definitely don't expect/unwrap here : ) 
either propagate or log the error here

explorer-api/src/helpers.rs
```diff
@@ -2,9 +2,27 @@
...
 
 pub(crate) fn best_effort_small_dec_to_f64(dec: Decimal) -> f64 {
     let num = dec.numerator().u128() as f64;
     let den = dec.denominator().u128() as f64;
     num / den
 }
+
+pub fn failed_ips_filepath() -> String {
+    let home_dir = env::var("HOME").unwrap_or_else(|_| "/tmp".to_string());
+    format!("{}/failed_ips.txt", home_dir)
```
jstuczyn: I'd perhaps use `join` method on `PathBuf` / `Path` instead

explorer-api/src/geo_ip/location.rs
```diff
@@ -42,6 +46,17 @@ pub(crate) struct Location {
...
 }
 
+pub(crate) struct FailedIpAddresses {
+    failed_ips: Arc<Mutex<HashSet<String>>>,
+}
+
+impl FailedIpAddresses {
+    pub fn new() -> Self {
+        FailedIpAddresses {
+            failed_ips: Arc::new(Mutex::new(HashSet::new())),
```
jstuczyn: You want one more thing here -> on startup it should check if the file already exists and load its content because on each restart you'll end up duplicating everything

explorer-api/src/geo_ip/location.rs
```diff
@@ -78,20 +99,36 @@ impl GeoIp {
...
+                Err(_) => {
+                    debug!("Fail to resolve IP address from {}:{}.", &address, p);
+                    if let Ok(mut failed_ips_guard) = self.failed_addresses.failed_ips.lock() {
+                        if failed_ips_guard.insert(address.to_string()) {
+                            append_ip_to_file(address);
+                        }
+                    } else {
+                        error!("Failed to acquire lock on failed_ips");
+                    }
+                    Err(GeoIpError::NoValidIP)
```
octol, line 142: This block looks the same as the one above so consider extract out to a method

explorer-api/src/geo_ip/location.rs
octol: Also, does it need to be in a Arc Mutex? the parent struct is already in a Arc?

jstuczyn: oh yeah, probably you don't need the `Arc` at all

tommyv1987: give me a second to make changes guys 😉

explorer-api/src/geo_ip/location.rs
```diff
@@ -78,20 +124,24 @@ impl GeoIp {
...
-                .ok_or_else(|| {
-                    error!("Fail to resolve IP address from {}:{}", &address, p);
-                    GeoIpError::NoValidIP
-                })?;
-            let ip = socket_addr.ip();
-            debug!("Internal lookup succeed, resolved ip: {}", ip);
-            Ok(ip)
+            match (address, p).to_socket_addrs() {
+                Ok(mut addrs) => {
+                    if let Some(socket_addr) = addrs.next() {
```
jstuczyn, line 129: you'd need to test it (but where would we find a qualified QA person? 🤔 ), but I think since this is an iterator, it might always return a `None` at the very end and thus you will always be hiting the `else` branch in every lookup.

so you could just do something along the lines of

```rust
for socket_addr in addr {
  let ip = socket_addr.ip();
  // do the rest 
}
```

octol (approved):

PR #4064, opened 2023-10-26
Run nym-ip-packet-router in gateway

octol:
# Description

NC-52

Run embedded ip-packet-router in gateway

- Add a hidden flag `--with-ip-packet-router` to the gateway `init` and `run` commands, mirroring `--with-network-requester`.
- Add ip_packet_router section to gateway config file: 

```
[ip_packet_router]
enabled = false
```

- Add hidden `setup-ip-packet-router` command that mirrors the existing `setup-network-requester`

Running with ip-packet-router enabled will currently not work.


gateway/src/commands/init.rs
```diff
@@ -79,9 +82,13 @@ pub struct Init {
     statistics_service_url: Option<url::Url>,
 
     /// Allows this gateway to run an embedded network requester for minimal network overhead
-    #[clap(long)]
+    #[clap(long, conflicts_with = "with_ip_forwarder")]
```
jstuczyn: should they be conflicting with each other? in principle, since they have different addresses, why can't we allow both?

gateway/src/commands/run.rs
```diff
@@ -15,39 +15,41 @@ use nym_node::error::NymNodeError;
 use std::net::IpAddr;
 use std::path::PathBuf;
 
+use super::helpers::OverrideIpForwarderConfig;
+
 #[derive(Args, Clone)]
 pub struct Run {
     /// Id of the gateway we want to run
-    #[clap(long)]
+    #[arg(long)]
```
jstuczyn, line 23: ok, what's the difference between `clap` and `arg`, because I think I've actually been recently changing them around...

gateway/src/node/client_handling/embedded_network_requester/mod.rs
```diff
@@ -28,6 +28,17 @@ impl LocalNetworkRequesterHandle {
         }
     }
 
+    // TODO: generalize this whole thing to be general. And change the name(s).
```
jstuczyn, line 31: just a comment so you wouldn't forget your own comment here : )

gateway/src/commands/run.rs
octol: `arg` is what the latest clap docs are using, so I assume that's the latest and greatest?

gateway/src/commands/init.rs
octol: In principle yes, in practice, for now I'm taking shortcuts elsewhere to assume this is strictly either

gateway/src/node/client_handling/embedded_network_requester/mod.rs
octol: yep, although that comment is for future PRs :)

jstuczyn (approved):

PR #4069, opened 2023-10-27
[feat] Enforcing forward travel

jstuczyn:
# Description

NC-112

This pull request introduces a currently optional feature for mixnodes and gateways for enforcing forward travel of mix packets. What it means is that, when enabled, nodes will refuse to accept connections from nodes that are not on the previous layer. Similarly, they will not forward any packets to nodes that are not on the next layer.

One question is whether we should perhaps enable it already by default or leave it optional and keep it in our back pocket for later.

Note that this is a naive implementation with bunch of limitations, but it's imo a good starting point.


gateway/src/config/old_config_v1_1_29.rs
```diff
@@ -164,6 +164,10 @@ impl From<ConfigV1_1_29> for Config {
                 stored_messages_filename_length: value.debug.stored_messages_filename_length,
                 message_retrieval_limit: value.debug.message_retrieval_limit,
                 use_legacy_framed_packet_version: value.debug.use_legacy_framed_packet_version,
+
+                // \/ ADDED
+                // enforce forward travel, et al.
+                ..Default::default() // /\ ADDED
```
octol: hmm ... can we get away with this? is all that conversion code in my PR not needed?

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -203,11 +206,25 @@ where
...
+    fn forward_packet(&self, mix_packet: MixPacket) -> Result<(), RequestHandlingError> {
+        let next_hop: SocketAddr = mix_packet.next_hop().into();
+
+        // TODO: another option is to move this filter
+        // (which is used by EVERY `ConnectionHandler`, so potentially hundreds of times)
+        // to packet forwarder where we could be filtering at the time of attempting to open new outbound connections
+        // However, in that case we wouldn't be able to return an error message to the client
+        if !self.inner.allowed_egress.is_allowed(next_hop.ip()) {
+            // TODO: perhaps this should get lowered in severity?
+            warn!("received an packet that was meant to get forwarded to {next_hop}, but this address does not belong to any node on the next layer - dropping the packet");
```
octol, line 218: yeah I'm leaning towards `info`

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -40,13 +40,13 @@ pub(crate) enum RequestHandlingError {
     #[error("Internal gateway storage error")]
     StorageError(#[from] StorageError),
 
-    #[error("Provided bandwidth IV is malformed - {0}")]
+    #[error("Provided bandwidth IV is malformed: {0}")]
```
octol, line 43: Since you're touching all strings, change to lowercase for first letter :)

gateway/src/node/mixnet_handling/receiver/listener.rs
```diff
@@ -42,6 +52,12 @@ impl Listener {
                 connection = tcp_listener.accept() => {
                     match connection {
                         Ok((socket, remote_addr)) => {
+                             if !self.allowed_ingress.is_allowed(remote_addr.ip()) {
+                                // TODO: perhaps this should get lowered in severity?
+                                warn!("received an incoming connection from {remote_addr}, but this address does not belong to any node on the previous layer - dropping the connection");
```
octol, line 57: yeah I think let's use INFO, since nothing is going wrong from the operators perspective. I'd imagine most operators run their nodes with RUST_LOG=warn and are only interested in stuff they need to be aware of

octol (approved):

mmsinclair:
Replaced by #4360

PR #4106, opened 2023-11-06
Quick addon on Exit Gateway setup

serinko:
# Description

Adding a tweak to change NR config to enable Exit policy for a GW with NR already connected.


mfahampshire (approved):
lgtm

PR #4115, opened 2023-11-07
Feature/increase default sdk surbs

### LINKED ISSUE #3627: Client running out of SURBs
Getting reports from users that the client is running out of SURBs

> What can we do when this happen on the nym-client service provider side ? 
> We send 100 SURB in each sendMessage 
> `2023-06-28T20:04:48.978Z WARN nym_client_core::client::replies::reply_controller > failed to request additional surbs... - Not enough reply SURBs to send the message. We have 0 available and require at least 1.`

mfahampshire:
# Description

This was originally to close: #3627 - as per [this comment]() whether it does close it is somewhat up for debate. Regardless still upping the number of SURBs (TS surbs especially) and making them uniform across sdks/clients. 

- [x] up default # of surbs 
     - [x] rust sdk 
     - [x] ts sdk 
- [x] fixed name of surb reply example to follow convention of other examples


mfahampshire:
I think there's a question of whether we want to enable the socks5 client to send them by default instead having to set via flag, its been a while since we introduced it, and I imagine most NRs will now be up to date enough to handle SURBs. 

You have to manually set the # of SURBs to send with `sendAnonymous` messages with the native client, so nothing to do there.

serinko:
It looks good to me. If the versions of NRs is the only question in the decision, I would say go with default `--use-reply-surbs` to be set as `true`.

jstuczyn (approved):

PR #4123, opened 2023-11-09
Disable poisson in ip-packet-router

### LINKED ISSUE #4130: Disable poisson in nym-ip-packet-router

octol:
# Description

Closes #4130 

Disable the poisson rate in the ip-packet-router by default


neacsu (approved):
We should probably disable it on both ends, gateway side and client side

PR #4127, opened 2023-11-10
IPR: add exit policy

### LINKED ISSUE #4129: Exit policy in nym-ip-packet-router

octol:
# Description

Closes: #4129 

Add exit policy packet check to `nym-ip-packet-router`.


service-providers/ip-packet-router/Cargo.toml
```diff
@@ -15,12 +15,15 @@ log = { workspace = true }
...
 nym-config = { path = "../../common/config" }
+nym-exit-policy = { path = "../../common/exit-policy" }
+nym-network-requester = { path = "../network-requester" }
 nym-sdk = { path = "../../sdk/rust/nym-sdk" }
 nym-service-providers-common = { path = "../common" }
 nym-sphinx = { path = "../../common/nymsphinx" }
 nym-task = { path = "../../common/task" }
 nym-wireguard = { path = "../../common/wireguard" }
 nym-wireguard-types = { path = "../../common/wireguard-types" }
+reqwest.workspace = true
```
jstuczyn, line 26: oh. i didn't know you could use that syntax, that's useful!

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -1,4 +1,9 @@
-use std::{net::IpAddr, path::Path};
+#![cfg_attr(not(target_os = "linux"), allow(dead_code))]
```
jstuczyn, line 1: hehe, nice

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -235,26 +253,34 @@ impl IpPacketRouter {
...
+        // reason to still forward them.
+        //
+        // For example, if we are running in a mode where we are only supposed to forward
+        // packets to a specific destination, we might want to forward them anyway.
+        //
+        // TODO: look into this
+        let (packet_type, src_addr, dst_addr, dst) = parse_packet(&reconstructed.message)?;
+
+        let dst_str = dst.map_or(dst_addr.to_string(), |dst| dst.to_string());
+        log::info!("Received packet: {packet_type}: {src_addr} -> {dst_str}");
```
jstuczyn, line 271: should we be using an `info` for every single ip packet received? that's a bit excessive imo

service-providers/ip-packet-router/src/request_filter/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+            .upstream_exit_policy_url
+            .as_ref()
+            .ok_or(IpPacketRouterError::NoUpstreamExitPolicy)?;
+        let policy_filter = ExitPolicyRequestFilter::new_upstream(upstream_url.clone()).await?;
+        Ok(RequestFilter {
+            inner: Arc::new(RequestFilterInner::ExitPolicy { policy_filter }),
+        })
+    }
+
+    pub(crate) async fn check_address(&self, address: &SocketAddr) -> bool {
```
jstuczyn, line 55: nitpicking: `SocketAddress` is a rather cheap `Copy` type so you might want to just take it by value instead

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -235,26 +253,34 @@ impl IpPacketRouter {
...
+
+        // Filter check
+        if let Some(dst) = dst {
+            if !self.request_filter.check_address(&dst).await {
+                let log_msg = format!("Failed filter check: {dst:?}");
+                log::warn!("{log_msg}");
```
jstuczyn: why not format it inside the log itself?

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -235,26 +253,34 @@ impl IpPacketRouter {
...
+
+                // TODO: send back a response here
```
jstuczyn, line 278: well, do we? isn't it the case that if the underlying client fails to get acks multiple in a row, it fail just give up? like say when you're trying to load a website in your browser with disabled internet, it will just give you no internet error at some point

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -235,26 +253,34 @@ impl IpPacketRouter {
...
+
+                return Err(IpPacketRouterError::AddressFailedFilterCheck { addr: dst });
             }
-        };
-        log::info!("Received packet: {src_addr} -> {dst_addr}");
+        } else {
+            // TODO: are we always allowing packets without port numbers?
+            log::warn!(
+                "Ignoring filter check for packet without port number! (TODO: is this correct?)"
```
jstuczyn, line 285: yeah, good question

service-providers/ip-packet-router/src/lib.rs
```diff
@@ -270,6 +296,47 @@ impl IpPacketRouter {
     }
 }
 
+fn parse_packet(
+    packet: &[u8],
+) -> Result<(&str, IpAddr, IpAddr, Option<SocketAddr>), IpPacketRouterError> {
```
jstuczyn: given your return type has 4 different things in it (and both being the same type), maybe you should wrap it on some struct instead?

service-providers/ip-packet-router/src/lib.rs
octol: No eventually we should rework this for sure. Cap the message to at most 1 per sec or something

service-providers/ip-packet-router/src/lib.rs
octol: Yep, but I was thinking that optionally we could be "polite" and send back a request that is parseable in in the nym-client to log that this was the reason there are no acks. Not sure, just an idea

jstuczyn (approved):

PR #4134, opened 2023-11-13
made taskclient shutdown log more verbose

### LINKED ISSUE #4105: [Feature Request] Add extra bit of info to `task client is being dropped` logging
`task client is being dropped` is a little vague: make the message more informative so that when people see it on shutdown it is expected

mfahampshire:
# Description

Closes: #4105


jstuczyn:
I'd maybe call it client shutdown than client disconnect, but from that, lgtm, 🚢 it!

jstuczyn (approved):

PR #4158, opened 2023-11-17
Feature/nymvisor

jstuczyn:
# Description

Closes: [NC-143](https://nymtech.atlassian.net/browse/NC-143)

Full description of the tool, its usage and purpose is available here: https://github.com/nymtech/nymsphere/pull/2

To work in its full capacity, it needs: https://github.com/nymtech/nym-websites/pull/143

[NC-143]: https://nymtech.atlassian.net/browse/NC-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/nymvisor/src/cli/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+mod add_upgrade;
+mod build_info;
+mod config;
+mod daemon_build_info;
+pub(crate) mod helpers;
+mod init;
+mod run;
+
+lazy_static! {
```
octol: use std::cell::OnceCell instead?

tools/nymvisor/src/cli/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        .collect::<Vec<_>>();
+
+    if instances.len() != 1 {
+        return Err(NymvisorError::NotSingleton {
+            instances: instances.len(),
+        });
+    }
+
+    // safety: that unwrap is fine as we've just checked we have 1 entry in the vector
+    #[allow(clippy::unwrap_used)]
```
octol, line 129: Move it inside the else branch above instead?

tools/nymvisor/src/cli/run.rs
```diff
@@ -0,0 +1,82 @@
...
+    info!("starting nymvisor for {}", config.daemon.name);
+
+    // TODO: experiment with other minimal runtimes, maybe futures::executor::LocalPool
+    //
+    // well, if the creation of the runtime failed, there isn't much we could do
+    #[allow(clippy::expect_used)]
+    let rt = runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .expect("failed to create the runtime");
```
octol: Why not return an error instead?

tools/nymvisor/src/cli/mod.rs
jstuczyn: it'd have a slightly different meaning then. if I moved it to the `else` branch I'd be checking if I have **at least** one instance. meanwhile I want to check I have **exactly** one instance

PR #4168, opened 2023-11-21
Add wireguard_mode flag to SDK

neacsu:


common/client-core/src/client/base_client/mod.rs
```diff
@@ -346,6 +357,12 @@ where
             if let Some(existing_client) = initialisation_result.authenticated_ephemeral_client {
                 existing_client.upgrade(packet_router, bandwidth_controller, shutdown)
             } else {
+                if wireguard_mode {
+                    gateway_config.gateway_listener = format!(
+                        "ws://{}:{}",
```
jstuczyn: this looks like a nasty return of our wss issues haha. but, I'm not sure this is what you want `start_gateway_client` is called after `init` has already been called -> meaning you have already used wrong ip address. 
By the looks of things you want to go layer up and overwrite gateway's ip address with the `WG_TUN_DEVICE_ADDRESS`. 
So perhaps make sure your client calls `GatewaySetup::New` with your gateway id and where your `available_gateways` has adjusted ip address for your node - kinda like what we used to be doing in wasm to cheat with tls

neacsu: Right, I moved the assignment higher up. The reason I didn't encounter this on init is because I was initing the client by hand before starting it in the vpn client, due to a circular dependency that is yet to be resolved. 
As to the wss, that won't be needed when using vpn mode, as the request would be on the local machine anyway.

jstuczyn (approved):

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -503,26 +528,57 @@ where
...
+                    "ws://{}:{}",
+                    WG_TUN_DEVICE_ADDRESS, DEFAULT_CLIENT_LISTENING_PORT
+                );
+                if let Err(e) = self
+                    .storage
+                    .gateway_details_store()
+                    .store_gateway_details(&PersistedGatewayDetails::Default(config))
+                    .await
+                {
+                    warn!("Could not switch to using wireguard more - {:?}", e);
```
octol: s/more/mode

octol (approved):

PR #4099, opened 2023-11-02
Explore running ci-docs on gh runner

octol:
# Description

Explore running the ci-docs workflow on gh runner to make it more predictable

PR #4172, opened 2023-11-23
Chore/clean config in docs runner

mfahampshire:
~~This PR is taking a different approach to the one linked above. That one is removing the need to do what is happening here - removing old configs - by performing the action on a fresh runner. The question is what we can do on the fresh runners and what we need to keep on our own machines. I am investigating this.~~

~~Bringing the approach from above referenced PR - building and deploying on a fresh machine each time.~~ 

UPDATE: running on its own machine works but is very slow. Exploring instead removing existing installs on our runner each time.

UPDATE2: best of both worlds: switching to 16 core GH runner for fast and ephemeral build env  
- [x] test full flow : ci -> cd -> vercel deploy
- [x] remove todos in `ci-docs`
- [x] add changes from `ci-docs` to `cd-docs`
- [x] remove `ci-docs-old` and `cd-docs-old` 

NB squash and merge


mfahampshire:
~~Looking @ https://github.com/nymtech/nym/pull/4099 as well to bring in~~

mfahampshire:
This PR is taking a different approach to the one linked above. That one is removing the need to do what is happening here - removing old configs - by performing the action on a fresh runner. The question is what we can do on the fresh runners and what we need to keep on our own machines. I am investigating this.

serinko (approved):

PR #4173, opened 2023-11-23
Change mixnode, gateway, network and nym-api binary licenses to GPLv3

mmsinclair:
Changes the licenses for the binaries that run Nym nodes and APIs to GPLv3

### DIFF
--- a/gateway/gateway-requests/Cargo.toml
+++ b/gateway/gateway-requests/Cargo.toml
@@ -1,5 +1,5 @@
 # Copyright 2020 - Nym Technologies SA <contact@nymtech.net>
-# SPDX-License-Identifier: GPL-3.0-only
+# SPDX-License-Identifier: Apache-2.0
 
 [package]
 name = "nym-gateway-requests"
--- a/gateway/gateway-requests/Cargo.toml
+++ b/gateway/gateway-requests/Cargo.toml
@@ -1,5 +1,5 @@
 # Copyright 2020 - Nym Technologies SA <contact@nymtech.net>
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: GPL-3.0-only
 
 [package]
 name = "nym-gateway-requests"
--- a/nym-api/nym-api-requests/src/coconut.rs
+++ b/nym-api/nym-api-requests/src/coconut.rs
@@ -1,5 +1,5 @@
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
-// SPDX-License-Identifier: GPL-3.0-only
+// SPDX-License-Identifier: Apache-2.0
 
 use cosmrs::AccountId;
 use getset::{CopyGetters, Getters};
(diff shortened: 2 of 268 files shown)


gateway/gateway-requests/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 # Copyright 2020 - Nym Technologies SA <contact@nymtech.net>
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: GPL-3.0-only
```
jstuczyn: you don't want to be changing this one, it's part of client library to interact with the gateway (and not the actual gateway). this comment applies to everything in `gateway-requests`

nym-api/nym-api-requests/src/coconut.rs
```diff
@@ -1,5 +1,5 @@
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
-// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: GPL-3.0-only
```
jstuczyn: same problem here as with gateway-requests

jstuczyn (changes requested):

octol:
IMHO we should add the license to the Cargo.toml files as well

gateway/gateway-requests/Cargo.toml
mmsinclair: Done

nym-api/nym-api-requests/src/coconut.rs
mmsinclair: Done

mmsinclair:
> IMHO we should add the license to the Cargo.toml files as well

Good plan, added this

jstuczyn (approved):

numbleroot:
Hi! Would it make sense to add the GPLv3 license file to the other licenses distributed as part of the Nym repository in the LICENSES folder?

timkuijsten:
Can I read somewhere about the reasons for the license change?

mmsinclair:
> Hi! Would it make sense to add the GPLv3 license file to the other licenses distributed as part of the Nym repository in the LICENSES folder?

thanks @numbleroot that was an oversight, will add it

mmsinclair:
> Can I read somewhere about the reasons for the license change?

Good question @timkuijsten - we are adding functionality to the gateways to accept wireguard connections, so that users can choose a less-private but faster mode for certain types of traffic, and we will be re-using some libraries that are GPLv3 licensed to manage tunnel devices, hence the change for the binaries.

Please note, the core of Nym's libraries will remain Apache 2.0, so that they can be easily used in software.

Also, bear in mind that we are part-way through smooshing all of Nym's binaries together into a single binary that offers all functionality that will simply be called `nym-node`. The `nym-node` binary will be GPLv3 licensed.

The standalone Nym clients (SOCKS5/websockets), will still remain Apache 2.0 licensed, so that they can be easily included in software.

Hope that clears things up 😁

PR #4197, opened 2023-11-29
Feature/kernel wg gw

neacsu:


octol (approved):
Maybe there are stuff in common/wireguard-types that we can now remove?

PR #4205, opened 2023-12-01
Overide number of mix hops separately per packet

octol:
# Description

Allow setting the number of mix hops separately for each packet. The use case is to allow the `nym-ip-packet-router` to specify this differently per each connected client.

## Not in scope

Support in `nym-client` or `nym-socks5-client`

### DIFF
--- a/common/client-core/src/client/inbound_messages.rs
+++ b/common/client-core/src/client/inbound_messages.rs
@@ -28,6 +28,7 @@ pub enum InputMessage {
         recipient: Recipient,
         data: Vec<u8>,
         lane: TransmissionLane,
+        mix_hops: Option<u8>,
     },
 
     /// Creates a message used for a duplex anonymous communication where the recipient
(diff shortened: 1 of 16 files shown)


common/client-core/src/client/inbound_messages.rs
```diff
@@ -28,6 +28,7 @@ pub enum InputMessage {
         recipient: Recipient,
         data: Vec<u8>,
         lane: TransmissionLane,
+        mix_hops: Option<u8>,
```
jstuczyn, line 31: just a side note: this enum is getting quite bloaty now, maybe we should rethink it in the future (not now!)

common/client-core/src/client/inbound_messages.rs
```diff
@@ -92,6 +93,27 @@ impl InputMessage {
...
+            message
+        }
+    }
+
+    pub fn new_regular_custom_hop(
+        recipient: Recipient,
+        data: Vec<u8>,
+        lane: TransmissionLane,
+        packet_type: Option<PacketType>,
+        mix_hops: Option<u8>,
```
jstuczyn: if it's "custom_hop" shouldn't this NOT be an option?

common/client-core/src/client/real_messages_control/acknowledgement_control/mod.rs
```diff
@@ -98,6 +101,8 @@ impl PendingAcknowledgement {
                 recipient_tag,
                 extra_surb_request,
             },
+            // Custom mix hops are not supported for anonymous messages
+            mix_hops: None,
```
jstuczyn, line 106: I wonder if they should (because in the long run I guess we want to be sending replies with surbs)

octol: In this particular location, the sender isn't in control of the number of mix hops though?

jstuczyn (approved):

PR #4325, opened 2024-01-15
Client packet counters

octol:
# Description

CT-4

Add counters for mixnet packets sent and received in the client.

### DIFF
--- a/common/client-core/src/client/cover_traffic_stream.rs
+++ b/common/client-core/src/client/cover_traffic_stream.rs
@@ -102,7 +102,7 @@ where
 // obviously when we finally make shared rng that is on 'higher' level, this should become
 // generic `R`
 impl LoopCoverTrafficStream<OsRng> {
-    pub fn new(
+    pub(crate) fn new(
         ack_key: Arc<AckKey>,
         average_ack_delay: Duration,
         mix_tx: BatchMixMessageSender,
--- a/common/client-core/src/client/cover_traffic_stream.rs
+++ b/common/client-core/src/client/cover_traffic_stream.rs
@@ -3,7 +3,7 @@
 
 use crate::client::mix_traffic::BatchMixMessageSender;
 use crate::client::topology_control::TopologyAccessor;
-use crate::client::COVER_PACKETS_SENT;
+// use crate::client::COVER_PACKETS_SENT;
 use crate::{config, spawn_future};
 use futures::task::{Context, Poll};
 use futures::{Future, Stream, StreamExt};
@@ -199,7 +199,6 @@ impl LoopCoverTrafficStream<OsRng> {
                 }
             }
         } else {
-            // COVER_PACKETS_SENT.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
             if self
                 .stats_tx
                 .send(PacketStatisticsEvent::CoverPacketSent)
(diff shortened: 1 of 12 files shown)


common/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -3,6 +3,7 @@
 
 use crate::client::mix_traffic::BatchMixMessageSender;
 use crate::client::topology_control::TopologyAccessor;
+// use crate::client::COVER_PACKETS_SENT;
```
jstuczyn: dead code

common/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -24,6 +25,8 @@ use tokio::time::{sleep, Sleep};
 #[cfg(target_arch = "wasm32")]
 use wasmtimer::tokio::{sleep, Sleep};
 
+use super::packet_statistics_control::{PacketStatisticsEvent, PacketStatisticsReporter};
```
jstuczyn: can you put this import before the target specific imports?

common/client-libs/gateway-client/src/client.rs
```diff
@@ -81,6 +81,9 @@ pub struct GatewayClient<C, St = EphemeralCredentialStorage> {
 
     /// Listen to shutdown messages.
     shutdown: TaskClient,
+
+    /// Number of websocket packets sent
+    sent_ws_packets: u64,
```
jstuczyn: if we're going to be having extra stats on the gateway client, how about wrap it in separate struct and also include sent/received packets + maybe also the data sizes?

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -259,6 +265,13 @@ where
 
         if let Err(err) = self.mix_tx.send(vec![next_message]).await {
             log::error!("Failed to send: {err}");
+        } else {
+            let event = if fragment_id.is_some() {
+                PacketStatisticsEvent::RealPacketSent
+            } else {
+                PacketStatisticsEvent::CoverPacketSent
```
jstuczyn: I wonder if it could be useful to separate cover messages generated from the cover traffic stream vs real traffic stream

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -259,6 +265,13 @@ where
 
         if let Err(err) = self.mix_tx.send(vec![next_message]).await {
             log::error!("Failed to send: {err}");
+        } else {
+            let event = if fragment_id.is_some() {
+                PacketStatisticsEvent::RealPacketSent
```
jstuczyn: also, for real packets, i think it might be cool to know what packet sized they used

jstuczyn (approved):

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: Well, I thought about it, but since we can't separate them when receiving I thought I'd keep it symmetric

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: Agree!

PR #4339, opened 2024-01-19
[DOC]: FAQ Updates

serinko:
# Description

Short mixnet-faq update


documentation/operators/src/faq/mixnodes-faq.md
```diff
@@ -1,22 +1,37 @@
...
+* [Nym Explorer](https://explorer.nymtech.net/)
+* [Guru Explorer](https://mixnet.explorers.guru/)
+* [ExploreNYM](https://explorenym.net/)
 
 ### What determines the rewards when running a Mix Node?
 
 The stake required for a Mix Node to achieve maximum rewards is called Mix Node saturation point. This is calculated from the staking supply (all circulating supply + part of unlocked tokens). The target level of staking is to have 50% of the staking supply locked in Mix Nodes.
 
-The node stake saturation point, which we denote by Nsat, is given by the stake supply, target level of staking divided by the number of rewarded (active) nodes. 
+The node stake saturation point, which we denote by Nsat, is given by the stake supply, target level of staking divided by the number of rewarded (active) nodes.
```
jstuczyn: active mixnode != rewarded mixnode.

we have following sets of mixnodes:
- bonded => self explanatory, essentially all of them
- rewarded => nodes that were selected to be in the rewarded set, i.e. will receive **some** rewards for this epoch (assuming non-zero performance), this can be further subdivided into:
  - active => top N nodes of the rewarded set (currently all of them, but doesn't have to be), nodes that will actually be used by the clients and iirc will also get better rewards
  - standby => bottom N nodes of the rewarded set - won't be used for mixing apart from testing
- inactive => nodes that haven't been selected to be in the rewarded set

documentation/operators/src/faq/mixnodes-faq.md
```diff
@@ -1,22 +1,37 @@
...
 
 ### Which VPS providers would you recommend?
 
 Consider in which jurisdiction you reside and where do you want to run a Mix Node. Do you want to pay by crypto or not and what are the other important particularities for your case? We always recommend operators to try to choose smaller and decentralised VPS providers over the most known ones controlling a majority of the internet. We receive some good feedback on these: Linode, Ghandi, Flokinet and Exoscale. Do your own research and share with the community.
 
-<!---### Why is a mix node setup on a self-hosted machine so tricky?--->
+### Why is a mix node setup on a self-hosted machine so tricky?
+
+We don't recommend this setup because it's really difficult to get a stable IPv6 from most of the ISPs.
```
jstuczyn: technically nodes don't need stable ipv6. what they need to be able to do is route ipv6 traffic

jstuczyn (approved):

serinko:
Addressed both points, done.

PR #4348, opened 2024-01-24
C++ FFI 

mfahampshire:
# Description

First version of C++ bindings for the Rust SDK with an example.


mmsinclair:
Looking good @mfahampshire

One minor change, please can you add Apache 2.0 license headers to all files like this - https://github.com/nymtech/nym/blob/develop/nym-api/src/main.rs#L1-L2

mmsinclair (approved):
lgtm, did some changes to the license headers

PR #4354, opened 2024-01-26
Feature/local gateway credential verification

jstuczyn:
# Description

Changes the behaviour of spending credentials so that they can be re-used between gateways, but still being restricted to a single client per gateway. In this model gateways are rewarded manually.


gateway/src/node/client_handling/websocket/connection_handler/coconut.rs
```diff
@@ -20,7 +20,10 @@ use std::collections::HashMap;
 use std::ops::Deref;
 use tokio::sync::{RwLock, RwLockReadGuard};
 
+pub(crate) const BANDWIDTH_PER_CREDENTIAL: u64 = 1024 * 1024 * 1024; // 1GB
```
jstuczyn, line 23: @mmsinclair you should probably verify this value

jstuczyn:
Superseded by https://github.com/nymtech/nym/pull/4397

PR #4366, opened 2024-01-31
Add config to run the block signing in monitoring mode only

mmsinclair:
When true, sending reward tokens is skipped and a zero hash will be returned and written to db.

### DIFF
--- a/nym-validator-rewarder/src/config/template.rs
+++ b/nym-validator-rewarder/src/config/template.rs
@@ -38,8 +38,11 @@ credential_verification = {{ rewarding.ratios.credential_verification }}
     
     
 [block_signing]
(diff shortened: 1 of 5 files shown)


nym-validator-rewarder/src/config/template.rs
```diff
@@ -38,9 +38,12 @@ credential_verification = {{ rewarding.ratios.credential_verification }}
     
     
 [block_signing]
-# Specifies whether credential issuance for block signing is enabled.
```
jstuczyn, line 41: I was probably tired when writing that 😅

jstuczyn (approved):

tommyv1987 (approved):

PR #4394, opened 2024-02-12
FFI share lib + initial uniffi-bindgen-go implementation

mfahampshire:
* split out blocking thread functions into shared library for use by Go and C/C++
* initial uniffi-bindgen-go bindings  

todo: 
- [x] add example for using Go import in non-local setting --> added info to readme


sdk/ffi/cpp/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "nym_cpp_ffi"
-version = "0.1.0"
+name = "nym-cpp-ffi"
+version = "0.1.1"
```
mmsinclair, line 3: Best not to mess with this too much, in case others use the package from their own cargo.toml files and a git import

sdk/ffi/go/ffi/bindings/bindings.go
mmsinclair, line 1: Wondering if we want this in source control, because it is generated? @jstuczyn what do you think?

sdk/ffi/go/README.md
```diff
@@ -0,0 +1,39 @@
+# Go FFI
+> ⚠️ This is an initial version of this library in order to give developers something to experiment with. If you use this code to begin testing out Mixnet integration and run into issues, errors, or have feedback, please feel free to open an issue; feedback from developers trying to use it will help us improve it. If you have questions feel free to reach out via our [Matrix channel](https://matrix.to/#/#dev:nymtech.chat).
+
+This repo contains: 
+* `lib.rs`: an initial version of bindings for interacting with the Mixnet via the Rust SDK from Go. These are essentially match statemtns wrapping imported functions from the `nym-ffi-shared` lib. 
+* `ffi/`: a directory containing: 
+  * the `bindings/` files generated using [`uniffi-bindgen-go`]()
+  * `main.go`: an example of using this library.
```
mmsinclair:
```suggestion
  * [`main.go`](./ffi/bindings/main.go): an example of using this library. 
```

It was hard to find the example file. A link would make it better?

mmsinclair: And just wondering if `sdk/ffi/go/ffi` should maybe change to `sdk/ffi/go/example`?

mmsinclair (changes requested):
Nice work Max, only a few thoughts

sdk/ffi/go/README.md
mfahampshire: Yeah its a bit of an annoying path - I wouldn't change to "example" though, as its the bindings files themselves that are generated.. so maybe `sdk/ffi/go/generated_bindings/`?

sdk/ffi/cpp/Cargo.toml
mfahampshire: My line of thought was that I'd do a version bump since it was being updated - although happy to not as well!

sdk/ffi/go/README.md
mfahampshire: Did a bit of a rearrange - i think its a bit clearer now

mmsinclair (approved):

PR #4398, opened 2024-02-13
Propagate gateway ws fd into sdk

neacsu:
# Description

In order to let the websocket gateway connection bypass the routing rules of non-wireguard vpn mode, the file descriptor of that connection needs to be exposed.


common/client-libs/gateway-client/src/socket_state.rs
```diff
@@ -37,9 +40,22 @@ type WsConn = JSWebsocket;
 
 type SplitStreamReceiver = oneshot::Receiver<Result<SplitStream<WsConn>, GatewayClientError>>;
 
+pub(crate) fn ws_fd(_conn: &WsConn) -> Option<RawFd> {
```
octol, line 43: _conn -> conn

common/client-core/src/client/base_client/mod.rs
```diff
@@ -103,6 +104,7 @@ pub struct ClientState {
     pub shared_lane_queue_lengths: LaneQueueLengths,
     pub reply_controller_sender: ReplyControllerSender,
     pub topology_accessor: TopologyAccessor,
+    pub gateway_ws_fd: Option<RawFd>,
```
octol, line 112: In the `ClientState`, maybe it's useful to wrap this in a slightly higher level `GatewayConnection` struct that could possibly also contain other useful info about the gateway connection state. Idk gateway address, connection state?

octol (approved):

common/client-libs/gateway-client/src/socket_state.rs
neacsu: I've left it like this so that it doesn't produce an error on wasm builds for unused argument

jstuczyn: maybe we should just target lock everything to do with file descriptors? It's not like we could even use them in wasm

neacsu: Yes, in wasm they would not appear. But I think it's a bit better to use `Option` instead, and just leave the wasm macros to as little code as possible, instead of applying it all the way up

PR #4408, opened 2024-02-19
Promethus is our friend

durch:
# Description

Integrate prometheus into packet statistics controller, remove manual rate calculations, we'll have prom handle that for us, whats left is exposing the endpoints in various clients. We can merge this now


common/client-core/src/client/packet_statistics_control.rs
```diff
@@ -479,11 +284,8 @@ impl PacketStatisticsControl {
                 },
                 _ = snapshot_interval.tick() => {
                     self.update_history();
-                    self.update_rates();
                 }
                 _ = report_interval.tick() => {
-                    self.report_rates();
```
octol: In the vpn cli client it's pretty useful to have the traffic data rates printed periodically, is there some way to get the equivalent functionality with this new setup?

durch: Yes, we can even put the packet rate calculations back, I wasn't sure how this was gonna end up eventually, will put something back

benedettadavico (approved):

PR #4409, opened 2024-02-19
sandbox nym-mixnode on OpenBSD

timkuijsten:
These patches [pledge(2)](https://man.openbsd.org/pledge.2) and [unveil(2)](https://man.openbsd.org/unveil.2) the nym-mixnode daemon so that it can be run with more confidence that it won't disrupt or interfere with other processes running on the same system. Roughly speaking, pledge is like Linux seccomp and unveil can be compared with [Landlock](https://docs.kernel.org/security/landlock.html).

I know this is a bit of a wild shot, but would the project be interested in incorporating patches like these?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5745,6 +5745,7 @@ dependencies = [
  "nym-types",
  "nym-validator-client",
  "opentelemetry",
+ "pledge",
  "rand 0.7.3",
  "serde",
  "serde_json",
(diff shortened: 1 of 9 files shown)


octol:
Sorry for the delay, I think this could be interesting. And maybe we can place it behind `#[cfg(target_os = openbsd)]`?

tommyv1987:
@octol do you think we should apply this to nym-nodes? as mixnode are now deprecated?

octol:
> @octol do you think we should apply this to nym-nodes? as mixnode are now deprecated?

I think it's an interesting thing yep, but I don't have a OpenBSD environment to experiment on currently

PR #4426, opened 2024-02-22
Fix windows build

neacsu:


common/client-libs/gateway-client/src/client.rs
```diff
@@ -33,14 +33,14 @@ use std::sync::Arc;
...
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(unix)]
 use std::os::fd::RawFd;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio::time::sleep;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio_tungstenite::connect_async;
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(not(unix))]
```
jstuczyn, line 43: so I guess we don't need the file descriptor in windows?

jstuczyn (approved):
.

common/client-libs/gateway-client/src/client.rs
neacsu: Yep, I think they don't even have that notion, and this fd is needed only for android for now

PR #4430, opened 2024-02-27
proof of concept: env-configurable mixnode

jstuczyn:
# Description

A simple proof of concept of to allow us to initialise (and run) our binaries by fully configuring them using environmental variables.

For now I've just implemented it for the mixnode, but it can be extended to other binaries like the gateway, nym-api, etc.


Bykow (approved):
Looks great ! Thanks Andrew !

Env names are long and descriptive, thanks for that too !

octol:
Closing since it's my understanding this functionality was merged as part of another PR

PR #4431, opened 2024-02-27
Add IPv6 support to IPPR 

neacsu:


common/ip-packet-requests/src/lib.rs
```diff
@@ -1,8 +1,17 @@
+use serde::{Deserialize, Serialize};
+use std::net::{Ipv4Addr, Ipv6Addr};
+
 pub mod codec;
 pub mod request;
 pub mod response;
 
-pub const CURRENT_VERSION: u8 = 3;
+pub const CURRENT_VERSION: u8 = 4;
```
jstuczyn, line 10: Imo it'd be nice to have a short comment here showing what changed between versions (just for our own future reference), like
version3: initial version
version4: ipv6 support

oh and btw. I'm just curious: since we bumped up the version by adding the extra feature, would we still be able to handle v3 requests? if not, imo it's still fine given we're not at the point of needing backwards compatibility

common/ip-packet-requests/src/lib.rs
```diff
@@ -1,8 +1,17 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct IPPair {
+    pub ipv4: Ipv4Addr,
+    pub ipv6: Ipv6Addr,
```
jstuczyn, line 15: maybe I'm missing something, but wouldn't this change require you to provide **both** ipv4 and ipv6 addresses for each request? shouldn't ipv6 be an option instead given it might not always be available?

service-providers/ip-packet-router/src/constants.rs
```diff
@@ -2,8 +2,10 @@ use std::time::Duration;
 
 // The interface used to route traffic
 pub const TUN_BASE_NAME: &str = "nymtun";
-pub const TUN_DEVICE_ADDRESS: &str = "10.0.0.1";
-pub const TUN_DEVICE_NETMASK: &str = "255.255.255.0";
+pub const TUN_DEVICE_ADDRESS_V4: &str = "10.0.0.1";
```
jstuczyn: would it perhaps make more sense to define those using concrete types? 
i.e. for example
```rust
pub const TUN_DEVICE_ADDRESS_V4: IpAddr = IpAddr::V4(Ipv4Addr::new(10,0,0,1));
```

this way you won't have to parse and unwrap anything

service-providers/ip-packet-router/src/util/generate_new_ip.rs
```diff
@@ -1,38 +1,54 @@
...
+use std::net::Ipv6Addr;
+use std::{collections::HashMap, net::Ipv4Addr};
 
-use crate::{constants::TUN_DEVICE_ADDRESS, mixnet_listener::ConnectedClient};
+use crate::constants::{TUN_DEVICE_ADDRESS_V4, TUN_DEVICE_ADDRESS_V6};
 
 // Find an available IP address in self.connected_clients
 // TODO: make this nicer
-fn generate_random_ip_within_subnet() -> Ipv4Addr {
+fn generate_random_ips_within_subnet() -> IPPair {
```
jstuczyn, line 9: is there any particular reason for having to use random ips rather than just using incremental values by keeping a counter?
i.e. first client would get `10.0.0.1`, the second client would get `10.0.0.2`, etc.

common/ip-packet-requests/src/lib.rs
jstuczyn: oh. I guess the tun device will give you one.
but I wonder if we might have a problem if the gateway's network stack does not support ipv6 (which might be the case for people running them on some crappy vps)

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -48,46 +52,63 @@ impl ConnectedClients {
...
+            IpAddr::V4(ip) => self.clients_ipv4_mapping.get_mut(ip),
+            IpAddr::V6(ip) => self.clients_ipv6_mapping.get_mut(ip),
+        }
     }
 
-    fn is_nym_address_connected(&self, nym_address: &Recipient) -> bool {
-        self.clients
-            .values()
-            .any(|client| client.nym_address == *nym_address)
+    async fn is_nym_address_connected(&self, nym_address: &Recipient) -> bool {
```
jstuczyn: if we rely on this method in multiple places, it might become a bottleneck down the line since we're having to go through **all** connected clients and then wait trying to read **each** value. could we maybe have a reverse map somewhere? (not neccessarily now in this PR, but at some point in the future)

jstuczyn (approved):

common/ip-packet-requests/src/lib.rs
neacsu: I've added the comment, it sounds useful for posterity.
No, there won't be any backwards compatibility, I haven't approached that since it wasn't really needed now. We'll have to make sure to redeploy all gateways.

common/ip-packet-requests/src/lib.rs
neacsu: That's a fair point, but I would say more for clients than for gateways, since we kind of expect IPv6 support from gateways, right? And for clients, from the types that currently work for the vpn (linux, macos, android), I think there might only be a small portion of legacy linux devices that don't have IPv6 support in their network stack. But even so, maybe making it an `Option` now rather then having to change the version later to support them is a safer approach.

service-providers/ip-packet-router/src/constants.rs
neacsu: Nice catch, I'll do that, thanks.

service-providers/ip-packet-router/src/util/generate_new_ip.rs
neacsu: I'm not sure why this is the approach @octol took, but we should talk and possibly change the approach here, as it can end up with inconsistent behaviour for users when close to reaching the limit, especially if we increase the size of the networks.

service-providers/ip-packet-router/src/mixnet_listener.rs
neacsu: Yep, you're right, this could be a bottleneck, and I think an unnecessary one too. I simplified the `Arc<RwLock>`, since they weren't needed everywhere anyway.

common/ip-packet-requests/src/lib.rs
neacsu: On second thought, I think this is not the place to modify this, because it only affects gateways, like you said, and I think those should have IPv6 support as a requirement. I will take a look at client side, in the other repo, as to not fail in case of lack of IPv6 support.

service-providers/ip-packet-router/src/util/generate_new_ip.rs
octol: I've seen other networking libs trying hard to avoid sequential client ids to avoid information leakage

PR #4435, opened 2024-02-28
Fix wasm build

neacsu:


neacsu:
The upgrade of `rustls` to latest `0.21.*` breaks wasm compatibility (https://github.com/rustls/rustls/pull/1532/files). `rustls 0.22.*`seems to re-enable it, by using another default crate instead of `ring`, but building that other crate (`aws-lc-rs`) for wasm looks to be non-trivial.

PR #4437, opened 2024-03-01
expire freepass bandwidth + stagger db flushes

jstuczyn:
# Description

This pull request slightly changes the behaviour of free passes:
- when the encoded expiration runs out, whatever remaining bandwidth client might have still held at the gateway is reset to 0
- when a freepass is active, no bandwidth vouchers or other free passes can be claimed

Furthermore, since we're going to be using a fair use policy for bandwidth, I made bandwdith flushing less aggressive, i.e. rather than updating the underlying database on **every** sphinx packet, I only do it on every claimed credential and then every 512kB data used or every 5ms, whichever happens first.


gateway/src/node/client_handling/websocket/shared_state.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+
+use crate::node::client_handling::websocket::connection_handler::coconut::CoconutVerifier;
+use crate::node::client_handling::websocket::connection_handler::BandwidthFlushingBehaviourConfig;
+use nym_crypto::asymmetric::identity;
+use std::sync::Arc;
+
+// I can see this being possible expanded with say storage or client store
+#[derive(Clone)]
+pub(crate) struct SharedHandlerState {
```
octol: With something called `Shared` I'd imagine the user expectation is that it's all behind an Arc? That is, if the boolean there is updated it's reflected in other shared instances as well?

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -274,16 +319,20 @@ where
             ));
         }
 
+        drop(aggregated_verification_key);
```
octol: imho having an explicit scope is preferable

octol (approved):
Nice :) 
left some suggestions

PR #4440, opened 2024-03-04
Add IPPair constructor

neacsu:


common/ip-packet-requests/src/lib.rs
```diff
@@ -15,6 +16,18 @@ pub struct IPPair {
     pub ipv6: Ipv6Addr,
 }
 
+impl IPPair {
```
octol: imho to match other convention of other related types, probably rename capitalization to `IpPair`?

octol (approved):

PR #4442, opened 2024-03-04
Bugfix/gateway registration

jstuczyn:
# Description

This PR allows for wider backwards compatibility between gateways and clients. If a client does not expect to be using any credentials, it will downgrade it's advertised protocol version to support communication with wider (older) range of gateways

### DIFF
--- a/gateway/gateway-requests/src/registration/handshake/state.rs
+++ b/gateway/gateway-requests/src/registration/handshake/state.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 - Nym Technologies SA <contact@nymtech.net>
+// Copyright 2020-2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::registration::handshake::error::HandshakeError;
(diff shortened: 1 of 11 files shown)


gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -200,35 +199,56 @@ impl<'a, S> State<'a, S> {
...
+                                types::RegistrationHandshake::HandshakePayload { data, .. } => {
+                                    Ok(data)
+                                }
+                                types::RegistrationHandshake::HandshakeError { message } => {
+                                    Err(HandshakeError::RemoteError(message))
+                                }
+                            };
+                        }
+                        Err(_) => {
+                            error!("Received a non-handshake message during the registration handshake! It's getting dropped. The received content was: '{msg}'");
```
jstuczyn, line 232: before this PR we were failing here without much explanation given. I found logging of the underlying message to be very helpful (since it includes the error message). ideally we'd just deserialize into a `Result` instead, but that'd require some refactoring

gateway/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -200,35 +199,56 @@ impl<'a, S> State<'a, S> {
         self.remote_pubkey = Some(remote_pubkey)
     }
 
-    pub(crate) async fn receive_handshake_message(&mut self) -> Result<Vec<u8>, HandshakeError>
+    pub(crate) async fn _receive_handshake_message(&mut self) -> Result<Vec<u8>, HandshakeError>
```
octol: Leading underscore?

octol (approved):

gateway/gateway-requests/src/registration/handshake/state.rs
jstuczyn: ah yes, it shouldn't be public

PR #4456, opened 2024-03-12
Feature Compact Ecash : Mega PR

simonwicky:
# Description

This is the base PR for the integration of Offline Compact Ecash, meant to replace the current coconut scheme. 
This integration is split into many stacked PRs to facilitate review. Note that they cannot be made independent and might thus not compile apart from the top one.

## PR breakdown : 
- DKG type adaptation : #4461
- TicketBook Issuance : #4457 
- Freepass Issuance : #4458 
- Credential spending : #4459 
- Final bits and bobs : #4580 

#4580 being the top of the stack, it should compile and represent the final state of the integration

# Checklist:

- [ ] added a changelog entry to `CHANGELOG.md`

PR #4457, opened 2024-03-12
Feature Compact Ecash : Issuance

simonwicky:
# Description

This PR is a part of the Ecash stack. The base of that stack is #4456 . Refer to the base's description for details.

This PR contains the necessary bits for TicketBook issuance, CLI and API side. It will probably not build on its own

# Checklist:


nym-api/migrations/20240208120000_issued_credential_ecash.sql
```diff
@@ -0,0 +1,19 @@
...
+-- particular ecash credential issued in this epoch
+CREATE TABLE issued_credential
+(
+    id                         INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    epoch_id                   INTEGER NOT NULL,
+    tx_hash                    VARCHAR NOT NULL UNIQUE,
+    bs58_partial_credential    VARCHAR NOT NULL,
+    bs58_signature             VARCHAR NOT NULL,
+    joined_private_commitments VARCHAR NOT NULL,
+    expiration_date            INTEGER NOT NULL
```
jstuczyn, line 18: I'm making this comment before looking at the other files (I have to start my review somewhere : P) so it might not make full sense, but if we're storing a date in the db, perhaps we should use proper timestamp type for it, i.e.
`expiration_date TIMESTAMP WITHOUT TIME ZONE NOT NULL`. `sqlx` works nicely with `time::OffsetDateTime` for that

nym-api/src/coconut/state.rs
```diff
@@ -104,7 +115,7 @@ impl State {
             request_body.tx_hash,
             blinded_signature,
             &encoded_commitments,
-            &request_body.public_attributes_plain,
+            request_body.expiration_date.try_into().unwrap(), //will fail in approx 290 billion years
```
jstuczyn, line 118: so why not just change the type on `BlindSignRequestBody` itself : ) either to `i64` or proper `OffsetDateTime` object

nym-api/src/coconut/state.rs
```diff
@@ -120,7 +131,7 @@ impl State {
                 blinded_signature,
                 signature,
                 encoded_commitments,
-                request_body.public_attributes_plain,
+                request_body.expiration_date.try_into().unwrap(), //will fail in approx 290 billion years
```
jstuczyn, line 134: ibid

nym-api/nym-api-requests/src/coconut/models.rs
```diff
@@ -54,49 +54,41 @@ impl VerifyCredentialResponse {
...
     /// Hash of the deposit transaction
     pub tx_hash: Hash,
 
     /// Signature on the inner sign request and the tx hash
     pub signature: identity::Signature,
 
-    pub public_attributes_plain: Vec<String>,
+    pub ecash_pubkey: PublicKeyUser,
+
+    pub expiration_date: u64,
```
jstuczyn, line 67: I'm always a fan of using strict types whenever possible. Would it make sense to use precise date with `OffsetDateTime` instead?

nym-api/nym-api-requests/src/coconut/models.rs
```diff
@@ -207,6 +199,32 @@ impl CosmosAddressResponse {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct PartialExpirationDateSignatureResponse {
+    pub signs: Vec<PartialExpirationDateSignature>,
```
jstuczyn: nitpicking: imo rename it to `signatures` rather than `signs`

common/credentials/src/coconut/utils.rs
```diff
@@ -3,15 +3,35 @@
...
-    aggregate_verification_keys, Signature, SignatureShare, VerificationKey,
+    aggregate_expiration_signatures, aggregate_indices_signatures, aggregate_verification_keys,
+    constants, setup, Base58, CoinIndexSignature, ExpirationDateSignature,
+    PartialCoinIndexSignature, PartialExpirationDateSignature, VerificationKeyAuth, Wallet,
 };
 use nym_validator_client::client::CoconutApiClient;
 
+pub fn today_timestamp() -> u64 {
+    let now_utc = Utc::now();
+    (now_utc.timestamp() - now_utc.num_seconds_from_midnight() as i64) as u64
```
jstuczyn: this logic will fail on a leap year : )

jstuczyn: maybe get current datetime and explicitly set h,m,s,ms,ns etc to 0?

oh. and use `time` rather than `chrono`.

nym-api/src/coconut/api_routes/mod.rs
```diff
@@ -215,12 +205,32 @@ pub async fn post_blind_sign(
...
     // get the transaction details of the claimed deposit
     debug!("getting transaction details from the chain");
     let tx = state
         .get_transaction(blind_sign_request_body.tx_hash)
         .await?;
 
+    //check expiration date validity
+    if blind_sign_request_body.expiration_date > cred_exp_date_timestamp() {
+        return Err(
+            CompactEcashError::ExpirationDate("Invalid expiration date".to_string()).into(),
```
jstuczyn, line 230: imo this shouldn't be returning errors from the crypto library. instead we should wrap them into something specific to nym-api requests. also, "invalid expiration date" is such an obscure error. perhaps create variant that says the specified expiration was too far into the future and add 2 fields with max allowed expiration and passed value

nym-api/src/coconut/state.rs
```diff
@@ -153,4 +164,82 @@ impl State {
...
+                        expiration_ts,
+                        &signing_key.keys.secret_key(),
+                    )
+                    .await)
+            }
+        }
+    }
+
+    //this one gives the signatures for a particular day. No cache because it's only gonna be used for recovery attempt and freepasses
+    pub async fn get_exp_date_signatures_timestamp(
```
jstuczyn, line 228: oh, but it should have a cache. otherwise somebody could call `/expiration-date-signatures` repeatedly ddosing nym-api because it has to spend a lot of time signing things

nym-api/src/coconut/state.rs
```diff
@@ -153,4 +164,82 @@ impl State {
...
+        &self,
+        timestamp: u64,
+    ) -> Result<Vec<ExpirationDateSignature>> {
```
jstuczyn, line 231: this might be a stupid question, but is nym-api supposed to give a signature to anyone without any verification? if so, what's even the point of those signatures

nym-api/src/coconut/state.rs
```diff
@@ -24,18 +34,18 @@ pub use nym_credentials::coconut::bandwidth::bandwidth_credential_params;
 pub struct State {
     pub(crate) client: Arc<dyn LocalClient + Send + Sync>,
     pub(crate) bandwidth_contract_admin: OnceCell<Option<AccountId>>,
-    pub(crate) mix_denom: String,
     pub(crate) coconut_keypair: KeyPair,
```
jstuczyn, line 37: well, it's no longer a coconut_keypair, is it now? : )

nym-api/src/coconut/state.rs
```diff
@@ -153,4 +164,82 @@ impl State {
             .aggregated_verification_key(epoch_id)
             .await
     }
+    pub async fn get_coin_indices_signatures(&self) -> Result<Vec<CoinIndexSignature>> {
+        let current_epoch = self.client.get_current_epoch().await?;
```
jstuczyn: get current_epoch out of `comm_channel` because that one is supposed to cache responses for queries from the chain

nym-api/src/coconut/state.rs
```diff
@@ -153,4 +164,82 @@ impl State {
...
+        match self
+            .coin_indices_sigs_cache
+            .get_signatures(current_epoch.epoch_id)
+            .await
+        {
+            Some(signatures) => Ok(signatures),
+            None => {
+                let ecash_params = setup(constants::NB_TICKETS);
```
jstuczyn, line 176: let's cache the params since they're never changing and curve operations are expensive (use `OnceLock` we were using for coconut params)

nym-api/src/coconut/state.rs
```diff
@@ -153,4 +164,82 @@ impl State {
...
+                let verification_key = self.verification_key(current_epoch.epoch_id).await?;
+                let maybe_keypair_guard = self.coconut_keypair.get().await;
+                let Some(keypair_guard) = maybe_keypair_guard.as_ref() else {
```
jstuczyn, line 179: check for coconut keys before performing any possible chain queries

nym-api/src/coconut/helpers.rs
```diff
@@ -50,15 +62,120 @@ impl CredentialRequest for FreePassRequest {
...
+            *signatures = Some(sign_expiration_date(secret_key, expected_exp_date));
+            self.epoch_id.store(expected_epoch_id, Ordering::Release);
+            self.expiration_date
+                .store(expected_exp_date, Ordering::Release);
+        }
+
+        signatures.clone().unwrap() // Either we or someone else update the signatures, so they must be there
+    }
+}
+}
```
jstuczyn, line 181: as clippy said 5 times, you have an extra `}` : P

nym-api/src/coconut/helpers.rs
```diff
@@ -50,15 +62,120 @@ impl CredentialRequest for FreePassRequest {
...
+        expected_epoch_id: u64,
+        expected_exp_date: u64,
+        secret_key: &SecretKeyAuth,
+    ) -> Vec<ExpirationDateSignature> {
+        let mut signatures = self.signatures.write().await;
+
+        //if this fails, it means someone else updated the signatures in the meantime
+        // => We don't have to update them, and we know they exist
+        // (this check can spare us some signing)
+        if self.epoch_id.load(Ordering::Acquire) != expected_epoch_id
```
jstuczyn: you can use `Ordering::Relaxed` or even possibly get rid of atomics completely given you're  holding an exclusive lock to `signatures` and you can't make any writes to those integers without holding the lock

nym-api/src/coconut/helpers.rs
```diff
@@ -50,15 +62,120 @@ impl CredentialRequest for FreePassRequest {
...
+            || self.expiration_date.load(Ordering::Acquire) != expected_exp_date
+        {
+            *signatures = Some(sign_expiration_date(secret_key, expected_exp_date));
```
jstuczyn, line 172: is it possible you might also need signatures for the previous epoch during an epoch transition?

nym-api/src/coconut/helpers.rs
```diff
@@ -50,15 +62,120 @@ impl CredentialRequest for FreePassRequest {
...
+        if self.epoch_id.load(Ordering::Acquire) != expected_epoch_id {
+            *signatures = Some(sign_coin_indices(
+                ecash_parameters,
+                verification_key,
+                secret_key,
+            ));
+            self.epoch_id.store(expected_epoch_id, Ordering::Release);
+        }
+
+        signatures.clone().unwrap() // Either we or someone else update the signatures, so they must be there
```
jstuczyn, line 123: assuming it was for the same epoch... I'm 99% sure that unwrap is fine, but I've got some weird feeling about it...

nym-api/src/coconut/deposit.rs
```diff
@@ -3,25 +3,20 @@
...
 pub async fn validate_deposit_tx(request: &BlindSignRequestBody, tx: TxResponse) -> Result<()> {
-    if request.public_attributes_plain.len()
-        != IssuanceBandwidthCredential::PUBLIC_ATTRIBUTES as usize
-    {
-        return Err(CoconutError::InconsistentPublicAttributes);
-    }
-
     // extract actual public attributes + associated x25519 public key
-    let deposit_value = find_tx_attribute(&tx, COSMWASM_DEPOSITED_FUNDS_EVENT_TYPE, DEPOSIT_VALUE)
+    // we're not using it anymore, but for legacy reasons it must be present
```
jstuczyn: well, we no longer have any legacy

nym-api/src/coconut/deposit.rs
```diff
@@ -3,25 +3,20 @@
...
     COSMWASM_DEPOSITED_FUNDS_EVENT_TYPE, DEPOSIT_ENCRYPTION_KEY, DEPOSIT_IDENTITY_KEY,
     DEPOSIT_INFO, DEPOSIT_VALUE,
 };
-use nym_credentials::coconut::bandwidth::voucher::BandwidthVoucherIssuanceData;
-use nym_credentials::coconut::bandwidth::IssuanceBandwidthCredential;
-use nym_crypto::asymmetric::identity;
 use nym_validator_client::nyxd::helpers::find_tx_attribute;
 use nym_validator_client::nyxd::TxResponse;
 
 pub async fn validate_deposit_tx(request: &BlindSignRequestBody, tx: TxResponse) -> Result<()> {
```
jstuczyn, line 15: so looking at it, technically the only think we really need to check is the identity key, isn't it? we don't need deposit amount, x25519 key or I'd argue the info string

common/credentials/src/coconut/bandwidth/voucher.rs
```diff
@@ -18,24 +16,24 @@ pub struct BandwidthVoucherIssuedData {
     /// the plain value (e.g., bandwidth) encoded in this voucher
     // note: for legacy reasons we're only using the value of the coin and ignoring the denom
     #[zeroize(skip)]
-    value: Coin,
+    value: u128,
```
jstuczyn, line 19: do we still need bandwidth vouchers?
Also, why did we change `Coin` to a `u128`?

common/credentials/src/coconut/bandwidth/voucher.rs
```diff
@@ -63,32 +57,29 @@ pub struct BandwidthVoucherIssuanceData {
 
 impl BandwidthVoucherIssuanceData {
     pub fn new(
-        value: impl Into<Coin>,
         deposit_tx_hash: Hash,
         signing_key: identity::PrivateKey,
         unused_ed25519: encryption::PrivateKey,
```
jstuczyn: since we have no legacy baggage, remove the unused things : P

common/credentials/src/coconut/bandwidth/voucher.rs
```diff
@@ -63,32 +57,29 @@ pub struct BandwidthVoucherIssuanceData {
...
     ) -> Self {
-        let value = value.into();
-        let value_prehashed = hash_to_scalar(value.amount.to_string());
+        let value = TICKET_BOOK_VALUE;
```
jstuczyn, line 57: I wonder, should the ticket book be a proper `Coin` given it corresponds to on-chain token deposits?

common/credentials/src/coconut/utils.rs
```diff
@@ -28,70 +48,157 @@ pub fn obtain_aggregate_verification_key(
...
+pub async fn obtain_coin_indices_signatures(
+    ecash_api_clients: &[CoconutApiClient],
+    verification_key: &VerificationKeyAuth,
+    threshold: u64,
+) -> Result<Vec<CoinIndexSignature>, Error> {
+    if ecash_api_clients.is_empty() {
+        return Err(Error::NoValidatorsAvailable);
+    }
+
+    let mut signatures: Vec<(u64, VerificationKeyAuth, Vec<PartialCoinIndexSignature>)> =
```
jstuczyn, line 112: I think I mentioned this in the first PR, but could we wrap those in a separate struct?

common/credentials/src/coconut/utils.rs
```diff
@@ -28,70 +48,157 @@ pub fn obtain_aggregate_verification_key(
...
+        Vec::with_capacity(ecash_api_clients.len());
+
+    let ecash_params = setup(constants::NB_TICKETS);
```
jstuczyn, line 115: make it a static global

common/credentials/src/coconut/utils.rs
```diff
@@ -28,70 +48,157 @@ pub fn obtain_aggregate_verification_key(
...
+    if signatures.len() < threshold as usize {
+        return Err(Error::NotEnoughShares);
+    }
+
+    //this takes care of validating partial signatures
+    aggregate_indices_signatures(&ecash_params, verification_key, &signatures)
+        .map_err(Error::CompactEcashError)
+}
+
 pub async fn obtain_aggregate_signature(
```
jstuczyn: just a suggestion, maybe rename it to `obtain_aggregate_wallet` given the return type is a `Wallet`?

common/credentials/src/coconut/utils.rs
```diff
@@ -28,70 +48,157 @@ pub fn obtain_aggregate_verification_key(
...
-    if shares.len() < threshold as usize {
+    if wallets.len() < threshold as usize {
         return Err(Error::NotEnoughShares);
     }
 
-    voucher.aggregate_signature_shares(&verification_key, &shares)
+    voucher.aggregate_signature_shares(&verification_key, &wallets, request)
 }
 
-pub(crate) mod scalar_serde_helper {
```
jstuczyn, line 77: do we no longer need to serialise scalars? if so, great (I'm making this comment after having reviewed only about half the files xd)

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -7,70 +7,44 @@ use crate::coconut::bandwidth::voucher::BandwidthVoucherIssuanceData;
...
-            BandwidthCredentialIssuanceDataVariant::Voucher(voucher) => voucher.value_plain(),
-            BandwidthCredentialIssuanceDataVariant::FreePass(freepass) => {
-                freepass.expiry_date_plain()
-            }
+            BandwidthCredentialIssuanceDataVariant::TicketBook(..) => CredentialType::TicketBook,
+            BandwidthCredentialIssuanceDataVariant::FreePass => CredentialType::FreePass,
         }
     }
 
     pub fn voucher_data(&self) -> Option<&BandwidthVoucherIssuanceData> {
```
jstuczyn, line 45: this comment is not limited to this function, but perhaps everything to do with bandwidth voucher, etc. should be renamed to more closely fit with ticket books? so say this should probably be `ticket_book_data` or something

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
 // all types of bandwidth credentials contain serial number and binding number
 #[derive(Zeroize, ZeroizeOnDrop, Serialize, Deserialize)]
 pub struct IssuanceBandwidthCredential {
```
jstuczyn, line 55: wait, so where does the wallet's secret `v` live? don't you need to keep it?

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
     variant_data: BandwidthCredentialIssuanceDataVariant,
 
-    /// type of the bandwdith credential hashed onto a scalar
-    #[serde(with = "scalar_serde_helper")]
-    type_prehashed: PublicAttribute,
+    ///ecash keypair related to the credential
+    ecash_keypair: KeyPairUser,
+
+    ///expiration_date of that credential
+    expiration_date: u64,
```
jstuczyn, line 63: can we keep it as concrete type?

common/credential-storage/src/storage.rs
```diff
@@ -30,9 +31,37 @@ pub trait Storage: Send + Sync {
     async fn consume_coconut_credential(
         &self,
         id: i64,
-        gateway_id: &str,
+    /// Inserts provided coin_indices_signatures into the database.
```
jstuczyn, line 34: you haven't closed the function definition (i.e. missing `) -> Result<(), Self::StorageError>;` bit)

also, why are we removing the gateway_id? iirc, we needed this for the purposes of tracking reuses of freepasses

common/credential-storage/src/storage.rs
```diff
@@ -30,9 +31,37 @@ pub trait Storage: Send + Sync {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `epoch_id`: Id of the epoch.
+    /// * `coin_indices_signatures` : The coin indices signatures for the epoch
+    async fn insert_coin_indices_sig(
+        &self,
+        epoch_id: String,
```
jstuczyn, line 42: why is epoch_id a string?

common/credential-storage/src/storage.rs
```diff
@@ -30,9 +31,37 @@ pub trait Storage: Send + Sync {
...
     ) -> Result<(), Self::StorageError>;
 
+    /// Check if coin indices signatures are present for a given epoch
+    ///
+    /// # Arguments
+    ///
+    /// * `epoch_id`: Id of the epoch.
+    async fn is_coin_indices_sig_present(
+        &self,
+        epoch_id: String,
```
jstuczyn, line 53: ibid.

common/credential-storage/src/storage.rs
```diff
@@ -30,9 +31,37 @@ pub trait Storage: Send + Sync {
...
+    ) -> Result<bool, Self::StorageError>;
+
+    /// Get coin_indices_signatures of a given epoch.
+    ///
+    /// # Arguments
+    ///
+    /// * `epoch_id`: Id of the epoch.
+    async fn get_coin_indices_sig(
+        &self,
+        epoch_id: String,
```
jstuczyn, line 63: ibid.

common/credential-storage/src/storage.rs
```diff
@@ -30,9 +31,37 @@ pub trait Storage: Send + Sync {
...
+    ) -> Result<CoinIndicesSignature, Self::StorageError>;
```
jstuczyn, line 64: I guess this should return `Result<Option<...>>` in case we don't have it in the db?

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE TABLE ecash_credentials
+(
+    id                     INTEGER                                                                  NOT NULL PRIMARY KEY AUTOINCREMENT,
+
+--     introduce a way for us to introduce breaking changes in serialization
+serialization_revision INTEGER                                                                      NOT NULL,
+
+--     the best we can do without enums
+    credential_type        TEXT CHECK ( credential_type IN ('TicketBook', 'FreeBandwidthPass') )    NOT NULL,
+    credential_data        BLOB                                                                     NOT NULL UNIQUE,
```
jstuczyn, line 21: since we have an opportunity for breaking changes now. `UNIQUE` was my super quick and hacky way of preventing people from importing the same underlying credential, but it caused a lot of headache on the vpn front...

could we instead remove that `UNIQUE` constraint on `credential_data` and introduce another column with something actually unique for that credential (and slap the constraint on that)? in the case of coconut it'd have been the serial number, but here I guess it might be the public key?

jstuczyn: pinging @octol since you've been on the receiving end of the `UNIQUE` issue : )

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
jstuczyn, line 1: can you run linter on that file justg to align all the fields, etc.

common/commands/src/utils.rs
```diff
@@ -123,6 +123,21 @@ impl CommonConfigsWrapper {
...
+    pub(crate) fn try_get_private_id_key(&self) -> anyhow::Result<PathBuf> {
+        match self {
+            CommonConfigsWrapper::NymClients(cfg) => Ok(cfg
+                .storage_paths
+                .inner
+                .keys
+                .private_identity_key_file
+                .clone()),
+            CommonConfigsWrapper::NymApi(_cfg) => {
+                todo!() //SW this will depend on the new network monitor structure. Ping @Drazen
```
jstuczyn, line 135: what's this one all about?

common/commands/src/utils.rs
```diff
@@ -225,4 +240,17 @@ impl UnknownConfigWrapper {
             bail!("no 'credentials_database_path' field present in the config")
         }
     }
+
+    pub(crate) fn try_get_private_id_key(&self) -> anyhow::Result<PathBuf> {
```
jstuczyn, line 244: why do we need it? we shouldn't only care about "proper" clients, no?

common/commands/src/coconut/issue_credentials.rs
```diff
@@ -16,20 +16,12 @@ pub struct Args {
     #[clap(long)]
     pub(crate) client_config: PathBuf,
 
-    /// The amount of utokens the credential will hold.
```
jstuczyn, line 19: would it make sense to perhaps allow to specify expiration date instead?

common/bandwidth-controller/src/acquire/mod.rs
```diff
@@ -5,30 +5,32 @@ use crate::error::BandwidthControllerError;
...
 
-pub async fn deposit<C>(client: &C, amount: Coin) -> Result<State, BandwidthControllerError>
+pub async fn deposit<C>(client: &C, client_id: &[u8]) -> Result<State, BandwidthControllerError>
 where
-    C: CoconutBandwidthSigningClient + Sync,
+    C: EcashSigningClient + Sync,
 {
     let mut rng = OsRng;
     let signing_key = identity::PrivateKey::new(&mut rng);
     let encryption_key = encryption::PrivateKey::new(&mut rng);
```
jstuczyn: I think we talked about it, but just making a mental note that after everything is ready to be merged, we should look at the top of the stack and see if we can remove that key

common/bandwidth-controller/src/acquire/mod.rs
```diff
@@ -5,30 +5,32 @@ use crate::error::BandwidthControllerError;
...
 
     let tx_hash = client
         .deposit(
-            amount.clone(),
-            CredentialType::Voucher.to_string(),
+            CredentialType::TicketBook.to_string(),
```
jstuczyn, line 32: do we actually need to explicitly state it's a ticket book given that freepass is also a ticketbook, just with different expiration? (unless im totally misremembering, because I do recall you were saying something about there being no difference between ticket books and freepasses)

common/bandwidth-controller/src/acquire/mod.rs
```diff
@@ -55,19 +57,48 @@ where
...
         .await?
         .ok_or(BandwidthControllerError::NoThreshold)?;
 
-    let coconut_api_clients = all_coconut_api_clients(client, epoch_id).await?;
+    let ecash_api_clients = all_ecash_api_clients(client, epoch_id).await?;
 
-    let signature =
-        obtain_aggregate_signature(&state.voucher, &coconut_api_clients, threshold).await?;
-    let issued = state.voucher.to_issued_credential(signature, epoch_id);
+    let verification_key = obtain_aggregate_verification_key(&ecash_api_clients)?;
```
jstuczyn, line 67: would it make sense to cache VK similarly to coin indices sigs given it should be constant for the [DKG] epoch?

common/commands/src/coconut/issue_credentials.rs
```diff
@@ -40,16 +32,24 @@ pub async fn execute(args: Args, client: SigningClient) -> anyhow::Result<()> {
...
 
-    let denom = &client.current_chain_details().mix_denom.base;
-    let coin = Coin::new(args.amount as u128, denom);
-
     let persistent_storage = initialise_persistent_storage(credentials_store).await;
-    utils::issue_credential(&client, coin, &persistent_storage, args.recovery_dir).await?;
+    let private_id_key: identity::PrivateKey = nym_pemstore::load_key(private_id_key)?;
+    utils::issue_credential(
+        &client,
+        &private_id_key.to_bytes(),
```
jstuczyn, line 48: so do I understand it right that user generates an ed25519 keypair and hashes its private key to generate the wallet keypair? 
shouldn't it be hashing public key instead so other parties could verify it, because I'd imagine it would never want to share private key lol

common/bandwidth-controller/src/acquire/mod.rs
```diff
@@ -5,30 +5,32 @@ use crate::error::BandwidthControllerError;
...
-use nym_validator_client::nyxd::Coin;
+use nym_validator_client::nyxd::contract_traits::EcashSigningClient;
 use rand::rngs::OsRng;
 use state::State;
 use zeroize::Zeroizing;
 
 pub mod state;
 
-pub async fn deposit<C>(client: &C, amount: Coin) -> Result<State, BandwidthControllerError>
+pub async fn deposit<C>(client: &C, client_id: &[u8]) -> Result<State, BandwidthControllerError>
```
jstuczyn, line 23: I wonder, could we make `client_id` a bit more type safe? it's kinda an open question i guess

common/credential-storage/src/backends/memory.rs
```diff
@@ -1,13 +1,14 @@
...
 
-use crate::models::{CredentialUsage, StoredIssuedCredential};
+use crate::models::CoinIndicesSignature;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
 #[derive(Clone)]
 pub struct CoconutCredentialManager {
     inner: Arc<RwLock<CoconutCredentialManagerInner>>,
+    coin_indices_sig: Arc<RwLock<Vec<CoinIndicesSignature>>>,
```
jstuczyn: don't create another field on it. there's a reason there's an `inner` field with all the locking and pointers around it : )

jstuczyn: change inner to something like
```rust
#[derive(Default)]
struct CoconutCredentialManagerInner {
    credentials: Vec<StoredIssuedCredential>,
    credential_usage: Vec<CredentialUsage>,
    coin_indices_sig: Vec<CoinIndicesSignature>,
    _next_id: i64,
}
```

jstuczyn: oh, and I guess rename it since its no longer coconut

common/credential-storage/src/backends/memory.rs
```diff
@@ -112,12 +114,40 @@ impl CoconutCredentialManager {
...
+    /// Get coin_indices_signatures of a given epoch.
+    ///
+    /// # Arguments
+    ///
+    /// * `epoch_id`: Id of the epoch.
+    pub async fn get_coin_indices_sig(&self, epoch_id: String) -> Option<CoinIndicesSignature> {
+        let sigs = self.coin_indices_sig.read().await;
+        sigs.iter().find(|s| s.epoch_id == epoch_id).cloned()
+    }
+
```
jstuczyn, line 156: as clippy says "this file contains an unclosed delimiter" : )

common/credential-storage/src/backends/sqlite.rs
```diff
@@ -113,4 +114,58 @@ impl CoconutCredentialManager {
...
+    pub async fn get_coin_indices_sig(
+        &self,
+        epoch_id: String,
+    ) -> Result<Option<CoinIndicesSignature>, sqlx::Error> {
+        sqlx::query_as!(
+            CoinIndicesSignature,
+            "SELECT * FROM coin_indices_signatures WHERE epoch_id = ?",
+            epoch_id
+        )
+        .fetch_optional(&self.connection_pool)
```
jstuczyn, line 168: if you're already fetching optional, what's the point of `is_coin_indices_sig_present`? this could be implicitly checked with checking for the option value

common/credential-storage/src/error.rs
```diff
@@ -18,4 +18,7 @@ pub enum StorageError {
 
     #[error("No unused credential in database. You need to buy at least one")]
     NoCredential,
+
+    #[error("No signatures for that epoch in the database")]
```
jstuczyn: what epoch? imo add that information to the enum
```rust
#[error("no signatures for epoch {epoch_id} in the database")]
NoSignatures {
  epoch_id: EpochId
}
```

common/credential-storage/src/models.rs
```diff
@@ -42,3 +43,9 @@ pub struct CredentialUsage {
     pub credential_id: i64,
     pub gateway_id_bs58: String,
 }
+
+#[derive(Clone)]
+pub struct CoinIndicesSignature {
+    pub epoch_id: String,
```
jstuczyn, line 49: why is this a String and not an integer?

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+CREATE TABLE coin_indices_signatures
+(
+    epoch_id                TEXT NOT NULL PRIMARY KEY,
```
jstuczyn, line 8: why no `epoch_id INTEGER NOT NULL PRIMARY KEY`?

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
```diff
@@ -0,0 +1,25 @@
...
+(
+    id                     INTEGER                                                                  NOT NULL PRIMARY KEY AUTOINCREMENT,
+
+--     introduce a way for us to introduce breaking changes in serialization
+serialization_revision INTEGER                                                                      NOT NULL,
+
+--     the best we can do without enums
+    credential_type        TEXT CHECK ( credential_type IN ('TicketBook', 'FreeBandwidthPass') )    NOT NULL,
+    credential_data        BLOB                                                                     NOT NULL UNIQUE,
+    epoch_id               INTEGER                                                                  NOT NULL,
```
jstuczyn, line 22: like in here : ) (references the top comment in this file)

common/credential-storage/src/models.rs
```diff
@@ -27,6 +27,7 @@ pub struct StoredIssuedCredential {
 
     pub epoch_id: u32,
     pub expired: bool,
+    pub consumed: bool,
```
jstuczyn, line 30: do we need it?

common/credential-storage/src/persistent_storage.rs
```diff
@@ -105,6 +106,37 @@ impl Storage for PersistentStorage {
...
+        epoch_id: String,
+        coin_indices_sig: String,
+    ) -> Result<(), StorageError> {
+        self.coconut_credential_manager
+            .insert_coin_indices_sig(epoch_id, coin_indices_sig)
+            .await?;
+        Ok(())
+    }
+
+    async fn is_coin_indices_sig_present(
```
jstuczyn, line 120: as I mentioned before, this is redundant of the `get` method returns an option. (unless we changed `fetch_optional()` to `fetch_one()`)

common/credential-utils/src/utils.rs
```diff
@@ -130,15 +130,25 @@ where
     let mut recovered_amount: u128 = 0;
     for voucher in recovery_storage.unconsumed_vouchers()? {
         let voucher_value = match voucher.typ() {
-            CredentialType::Voucher => voucher.get_bandwidth_attribute(),
+            CredentialType::TicketBook => voucher.value(),
+            CredentialType::Voucher => {
```
jstuczyn, line 134: do we even want to acknowledge existence of `Voucher` type? why not just purge it : P

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
 
 impl IssuanceBandwidthCredential {
-    pub const PUBLIC_ATTRIBUTES: u32 = 2;
-    pub const PRIVATE_ATTRIBUTES: u32 = 2;
-    pub const ENCODED_ATTRIBUTES: u32 = Self::PUBLIC_ATTRIBUTES + Self::PRIVATE_ATTRIBUTES;
-
     pub fn default_parameters() -> Parameters {
-        // safety: the unwrap is fine here as Self::ENCODED_ATTRIBUTES is non-zero
-        Parameters::new(Self::ENCODED_ATTRIBUTES).unwrap()
+        setup(constants::NB_TICKETS)
```
jstuczyn, line 68: we talked about it, this should probably use the global params from somewhere

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
-        let type_prehashed = hash_to_scalar(variant_data.info().to_string());
-
-        let params = bandwidth_credential_params();
-        let serial_number = params.random_scalar();
-        let binding_number = params.random_scalar();
+        let params = bandwidth_credential_params().grp();
+        let ecash_keypair = if let Some(id) = identifier {
+            generate_keypair_user_from_seed(params, id)
+        } else {
+            generate_keypair_user(params)
```
jstuczyn, line 81: is there any situation under which you wouldn't use the seed?

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
+        Self::new(
+            BandwidthVoucherIssuanceData::new(deposit_tx_hash, signing_key, unused_ed25519),
+            Some(identifier),
+            cred_exp_date_timestamp(),
+        )
     }
 
-    pub fn blind_serial_number(&self) -> BlindedSerialNumber {
-        (bandwidth_credential_params().gen2() * self.serial_number).into()
+    pub fn new_freepass(timestamp: u64) -> Self {
```
jstuczyn, line 103: use `OffsetDateTime` if possible

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
-    pub fn get_public_attributes(&self) -> Vec<&PublicAttribute> {
-        vec![self.variant_data.public_value(), &self.type_prehashed]
-    }
-
-    pub fn get_plain_public_attributes(&self) -> Vec<String> {
-        vec![
-            self.variant_data.public_value_plain(),
-            self.typ().to_string(),
-        ]
+    pub fn expiration_date(&self) -> u64 {
```
jstuczyn, line 129: ibid.

common/credentials/src/coconut/bandwidth/issuance.rs
```diff
@@ -79,141 +53,141 @@ impl BandwidthCredentialIssuanceDataVariant {
...
-        self.variant_data.public_value_plain()
+    pub fn value(&self) -> u128 {
+        if let BandwidthCredentialIssuanceDataVariant::TicketBook(data) = &self.variant_data {
+            data.value()
+        } else {
+            0_u128
+        }
+    }
+
+    pub fn check_expiration_date(&self) -> bool {
```
jstuczyn, line 145: wait, I'm a bit confused. why are we storing the same expiration date on 2 different fields?

common/credentials/src/coconut/bandwidth/issued.rs
```diff
@@ -2,117 +2,97 @@
...
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 pub const CURRENT_SERIALIZATION_REVISION: u8 = 1;
 
 #[derive(Debug, Zeroize, Serialize, Deserialize)]
 pub enum BandwidthCredentialIssuedDataVariant {
-    Voucher(BandwidthVoucherIssuedData),
-    FreePass(FreePassIssuedData),
+    TicketBook(BandwidthVoucherIssuedData),
+    FreePass,
```
jstuczyn, line 26: huh, do we not need any data for the freepasses?

common/credentials/src/coconut/bandwidth/issued.rs
```diff
@@ -2,117 +2,97 @@
...
-    Parameters, PrivateAttribute, PublicAttribute, Signature, VerificationKey,
+    constants, date_scalar, CoinIndexSignature, ExpirationDateSignature, Parameters, PayInfo,
+    SecretKeyUser, VerificationKeyAuth, Wallet,
 };
 use nym_validator_client::nym_api::EpochId;
 use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 pub const CURRENT_SERIALIZATION_REVISION: u8 = 1;
```
jstuczyn, line 21: bump it up because `v1` was for all the coconut stuff

common/credentials/src/coconut/bandwidth/issued.rs
```diff
@@ -137,6 +117,27 @@ impl IssuedBandwidthCredential {
         CURRENT_SERIALIZATION_REVISION
     }
 
+    pub fn expiration_date(&self) -> u64 {
+        self.expiration_date
+    }
+
+    pub fn expiration_date_formatted(&self) -> OffsetDateTime {
+        //SAFETY : expiration date is encoded as a u64 but it is a unix timestamp. The unwrap is guaranteed to succeed for at least 290 million more years
```
jstuczyn, line 125: .... unless somebody enters malformed data since the constructor is public

jstuczyn: just keep everything as an OffsetDateTime from the start

common/credentials/src/coconut/bandwidth/issued.rs
```diff
@@ -2,117 +2,97 @@
...
 
     /// Specifies the (DKG) epoch id when this credential has been issued
     epoch_id: EpochId,
+
+    ///secret ecash key used to generate this wallet
+    ecash_secret_key: SecretKeyUser,
+
+    ///signatures on expiration dates used to spend tickets
+    #[zeroize(skip)]
+    exp_date_signatures: Vec<ExpirationDateSignature>,
```
jstuczyn, line 74: how many of those do we have? because the whole `IssuedBandwidthCredential` will need to be serialised and given back to the user. the old credentials were like 250bytes so could fit on a QR code. ideally so would this one

common/credentials/src/coconut/bandwidth/issued.rs
```diff
@@ -137,6 +117,27 @@ impl IssuedBandwidthCredential {
...
+    pub fn expiration_date_formatted(&self) -> OffsetDateTime {
+        //SAFETY : expiration date is encoded as a u64 but it is a unix timestamp. The unwrap is guaranteed to succeed for at least 290 million more years
+        OffsetDateTime::from_unix_timestamp(self.expiration_date.try_into().unwrap()).unwrap()
+    }
+
+    pub fn expired(&self) -> bool {
+        self.expiration_date < today_timestamp()
+    }
+
+    pub fn exp_date_sigs(&self) -> Vec<ExpirationDateSignature> {
```
jstuczyn, line 133: do we have to clone it?

jstuczyn (changes requested):

nym-api/migrations/20240208120000_issued_credential_ecash.sql
simonwicky: This ties up with the whole timestamp handling case. The underlying library uses `u64`, so I decided to handle `u64` as much as possible. We can debate about that, but ultimately, any fix will happen right before merging as it will affect the whole stack. 
TBD

nym-api/src/coconut/state.rs
simonwicky: See earlier comments on timestamp handling. 
TBD

nym-api/src/coconut/state.rs
simonwicky: See earlier comments on timestamp handling.
TBD

nym-api/nym-api-requests/src/coconut/models.rs
simonwicky: See earlier comments on timestamp handling.
TBD

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: See earlier comments on timestamp handling.
TBD

common/commands/src/coconut/issue_credentials.rs
simonwicky: The private ID key is hashed into the private wallet key. If you hash the public ID key, then your private wallet key isn't really private. 
The goal is to tie the wallet secret key to something more or less painful to change because it is the core of double spending protection.

common/credential-storage/src/models.rs
simonwicky: A ticketbook is consumed when there's no tickets left. The number of tickets left is within the credential data. Having that field lets us check if it's usable without getting and deserialising the whole credential.

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: See earlier comments on timestamp handling.
TBD

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: See earlier comments on timestamp handling.
TBD

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: As many as days where this credential will be valid. 30 days => 30 signatures

nym-api/src/coconut/state.rs
simonwicky: Note that on paper, they are available at anytime, so we can imagine other ways on distributing those. You could fetch them at spend time, but you would reveal your expiration date

nym-api/src/coconut/api_routes/mod.rs
simonwicky: This variant will be changed as part of the whole crypto lib error revamp, so I will make that change then

nym-api/src/coconut/state.rs
simonwicky: Indeed. We can't cache everything though. What do you propose?

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: Note that on paper, they are available at anytime, so we can imagine other ways on distributing those. You could fetch them at spend time, but you would reveal your expiration date

nym-api/src/coconut/state.rs
simonwicky: Crypto paper says yes. I don't speak crypto fluently enough to give you the explanation though :(

nym-api/src/coconut/state.rs
simonwicky: Indeed. I was trying to minimise the changes. To be fixed later down the line

nym-api/src/coconut/state.rs
simonwicky: They will be static following some changes in the underlying crypto library. I'll change that then

nym-api/src/coconut/state.rs
simonwicky: I'm not sure what I'm missing. I'm not querying the chain here, am I?

nym-api/src/coconut/helpers.rs
simonwicky: Artefact of the PR stack, it does not exist higher up

nym-api/src/coconut/helpers.rs
simonwicky: These values are shared between threads so not having the atomics will create race conditions no?
I will change the `Ordering` though

nym-api/src/coconut/helpers.rs
simonwicky: Yes indeed. Those signatures' epoch needs to match the underlying credential.
Let's discuss about signatures distribution altogether

nym-api/src/coconut/helpers.rs
simonwicky: `signatures` is `None` only at startup, to avoid useless computations. The first call to that `fn` WILL set `signatures`, and it is never put to `None` anywhere after that, so there will be something

nym-api/src/coconut/deposit.rs
simonwicky: Tied to the contract interface, will be changed with the contract interface on #4456

nym-api/src/coconut/deposit.rs
simonwicky: Same as before, this will fixed with the contract in #4456

common/credentials/src/coconut/bandwidth/voucher.rs
simonwicky: About bandwidth vouchers, you tell me. What are they useful for?
We didn't used the denomination from the coin. What was the legacy reason?

common/credentials/src/coconut/utils.rs
simonwicky: Yes, TODO later down the line, when it won't cause any PR stack problem

common/credentials/src/coconut/utils.rs
simonwicky: Mentioned earlier, this will happen, but changes overlap #4456. Will be changed as part of that

nym-api/src/coconut/state.rs
simonwicky: Didn't know it existed. Fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

common/credentials/src/coconut/utils.rs
simonwicky: Fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

nym-api/nym-api-requests/src/coconut/models.rs
simonwicky: Fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

nym-api/src/coconut/helpers.rs
simonwicky: `Ordering` fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

nym-api/src/coconut/deposit.rs
simonwicky: Fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

nym-api/src/coconut/deposit.rs
simonwicky: Removed legacy stuff in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311
Rest will depend on the contract changes from #4456

common/credentials/src/coconut/bandwidth/voucher.rs
simonwicky: Fixed in 6ac4627c99986a865c6d0e3b9b82ac32ac2b5311

common/credentials/src/coconut/utils.rs
simonwicky: We don't, there's no scalar attributes exposed anymore

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: TBD. I only adapted to existing structure

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Inside the `Wallet` object, it's not exposed

common/credential-storage/src/storage.rs
simonwicky: Artefact of the PR stack, my bad.
We don't need the gateway_id anymore, because ticket book are by essence multi-use credential. You don't want to be able to spend the whole ticket book at a gateway

common/credential-storage/src/storage.rs
simonwicky: Sqlite doesn't support `u64` and this avoids conversion problems

common/credential-storage/src/storage.rs
simonwicky: Sqlite doesn't support `u64` and this avoids conversion problems

common/credential-storage/src/storage.rs
simonwicky: Sqlite doesn't support `u64` and this avoids conversion problems

common/credential-storage/src/storage.rs
simonwicky: Since we don't really care if we have it or if there's a problem with the db, I integrated the `Option` part in the `Result`

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: The public key will not be unique, but `credential_data` will though. If you have the same data, you have the same credential. 
Waiting on @octol's input for that one

common/commands/src/utils.rs
simonwicky: The Nym API currently has a network monitor which is a standard Nym-client and would require credentials. Since @durch is working on a new one, I didn't want to implement this for nothing.

common/commands/src/utils.rs
simonwicky: Anything that connect to the mixnet will need a credential though. TBH, I'm not sure of the actual use of that `UnknownConfigWrapper`

common/commands/src/coconut/issue_credentials.rs
simonwicky: Not really, our credentials will have a fixed validity period (at least in the current code state) and changing that adds quite some changes elsewhere

common/bandwidth-controller/src/acquire/mod.rs
simonwicky: We can remove it, but we need to change the contract interface along that. To be done at #4456 stage

common/bandwidth-controller/src/acquire/mod.rs
simonwicky: The difference for now is that a freepass is not backed by a deposit, which make handling them slightly different, only because there's no rewards for them. 
It's also a copycat of the late `Voucher`-`Freepass` distinction

common/bandwidth-controller/src/acquire/mod.rs
simonwicky: I guess so, we can discuss that with the whole signatures distribution stuff

common/bandwidth-controller/src/acquire/mod.rs
simonwicky: I specified `client_id` but in reality it's more a seed than anything else. If anything, it should be something that can be hashed.

common/credential-storage/src/backends/memory.rs
simonwicky: PR stack building artefact

common/credential-storage/src/backends/sqlite.rs
simonwicky: Signatures are needed at spend time. 
When creating a credential, a client checks if he has the corresponding signatures. It he does, he doesn't need them right away, so let's not load them here (they aren't really small).
The `fetch_optional` also help differentiating with other database errors

common/credential-storage/src/models.rs
simonwicky: Ties up with some comments above about `u64` and sqlite

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: Same as above

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: I copied that from elsewhere. And the conversion uses an `expect`
Let's discuss on that

common/credential-storage/src/persistent_storage.rs
simonwicky: Signatures are quite big, and at credential issuance, we only want to check if we have them, without fetching them

common/credential-utils/src/utils.rs
simonwicky: I kept backward compatibility as much as possible

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Yep, that will be done on the #4456 step

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Network monitor maybe?

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: Freepasses have no informations that Ticketbooks do not have. The reverse isn't true

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: We're still using the same serializer though?

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: Let's discuss this with the whole timestamp keeping topic

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: We don't. It affect spending, so let's keep that for later

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: This method is used if we need to recover the deposit. It used to modify the expiration date, but it doesn't anymore. I'm gonna rework it a bit in 3bceecb6b8e48fd3eb719f5c40ca8321105d2357

common/credential-storage/src/error.rs
simonwicky: Fixed in 3bceecb6b8e48fd3eb719f5c40ca8321105d2357

common/credential-storage/src/backends/memory.rs
simonwicky: The outer struct name will be renamed later.
Fixed in 3bceecb6b8e48fd3eb719f5c40ca8321105d2357

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: Can't find a linter that doesn't change the whole format. 
So, maybe fixed in 3bceecb6b8e48fd3eb719f5c40ca8321105d2357

common/credentials/src/coconut/utils.rs
simonwicky: Fixed in 3bceecb6b8e48fd3eb719f5c40ca8321105d2357

common/bandwidth-controller/src/acquire/mod.rs
simonwicky: Fixed in 757dc2880ce4b2233062dc019d8ac7a0a95fe7eb

nym-api/src/coconut/state.rs
simonwicky: Fixed in ebd882992abaa333d3f6a3d61411526ac4830a14

common/credentials/src/coconut/utils.rs
simonwicky: Fixed in ebd882992abaa333d3f6a3d61411526ac4830a14

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Fixed in ebd882992abaa333d3f6a3d61411526ac4830a14

nym-api/src/coconut/api_routes/mod.rs
simonwicky: Fixed in 5972077342e1ba29bd21b9c12e7f2db875860939

common/credentials/src/coconut/utils.rs
simonwicky: Fixed in 2130ee14647238714d4fb060cd1cc839fdea03df

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: Fixed in e5d4e31fc22fc36493b04cdb07bc805ecc975ae0

nym-api/src/coconut/state.rs
jstuczyn: but you are, in ` let verification_key = self.verification_key(current_epoch.epoch_id).await?;`

common/credential-storage/src/storage.rs
jstuczyn: then store it as u32 and cast it. way better than string

nym-api/migrations/20240208120000_issued_credential_ecash.sql
simonwicky: TO FIX : timestamps

nym-api/src/coconut/state.rs
simonwicky: TO FIX : timestamps

nym-api/nym-api-requests/src/coconut/models.rs
simonwicky: TO FIX : timestamps

nym-api/src/coconut/state.rs
simonwicky: Open question

nym-api/src/coconut/state.rs
simonwicky: TO FIX

nym-api/src/coconut/state.rs
simonwicky: Isn't that from the communication channel which cache the chain? (see a couple of comments above)

nym-api/src/coconut/helpers.rs
simonwicky: Nothing to fix

nym-api/src/coconut/helpers.rs
simonwicky: Won't fix

common/credentials/src/coconut/bandwidth/voucher.rs
simonwicky: Why should it be?

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: TODO after merge?

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: TO FIX : timestamps

common/credential-storage/src/storage.rs
simonwicky: TO FIX : string in storage

common/credential-storage/src/storage.rs
simonwicky: TO FIX : string in storage

common/credential-storage/src/storage.rs
simonwicky: TO FIX : string in storage

common/credential-storage/src/backends/memory.rs
simonwicky: Nothing to fix

common/credential-storage/src/models.rs
simonwicky: TO FIX : string in storage

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: TO FIX : string in storage

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: Nothing to fix for that, will fix the others

common/credential-utils/src/utils.rs
simonwicky: TO FIX : scrape backward compatibility

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: TO FIX : timestamps

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: TO FIX : timestamps

common/credential-storage/src/storage.rs
simonwicky: Fixed in 9cbae803f579a01780a0bd21af3e3a410ca25e1d

common/credential-storage/src/storage.rs
simonwicky: Fixed in 9cbae803f579a01780a0bd21af3e3a410ca25e1d

common/credential-storage/src/storage.rs
simonwicky: Fixed in 9cbae803f579a01780a0bd21af3e3a410ca25e1d

common/credential-storage/src/models.rs
simonwicky: Fixed in 9cbae803f579a01780a0bd21af3e3a410ca25e1d

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
simonwicky: Fixed in 9cbae803f579a01780a0bd21af3e3a410ca25e1d

nym-api/src/coconut/state.rs
jstuczyn: maybe some ephemeral cache... I will need to ponder about it. let's leave it towards the end.

nym-api/src/coconut/state.rs
jstuczyn: sure, but you might get unlucky and hit the stale entry thus having to query the chain directly.

thus if you check for keys beforehand, you might avoid the query : )

common/credentials/src/coconut/bandwidth/voucher.rs
jstuczyn: so we started off by using just the value of the coin. 
then I've done bunch of refactoring of those types and changed the internal types to `Coin` (didn't touch the contract yet) since it'd be better to be explicit about it and use the full coin value
but run out of time to make changes in the contract

common/credentials/src/coconut/bandwidth/voucher.rs
jstuczyn: why shouldn't it be? : )
it's better to be explicit in case we have some weird edge case where somebody decides to send idk. `nyx` rather than `nym`

common/credentials/src/coconut/bandwidth/issuance.rs
jstuczyn: yeah maybe after merge or towards the end, we'll see

common/credential-utils/src/utils.rs
simonwicky: Fixed in c6c3b2f644b6c608f9f90ea3bd75aded0d45d01e

common/credential-storage/migrations/20241104120001_add_ecash_tables.sql
jstuczyn: i think it might be. if not, I will fix it before merging : )

common/commands/src/coconut/issue_credentials.rs
jstuczyn: as discused on element, we might want to consider generating a separate long-term ecash "key" instead

common/credentials/src/coconut/bandwidth/issued.rs
jstuczyn: it's more to indicate the structure of the thing being serialised. for example if we added/renamed a field, the version should get bumped

common/credentials/src/coconut/bandwidth/issued.rs
jstuczyn: just thinking aloud here, we might want to have 2 different types of "serialised" credential. one with the signatures, one without.

nym-api/migrations/20240208120000_issued_credential_ecash.sql
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

nym-api/src/coconut/state.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

nym-api/src/coconut/state.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

nym-api/nym-api-requests/src/coconut/models.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

common/credentials/src/coconut/bandwidth/issuance.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

common/credentials/src/coconut/bandwidth/issued.rs
simonwicky: Fixed in 7bc7a8dde1e5c49778ba132e454a138dd1f0fdea

nym-api/src/coconut/state.rs
simonwicky: Fixed in 285e17ca0cdbf15af680ef82d62ff0480521284a

common/credentials/src/coconut/bandwidth/voucher.rs
simonwicky: The contract only accepts a fixed amount of `nym` though

jstuczyn:

PR #4458, opened 2024-03-12
Feature Compact Ecash : Freepass issuance

simonwicky:
# Description

This PR is a part of the Ecash stack. The base of that stack is https://github.com/nymtech/nym/pull/4456 . Refer to the base's description for details.

This PR contains the necessary bits for Freepass issuance, CLI and API side. It will probably not build on its own

# Checklist:


common/commands/src/coconut/generate_freepass.rs
```diff
@@ -55,17 +59,17 @@ pub struct Args {
 
 async fn get_freepass(
     api_clients: Vec<CoconutApiClient>,
-    aggregate_vk: &VerificationKey,
+    aggregate_vk: &VerificationKeyAuth,
     threshold: u64,
     epoch_id: EpochId,
     signing_account: &AccountData,
-    expiration_date: OffsetDateTime,
+    expiration_date_ts: u64,
```
jstuczyn, line 66: why not keep the proper timestamp given that's what our CLIs and APIs will be accepting? in particular the vpn api

common/commands/src/coconut/generate_freepass.rs
```diff
@@ -55,17 +59,17 @@ pub struct Args {
...
 ) -> anyhow::Result<IssuedBandwidthCredential> {
-    let issuance_pass = IssuanceBandwidthCredential::new_freepass(Some(expiration_date));
+    let issuance_pass = IssuanceBandwidthCredential::new_freepass(expiration_date_ts);
     let signing_data = issuance_pass.prepare_for_signing();
 
-    let credential_shares = Arc::new(tokio::sync::Mutex::new(Vec::new()));
-
+    let wallet_shares = Arc::new(tokio::sync::Mutex::new(Vec::new()));
+    let signatures_shares = Arc::new(tokio::sync::Mutex::new(Vec::new()));
```
jstuczyn: it's a tiny thing but imo just keep a single mutex. no point in having two of them since they're always used at the same time anyway

common/commands/src/coconut/generate_freepass.rs
```diff
@@ -142,13 +176,18 @@ pub async fn execute(args: Args, client: SigningClient) -> anyhow::Result<()> {
         None => OffsetDateTime::from_unix_timestamp(args.expiration_timestamp.unwrap())?,
     };
 
-    let now = OffsetDateTime::now_utc();
+    //SAFETY : positive timestamp, so conversion to unsigned will not fail
```
jstuczyn: but since we're returning an `anyhow::Result<()>` you might as well use `?` operator and remove the unwrap completely

common/credentials/src/coconut/bandwidth/freepass.rs
```diff
@@ -1,105 +1,27 @@
...
 
-pub const MAX_FREE_PASS_VALIDITY: Duration = Duration::WEEK; // 1 week
-
 #[derive(Debug, Zeroize, ZeroizeOnDrop, Serialize, Deserialize)]
-pub struct FreePassIssuedData {
-    /// the plain validity value of this credential expressed as unix timestamp
-    #[zeroize(skip)]
-    expiry_date: OffsetDateTime,
-}
+pub struct FreePassIssuedData {}
```
jstuczyn, line 12: doesn't freepass have an explicit expiration date?

common/credentials/src/coconut/bandwidth/freepass.rs
```diff
@@ -1,105 +1,27 @@
...
-pub struct FreePassIssuanceData {
-    /// the plain validity value of this credential expressed as unix timestamp
-    #[zeroize(skip)]
-    expiry_date: OffsetDateTime,
-
-    // the expiry date, as unix timestamp, hashed into a scalar
-    #[serde(with = "scalar_serde_helper")]
-    expiry_date_prehashed: PublicAttribute,
-}
+pub struct FreePassIssuanceData {}
```
jstuczyn, line 14: and we don't need anything to issue them?

I guess then maybe we could remove `BandwidthCredentialIssuedDataVariant` and `BandwidthCredentialIssuanceDataVariant` and flatten the Issued/Issuance credential variants

nym-api/nym-api-requests/src/coconut/models.rs
```diff
@@ -135,7 +135,7 @@ pub struct FreePassRequest {
     // secp256k1 key associated with the admin account
```
jstuczyn, line 135: so isn't the inner request actually the same one as for "normal" credential? (minus nonce + signature)

nym-api/src/coconut/api_routes/mod.rs
```diff
@@ -94,7 +52,16 @@ pub async fn post_free_pass(
     debug!("Received free pass sign request");
     trace!("body: {:?}", freepass_request_body);
 
-    validate_freepass_public_attributes(&freepass_request_body)?;
+    //check expiration date validity
+    if freepass_request_body.expiration_date > freepass_exp_date_timestamp() {
+        return Err(CoconutError::TooLongFreePass {
+            expiry_date: OffsetDateTime::from_unix_timestamp(
+                //safety : expiration date is a unix timestamp so these unwraps will not fail for 290 million years
+                freepass_request_body.expiration_date.try_into().unwrap(),
```
jstuczyn: absolutely not. the request comes from an user with no sanitisation. so a malicious client could crash the nym-api by providing an invalid timestamp.

jstuczyn (changes requested):

common/commands/src/coconut/generate_freepass.rs
simonwicky: This ties up with the whole timestamp handling case. The underlying library uses u64, so I decided to handle u64 as much as possible. We can debate about that, but ultimately, any fix will happen right before merging as it will affect the whole stack.
TBD

common/credentials/src/coconut/bandwidth/freepass.rs
simonwicky: It's not needed in ecash, as any ticket book has an expiration date. The expiration date that was there was moved up in the struct that is using `FreePassIssuedData`.
Freepasses have nothing that normal ticket books do not have

common/credentials/src/coconut/bandwidth/freepass.rs
simonwicky: To be fair I was unsure about those two new structs so I guess we could.
Freepasses are ticket books that aren't backed by a deposit, so yeah, we don't need anything special

nym-api/nym-api-requests/src/coconut/models.rs
simonwicky: Yes

nym-api/src/coconut/api_routes/mod.rs
simonwicky: Fixed in fa1cf7106c7a178fad8d11a99d3ca0a15e09376a

common/commands/src/coconut/generate_freepass.rs
simonwicky: Fair point. Fixed in fa1cf7106c7a178fad8d11a99d3ca0a15e09376a

common/commands/src/coconut/generate_freepass.rs
simonwicky: Fixed in 0090c6954d5668f225af83f58bd91e7311ce077b

common/commands/src/coconut/generate_freepass.rs
simonwicky: TO FIX : timestamps

simonwicky: Fixed in f2cfee9ee8605d3cc8dd3bf25a119aa38aa37acd

jstuczyn:

PR #4461, opened 2024-03-12
Feature Compact Ecash : DKG types adaptation

simonwicky:
# Description

This PR is the second in the Ecash stack. The base of that stack is https://github.com/nymtech/nym/pull/4456 . Refer to the base's description for details.

This PR mainly does trivial type changes from coconut types to ecash types

# Checklist:

### DIFF
--- a/nym-api/src/coconut/keys/mod.rs
+++ b/nym-api/src/coconut/keys/mod.rs
@@ -1,8 +1,8 @@
 // Copyright 2022-2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use nym_coconut::VerificationKey;
 use nym_coconut_dkg_common::types::EpochId;
+use nym_compact_ecash::VerificationKeyAuth;
 use nym_dkg::Scalar;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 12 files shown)


nym-api/src/coconut/keys/mod.rs
```diff
@@ -64,11 +64,11 @@ impl KeyPair {
         }
     }
 
-    pub async fn verification_key(&self) -> Option<RwLockReadGuard<'_, VerificationKey>> {
-        RwLockReadGuard::try_map(self.get().await?, |maybe_keypair| {
-            maybe_keypair.as_ref().map(|k| k.keys.verification_key())
-        })
-        .ok()
+    pub async fn verification_key(&self) -> Option<VerificationKeyAuth> {
```
jstuczyn, line 67: The reason I was returning `RwLockReadGuard` was to avoid key cloning for simple reads. And the key is quite complex -> it's bunch of curve points.

I think the issue here comes from the fact that `keys.verification_key()` in the crypto code does key cloning. Not sure if I left this comment in the previous PR

nym-api/src/coconut/keys/persistence.rs
```diff
@@ -3,17 +3,17 @@
...
 use std::mem;
 
 impl PemStorableKey for KeyPairWithEpoch {
     // that's not the best error for this, but it felt like an overkill to define a dedicated struct just for this purpose
-    type Error = CoconutError;
+    type Error = CompactEcashError;
 
     fn pem_type() -> &'static str {
-        "COCONUT KEY WITH EPOCH"
+        "COCONUT KEY WITH EPOCH" // avoid the invalidation of already present key
```
jstuczyn, line 16: I don't think we want to be hacking our compatibility with coconut keys this way. 
I'd say, define a brand new pem type with brand new serialisation (in case we changed something in the serde crypto code in the previous PR).
Then, on nym-api startup (or coconut module creation/key load) see if you have old coconut keys. if so, convert them into ecach keys via `From` trait

jstuczyn (changes requested):
One major note is that I'd say we shouldn't replace the key types in place because we can't guarantee 1-1 serialisation compatibility between coconut and ecash keys (especially if we want to be messing with the previous PR). Instead just implement conversion from coconut keys into ecash keys and as an extra bonus, we'll get rid of any legacy dependencies (like the 'wrong' PEM type)

jstuczyn:

PR #4463, opened 2024-03-12
Feature/multigateway config

jstuczyn:
# Description

Closes: CT-80

tl;dr:

This PR changes client storage and behaviour to more easily allow usage of multiple gateways. Assume there exist 2 gateways in the system, `A` and `B`. Prior this change you'd have to `init` unique clients with different `id` for both and you'd end up with different address prefixes, i.e. `aaa.bbb@A` and `ccc.ddd@B`. It was also more annoying to deal with credentials, in particular free passes. Now all of this data is stored under a single client `id`. So now client has a single identity key, encryption key, ack key and **multiple** shared gateway keys.

On paper this is quite a straightforward and simple change, but because it had to touch one of the core client bits it required bunch of changes. Fortunately, most of them were due to config migrations and backwards compatibility so the number of lines changed is heavily biased because I had to make **identical** changes to socks5, native client, NR and IPR. But as the result of that, there's going to be a lot of QA work needed to check for regression.

From the top of my head, the following need to be tested:

full client regression with backwards compatibility, i.e. `init`, "re-init", `run`, et. al:
- native
- socks5 (client + nym-connect)
- NR
- IPR
- rust SDK
- wasm (that one is **not** backwards compatible; it should throw you an error when you try to use existing data)
- also vpn related things because ive changed few tiny bits with wireguard

new features (for all clients):
- ability to register with multiple gateways (via `init` - now that I write this, I wonder if there should be a dedicated method for that...)
- ability to switch currently active gateway (via `switch-gateway`)
- `list-gateways` command
- clients should now automagically re-use freepasses with multiple gateways, meaning once you import a freepass and spend it with gateway `A` and then switch active gateway to `B`, you will be able to use it without having to manually mess with the storage (the same is **not** true for bandwidth vouchers; those are globally unique)


octol (approved):
Another beast of a PR I see :)

PR #4464, opened 2024-03-13
removed the usage of allow.list

jstuczyn:
# Description

Closes: CT-92

As the title says, this PR completely removes allow.list in favour of the exit policy.
Note that this PR will remain a draft until https://github.com/nymtech/nym/pull/4463 is merged since I need to introduce config migrations for network requester and I don't want to cause myself even more merge conflicts...
I've deprecated relevant fields so clippy will yell about their usage, but that's fine.


octol (approved):

tommyv1987:
LGTM - next stop smoosh

PR #4465, opened 2024-03-13
Feature Noise XKpsk3 : Fix localnet

simonwicky:
# Description

This PR brings small QoL changes to NymTopology, adds the `--custom-mixnet` option to mix nodes and gateways and adapt the `localnet_start.sh` script.

# Checklist:

### DIFF
--- a/common/topology/src/lib.rs
+++ b/common/topology/src/lib.rs
@@ -119,11 +119,15 @@ pub struct NymTopology {
 }
 
 impl NymTopology {
-    pub fn new(mixes: BTreeMap<MixLayer, Vec<mix::Node>>, gateways: Vec<gateway::Node>) -> Self {
+    pub fn new(
+        mixes: BTreeMap<MixLayer, Vec<mix::Node>>,
+        gateways: Vec<gateway::Node>,
+        described_nodes: Vec<DescribedNymNode>,
+    ) -> Self {
         NymTopology {
             mixes: mixes.clone(),
             gateways: gateways.clone(),
-            described_nodes: Vec::new(),
+            described_nodes: described_nodes.clone(),
         }
     }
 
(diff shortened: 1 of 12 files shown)


common/topology/src/serde.rs
```diff
@@ -181,7 +192,7 @@ pub struct SerializableGateway {
     // optional ip address in the case of host being a hostname that can't be resolved
     // (thank you wasm)
     #[cfg_attr(feature = "wasm-serde-types", tsify(optional))]
-    #[serde(alias = "explicit_ip")]
+    #[serde(alias = "explicit_ips")]
```
octol, line 196: Is this correct?

simonwicky: I changed that to match the changes I did on that variable name earlier. 
It wasn't used by the localnet script, so it wasn't creating any problems. I don't know where else it's used though.

octol (approved):

common/topology/src/lib.rs
```diff
@@ -119,11 +119,15 @@ pub struct NymTopology {
 }
 
 impl NymTopology {
-    pub fn new(mixes: BTreeMap<MixLayer, Vec<mix::Node>>, gateways: Vec<gateway::Node>) -> Self {
+    pub fn new(
+        mixes: BTreeMap<MixLayer, Vec<mix::Node>>,
+        gateways: Vec<gateway::Node>,
+        described_nodes: Vec<DescribedNymNode>,
```
jstuczyn, line 125: unless I'm missing something, this new field doesn't seem to be used anywhere? you just assign it and kinda leave it

simonwicky: I'm using it here to get nodes' key: https://github.com/nymtech/nym/blob/1663f75b31626627db9895139024f8758afa2bdc/common/topology/src/lib.rs#L203
I need the description during the update phase, afterwards I only need a list of all nodes (including inactive and blacklisted one).
I was using https://github.com/nymtech/nym/blob/1663f75b31626627db9895139024f8758afa2bdc/common/topology/src/lib.rs#L142 to assign it before, but it wasn't set everywhere because of that and it was leading to problems.

PR #4474, opened 2024-03-15
Support running both NR and IPR

octol:
# Description

Support running both the network-requester and the ip packet router at the same time on a gateway

Some duplication going on in starting up the embedded client. I claim rule of three


gateway/src/main.rs
```diff
@@ -42,7 +41,7 @@ struct Cli {
 }
 
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+async fn main() -> anyhow::Result<()> {
```
jstuczyn, line 44: hehe, nice one

jstuczyn (approved):

PR #4492, opened 2024-03-21
Fix TLS connection error

octol:
Fix the TLS connection issue ` TLS support not compiled in` by making sure we use the `rustls-tls-native-roots` feature flag to tungstenite in gateway-client and client-core.

The fix can be checked with
```sh
cargo run -p nym-client -- init --id sdaf1 --gateway EBT8jTD8o4tKng2NXrrcrzVhJiBnKpT1bJy5CMeArt2w
```

which failed before and now succeeds

### DIFF
--- a/common/client-libs/gateway-client/src/socket_state.rs
+++ b/common/client-libs/gateway-client/src/socket_state.rs
@@ -44,9 +44,7 @@ pub(crate) fn ws_fd(_conn: &WsConn) -> Option<RawFd> {
     #[cfg(unix)]
     match _conn.get_ref() {
         MaybeTlsStream::Plain(stream) => Some(stream.as_raw_fd()),
-        &_ => unreachable!(
-            "If tls features are enabled, the inner stream needs to be unpacked into raw fd"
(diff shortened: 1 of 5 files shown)


common/client-libs/gateway-client/src/socket_state.rs
```diff
@@ -44,9 +44,7 @@ pub(crate) fn ws_fd(_conn: &WsConn) -> Option<RawFd> {
     #[cfg(unix)]
     match _conn.get_ref() {
         MaybeTlsStream::Plain(stream) => Some(stream.as_raw_fd()),
-        &_ => unreachable!(
-            "If tls features are enabled, the inner stream needs to be unpacked into raw fd"
-        ),
+        &_ => None,
```
octol, line 47: Will this impact Android @neacsu you think?

octol: Tested and seems to be fine, at least no regression

tommyv1987 (approved):

PR #4500, opened 2024-03-25
Feature/one binary to rule them all

jstuczyn:
ok. this PR. it's quite a doozy. Let's get some things out of the way first:
- for purposes of our favourite ticketing system, it closes: CT-101
- https://github.com/nymtech/nym/pull/4463 needs to be merged first. I branched off that branch so that I wouldn't need to resolve extra merge conflicts later since NR+IPR configs changed
- EDIT: to resolve even more future merge conflicts, I rebased on top of https://github.com/nymtech/nym/pull/4464 so that branch also needs to be merged prior
- https://github.com/nymtech/nym/pull/4430 is now completely redundant as `nym-node` can be fully configured with env

what does this PR do?

it's the first grand step (if I can say so myself) towards the project smoosh ™️ 
It completely deprecates standalone mixnode and gateway binaries in favour of the single `nym-node` process. What does it mean in practice?
- devs should no longer be adding features directly to mixnodes/gateways. instead everything should go to nym-node.
- mixnode `init` command has been disabled
- gateway `init` command has been disabled
- all other mixnode and gateway commands prompt you to migrate into `nym-node`
- there exist `nym-node migrate mixnode` and `nym-node migrate gateway` commands that get your existing configuration, data and keys and convert them into a nym-node
- currently `nym-node` runs in one of 3 modes: "mixnode", "entry-gateway" or "exit-gateway" (will be changed later); it means it runs as either a mixnode, a plain gateway or gateway with NR+IPR
- it always runs full blown HTTP API with all endpoints (there's still some cleanup there to do, but it can be done later)
- speaking of HTTP API, all endpoints have full swagger support, you can even pass the required bearer token for metrics `Authorization` header via swagger
- `nym-node`, unlike `mixnode` or `gateway` doesn't have separate `init` and `run` commands. instead if node has already been configured `run` will, well, just run the process. otherwise it will perform first time setup before continuing to run. this should make our docker runners happier. you can experiment with `--deny-init`, `--init-only` and `--write-changes` flags (https://github.com/nymtech/nym/blob/82d5711c8dd3cabf7fe56f88b8f445c17226120a/nym-node/src/cli/commands/run/args.rs#L21-L54) to replicate old behaviour.
- also for docker runners, as briefly mentioned on top, every single cli argument has an equivalent environmental variable. this should make @Bykow happy too
- I've tried to introduce sane defaults for every single argument so it's now possible to run a brand new node with `./nym-node run` and it will just run! (in mixnode mode by default)


common/client-core/config-types/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "nym-client-core-config-types"
```
octol, line 2: Just a personal taste comment: imho it's nice to group related crates in a subdir, but nested crates feels iffy. Just like nested workspaces

common/client-core/config-types/src/old/v2.rs
```diff
@@ -124,9 +137,9 @@ pub struct TrafficV1_1_20 {
     pub secondary_packet_size: Option<PacketSize>,
 }
 
-impl From<TrafficV1_1_20> for TrafficV1_1_20_2 {
-    fn from(value: TrafficV1_1_20) -> Self {
-        TrafficV1_1_20_2 {
+impl From<TrafficV2> for TrafficV3 {
```
octol, line 140: This feels cleaner than before, the naming that is. Nice :)

gateway/src/main.rs
```diff
@@ -52,11 +47,12 @@ async fn main() -> anyhow::Result<()> {
     }
 
     commands::execute(args).await.map_err(|err| {
-        if atty::is(atty::Stream::Stdout) {
+        if std::io::stdout().is_terminal() {
```
octol, line 50: I did not know about this, excellent :)

octol (approved):
This PR excites me!

gateway/src/main.rs
jstuczyn: rust 1.70 ftw : )

common/client-core/config-types/src/old/v2.rs
jstuczyn: Yep. Especially since we stopped properly versioning the client-core crate

common/client-core/config-types/Cargo.toml
jstuczyn: Yeah, I'm not the biggest fan of it but couldn't think of anything much better without other drawbacks

benedettadavico:
Wonderful work here! 🚀
As discussed, the part below is not part of this PR:  
every single cli argument has an equivalent environmental variable

PR #4506, opened 2024-03-27
Switch to IntCounter and Gauge

durch:


octol (approved):

jstuczyn (changes requested):
plz remember to change those back to integers: https://github.com/nymtech/nym/blob/e67d3d816ccdda8e57df6154c286c8cbedf4a5c1/mixnode/src/node/node_statistics.rs#L149-L163

jstuczyn (approved):

PR #4515, opened 2024-04-04
Handle selected gateway in rust sdk

octol:
# Description

CT-121

Rework gateway setup logic in the Rust SDK to take into account the new ability to register multiple gateways and select one as the active one.

For testing: we need to confirm that this doesn't affect the network-requester and ip-packet-router, which both use the rust SDK

### DIFF
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -1,6 +1,8 @@
 // Copyright 2022-2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+use self::storage::helpers::set_active_gateway;
+
 use super::packet_statistics_control::PacketStatisticsReporter;
 use super::received_buffer::ReceivedBufferMessage;
 use super::topology_control::geo_aware_provider::GeoAwareTopologyProvider;
@@ -614,7 +616,21 @@ where
             store_client_keys(keys, key_store).await?;
         }
 
-        setup_gateway(setup_method, key_store, details_store).await
+        let init_result = setup_gateway(setup_method, key_store, details_store).await?;
+
+        // Should we really set active gateway here? Was it an omission to not set it there before?
+        // Or was it omitted on purpose?
+        set_active_gateway(
+            details_store,
+            &init_result
+                .gateway_registration
+                .details
+                .gateway_id()
+                .to_base58_string(),
+        )
+        .await?;
+
+        Ok(init_result)
     }
 
     pub async fn start_base(mut self) -> Result<BaseClient, ClientCoreError>
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -1,8 +1,6 @@
 // Copyright 2022-2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-use self::storage::helpers::set_active_gateway;
-
 use super::packet_statistics_control::PacketStatisticsReporter;
 use super::received_buffer::ReceivedBufferMessage;
 use super::topology_control::geo_aware_provider::GeoAwareTopologyProvider;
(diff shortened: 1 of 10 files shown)


common/client-core/src/client/base_client/mod.rs
```diff
@@ -635,6 +639,21 @@ where
         )
         .await?;
 
+        if is_new_gateway {
+            storage::helpers::set_active_gateway(
```
octol: @jstuczyn

common/client-core/src/client/base_client/mod.rs
```diff
@@ -614,7 +616,21 @@ where
             store_client_keys(keys, key_store).await?;
         }
 
-        setup_gateway(setup_method, key_store, details_store).await
+        let init_result = setup_gateway(setup_method, key_store, details_store).await?;
+
+        // Should we really set active gateway here? Was it an omission to not set it there before?
+        // Or was it omitted on purpose?
+        set_active_gateway(
```
octol: Is this correct? It's not needed for the SDK changes, but it seemed like it should be here? @jstuczyn

jstuczyn: i think there might be situation where you'd want to register with new gateway but not set it as active immediately, it should be up to the caller to decide that imo

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -241,13 +245,15 @@ where
...
-        let client = DisconnectedMixnetClient::new(self.config, self.socks5_config, self.storage)?
-            .custom_gateway_transceiver(self.custom_gateway_transceiver)
-            .custom_topology_provider(self.custom_topology_provider)
-            .custom_shutdown(self.custom_shutdown)
-            .wireguard_mode(self.wireguard_mode)
-            .wait_for_gateway(self.wait_for_gateway)
-            .force_tls(self.force_tls);
+        let mut client =
+            DisconnectedMixnetClient::new(self.config, self.socks5_config, self.storage)?;
+
```
jstuczyn, line 257: any reason for moving away from the builder? i dont mind, just curious

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -438,67 +458,64 @@ where
...
     /// This function will return an error if you try to re-register when in an already registered
     /// state.
-    pub async fn register_and_authenticate_gateway(&mut self) -> Result<()> {
+    pub async fn setup_gateway(&mut self) -> Result<()> {
         if !matches!(self.state, BuilderState::New) {
             return Err(Error::ReregisteringGatewayNotSupported);
         }
 
-        log::debug!("Registering with gateway");
+        self.print_all_registered_gateway_identities().await;
```
jstuczyn, line 484: should this always be printed by default?

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -438,67 +458,64 @@ where
...
+            // When a user-chosen gateway exists and matches the active one.
+            (Some(user_chosen_gateway), Some(active_gateway))
+                if &active_gateway.to_base58_string() == user_chosen_gateway =>
+            {
+                GatewaySetup::MustLoad { gateway_id: None }
+            }
+            // When a user-chosen gateway exists but there's no active gateway, or it doesn't match the active one.
+            (Some(_), _) => self.new_gateway_setup().await?,
+            // When no user-chosen gateway exists but there's an active gateway.
+            (None, Some(_)) => GatewaySetup::MustLoad { gateway_id: None },
```
jstuczyn, line 513: seems you can combine this with the first branch, because you always create a `MustLoad` setup when you have an active gateway.

it sounds like you want to set to `MustLoad` when you either: have an active gateway or the user chosen gateway exists in the storage (but is not necessarily active)

octol: Renamed to set_active_gateway_if_previously_registered above to clarify the flow

jstuczyn (approved):

benedettadavico (approved):

PR #4534, opened 2024-04-10
Explorer NextJS Rebuild

fmtabbara:
# Description

Rebuild the entire Explorer application using Next JS.

The entire app has been rebuilt using the NextJS + App Router. Almost all components have been reused with a few tweaks to some of them. The original frontend API is still being used, but as a further improvement when time allows, this could be refactored to make use of NextJS router handlers.

The hover/open/close desktop nav has been removed - There seemed to be a lot of code for a feature that isn't necessarily required. In desktop mode there is more than enough room to accommodate an "always open" menu.

All tables have been upgraded to use Material React Table https://www.material-react-table.com. This library has a comprehensive built in sorting and filtering system.

Closes: Jira TX-4

# Checklist:


mmsinclair (approved):
🎉

PR #4545, opened 2024-04-17
Phase out curve25519 for x25519 and ed25519

simonwicky:
# Description

Replace `curve25519-dalek` by `ed25519-dalek` and `x25519-dalek`.
Also upgrade those crate, `ed25519` to 2.1 and `rand` to 0.8

WARNING : This PR will break compatibility. Updated used sphinx library uses x25519 scalar clamping vs scalar reduction that is currently done

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4545)

### DIFF
--- a/nym-api/src/coconut/tests/fixtures.rs
+++ b/nym-api/src/coconut/tests/fixtures.rs
@@ -27,11 +27,6 @@ pub fn test_rng(seed: [u8; 32]) -> ChaCha20Rng {
     ChaCha20Rng::from_seed(seed)
 }
 
-pub fn test_rng_07(seed: [u8; 32]) -> rand_chacha_02::ChaCha20Rng {
-    use rand_chacha_02::rand_core::SeedableRng;
-    rand_chacha_02::ChaCha20Rng::from_seed(seed)
-}
-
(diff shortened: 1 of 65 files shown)


nym-api/src/coconut/tests/fixtures.rs
```diff
@@ -48,7 +43,7 @@ pub fn dealer_fixture(mut rng: &mut ChaCha20Rng, id: NodeIndex) -> DealerDetails
     rng.fill_bytes(&mut secondary_seed);
```
jstuczyn: you can remove the above lines btw since we don't need the secondary rng now : )

simonwicky: Removed in https://github.com/nymtech/nym/pull/4545/commits/dc5a640ad596a3c1d39bd61fa63f3cffdf02a436

simonwicky:
This PR needs https://github.com/nymtech/sphinx/pull/94 to be merged in so dependency to sphinx can be updated and properly pull the version from crates.io

simonwicky:
Sphinx's crate new version was released, so this PR is now ready

PR #4569, opened 2024-05-02
Include nym-node endpoint

tommyv1987:
# Description

Fixes the newly updated endpoints for nym-node for the explorer - for now until everyone has updated, use the `/stats` then query for the new `/api/v1/`

### DIFF
--- a/explorer-api/src/mix_node/models.rs
+++ b/explorer-api/src/mix_node/models.rs
@@ -98,25 +98,30 @@ pub(crate) struct NodeDescription {
     pub(crate) location: String,
 }
 
-#[derive(Serialize, Clone, Deserialize, JsonSchema)]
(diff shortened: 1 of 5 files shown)


explorer-api/src/mix_node/models.rs
```diff
@@ -98,25 +98,30 @@ pub(crate) struct NodeDescription {
...
     )]
     update_time: SystemTime,
-
     #[serde(
         serialize_with = "humantime_serde::serialize",
         deserialize_with = "humantime_serde::deserialize"
     )]
     previous_update_time: SystemTime,
 
+    #[serde(alias = "received_since_startup", default)]
```
jstuczyn: imo remove the `default` because it might conceal weird failures as in principle with the alias we should be able to deserialize both responses

tommyv1987:
going to close this in favour of work soon to be completed

PR #4573, opened 2024-05-03
upgraded axum and related deps to the most recent version

jstuczyn:
For testing just make sure all routes still work as they did before.
I don't see why anything would have changed, but code works in mysterious ways...

note: I've also temporarily disabled all libp2p examples due to outdated deps (@mfahampshire  @serinko )


gateway/src/http/mod.rs
```diff
@@ -295,12 +295,21 @@ impl<'a> HttpApiBuilder<'a> {
...
-            .build_server(&self.gateway_config.http.bind_address)?
-            .with_task_client(task_client);
-        tokio::spawn(async move { server.run().await });
+        tokio::spawn(async move {
+            let server = match router.build_server(&bind_address).await {
+                Ok(server) => server.with_task_client(task_client),
+                Err(err) => {
+                    error!("failed to create http server: {err}");
+                    // this will cause global shutdown
+                    drop(task_client);
```
octol: Does it not work without it? Was thinking it would be dropped at the end of the scope anyway?

mixnode/src/node/http/mod.rs
```diff
@@ -104,11 +104,19 @@ impl<'a> HttpApiBuilder<'a> {
...
-            .with_task_client(task_client);
-        tokio::spawn(async move { server.run().await });
+
+        tokio::spawn(async move {
+            let server = match router.build_server(&bind_address).await {
+                Ok(server) => server.with_task_client(task_client),
+                Err(err) => {
+                    error!("failed to create http server: {err}");
+                    // this will cause global shutdown
+                    drop(task_client);
```
octol: Same as other comment?

octol (approved):

gateway/src/http/mod.rs
jstuczyn: good point. I think I just wanted to be explicit to state that creation of the server will cause everything else to shutdown, but I think you're right. it's a bit redundant

PR #4591, opened 2024-05-13
Use rfc3339 for last_polled in described nym-api endpoint

octol:
# Description

Currently there is an issue where the validator-client can't parse the nym-api response for the described endpoint, in particular the `latest_polled` field that was recently added.

Fix this by making the field use rfc3339

NOTE: this will require upgrading nym-api and everything that depends on the described endpoint

This can be tested locally using the sdk examples

```sh
~/src/nym/nym/sdk/rust/nym-sdk
❯ cargo run --example simple
```

```
thread 'main' panicked at sdk/rust/nym-sdk/examples/simple.rs:9:64:
called Result::unwrap() on an Err value: ClientCoreError(ValidatorClientError(NymAPIError { source: ReqwestClientError { source: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, 
```

### DIFF
--- a/nym-api/nym-api-requests/src/models.rs
+++ b/nym-api/nym-api-requests/src/models.rs
@@ -414,7 +414,9 @@ const fn unix_epoch() -> OffsetDateTime {
 
 // for all intents and purposes it's just OffsetDateTime, but we need JsonSchema...
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-pub struct OffsetDateTimeJsonSchemaWrapper(#[serde(default = "unix_epoch")] pub OffsetDateTime);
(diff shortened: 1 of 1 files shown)


jstuczyn (approved):

tommyv1987:
**Testing Completion Date:** 15th May 2024

**Testing Steps Performed:**
1. update a nym-api to the binary built from this branch, then restart the api
2. and check the journalctl for errors message
3. connected via client and could not see the error messages, this is backwards compatible

**Validation:**
- [X] Test Passed

**Notes (if any):**
[Provide any additional observations or relevant information here.]

[Attach any relevant screenshots, logs, or documents.]

PR #4603, opened 2024-05-20
Purge name service and service provider directory contracts

octol:
This is a compiler assisted purge of the `nym-name-service` and `nym-service-provider-directory` contracts that were never deployed on mainnet, and will anyhow be superseded by the new mixnode directory that is being worked on.

It works insofar that it compiles, we need to deploy and test this on non-mainnet before merging in

- Purge `nym-name-service` contract
- Purge `nym-name-service-common`
- Purge `nym-service-provider-directory` contract
- Purge `nym-service-provider-directory-common`
- Remove everywhere name-service contract is used
- Remove everywhere sp contract is used


jstuczyn (approved):
🔥  nice one!

this is more of a testing note: since we removed it everywhere, we need to check that binaries don't start complaining that they're gone. for example anything using the env files (since we've always been putting it there) or the `/api/v1/network` endpoints on our nym API since it was returning that data

tommyv1987:
**Testing Completion Date:** [12/06/2024]

**Testing Steps Performed:**
1. Check references to everything named SP and Ephemera and removed any additional references

**Validation:**
- ✅  Test Passed
- [ 🔴 ] Issues Found 

**Notes (if any):**
N/A

N/A

PR #4605, opened 2024-05-20
Feature/nym node api location

jstuczyn:
# Description

this PR introduces new endpoint on nym-node `/api/v1/auxiliary-details` to expose any additional information. currently its just the location. nym-api will then query all nodes for that information and put it in the self-described endpoint.

@serinko note on how people can add it (those options are already available btw):
1. update the `location` field under `[host]` section of `config.toml`
2. (for new nodes) initialise the node with `--location` flag, where they have to provide the country info. either full country name (e.g. 'Poland'), two-letter alpha2 (e.g. 'PL'), three-letter alpha3 (e.g. 'POL') or three-digit numeric-3 (e.g. '616') can be provided.
3. (for existing nodes) if they don't want to mess with the `config.toml`, they can use exactly the same `--location` argument as above, but they need to make sure to also provide `--write-changes` (or `-w`) flag to persist those changes


benedettadavico:
<node-url>:8080/api/v1/auxiliary-details returns the location information, after it has been added either by `--location` in init, or ar run time alongside `-w`. Or if it is manually changed in the config.toml 

The same info (auxiliary-details) is now also seen in the gateways/described api endpoints

benedettadavico (approved):

PR #4606, opened 2024-05-20
Python CLI to self diagnose Nym nodes

serinko:
# Description

Python CLI for devrel and operators, simplifying diagnose of a any node, especially if ran side by side with `nym-gateway-probe`.

`query_stats` command runs over all possible API checks based on nodes identity key. An example like:
`./node_api_check.py query_stats <IDENTITY_KEY>` will return all stats about any bonded node. 

Optional flags: 
- `--output <PATH>` - export output to a txt file
- `--markdown` - will output stats in markdown format
- `--no_routing_history` - excludes sometimes lengthy routing history from output

`version_count` command returns a table of node summary in the given version. For example `./node_api_check.py version_count <VERSION_1> <VERSION_2> <VERSION_3> --markdown` will return a neat markdown table counting all nodes in version 1, 2, and 3. Use Nym binary versioning (ie `1.1.3`) separated by space.

## Tasks

- [x] Checks whether the nodes is a `mixnode` or a `gateway` or the identity keys doesn't exist
- [x] Queries and prints all the associated endpoint stats
    - [x] pulls IP to run all swagger checks
    - [x] Add check if https/http and associate a port to it
- [x] Output to file for easier debugging
- [x] Version counter 

**Future steps:**

- [ ] Simplify by using nym-node APIs when they are released
- [ ] Replace hardcoded [`api_endpoints.json`](https://github.com/nymtech/nym/blob/feature/node-api-check/scripts/api_endpoints.json) and use endpoints from dynamic [`openapi.json`](https://validator.nymtech.net/api/v1/openapi.json)

### DIFF
--- a/scripts/node_api_check.py
+++ b/scripts/node_api_check.py
@@ -25,23 +25,37 @@ class MainFunctions:
         print(f"Version = {version}")
         if mix_id:
             print(f"Mix ID = {mix_id}")
-        #api_data = self.format_dataframe(api_data)
         print("\n\nNODE RESULTS FROM UNFILTERED QUERY\n")
-        self.print_neat_dict(node_dict)
+        if args.markdown:
+            node_markdown = self._dataframe_to_markdown(node_df, ["RESULT"], ["API EDNPOINT"])
+            print(node_markdown, "\n")
+        else:
+            self.print_neat_dict(node_dict)
         print(f"\n\nNODE RESULTS FROM {self.api_url.upper()}\n")
-        self.print_neat_dict(api_data)
-#        print(node_df.T.to_markdown(), "\n")
-#        print(api_data.to_markdown(), "\n")
+        if args.markdown:
+            api_df = self._json_to_dataframe(api_data)
+            node_markdown = self._dataframe_to_markdown(api_df, ["RESULT"], ["API EDNPOINT"])
+            print(node_markdown, "\n")
+        else:
+            self.print_neat_dict(api_data)
         if swagger_data:
             print(f"\n\nNODE RESULTS FROM SWAGGER PAGE\n")
-            self.print_neat_dict(swagger_data)
+            if args.markdown:
+                swagger_df = self._json_to_dataframe(swagger_data)
+                node_markdown = self._dataframe_to_markdown(swagger_df, ["RESULT"], ["API EDNPOINT"])
+                print(node_markdown, "\n")
+            else:
+                self.print_neat_dict(swagger_data)
         if routing_history:
             print(f"\n\nNODE UPTIME HISTORY\n")
-            self.print_neat_dict(routing_history)
-#            swagger_data = self.format_dataframe(swagger_data)
-#            routing_history = self.format_dataframe(routing_history)
-#            print(swagger_data.to_markdown())
-#            print(routing_history.to_markdown())
+            if args.markdown:
+                routing_history_df = self._json_to_dataframe(routing_history)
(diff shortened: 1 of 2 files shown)


scripts/node_api_check.py
```diff
@@ -0,0 +1,131 @@
...
+            host = str(node_series["gateway_bond.gateway.host"])
+            api_data["API ENPOINTS"] = "RESPONSE"
+            for key in enpoints:
+                endpoint = key.replace("{identity}", identity)
+                url = f"{self.api_url}{endpoint}"
+                value = r.get(url)
+                api_data[endpoint] = value
+            swagger_data["SWAGGER ENDPOINTS"] = "RESPONSE"
+            for key in swagger:
+                swagger_url = f"https://{host}:8080/api/v1{key}"
```
jstuczyn: this will only work if gateway has bonded with a hostname (you can't do `https` over an ip address*). 
for reference, what we're doing in nym-api is this:
```rust

    // first try the standard port in case the operator didn't put the node behind the proxy,
    // then default https (443)
    // finally default http (80)
    let addresses_to_try = vec![
        format!("http://{gateway_host}:{DEFAULT_NYM_NODE_HTTP_PORT}"),
        format!("https://{gateway_host}"),
        format!("http://{gateway_host}"),
    ];
```

\* unless you consider some very special edge cases that would almost definitely not apply here

scripts/node_api_check.py
```diff
@@ -0,0 +1,131 @@
...
+        if mode == "gateway":
+            host = str(node_series["gateway_bond.gateway.host"])
+            api_data["API ENPOINTS"] = "RESPONSE"
+            for key in enpoints:
+                endpoint = key.replace("{identity}", identity)
+                url = f"{self.api_url}{endpoint}"
+                value = r.get(url)
+                api_data[endpoint] = value
+            swagger_data["SWAGGER ENDPOINTS"] = "RESPONSE"
+            for key in swagger:
```
jstuczyn: just to make sure, you're iterating over nym-node endpoints here, right? and **not** nym-api?

scripts/node_api_check.py
```diff
@@ -0,0 +1,131 @@
...
+                swagger_url = f"https://{host}:8080/api/v1{key}"
+                value = r.get(url)
+                swagger_data[key] = value
+
+        elif mode == "mixnode":
+            mix_id = int(node_series["mixnode_details.bond_information.mix_id"])
```
jstuczyn: if the mixnode is running as nym-node you can query for the same endpoints as the gateway

scripts/node_api_check.py
serinko: I will check this out, thank you. Still refining the basis there

scripts/node_api_check.py
serinko: I iterate over validator.nymtech.net/api/v1 and over swagger endpoints for gws as below. Should I add some? My plan is to eventually dump api_json_file and use the online openapi.json as a source.

scripts/node_api_check.py
serinko: Python doesn't have let as far as I am concerned, Fixed it with try-except.

scripts/node_api_check.py
serinko: you mean swagger endpoint?

serinko:
I found out that it returns error when running over a blacklisted GW, This is a crucial problem, I will get back to it this week again.

benedettadavico (approved):

PR #4607, opened 2024-05-21
Update crypto and rand crates

simonwicky:
# Description

Update sphinx crate to `0.1.1` along with 25519 crates and `rand` crates


simonwicky:
⚠️  This PR contains a test failure due to the update here : https://github.com/nymtech/nym/blob/b4a0487a41375167b2f481c00917b957b9f89789/common/crypto/src/asymmetric/encryption/mod.rs#L353-L358

This is due a change in `x25519-dalek` from `1.1.1` to `2`. 
Crypto operations should be identical, but the byte representation has changed (sphinx clamps at creation, x25519 clamps at use). This cannot be changed in the sphinx crate without breaking changes. 
There is a good chance that this failure doesn't impact anything else, but it has to be tested to see. 
A mix of old and new clients with a mix of old and new mixnodes should do

jstuczyn (approved):

PR #4610, opened 2024-05-22
New Network Monitor

durch:
# Description

Monitors the Nym network by sending itself packages across the mixnet.

Network monitor is running two tokio tasks, one manages mixnet clients and another manages monitoring itself. Monitor is designed to be driven externally, via an HTTP api. This means that it does not do any monitoring unless driven by something like [`locust`](https://locust.io/). This allows us to tailor the load externally, potentially distributing it across multiple monitors.

Includes a dockerised setup for automatically spinning up monitor and driving it with locust.


nym-network-monitor/src/http.rs
```diff
@@ -0,0 +1,106 @@
...
+                info!("received shutdown");
+            }
+            res = server_future => {
+                info!("the http server has terminated");
+                if let Err(err) = res {
+                    info!("with the following error: {err}");
+                    return Err(err.into())
+                }
+            }
+        }
```
dynco-nym: This looks like a forced shutdown of axum server: when `shutdown_future` branch of `select` completes, other branch (`server_future`) is cancelled. 

Besides, there is a more ergonomic way to use a cancellation token future with Axum server, where you don't need to manually "cancel" server branch after you await the cancellation token:

```rust
axum::serve(listener, app).with_graceful_shutdown(self.cancel.cancelled())
```


`.with_graceful_shutdown` accepts a future which, when resolved, starts graceful shutdown. 

`CancellationToken::cancelled()` offers just that kind of future so they work hand in hand. 

This way, you just issue a cancel signal on the token, and `tokio` will signal axum server to initiate a (graceful) shutdown.

durch: Nice one!

nym-network-monitor/src/http.rs
```diff
@@ -81,25 +80,15 @@ impl HttpServer {
             .route("/v1/received", get(recv_handler));
         let listener = tokio::net::TcpListener::bind(self.listener).await?;
 
-        let shutdown_future = self.cancel.cancelled();
-        let server_future = axum::serve(listener, app).into_future();
+        let server_future = axum::serve(listener, app).with_graceful_shutdown(async move {
+            self.cancel.cancelled().await;
+        });
```
dynco-nym: suggestion:
```rust
let server_future =
            axum::serve(listener, app).with_graceful_shutdown(self.cancel.cancelled_owned());
```

nym-network-monitor/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    info!("Waiting for message (ctrl-c to exit)");
+
+    ctrl_c().await?;
+    info!("Received kill signal, shutting down, submitting final batch of metrics");
+
+    submit_metrics().await?;
+
+    cancel_token.cancel();
+    Ok(())
```
dynco-nym: Sequentially, what happens here is:
- capture interrupt
- submit metrics
- cancel token -> sends shutdown signal to background task running axum server
- immediately return from main

Returning from `main` here [shuts down the runtime, which cancels all background tasks](https://docs.rs/tokio/latest/tokio/task/#cancellation). This means there is almost no time for cancel signal to be respected by axum server and it may be shutting down forcefully.

If you care about this, you could await on server's `JoinHandle` after giving it a shutdown signal (with a possible time out)

All of this is under assumption you care about shutting it down gracefully. Maybe you don't care about what happens after
```rust
submit_metrics().await?;
```
in which case you can disregard this. 

But then, unless I'm mistaken, signalling via cancellation token is pointless - you can just spawn axum task and forget about it - it will get cancelled anyway when main terminates.

durch: All good comments, will address them :)

common/client-core/src/client/packet_statistics_control.rs
```diff
@@ -486,7 +486,7 @@ impl PacketStatisticsControl {
             // Check what the number of retransmissions was during the recording window
             if let Some((_, start_stats)) = self.history.front() {
                 let delta = self.stats.clone() - start_stats.clone();
-                log::info!(
+                log::debug!(
```
octol, line 489: Nice, I've been meaning to bump these down to debug, and instead emit events that the vpn client can listen for so we don't pollute the log output

common/nymsphinx/chunking/src/lib.rs
```diff
@@ -22,6 +27,118 @@ pub mod fragment;
...
+    pub fn header(&self) -> FragmentHeader {
+        self.header.clone()
+    }
+
+    pub fn at(&self) -> u64 {
+        self.at
+    }
+}
+
+lazy_static::lazy_static! {
```
octol: nit: consider using `std::cell::OnceCell`

common/topology/src/lib.rs
```diff
@@ -116,13 +118,40 @@ impl Display for NetworkAddress {
...
     gateways: Vec<gateway::Node>,
 }
 
 impl NymTopology {
+    pub async fn new_from_env() -> Result<Self, NymTopologyError> {
+        let api_url = std::env::var(NYM_API)?;
+
+        info!("Generating topology from {}", api_url);
+
+        let mixnodes = reqwest::get(&format!("{}/v1/mixnodes", api_url))
```
octol, line 133: Probably a too heavy dependency to pull in nym-validator-client, but still, bare string endpoints seems like it could be fragile?

octol (approved):
Looks very nice :)
Possible Andrew wants to have a peek at the sphinx changes?

common/topology/src/lib.rs
durch: Yes, I was thinking through this one, I might refactor the api endpoints out, so they're lighter to pull in

common/nymsphinx/chunking/src/lib.rs
jstuczyn: yeah, with current changes to rust, you no longer need to use `lazy_static`, in 99.9% cases `OnceLock`/`LazyLock` (or the `Cell` variants if you don't need thread-safety) are sufficient

common/client-libs/validator-client/Cargo.toml
```diff
@@ -90,4 +90,6 @@ default = ["http-client"]
 http-client = ["cosmrs/rpc"]
 generate-ts = []
 contract-testing = ["nym-mixnet-contract-common/contract-testing"]
-
+# Features below are added to make clippy happy, it seems like they're unused we should remove them
+tendermint-rpc-http-client = ["tendermint-rpc/http-client"]
```
jstuczyn, line 94: but I'm not 100% sure if adding those features will break the wasm client build. I don't remember if wasm didn't like the types or just the actual client (when used)

common/nymsphinx/chunking/src/fragment.rs
```diff
@@ -324,6 +338,20 @@ pub(crate) struct FragmentHeader {
 }
 
 impl FragmentHeader {
+    pub fn seed(&self) -> i32 {
+        let mut seed = self.id;
+        seed = seed.wrapping_mul(self.total_fragments as i32);
+        seed = seed.wrapping_mul(self.current_fragment as i32);
```
jstuczyn, line 344: I don't remember from the top of my head, but aren't we counting the fragments from 0, meaning for bunch of `FragmentHeader`, `seed()` will return 0?

common/nymsphinx/chunking/src/lib.rs
```diff
@@ -22,6 +27,118 @@ pub mod fragment;
...
+    pub static ref FRAGMENTS_RECEIVED: DashMap<i32, Vec<ReceivedFragment>> = DashMap::new();
+    pub static ref FRAGMENTS_SENT: DashMap<i32, Vec<SentFragment>> = DashMap::new();
+}
+
+#[macro_export]
+macro_rules! now {
+    () => {
+        match std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH) {
```
jstuczyn, line 123: nitpicking: we're using `OffsetDateTime` for all time-keeping everywhere in the codebase, could we perhaps use that here too?

common/nymsphinx/chunking/src/lib.rs
```diff
@@ -22,6 +27,118 @@ pub mod fragment;
...
+    }
+
+    pub fn at(&self) -> u64 {
+        self.at
+    }
+}
+
+lazy_static::lazy_static! {
+    pub static ref FRAGMENTS_RECEIVED: DashMap<i32, Vec<ReceivedFragment>> = DashMap::new();
+    pub static ref FRAGMENTS_SENT: DashMap<i32, Vec<SentFragment>> = DashMap::new();
```
jstuczyn: are we not possibly risking OOM-ing if the client runs for long enough? it doesn't look like we're clearing those maps

common/nymsphinx/src/preparer/mod.rs
```diff
@@ -192,9 +196,18 @@ pub trait FragmentPreparer {
         packet_type: PacketType,
         mix_hops: Option<u8>,
     ) -> Result<PreparedFragment, NymTopologyError> {
+        debug!("Preparing chunk for sending");
         // each plain or repliable packet (i.e. not a reply) attaches an ephemeral public key so that the recipient
         // could perform diffie-hellman with its own keys followed by a kdf to re-derive
         // the packet encryption key
+
+        let seed = fragment.seed().wrapping_mul(self.nonce());
+        let mut rng = ChaCha8Rng::seed_from_u64(seed as u64);
```
jstuczyn: I'm a bit afraid of rather low entropy of this `rng`, especially if "normal" client packets return constant `1` `nonce` -> could we make that one also return a proper pseudorandom value instead?

common/topology/src/lib.rs
jstuczyn: it feels like those functionalities shouldn't live in the `common/topology` crate, because it shouldn't really be doing any network calls. imo it should be in separate crate.
maybe we could even define some `trait TopologyExt` to have those methods directly on the type should we really want it

common/types/src/monitoring.rs
```diff
@@ -0,0 +1,122 @@
+use std::{collections::HashSet, sync::LazyLock, time::SystemTime};
+
+use nym_crypto::asymmetric::identity::{PrivateKey, PublicKey, Signature};
+use nym_mixnet_contract_common::MixId;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+static NETWORK_MONITORS: LazyLock<HashSet<String>> = LazyLock::new(|| {
+    let mut nm = HashSet::new();
+    nm.insert("5VsPyLbsBCq9PAMWmjKkToteVAKNabNqex6QwDf5fWzt".to_string());
```
jstuczyn, line 10: what's with this hardcoded value here?

common/types/src/monitoring.rs
```diff
@@ -0,0 +1,122 @@
...
+            identity,
+            owner,
+            reliability,
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+pub struct GatewayResult {
+    pub identity: String,
```
jstuczyn, line 42: it will save me a lot of headache in rebasing if you also included `mix_id` / `node_id` here

maybe just have a 

```rust
pub struct NodeResult {
    pub node_id: MixId,
    pub identity: String,
    pub owner: String,
    pub reliability: u8,
}
```

and it would just temporarily be `0` for gateways : )

because the idea is to unify gateways and mixnodes and remove as many "node-specific" features as possible

common/types/src/monitoring.rs
```diff
@@ -0,0 +1,122 @@
...
+        let mut msg = match serde_json::to_vec(results) {
+            Ok(msg) => msg,
+            Err(_) => return Vec::new(),
+        };
+        msg.extend_from_slice(&timestamp.to_le_bytes());
+        msg
+    }
+
+    pub fn timely(&self) -> bool {
+        let now = SystemTime::now()
```
jstuczyn, line 85: nitpick: same comment about time-keeping

common/types/src/monitoring.rs
```diff
@@ -0,0 +1,122 @@
...
+    signature: String,
+    signer: String,
+    timestamp: i64,
+}
+
+impl MonitorMessage {
+    fn message_to_sign(results: &MonitorResults, timestamp: i64) -> Vec<u8> {
+        let mut msg = match serde_json::to_vec(results) {
+            Ok(msg) => msg,
+            Err(_) => return Vec::new(),
```
jstuczyn: if serialisation fails, you still want to attach the timestamp so that somebody couldn't replay your signature on an empty message

common/types/src/monitoring.rs
```diff
@@ -0,0 +1,122 @@
...
+        MonitorMessage {
+            results,
+            signature: signature.to_base58_string(),
+            signer: public_key.to_base58_string(),
+            timestamp,
+        }
+    }
+
+    pub fn from_allowed(&self) -> bool {
+        NETWORK_MONITORS.contains(&self.signer)
```
jstuczyn, line 112: should those be maybe defined in env/config?

nym-api/migrations/20240524101449_monitor_v2_tables.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE mixnode_details_v2
+(
+    id       INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    mix_id   INTEGER NOT NULL UNIQUE,
+    owner    VARCHAR NOT NULL,
```
jstuczyn, line 5: mmm. future merge conflicts : )

if you want to make my life easier, could you add `node_id` to `gateway_details` and remove `owner` column from both gateways and mixnodes?
if that's too much work on your end (because it's all over the place), don't worry. I've already handled migration from the old `mixnode_details` and `gateway_details` so re-applying them shouldn't be too much of a problem

nym-network-monitor/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+use tokio::sync::OnceCell;
+use tokio::{signal::ctrl_c, sync::RwLock};
+use tokio_util::sync::CancellationToken;
+
+static NYM_API_URL: LazyLock<String> = LazyLock::new(|| {
+    std::env::var(NYM_API).unwrap_or_else(|_| panic!("{} env var not set", NYM_API))
+});
+
+static MIXNET_TIMEOUT: OnceCell<u64> = OnceCell::const_new();
+static TOPOLOGY: OnceCell<NymTopology> = OnceCell::const_new();
```
jstuczyn, line 30: don't you need to periodically change it as new nodes bond/old ones unbond?

common/nymsphinx/chunking/src/fragment.rs
durch: Oh crap, amazing find, I'm gonna add a `+1` in there somewhere

common/nymsphinx/chunking/src/lib.rs
durch: Not sure if we have the `time` crate in there, I'll change it if we have it

common/nymsphinx/chunking/src/lib.rs
durch: We're not, this is me being lazy for now, I'll see if I can add it somewhere lazyly

nym-network-monitor/src/main.rs
durch: The idea is that monitor is a short lived process (shorter then the epoch) that bootstraps its topology every run

common/nymsphinx/chunking/src/fragment.rs
durch:
```
/// ID associated with `FragmentSet` to which this particular `Fragment` belongs.
/// Its value is restricted to (0, i32::MAX].
/// Note that it *excludes* 0, but *includes* i32::MAX.
/// This allows the field to be represented using 31 bits.
```

We're good here after all

common/nymsphinx/chunking/src/lib.rs
durch: We're not pulling in time, I like the std lib way of going about it, I'd rather leave it in, maybe migrate towards SystemTime as we go forward, one less dependency :)

common/nymsphinx/chunking/src/lib.rs
durch: Went with `LazyLock`, so that it is initialised on first access

common/nymsphinx/chunking/src/lib.rs
durch: Done

common/topology/src/lib.rs
durch: Just going to pull in the routes, no sense in complicating things

dynco-nym (approved):
so graceful now, approved :)

common/types/src/monitoring.rs
durch: Thats the current public key of the network monitor that is allowed to submit stuff

PR #4612, opened 2024-05-23
v1.77 Update ci-contracts-upload-binaries.yml

benedettadavico:


tommyv1987:
@benedettadavico do you need this now that it's in develop?

PR #4614, opened 2024-05-23
update mixfetch docs

mfahampshire:
* fixed broken examples with new infra addresses
* added info on gateway filtering via habourmaster


tommyv1987 (approved):
LGTM

PR #4615, opened 2024-05-23
Refine wireguard on gateway side

neacsu:
# Description

Prepare wireguard gateway for future work on the registration of peers.


gateway/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "nym-gateway"
 license = "GPL-3.0"
-version = "1.1.36"
+version = "1.1.37"
```
octol: I guess we just need to remember to not bump again during the release process

gateway/src/error.rs
```diff
@@ -63,6 +63,17 @@ pub enum GatewayError {
...
     },
 
+    #[error(
+        "failed to load config file for wireguard (gateway-id: '{id}') using path '{}'. detailed message: {source}",
+        path.display()
+    )]
+    WireguardConfigLoadFailure {
+        id: String,
+        path: PathBuf,
+        #[source]
```
octol: if the field is named `source` then you don't need the attribute

octol (approved):

gateway/Cargo.toml
neacsu: I can leave it to 36 actually, initially I made the change because I was putting the update procedure in the gateway, but now that the PR is focused on nym-node, it can just stay to 36

gateway/src/error.rs
neacsu: Good point. Then I should probably remove the attribute for the other error types as well.

common/types/src/gateway.rs
```diff
@@ -171,3 +171,25 @@ impl fmt::Display for GatewayIpPacketRouterDetails {
         writeln!(f, "\taddress: {}", self.address)
     }
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GatewayWireguardDetails {
```
jstuczyn: this doesn't seem to be used anywhere? what's the point of that then?

gateway/src/config/mod.rs
```diff
@@ -85,10 +85,6 @@ pub struct Config {
 
     pub gateway: Gateway,
 
-    // currently not really used for anything useful
-    #[serde(default)]
```
jstuczyn, line 89: because we have `#[serde(deny_unknown_fields)]` on that struct, after you remove `pub wireguard: Wireguard,` all gateways that have had this section in their config files will fail to load. you might have to either write a config migration or maybe we could somehow expedite the process of removing standalone gateway binaries to bypass that issue completely

gateway/src/node/mod.rs
```diff
@@ -565,21 +571,28 @@ impl<St> Gateway<St> {
...
+
         if let Err(source) = shutdown.wait_for_shutdown().await {
             // that's a nasty workaround, but anyhow errors are generally nicer, especially on exit
             return Err(GatewayError::ShutdownFailure { source });
         }
         #[cfg(all(feature = "wireguard", target_os = "linux"))]
-        if let Some(wg_api) = wg_api {
-            defguard_wireguard_rs::WireguardInterfaceApi::remove_interface(&wg_api)?;
-        }
+        defguard_wireguard_rs::WireguardInterfaceApi::remove_interface(&wg_api)?;
```
jstuczyn: you might have to create some temporary struct here with a `Drop` implementation that would wrap around `wg_api` and that removes the wireguard interface.
i.e. something like
```rust
struct Foo {
wg_api: WhateverType
}

impl Drop for Foo {
  fn drop(&mut self) {
     if let Err(err) = defguard_wireguard_rs::WireguardInterfaceApi::remove_interface(&self.wg_api) {
       // I guess log it or something?
    }
  }
}
```

otherwise if we experience any failures before getting here, like if we fail to catch shutdown: `if let Err(source) = shutdown.wait_for_shutdown().await {`, we'd exit before removing the interface which I doubt is desirable

nym-node/src/cli/commands/migrate.rs
```diff
@@ -606,6 +597,8 @@ async fn migrate_gateway(mut args: Args) -> Result<(), NymNodeError> {
         .await?;
     }
 
+    crate::node::WireguardData::initialise(&config.wireguard)?;
```
jstuczyn, line 602: we should also call this inside `migrate_mixnode(...)`, because it seems the function just generates new keypair. 
and we want `nym-node` to always have all the data it needs for any mode.

nym-node/src/config/persistence.rs
```diff
@@ -366,11 +370,23 @@ impl ExitGatewayPaths {
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct WireguardPaths {
-    // pub keys:
+    pub private_diffie_hellman_key_file: PathBuf,
```
jstuczyn, line 373: since we now introduced new keys and changes in nym-node config. we might have to write some migrations in `try_upgrade_config` (https://github.com/nymtech/nym/blob/4c5351ba60fd63b595e4da64d77dad5e3a0a5ab8/nym-node/src/config/upgrade_helpers.rs#L9). something like we already had for mixnodes and gateways. 
existing nodes should generate the missing keys and update the config structure

common/wireguard/src/lib.rs
```diff
@@ -3,40 +3,37 @@
...
+
+    let mut peers = vec![];
+    for peer_client in wireguard_data.client_registry().iter() {
+        let mut peer = Peer::new(Key::new(peer_client.pub_key.to_bytes()));
+        let peer_ip_mask = IpAddrMask::new(peer_client.private_ip, 32);
+        peer.set_allowed_ips(vec![peer_ip_mask]);
+        peers.push(peer);
+    }
 
     let ifname = String::from("wg0");
```
jstuczyn: can we guarantee `"wg0"` will always be available to use. should we have some fallback?

common/types/src/gateway.rs
neacsu: At some point I wanted to include it in the gateway details, but since it's now only in nym-node, maybe it doesn't make sense anymore. If we want to include it later, we can re-add it, for now I'll just delete this code.

common/wireguard/src/lib.rs
neacsu: I've used a nym specific one, so there shouldn't be any problems now.

gateway/src/config/mod.rs
neacsu: I had the impression I didn't need to write the upgrade, didn't notice that attribute. I'll push one now

gateway/src/node/mod.rs
neacsu: Yeah, I was looking now at this code and was thinking of something similar for the next PR. But since you've mentioned it, no reason to not do it here. I'll push a commit for it

jstuczyn (approved):

PR #4616, opened 2024-05-24
Nym API endpoint to submit monitoring results

durch:


nym-api/src/node_status_api/routes.rs
```diff
@@ -29,6 +30,24 @@ use crate::node_status_api::models::ErrorResponse;
 use crate::storage::NymApiStorage;
 use crate::NymContractCache;
 
+#[openapi(tag = "status")]
+#[post("/submit", data = "<results>")]
+pub(crate) async fn submit_monitoring_results(
+    results: Json<Vec<MixnodeResult>>,
+    storage: &State<NymApiStorage>,
+) -> Result<(), ErrorResponse> {
+    match storage.manager.submit_mixnode_statuses_v2(results.0).await {
```
jstuczyn: should we have some bearer token or something to ensure random people can't pollute the database and we only accept the data from our trusted agents?

durch: Yes, we'll do the whole register with hmac thing we also have/had for wg

PR #4620, opened 2024-05-28
Handle v6 and v7 requests in the IPR, but reply with v6

octol:
Teach the IPR to read both v6 and v7 requests, but always reply with v6. This is to prepare for bumping to v7 and signed connect/disconnect messages.

Follow up PRs will add
- Verify signature
- Send v7 in client with signatures included

### DIFF
--- a/common/ip-packet-requests/src/v7/request.rs
+++ b/common/ip-packet-requests/src/v7/request.rs
@@ -1,10 +1,14 @@
+use nym_crypto::asymmetric::identity;
 use nym_sphinx::addressing::clients::Recipient;
 use serde::{Deserialize, Serialize};
 use time::OffsetDateTime;
 
 use crate::{make_bincode_serializer, IpPair};
 
-use super::VERSION;
+use super::{
+    signature::{SignatureError, SignedRequest},
+    VERSION,
+};
 
 fn generate_random() -> u64 {
     use rand::RngCore;
@@ -236,6 +240,25 @@ pub struct SignedStaticConnectRequest {
     pub signature: Option<Vec<u8>>,
 }
 
+impl SignedRequest for SignedStaticConnectRequest {
+    fn identity(&self) -> &identity::PublicKey {
+        self.request.reply_to.identity()
+    }
+
+    fn request(&self) -> Result<Vec<u8>, SignatureError> {
+        self.request
+            .to_bytes()
+            .map_err(|error| SignatureError::RequestSerializationError {
+                message: "failed to serialize request to binary".to_string(),
+                error,
+            })
+    }
+
+    fn signature(&self) -> Option<&Vec<u8>> {
+        self.signature.as_ref()
+    }
+}
+
 // A dynamic connect request is when the client does not provide the internal IP address it will use
 // on the ip packet router, and instead requests one to be assigned to it.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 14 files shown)


octol:
For testing: deploy exit gateway from this PR and check that old vpn clients can still connect to it as exit gateway

common/ip-packet-requests/src/v7/request.rs
```diff
@@ -298,6 +350,29 @@ pub struct SignedDisconnectRequest {
     pub signature: Option<Vec<u8>>,
 }
 
+impl SignedRequest for SignedDisconnectRequest {
+    fn identity(&self) -> &identity::PublicKey {
+        self.request.reply_to.identity()
```
jstuczyn, line 355: while it's fine now, i feel it will be a problem in the long-term when this field goes away due to our favourite SURBs

common/ip-packet-requests/src/v7/signature.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn request(&self) -> Result<Vec<u8>, SignatureError>;
+
+    fn signature(&self) -> Option<&Vec<u8>>;
+
+    fn timestamp(&self) -> time::OffsetDateTime;
+
+    fn verify(&self) -> Result<(), SignatureError> {
+        if let Some(signature) = self.signature() {
+            // First check that the request is recent enough
+            if time::OffsetDateTime::now_utc() - self.timestamp() > Duration::from_secs(10) {
```
jstuczyn: is 10s enough?

common/ip-packet-requests/src/v7/signature.rs
```diff
@@ -0,0 +1,66 @@
...
+                return Err(SignatureError::RequestOutOfDate);
+            }
+
+            let request_as_bytes = self.request()?;
+            let signature = identity::Signature::from_bytes(signature).map_err(|error| {
```
jstuczyn: why not just attach the proper `Signature` type on the request directly and avoid the parsing? iirc it has all the `serde` stuff already

common/ip-packet-requests/src/v7/signature.rs
```diff
@@ -0,0 +1,66 @@
...
+                SignatureError::SignatureParseError {
+                    message: "failed to parse signature".to_string(),
+                    error,
+                }
+            })?;
+
+            self.identity()
+                .verify(request_as_bytes, &signature)
```
jstuczyn: does the `request_as_bytes` also include the timestamp?

common/ip-packet-requests/src/v7/request.rs
octol: For sure, once we support SURBs a lot of stuff will need to be updated and changed

common/ip-packet-requests/src/v7/signature.rs
octol: Yep it should be included

common/ip-packet-requests/src/v7/signature.rs
octol: Good point, let me investigate and get back to you

octol: Updated!

common/ip-packet-requests/src/v7/signature.rs
octol: I'm thinking that if the latency between the client and IPR is anywhere close to 10 sec, IP packets over mixnet will not be usable anyway

jstuczyn (approved):

PR #4622, opened 2024-05-28
Update kernel peers on peer modification

neacsu:


common/wireguard-types/Cargo.toml
```diff
@@ -13,13 +13,16 @@ license.workspace = true
 [dependencies]
 base64 = { workspace = true }
 dashmap = { workspace = true }
+futures = { workspace = true }
 log = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 thiserror = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "net", "signal"] } # async runtime
```
jstuczyn: I'm just judging based on the crate name(I haven't reviewed this PR, just looked at CI errors) - this import probably causes the wasm-sdk to fail to build since it can't use tokio runtime.
You might probably want to leave this crate with types only and move anything extra to another crate

neacsu: Removing the features was enough to make it build for wasm

nym-node/nym-node-http-api/src/router/api/v1/gateway/client_interfaces/wireguard/mod.rs
```diff
@@ -214,11 +219,11 @@ mod test {
         assert_eq!(response.status(), StatusCode::OK);
         assert!(!registration_in_progress.is_empty());
 
-        let ClientRegistrationResponse::PendingRegistration {
+        let ClientRegistrationResponse::PendingRegistration(RegistrationData {
             nonce,
             gateway_data,
             wg_port: 8080,
-        } = serde_json::from_slice(&to_bytes(response.into_body(), usize::MAX).await.unwrap())
+        }) = serde_json::from_slice(&to_bytes(response.into_body(), usize::MAX).await.unwrap())
```
octol, line 226: Do these two unwraps make it possible to crash the gateway with a malformed request?

neacsu: This is inside a unit test so no problem of crash here

nym-node/nym-node-http-api/src/router/api/v1/gateway/client_interfaces/wireguard/client_registry.rs
```diff
@@ -11,33 +11,59 @@ use axum::extract::{Path, Query, State};
...
 
+fn remove_from_registry(
+    state: &WireguardAppStateInner,
+    remote_public: &PeerPublicKey,
+    gateway_client: &GatewayClient,
+) -> Result<(), RequestError> {
+    state
+        .wireguard_gateway_data
+        .remove_peer(gateway_client)
+        .map_err(|err| RequestError::from_err(err, StatusCode::INTERNAL_SERVER_ERROR))?;
```
jstuczyn, line 27: are we sure we want to be returning the error immediately here? because if we fail to remove the peer for some weird reason, the public key will still be stuck in the `client_registry`

jstuczyn (approved):

nym-node/nym-node-http-api/src/router/api/v1/gateway/client_interfaces/wireguard/client_registry.rs
neacsu: Once the `client_registry` will be made persistent, we will have an error path there as well, so we'd need to decide which one we regard as a source of truth, so that it can be the first operation. Right now, my thinking is that, since kernel data is the only one that can fail, if it's not removed from that list, there's no reason to remove it from the registry either, in case the weird reason disappears later.
tl;dr I agree the removal error case needs to be carefully ordered here, but for now, as `client_registry` doesn't do much, it's not so important.

tommyv1987:
Testing Completion Date: [12/06/2024]

Testing Steps Performed:

This is a part of a bigger ticket, but initial testing has proven to shown that launching nym-nodes (entry and exit gateways) in WG enable mode to be working
I will use this template for the other related WG tickets in this release milestone.
#4636 

Validation:

✅ Test Passed
[ 🔴 ] Issues Found
Notes (if any):
N/A

Attachments (if applicable):
N/A

PR #4634, opened 2024-06-10
impl AsyncRead, AsyncWrite for native MixnetClient

durch:
Heavily inspired by what @mfahampshire did for mix-tcp, folds AsyncRead functionality into the MixnetClient.

TODO:

- [x] Sink
- [x] AsyncWrite
- [ ] Add examples

Interesting changes in `wasm`, due to `Sink`, we now need `&mut` for sending, so had to wrap stuff in `RwLock` in `wasm` client and `mixfetch`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4634)

### DIFF
--- a/common/client-core/src/client/inbound_messages.rs
+++ b/common/client-core/src/client/inbound_messages.rs
@@ -1,4 +1,3 @@
-use hyper::body::Buf;
 // Copyright 2020-2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 use nym_sphinx::addressing::clients::Recipient;
@@ -8,6 +7,7 @@ use nym_sphinx::params::PacketType;
 use nym_task::connections::TransmissionLane;
 use serde::{Deserialize, Serialize};
 use tokio_util::{
+    bytes::Buf,
     bytes::BytesMut,
     codec::{Decoder, Encoder},
 };
--- a/common/client-core/src/client/inbound_messages.rs
+++ b/common/client-core/src/client/inbound_messages.rs
@@ -1,16 +1,20 @@
+use hyper::body::Buf;
 // Copyright 2020-2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
-
 use nym_sphinx::addressing::clients::Recipient;
 use nym_sphinx::anonymous_replies::requests::AnonymousSenderTag;
 use nym_sphinx::forwarding::packet::MixPacket;
 use nym_sphinx::params::PacketType;
 use nym_task::connections::TransmissionLane;
+use serde::{Deserialize, Serialize};
+use tokio_util::{bytes::BytesMut, codec::{Decoder, Encoder}};
+
+use crate::error::ClientCoreError;
 
 pub type InputMessageSender = tokio_util::sync::PollSender<InputMessage>;
 pub type InputMessageReceiver = tokio::sync::mpsc::Receiver<InputMessage>;
 
-#[derive(Debug)]
+#[derive(Serialize, Deserialize, Debug)]
 pub enum InputMessage {
     /// Fire an already prepared mix packets into the network.
     /// No guarantees are made about it. For example no retransmssion
@@ -196,5 +200,48 @@ impl InputMessage {
             | InputMessage::Premade { lane, .. } => lane,
             InputMessage::MessageWrapper { message, .. } => message.lane(),
         }
+    }
+}
+
+
+// TODO: Tests
+pub struct InputMessageCodec;
+
+impl Encoder<InputMessage> for InputMessageCodec {
+    type Error = ClientCoreError;
+
+    fn encode(&mut self, item: InputMessage, buf: &mut BytesMut) -> Result<(), Self::Error> {
+        let encoded = bincode::serialize(&item).expect("failed to serialize InputMessage");
+        let encoded_len = encoded.len() as u32;
+        let mut encoded_with_len = encoded_len.to_le_bytes().to_vec();
+        encoded_with_len.extend(encoded);
+        buf.reserve(encoded_with_len.len());
+        buf.extend_from_slice(&encoded_with_len);
+        Ok(())
+    }
+}
+
+impl Decoder for InputMessageCodec {
+    type Item = InputMessage;
+    type Error = ClientCoreError;
+
+    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if buf.len() < 4 {
+            return Ok(None);
+        }
+
+        let len = u32::from_le_bytes([buf[0], buf[1], buf[2], buf[3]]) as usize;
+        if buf.len() < len + 4 {
+            return Ok(None);
+        }
+
+        let decoded = match bincode::deserialize(&buf[4..len]) {
+            Ok(decoded) => decoded,
+            Err(_) => return Ok(None)
+        };
+
+        buf.advance(len + 4);
+
+        Ok(Some(decoded))
     }
 }
--- a/common/client-core/src/client/inbound_messages.rs
+++ b/common/client-core/src/client/inbound_messages.rs
@@ -7,7 +7,10 @@ use nym_sphinx::forwarding::packet::MixPacket;
 use nym_sphinx::params::PacketType;
 use nym_task::connections::TransmissionLane;
 use serde::{Deserialize, Serialize};
-use tokio_util::{bytes::BytesMut, codec::{Decoder, Encoder}};
+use tokio_util::{
+    bytes::BytesMut,
+    codec::{Decoder, Encoder},
+};
 
 use crate::error::ClientCoreError;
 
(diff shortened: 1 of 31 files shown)


durch:
We can keep this one open for the other changes (AsyncWrite + Sink) as well, it will be easier to polish wholesale

jstuczyn:
it's definitely not in the scope for this feature, but I wonder if conceptually it would make sense for `AsyncRead` and `AsyncWrite` to completely bypass message reconstructing and operate on raw bytes instead where it's up to the caller to deal with them as they see fit

common/client-core/src/client/inbound_messages.rs
```diff
@@ -197,4 +208,50 @@ impl InputMessage {
             InputMessage::MessageWrapper { message, .. } => message.lane(),
         }
     }
+
+    pub fn serialized_size(&self) -> u64 {
+        bincode::serialized_size(self).expect("failed to get serialized InputMessage size") + 4
```
jstuczyn: maybe it's because it's early, but it took me a while to figure out why we're adding the "4" here. could you add a comment for future reference that it's length prefixed encoding?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -190,6 +214,44 @@ impl MixnetClient {
         // note: it's important to take ownership of the struct as if the shutdown is `TaskHandle::External`,
         // it must be dropped to finalize the shutdown
     }
+
+    // fn read_buffer_to_slice(
```
jstuczyn: what's with this?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -51,6 +56,24 @@ pub struct MixnetClient {
 
     // internal state used for the `Stream` implementation
     _buffered: Vec<ReconstructedMessage>,
+
+    // internal state used for the `AsyncRead` implementation
+    _read: ReadBuffer,
```
jstuczyn, line 61: is it "safe" to have separate states for both `Stream` and `AsyncRead`? 
what would have happened if somebody attempted to use them interchangeably? 
i.e. called `next()` few times followed by `read()`. would we have corrupted client's data and lost messages? if so, should we perhaps put some safeguards for that?

wasm/client/src/helpers.rs
```diff
@@ -48,10 +50,11 @@ pub(crate) trait InputSender {
     fn send_messages(&self, messages: Vec<InputMessage>) -> Promise;
 }
 
-impl InputSender for Arc<ClientInput> {
+impl InputSender for Arc<RwLock<ClientInput>> {
```
jstuczyn, line 53: for sanity check, have you actually tried to run the wasm code with those changes?
imo the simplest way, without dealing with bundling et al. would be to go to `wasm/client`, do `make build-debug-dev` and finally run the `internal-dev` project. though I'm not 100% sure how rotten it is at this point, so you might have to make tiny changes in `worker.js`

wasm/client/src/helpers.rs
```diff
@@ -48,10 +50,11 @@ pub(crate) trait InputSender {
...
     fn send_message(&self, message: InputMessage) -> Promise {
         let this = Arc::clone(self);
         future_to_promise(async move {
+            let mut this = this.write().await;
```
jstuczyn, line 57: this limitation is a bit annoying, it means that whenever you want to send a message and waiting for the whole pipeline to go through, you can't receive anything : /

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -190,6 +214,44 @@ impl MixnetClient {
...
+        cx: &mut Context<'_>,
+    ) -> Poll<tokio::io::Result<()>> {
+        if self._read.buffer.len() < buf.capacity() {
+            // let written = self._read.buffer.len();
+            buf.put_slice(&self._read.buffer);
+            self._read.clear();
+            Poll::Ready(Ok(()))
+        } else {
+            let written = buf.capacity();
+            buf.put_slice(&self._read.buffer[..written]);
```
jstuczyn: nit: you could instead use `buf.put_slice(&self_read.buffer.split_off(written))` then you wouldn't need to worry about manually advancing the buffer

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -198,6 +260,195 @@ pub struct MixnetClientSender {
...
+
+        let msg = match self.as_mut().poll_next(cx) {
+            Poll::Ready(Some(msg)) => msg,
+            Poll::Ready(None) => return Poll::Ready(Ok(())),
+            Poll::Pending => return Poll::Pending,
+        };
+
+        // let mut buffer = BytesMut::new();
+
+        codec.encode(msg, &mut self._read.buffer).unwrap();
```
jstuczyn: is it possible for that codec to fail? 
if not, could you make a comment on why? `unwrap()` are always scary : )

jstuczyn: but speaking of the codec, whoever wants to use `AsyncRead` on the `MixnetClient` will get `bincode`-serialised data which might not be very desirable or expected. as an user i'd expect to receive pure mixnet bytes.

what if we had two different and configurable ways of encoding that message, "advanced" and "simple" (default). "advanced" would just do whatever is happening right now, but simple would just strip the `sender_tag` out of `ReconstructedMessage` and just propagate the raw `message : Vec<u8>` instead. what do you think?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -198,6 +260,195 @@ pub struct MixnetClientSender {
...
+
+        // = buffer.to_vec();
+
+        self.read_buffer_to_slice(buf, cx)
+    }
+}
+
+impl AsyncWrite for MixnetClient {
```
jstuczyn, line 277: imo `AsyncWrite` shouldn't be implemented on `MixnetClient` directly because conceptually it doesn't make a lot of sense. you're really just sending full messages as opposed to bytes. there should be some intermediate structure, like

```rust
struct MixnetClientAsyncWrite{
  recipient: Recipient
}
```

so then you'd have something like `fn async_write(&self, recipient: Recipient) -> MixnetClientAsyncWrite` on the client where you could just write your bytes your heart's content without having to do any encoding

jstuczyn: now, for any "higher" level applications that might need `AsyncRead + AsyncWrite` maybe we could just make a wrapper like
```rust
struct Wrapper {
  inner: MixnetClient,
  // or whatever fields we need
  write_recipient: Recipient,
}
```
with maybe even `Deref` trait for easier use

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -198,6 +260,195 @@ pub struct MixnetClientSender {
...
+    }
+
+    fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>> {
+        self.sender()
+            .poll_close_unpin(cx)
+            .map_err(|_| Error::MessageSendingFailure)
+    }
+}
+
+// TODO: there should be a better way of implementing Sink and AsyncWrite over T: MixnetMessageSender
```
jstuczyn, line 357: I second that comment, although can't immediately think of anything

common/client-core/src/client/inbound_messages.rs
durch: Great catch, I actually wanted to replace that with `const` and forgot

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: I think so, since AsyncRead, just uses Stream under the hood, worst case that comes to mind is that you miss a message if you mess up AsyncRead. And since both traits require `&mut` you can't use both at the same exact time.

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: Thats from when it implemented futures::AsyncRead, gone

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: Yes! I've actually done the thing on the InputMessage you suggested, and added another codec that takes a Receiver as a parameter, was just thinking of how to get similar ergonomics for the ReconstructedMessage

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: Thats a very good point, but I wanted to avoid adapters here, I'm gonna give it some more though, maybe with the new InputMessageCodec we can get to just sending bytes in, and they discretley get converted to messages...

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: > is it possible for that codec to fail? if not, could you make a comment on why? `unwrap()` are always scary : )

Ouch, fixed that

wasm/client/src/helpers.rs
durch: Agree, I didn't spend a lot of time on the WASM part, I'll revisit these. One could not await on the JS level, ie use a Promise::all or something like that

PR #4636, opened 2024-06-10
Add generic wg private network routing

neacsu:
As defguard wireguard only allows for peer routing modifications, we will configure the entire wireguard private network to be routed to the wg device.

Configuring per peer is also not desirable, as the interface doesn't allow removing routes, so unused ip routing won't be cleaned until gateway restart (and it would also pollute to routing table with a lot of rules when many peers are added)


octol (approved):

tommyv1987:
**Testing Completion Date:** [12/06/2024]

**Testing Steps Performed:**
1. This is a part of a bigger ticket, but initial testing has proven to shown that launching nym-nodes (entry and exit gateways) in WG enable mode to be working

I will use this template for the other related WG tickets in this release milestone.

**Validation:**
- ✅  Test Passed
- [ 🔴 ] Issues Found 

**Notes (if any):**
N/A

N/A

PR #4643, opened 2024-06-11
Remove nym-connect directory

octol:
Since the nym-vpn has superseded nym-connect, remove nym-connect from the repo


jstuczyn (approved):

Cargo.toml
```diff
@@ -140,9 +140,6 @@ exclude = [
     "explorer",
     "contracts",
     "nym-wallet",
-    "nym-connect/mobile/src-tauri",
-    "nym-connect/desktop",
-    "nym-vpn/ui/src-tauri",
```
jstuczyn, line 145: why are we removing `nym-vpn`? why was it even there?

octol: old leftover, cargo probably doesn't check the existence of it

PR #4654, opened 2024-06-14
renamed 'accept-toc' flag and fields into explicit 'accept-operator-terms-and-conditions'

jstuczyn:
this PR makes the `accept-toc` flag more explicit

### DIFF
--- a/nym-node/src/cli/commands/run/mod.rs
+++ b/nym-node/src/cli/commands/run/mod.rs
@@ -46,9 +46,9 @@ pub(crate) async fn execute(mut args: Args) -> Result<(), NymNodeError> {
     let bonding_info_path = args.bonding_information_output.clone();
     let init_only = args.init_only;
     let local = args.local;
-    let accepted_toc = args.accept_operator_terms;
+    let accepted_operator_terms_and_conditions = args.accept_operator_terms_and_conditions;
 
-    if !accepted_toc {
+    if !accepted_operator_terms_and_conditions {
         warn!("you don't seem to have accepted the terms and conditions of a Nym node operator");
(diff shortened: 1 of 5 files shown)


nym-node/src/cli/commands/run/mod.rs
```diff
@@ -46,9 +46,9 @@ pub(crate) async fn execute(mut args: Args) -> Result<(), NymNodeError> {
     let bonding_info_path = args.bonding_information_output.clone();
     let init_only = args.init_only;
     let local = args.local;
-    let accepted_toc = args.accept_operator_terms;
+    let accepted_operator_terms_and_conditions = args.accept_operator_terms_and_conditions;
 
-    if !accepted_toc {
+    if !accepted_operator_terms_and_conditions {
         warn!("you don't seem to have accepted the terms and conditions of a Nym node operator");
         warn!("please familiarise yourself with <https://nymtech.net/terms-and-conditions/operators/v1.0.0> and run the binary with '--accept-toc' flag if you agree with them");
```
tommyv1987: --accept-operator-terms-and-conditions
          Explicitly specify whether you agree with the terms and conditions of a nym node operator as defined at <https://nymtech.net/terms-and-conditions/operators/v1.0.0> [env: NYMNODE_ACCEPT_OPERATOR_TERMS=]
          
          @jstuczyn this will need to change here sir `--accept-operator-terms-and-conditions`

jstuczyn: fixed in https://github.com/nymtech/nym/pull/4654/commits/1c3140898fafdf32f4ae05b91a972f765c64d8f6

benedettadavico (approved):

PR #4667, opened 2024-06-18
Add authenticator

neacsu:


gateway/src/node/mod.rs
```diff
@@ -222,20 +236,31 @@ impl<St> Gateway<St> {
...
         &mut self,
+        opts: &LocalAuthenticatorOpts,
         shutdown: TaskClient,
     ) -> Result<Arc<nym_wireguard::WgApiWrapper>, Box<dyn std::error::Error + Send + Sync>> {
         if let Some(wireguard_data) = self.wireguard_data.take() {
+            let authenticator_server = nym_authenticator::Authenticator::new(
+                opts.config.clone(),
+                wireguard_data.inner.clone(),
+            )
+            .with_shutdown(shutdown.clone());
```
octol: Possibly you want to use TaskClient::fork() to make the name unique in the logs

service-providers/authenticator/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+version = "0.1.0"
+authors.workspace = true
+repository.workspace = true
+homepage.workspace = true
+documentation.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+fastrand = "2"
```
octol: Move it to workspace?

common/authenticator-requests/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+authors.workspace = true
+repository.workspace = true
+homepage.workspace = true
+documentation.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+bincode = { workspace = true }
+rand = "0.8.5"
```
octol: Use workspace dependency?

service-providers/authenticator/src/cli/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+            nym_config::parse_urls,
+        )
+        .with_optional_base(
+            BaseClientConfig::with_disabled_credentials,
+            args.enabled_credentials_mode.map(|b| !b),
+        )
+}
+
+pub(crate) async fn execute(args: Cli) -> Result<(), AuthenticatorError> {
+    let bin_name = "nym-ip-packet-router";
```
octol: copy paste name?

octol (approved):

service-providers/authenticator/src/mixnet_listener.rs
```diff
@@ -0,0 +1,275 @@
...
+        };
+        if let Some(gateway_client) = gateway_client_opt {
+            self.remove_from_registry(&remote_public, &gateway_client)?;
+        }
+        let mut private_ip_ref = self
+            .free_private_network_ips
+            .iter_mut()
+            .filter(|r| **r)
+            .choose(&mut thread_rng())
+            .ok_or(AuthenticatorError::NoFreeIp)?;
```
octol, line 163: If this line triggers an error, does that mean we end up in a half-registered state with the IP claimed but not really finished?

common/authenticator-requests/Cargo.toml
neacsu: Fixed it here and in `ip-packet-requests` too

service-providers/authenticator/src/cli/mod.rs
neacsu: Good catch

service-providers/authenticator/src/mixnet_listener.rs
neacsu: That doesn't happen here, the maximum that could happen is that they get deregistered from the previous IP they got allocated and don't get to have a new one, so they need to try later. But the problem you're asking does appear actually, with half-registered states that might never actually concluded (because of errors or bad actors). I've pushed a commit that also removes those stale registration attempts, so that they don't accumulate over time.

PR #4677, opened 2024-06-26
Create UserAgent that can be passed from the binary to the nym api client

octol:
- Support setting `UserAgent` for the validator client
- Support setting `UserAgent` in the SDK `MixnetClient`
- Set `UserAgent` when getting the list of gateways and topology in
    - `nym-client`
    - `nym-socks5-client`
    - Standalone `ip-packet-router`

Example of the user agent sent:
`nym-client/1.1.36/x86_64-unknown-linux-gnu/e18bb70`

### DIFF
--- a/common/bin-common/src/build_information/mod.rs
+++ b/common/bin-common/src/build_information/mod.rs
@@ -5,7 +5,6 @@
 // and be used by our smart contracts
 
 use serde::{Deserialize, Serialize};
-use serde_json::de::StrRead;
 use std::fmt::{Display, Formatter};
 
 #[derive(Debug)]
--- a/common/bin-common/src/build_information/mod.rs
+++ b/common/bin-common/src/build_information/mod.rs
@@ -44,6 +44,10 @@ pub struct BinaryBuildInformation {
     /// Provides the cargo debug mode that was used for the build.
     // NOTE: keep the old name cargo_profile instead of cargo_debug for backwards compatibility
     pub cargo_profile: &'static str,
+
+    // VERGEN_CARGO_TARGET_TRIPLE
+    /// Provides the cargo target triple that was used for the build.
+    pub cargo_triple: &'static str,
 }
 
 impl BinaryBuildInformation {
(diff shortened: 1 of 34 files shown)


common/bin-common/src/build_information/mod.rs
```diff
@@ -156,6 +163,15 @@ pub struct BinaryBuildInformationOwned {
     /// Provides the cargo debug mode that was used for the build.
     // NOTE: keep the old name cargo_profile instead of cargo_debug for backwards compatibility
     pub cargo_profile: String,
+
+    // VERGEN_CARGO_TARGET_TRIPLE
+    /// Provides the cargo target triple that was used for the build.
+    #[serde(default = "unknown")]
+    pub cargo_triple: String,
```
octol, line 170: So, this thing here might impact backwards compatibility

octol:
Testing note: this thing will affect nym-api responses, so we need to confirm that it doesn't affect IPR and other clients ability to parse responses

common/bin-common/src/build_information/mod.rs
jstuczyn: I think it should be fine. since you've added the `default`, old code should still be deserialisable

common/http-api-client/src/lib.rs
```diff
@@ -283,6 +287,9 @@ impl Client {
         E: Display + DeserializeOwned,
     {
         let res = self.send_get_request(path, params).await?;
+        // dbg!(&res);
```
jstuczyn: remove dead code

common/http-api-client/src/lib.rs
```diff
@@ -14,6 +14,10 @@ use url::Url;
 
 pub use reqwest::IntoUrl;
 
+pub use user_agent::UserAgent;
+
+mod user_agent;
```
jstuczyn, line 19: shouldn't this be public?

jstuczyn (approved):

common/http-api-client/src/lib.rs
octol: Currently not needed, so thought I'd leave it private for now

tommyv1987:
Testing notes:
Used the nym-vpn-cli to test this, and we can visibly see the userAgent, no issues with the comments mentioned above.
image

Connected with no problems

PR #4682, opened 2024-06-28
Update setup.md

stefanopier:


serinko:
@stefanopier What is the reason of removing this part?

stefanopier:
sorry for late reply
I noticed that there was a duplicate entry, and I just checked online and it's still there: on the mixnode section the "Run the node with custom --id without initialising:" command is duplicated but the second one is for an entry-gateway run, that is listed above and may dangerous if someone copies that the command without reading, leading to run a gw instead of a mixnode.
Or I didn't get that at all, in that case please ignore this PR;)
Screenshot 2024-07-23 114345

serinko:
I see, thank you for flagging!

serinko (approved):

PR #4684, opened 2024-07-01
Add mixnodes to self describing api cache

durch:
+ abstracts getting the self describing info a bit
+ adds mixnodes to the cache refresher as well
+ adds `role` field to the NodeDescription struct, to be able to distinguish between mixnodes and gateways
+ Switched to using `NodeStatusCache` instead of `ContractCache`

### DIFF
--- a/nym-api/nym-api-requests/src/models.rs
+++ b/nym-api/nym-api-requests/src/models.rs
@@ -9,7 +9,7 @@ use nym_mixnet_contract_common::mixnode::MixNodeDetails;
 use nym_mixnet_contract_common::reward_params::{Performance, RewardingParams};
 use nym_mixnet_contract_common::rewarding::RewardEstimate;
 use nym_mixnet_contract_common::{
-    GatewayBond, IdentityKey, Interval, MixId, MixNode, Percent, RewardedSetNodeStatus,
+    GatewayBond, IdentityKey, Interval, MixId, MixNode, MixNodeBond, Percent, RewardedSetNodeStatus,
 };
 use nym_node_requests::api::v1::node::models::{AuxiliaryDetails, BinaryBuildInformationOwned};
 use schemars::gen::SchemaGenerator;
@@ -621,6 +621,21 @@ pub struct DescribedGateway {
 impl From<GatewayBond> for DescribedGateway {
     fn from(bond: GatewayBond) -> Self {
         DescribedGateway {
+            bond,
+            self_described: None,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
+pub struct DescribedMixNode {
+    pub bond: MixNodeBond,
+    pub self_described: Option<NymNodeDescription>,
+}
+
+impl From<MixNodeBond> for DescribedMixNode {
+    fn from(bond: MixNodeBond) -> Self {
+        DescribedMixNode {
             bond,
             self_described: None,
         }
--- a/nym-api/nym-api-requests/src/models.rs
+++ b/nym-api/nym-api-requests/src/models.rs
@@ -607,6 +607,15 @@ pub struct NymNodeDescription {
 
     // for now we only care about their ws/wss situation, nothing more
     pub mixnet_websockets: WebSockets,
+
+    pub role: NymNodeRole,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum NymNodeRole {
+    Mixnode,
+    Gateway,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
--- a/nym-api/nym-api-requests/src/models.rs
+++ b/nym-api/nym-api-requests/src/models.rs
@@ -1,6 +1,7 @@
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::nym_nodes::NodeRole;
 use crate::pagination::PaginatedResponse;
 use cosmwasm_std::{Addr, Coin, Decimal};
 use nym_mixnet_contract_common::families::FamilyHead;
@@ -608,14 +609,7 @@ pub struct NymNodeDescription {
     // for now we only care about their ws/wss situation, nothing more
     pub mixnet_websockets: WebSockets,
 
-    pub role: NymNodeRole,
-}
-
-#[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
-#[serde(rename_all = "lowercase")]
-pub enum NymNodeRole {
-    Mixnode,
-    Gateway,
+    pub role: NodeRole,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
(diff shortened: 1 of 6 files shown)


nym-api/src/node_describe_cache/mod.rs
```diff
@@ -79,17 +78,16 @@ impl NodeDescriptionProvider {
...
-    let gateway_host = &gateway.host;
-
     // first try the standard port in case the operator didn't put the node behind the proxy,
     // then default https (443)
     // finally default http (80)
     let addresses_to_try = vec![
-        format!("http://{gateway_host}:{DEFAULT_NYM_NODE_HTTP_PORT}"),
-        format!("https://{gateway_host}"),
-        format!("http://{gateway_host}"),
+        format!("http://{host}:{DEFAULT_NYM_NODE_HTTP_PORT}"),
```
jstuczyn, line 88: if you're using the same list of addresses for both gateways and mixnodes, you will have to add port `8000` (`DEFAULT_HTTP_API_LISTENING_PORT`) because old, migrated, mixnodes will most likely be using that one instead; or if you want to be fancy, you could add an extra argument `port: Option<u16>` since mixnodes will have announced this one to the directory server

nym-api/nym-api-requests/src/models.rs
```diff
@@ -607,6 +607,15 @@ pub struct NymNodeDescription {
 
     // for now we only care about their ws/wss situation, nothing more
     pub mixnet_websockets: WebSockets,
+
+    pub role: NymNodeRole,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum NymNodeRole {
```
jstuczyn: we already have a role enum defined: https://github.com/nymtech/nym/blob/develop/nym-api/nym-api-requests/src/nym_nodes.rs#L73 maybe we could somehow refactor it a tiny bit and reuse it here too?

nym-api/src/node_describe_cache/mod.rs
```diff
@@ -209,23 +236,48 @@ impl CacheItemProvider for NodeDescriptionProvider {
     }
 
     async fn try_refresh(&self) -> Result<Self::Item, Self::Error> {
```
jstuczyn, line 249: I don't remember from the top of my head where this information is persisted, but isn't it the case that the current described gateways are served via `/gateways/described` meaning they'd now confusingly also include mixnodes?

nym-api/src/node_describe_cache/mod.rs
mmsinclair: doesn't it still need to check all of those? Because it is just a `nym-node` that can sometimes be a gateway and sometimes a mixnode?

jstuczyn: - brand new nym-node will be on `8080` (or `80`/`443` if the user bothered to setup reverse proxy)
- migrated gateway will also be on `8080`
- migrated mixnode, however, will use whatever they have configured before. by default it was `8000` but the "true" value is stored in the mixnet contract

nym-api/src/node_describe_cache/mod.rs
durch: Will add an endpoint and return via Role filtering, will also add one for returning all

nym-api/src/node_describe_cache/mod.rs
durch: @jstuczyn could we count on what the mixnodes announce blindly, ie remove the DEFAULT_PORT, and only leave what is announced and `80/443`

nym-api/nym-api-requests/src/models.rs
durch: We can probably guess if its an exit gateway if its running IPR, so we can assign gateway roles off that

durch: Actually not really, since they could plausibly be Entry as well, @jstuczyn any ideas on inferring these roles from data available to the `nym-api`

nym-api/src/node_describe_cache/mod.rs
```diff
@@ -194,51 +197,104 @@ async fn get_gateway_description(
...
+    }
 }
 
 #[async_trait]
 impl CacheItemProvider for NodeDescriptionProvider {
     type Item = HashMap<IdentityKey, NymNodeDescription>;
     type Error = NodeDescribeCacheError;
 
-    async fn wait_until_ready(&self) {
-        self.contract_cache.wait_for_initial_values().await
```
jstuczyn: don't we want to wait for until we have some initial values (if not contract_cache, then the `node_status_cache`)? otherwise, on first refresh we won't get anything and our endpoints are going to be returning empty data until it refreshes for the second time, which from the top of my head will happen after few hours

nym-api/src/node_describe_cache/mod.rs
```diff
@@ -112,28 +115,28 @@ async fn try_get_client(
...
 
     let host_info =
         client
             .get_host_information()
             .await
             .map_err(|err| NodeDescribeCacheError::ApiFailure {
-                gateway: gateway.identity_key.clone(),
+                gateway: data.identity_key().to_string(),
                 source: err,
             })?;
```
jstuczyn, line 136: fyi. while I don't think I've put it on the client (because I didn't need it at the time), you could query for `self.get_json_from(routes::api::v1::roles_absolute()).await` to get info whether the node is a mixnode/gateway/runs IPR/NR from the node itself

nym-api/src/node_describe_cache/mod.rs
```diff
@@ -248,13 +304,13 @@ impl CacheItemProvider for NodeDescriptionProvider {
 #[allow(dead_code)]
 pub(crate) fn new_refresher(
     config: &config::TopologyCacher,
-    contract_cache: NymContractCache,
+    node_status_cache: NodeStatusCache,
```
jstuczyn: any reason we switched from using raw contract data for figuring out what to query in favour of the fat node status cache with all the inclusion probabilities, performance, families, etc?

durch: > any reason we switched from using raw contract data for figuring out what to query in favour of the fat node status cache with all the inclusion probabilities, performance, families, etc?

Yes, we need the annotated bonds to get the layer info for mixnodes, so that we can reuse the NodeRole thing we already have

nym-api/src/node_describe_cache/mod.rs
durch: If we could then go back to contract cache, that would be a win, would that be another http request towards the node then, if so I'd rather avoid it, I'm worried about how long this will take as is...

nym-api/src/node_describe_cache/mod.rs
durch: I'm worried about blocking here, do we know how long this might take, I'd rather return nothing until we have something, let the clients deal with empty stuff, then again if this is fast then we can just block, but I'm worried it won't be fast

jstuczyn (approved):

benedettadavico:
Testing Completion Date: [16/07/2024]

Testing Steps Performed:
Called the new `/mixnodes/described` endpoint as well as the existing `/gateways/described` endpoint and verified that the data returned for each was correct based on the settings that different nodes have when they are setup. 
For gateway endpoint, the “role” for now does not differentiate between entry and exit gateways, this will be implemented in the future.
 
Validation:
✅ Test Passed

PR #4692, opened 2024-07-03
Max/try fix doc search

mfahampshire:
- [x] fix `.hds` for `/docs/` 
- [ ] ~~cont try and fix search~~
- [x] removed button for the moment  
- [x] remove extra logging in `post_process`

something is borked with the search, for the moment removing it and will come back to it when we have time to do a deeper docs overhaul @serinko 

since I was already messing around I quickly whacked a few theming updates in here to fix the strange padding issue that we have had for a little while, and tidied up the header


serinko (approved):
LGTM!

PR #4698, opened 2024-07-09
Migrate nym-api HTTP server from rocket to axum

dynco-nym:
# Changes
- start `axum` server alongside `rocket`
- migrate all routes
- write simple API tests that migrated APIs return 200

# TODO
- serve swagger
- migrate logging to tracing
- remove all rocket code

# Notes to reviewers
I left all `rocket` routes alongside new `axum` ones
- `rocket` files are named `routes.rs` or `routes_deprecated.rs`
  + they may show whole file green diffs but it's only rename
- `axum` files are named `handlers.rs`
- merging this into a feature branch: will merge into develop once complete migration is done

Run server
```
cargo run --package nym-api run
```

Simple API tests
```
./nym-node/tests/test_apis.sh
```

Good starting point: compare 
- `start_nym_api_tasks()` to 
- `start_nym_api_tasks_v2()`


common/bin-common/src/logging/mod.rs
```diff
@@ -41,6 +41,8 @@ pub fn setup_logging() {
         .filter_module("tokio_tungstenite", log::LevelFilter::Warn)
         .filter_module("handlebars", log::LevelFilter::Warn)
         .filter_module("sled", log::LevelFilter::Warn)
+        .filter_module("axum", log::LevelFilter::Info)
+        .filter_module("nym_api", log::LevelFilter::Info)
```
jstuczyn: why are we filtering nym-api globally?

nym-api/src/main.rs
```diff
@@ -54,6 +58,7 @@ async fn main() -> Result<(), anyhow::Error> {
     }}
 
     setup_logging();
+    // TODO dz replace with tracing logger once rocket is eliminated from code
```
jstuczyn, line 61: yes please : )

nym-api/src/support/http/static_routes.rs
jstuczyn, line 1: I think this file should live in  the `nym-api-requests` crate as the client will also need to be able to refer to those

nym-api/src/support/http/mod.rs
```diff
@@ -52,33 +63,26 @@ pub(crate) async fn setup_rocket(
         "/network" => network_routes(&openapi_settings),
         "/api-status" => api_status_routes(&openapi_settings),
         "/ecash" => ecash::routes_open_api(&openapi_settings, config.coconut_signer.enabled),
-        "" => nym_node_routes(&openapi_settings),
+        "" => nym_node_routes_deprecated(&openapi_settings),
```
jstuczyn, line 66: just for pure curiosity, why did you rename those to deprecated? I mean you're not wrong, because I did just deprecate them on my own branch, but how did you know : )

nym-api/src/support/http/mod.rs
```diff
@@ -154,3 +151,50 @@ fn setup_cors() -> Result<Cors> {
 
     Ok(cors)
 }
+
+#[allow(dead_code)]
+// TODO dz Cannot combine `Access-Control-Allow-Credentials: true` with `Access-Control-Allow-Headers: *`
```
jstuczyn: imo just remove the allow credentials then because we really don't care about those. we just needed the headers; the reason I had the credentials on my branch is that it was the default behaviour from rocket's cors setup

nym-api/src/support/http/mod.rs
```diff
@@ -154,3 +151,50 @@ fn setup_cors() -> Result<Cors> {
...
+        .route("/", axum::routing::get(|| async {axum::response::Redirect::permanent("/swagger")}))
+        .route("/swagger", axum::routing::get(hello))
+        .merge(circulating_supply_routes())
+        .merge(nym_contract_cache_routes())
+        .merge(node_status_routes(network_monitor))
+        .merge(nym_network_routes())
+        .merge(status::handlers::api_status_routes())
+        .merge(nym_node_routes())
+        .merge(nym_node_routes_unstable())
+        // CORS layer needs to be "outside" of routes
```
jstuczyn, line 188: what about the elephant in the room also known as ecash? : )

nym-api/src/support/storage/mod.rs
```diff
@@ -64,6 +64,7 @@ impl NymApiStorage {
         Ok(storage)
     }
 
+    #[deprecated(note = "TODO dz: obsolete because it's used for Rocket")]
     pub(crate) fn stage(storage: NymApiStorage) -> AdHoc {
```
jstuczyn, line 68: you might as well then remove it completely because the only thing this fairing does is to manage the state; so in your temporary rocket builder juts add explicit `rocket.manage(storage)`

nym-api/src/support/http/helpers.rs
```diff
@@ -1,11 +1,72 @@
...
+use tracing::info;
 
 #[derive(Serialize, Deserialize, FromForm, Debug, JsonSchema)]
 pub struct PaginationRequest {
     pub page: Option<u32>,
     pub per_page: Option<u32>,
 }
+
+// TODO dz move this fn to a "common" place so this can be exported ?
+pub async fn logger(
```
jstuczyn: I swear I already put this code alongside other common axum helpers in http-common crate on my initial axum-migration branch 🤔

nym-api/src/support/config/template.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+// TODO dz any changes to this ?
```
jstuczyn, line 4: yes. as I had on my branch, you need to put the bind address into the config because rocket pulls it from `Rocket.toml` which obviously won't exist

nym-api/src/support/config/mod.rs
```diff
@@ -234,6 +246,13 @@ pub struct Base {
     #[zeroize(skip)]
     pub local_validator: Url,
 
+    // TODO dz Andrew added this here ?!
```
jstuczyn, line 249: I did : ) now we also need it in the template

nym-api/src/node_status_api/cache/mod.rs
```diff
@@ -41,12 +41,13 @@ pub struct NodeStatusCache {
...
 impl NodeStatusCache {
     /// Creates a new cache with no data.
-    fn new() -> NodeStatusCache {
+    pub(crate) fn new() -> NodeStatusCache {
         NodeStatusCache {
             inner: Arc::new(RwLock::new(NodeStatusCacheData::new())),
         }
     }
 
+    #[deprecated(note = "TODO dz: obsolete because it's used for Rocket")]
```
jstuczyn, line 50: the same comment I had with the storage, just remove it, because it doesn't do much apart from adding it to the managed states

nym-api/src/main.rs
```diff
@@ -107,7 +112,11 @@ async fn start_nym_api_tasks(config: Config) -> anyhow::Result<ShutdownHandles>
     let nym_contract_cache_state = rocket.state::<NymContractCache>().unwrap();
     let node_status_cache_state = rocket.state::<NodeStatusCache>().unwrap();
     let circulating_supply_cache_state = rocket.state::<CirculatingSupplyCache>().unwrap();
-    let maybe_storage = rocket.state::<NymApiStorage>();
+    let storage = if let Some(storage) = rocket.state::<NymApiStorage>() {
```
jstuczyn, line 115: I think that thing can backfire in the future. can we just ensure that the storage always exists when setting up the rocket? though I guess since ultimately we'll remove the rocket, maybe it's fine as the stepping stone

nym-api/src/ecash/api_routes/aggregation_axum.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::sync::Arc;
+use time::Date;
+
+/// routes with globally aggregated keys, signatures, etc.
+pub(crate) fn aggregation_routes(ecash_state: Arc<EcashState>) -> Router<AxumAppState> {
+    Router::new()
+        .route(
+            "/master-verification-key:epoch_id",
+            axum::routing::get({
+                let ecash_state = Arc::clone(&ecash_state);
```
jstuczyn, line 26: I'm confused, why are we explicitly passing the `EcashState` in an `Arc` rather than using the state extractor like you're already doing with other routes?

jstuczyn: ah. I think I know why, because under some conditions the routes wouldn't exist... 
🤔 
if it would make your life easier (and I think it would make the code nicer), what do you think of managing `Option<EcashState>`
and then simply at the beginning of each request have something along the lines of 
```rust
let Some(ecash_state) = ecash_state else {
  // return an error response, idk a 400 maybe?
}
```

jstuczyn: sorry, I'm using rocket wording here ; D
by managing I meant putting it inside your `AxumAppState`

nym-api/src/node_status_api/handlers.rs
```diff
@@ -0,0 +1,585 @@
...
+    GatewayCoreStatusResponse, GatewayStatusReportResponse, GatewayUptimeHistoryResponse,
+    GatewayUptimeResponse, InclusionProbabilityResponse, MixNodeBondAnnotated,
+    MixnodeCoreStatusResponse, MixnodeStatusReportResponse, MixnodeStatusResponse,
+    MixnodeUptimeHistoryResponse, RewardEstimationResponse, StakeSaturationResponse,
+    UptimeResponse,
+};
+use nym_mixnet_contract_common::MixId;
+
+pub(crate) fn node_status_routes(network_monitor: bool) -> Router<AxumAppState> {
+    // in the minimal variant we would not have access to endpoints relying on existence
```
jstuczyn, line 29: again, if it would make your life and routing easier, perhaps always expose **all** the routes, but simply immediately return an error in the handler if one of the conditions is not fulfilled, such as in this case network monitor not being enabled

jstuczyn: axum routes don't have to behave exactly 100% identical to what we have in rocket

common/bin-common/src/logging/mod.rs
dynco-nym: for testing, removed

nym-api/src/support/http/static_routes.rs
dynco-nym: will move it in the next PR (serving swagger docs) where I make some changes to this file

nym-api/src/support/http/mod.rs
dynco-nym: done

nym-api/src/support/http/mod.rs
dynco-nym: ecash routes were added in separate commit

nym-api/src/support/storage/mod.rs
dynco-nym: valid point, but I'll leave the function to avoid too much refactoring of rocket code, which will be removed anyway

nym-api/src/main.rs
dynco-nym: Your suggestion over chat was to always init storage: could you elaborate how would it backfire?

nym-api/src/node_status_api/handlers.rs
dynco-nym: > always expose all the routes, but simply immediately return an error in the handler if one of the conditions is not fulfilled

Doesn't it make much more sense to not expose routes which shouldn't be there and not return unnecessary 4xx/5xx ?

 It's better from both code and API design perspective

nym-api/src/ecash/api_routes/aggregation_axum.rs
dynco-nym: `Option<State>`  was my initial approach, but... 

Why choose to add an `if` condition for an `Option` you know shouldn't be there, if you can instead encode that info into the type system?

This avoids developer error if a future me or you introduces a change and forgets to initialize an `Option`. 

With my code, it's simply not possible because route either is or isn't there. And if it is, it has access to correct state.

nym-api/src/support/http/helpers.rs
dynco-nym: yep, moved

nym-api/src/main.rs
octol: I mean will we not remove rocket anyway as part of merging this into develop? Seems like why bother with this then?

octol (approved):

PR #4703, opened 2024-07-10
Add upgrades to nym-node for authenticator changes

neacsu:


gateway/src/node/mod.rs
```diff
@@ -565,12 +619,18 @@ impl<St> Gateway<St> {
 
         #[cfg(feature = "wireguard")]
         let _wg_api = if let Some(opts) = self.authenticator_opts.clone() {
```
jstuczyn: nitpicking: you don't have to clone it, you can just use `as_ref()` since `start_authenticator` takes a reference anyway

nym-node/src/config/entry_gateway.rs
jstuczyn, line 1: just making sure: when you migrate into a nym-node, or as a matter of fact, initialise a fresh nym-node, you don't need to setup any persistent data for the authenticator like we do with the network requester or the IPR?

jstuczyn (approved):

jstuczyn:
I've also noticed the authenticator has the following in its config: `pub authenticator_description: PathBuf`, do we really care about it?

gateway/src/node/mod.rs
dynco-nym: clippy could capture things like this, do we run it in the CI?

neacsu: The reason it needs to be like this and clippy didn't say anything is because the `opts` of `self` is used in a function that also takes `self` as mutable. But I've moved things a bit so that `opts` is unwrapped inside the function, which is a bit more natural.

neacsu:
> I've also noticed the authenticator has the following in its config: `pub authenticator_description: PathBuf`, do we really care about it?

You're right, I don't think we need it, I removed it

nym-node/src/config/entry_gateway.rs
neacsu: I do setup persistent data i.e. the keys for fresh nym-node. For migration, we assumed that authenticator will only be supported for nym-nodes, since we were in the process of deprecating the other binaries, so there is no migration of keys happening, since there shouldn't be any authenticator on nym-gateway binary

tommyv1987:
**Testing Completion Date:** 23/07/2024

**Testing Steps Performed:**
1. Reviewed the changes in the `gateway/src/error.rs` and `gateway/src/node/mod.rs` files.
2. Verified the new error enum `AuthenticatorStartupFailure` was added to `GatewayError`.
3. Confirmed the implementation of the `StartedAuthenticator` struct and its usage in the `start_authenticator` function.
4. Ran the updated code in the canary environment.
5. Monitored the canary environment for any issues or errors related to the changes.

**Validation:**
- [x] Test Passed

**Notes (if any):**
The test was successful. The changes for authenticator upgrades worked as expected in the canary environment. No issues were observed. This has not been fully rolled out for mainnet yet, as the metering ticket needs to be completed.

No attachments provided.

PR #4706, opened 2024-07-11
removed mixnode/gateway config migration code and disabled cli without explicit flag

jstuczyn:
gateway and mixnode commands now won't do anything without explicit `--force-run` to bypass the deprecation.

the next step, in say a month or so, is to completely remove all `cli` related things.


octol (approved):

benedettadavico:
Testing Completion Date: 30/07/2024

Testing Steps Performed:

- Verified that the nym-gateway binary and nym-mixnode binary commands return the _error message_ stating to update to nym-node
- checked that when adding the `--force-run` flag, it still allows the command to be run (aside from `init` which has been removed) and the message stating to update to nym-node is a _warning_ now
- Checked nym-node is not affected
- Reviewed the changes in the PR

Validation:
- [x ] Test Passed

PR #4713, opened 2024-07-17
Update setup.md

edcalderon:
remove cus "--accept-operator-terms-and-condition " error: unexpected argument '--accept-operator-terms-and-conditions' found

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4713)


serinko:
On which version of node did you had that issue?

serinko:
It seems to me like you ran the command on an version older than `nym-node 1.1.3`.

tommyv1987:
this is outdated now and all operators are required to run with the tnc flags

PR #4723, opened 2024-07-23
Handle clients with different versions in IPR

octol:
Allow the IPR to handle clients connecting both using v6 and v7, independently.

The motivation is that we want to be able to roll out a API version change gradually for vpn clients without breaking backwards compatibility. The main feature on the new v7 format that is not yet used, is that it adds signatures for connect/disconnect.


neacsu (approved):

benedettadavico:
Testing Completion Date: 31/07/2024

Testing Steps Performed:

Ran the same command (using same gateways deployed from this PR) on different versions of the nym-vpn-cli. 
Example: 
`~/nym-vpn-core-v0.1.10_macos_universal ❯ sudo -E ./nym-vpn-cli -c ../qa.env run --entry-gateway-id $entry_gateway --exit-gateway-id $exit_gateway --enable-two-hop`

`~/nym-vpn-core-v0.1.11-dev_macos_universal ❯ sudo -E ./nym-vpn-cli -c ../qa.env run --entry-gateway-id $entry_gateway --exit-gateway-id $exit_gateway --enable-two-hop`

Validation:
[x ] Test Passed

PR #4623, opened 2024-05-28
Feature Compact Ecash : The One PR

simonwicky:
Ecash PR after review

One PR to rule them all, 
One PR to find them, 
One PR to bring them all, 
and in `develop` merge them

PR #4729, opened 2024-07-24
Ania/ticket types

aniampio:
Introduced a new public attribute for the ticket type (t_type). This attribute is known to validators during the issuance phase and revealed to the entry/exit node during the spending phase. The addition of the t_type attribute requires only small modifications across the ticket issuance and spending process.

1. Ticket Book Request and Issuance:
- The t_type is now included in the ticket book request as a public attribute.
- During the issuance phase, validators sign the t_type (without blinding). This ensures that the ticket type is verifiable as part of the issuance process.

2. Spending Function:
- Payment Field Addition: The t_type is added as a field in the payment structure within the spend function.
- Signature Validity Check: The spend verification process now includes a modification to the ticket's signature validity check to account for the ticket type.


jstuczyn:
cherry-picked into https://github.com/nymtech/nym/pull/4623

PR #4730, opened 2024-07-25
Fix tokio error in 1.39

octol:
Bump tokio to 1.39.2, skipping the issue with 1.39.1


Cargo.toml
```diff
@@ -277,11 +277,11 @@ tar = "0.4.40"
...
 time = "0.3.30"
-tokio = "1.33.0"
-tokio-stream = "0.1.14"
-tokio-test = "0.4.2"
+tokio = "1.39.1"
+tokio-stream = "0.1.15"
+tokio-test = "0.4.4"
 tokio-tungstenite = { version = "0.20.1" }
-tokio-util = "0.7.10"
+tokio-util = "0.7.11"
```
dynco-nym: why not specify only minor, without patch versions?

dynco-nym (approved):
lgtm :+1:

Cargo.toml
octol: There is no good reason, just inertia. But it's a good point, I think I will start explicitly skip patch versions for >1.0 in the future

PR #4733, opened 2024-07-26
Fix version 1 not having template correspondent initially

neacsu:


benedettadavico (approved):

benedettadavico:
- Tested updating an old nym-node version and ensuring it did not throw any errors.

PR #4734, opened 2024-07-28
[DOCs/operators]: Syntax fix

serinko:
This PR clarifies to operators that they have to use `sudo` prefix or run commands from a root shell to setup their VPS with system tools. Besides that it adds `8080` to ports opened by default for all types of nodes.


serinko:
@mfahampshire If you could please run CI/CD and deploy vercel when merged. Thank you.

mfahampshire (approved):

PR #4742, opened 2024-08-01
Fix clippy for beta toolchain

octol:
- Fix beta toolchain clippy by removing unused module
- Add nym- prefix to serde-common crate
- Remove ignored `default-features = false` attribute for workspace dependency


jstuczyn (approved):

jstuczyn:
given it's unused and it's also rocket specific, can we just remove it?

octol:
> given it's unused and it's also rocket specific, can we just remove it?

Sounds like a plan, removing the whole module

jstuczyn (approved):

PR #4744, opened 2024-08-02
Nym Delegation Program Rules

JohnSmith-PrivacyRebels:
These are proposed Rules to govern DP and keep everything in one place rather than fragmented on various platforms.


sudonym2 (changes requested):
Can we please remove 4/e? It is a duplicate of 4/h.

JohnSmith-PrivacyRebels:
Done, 4e removed.

tommyv1987:
@sudonym2 what do we want to do with this ticket?

PR #4755, opened 2024-08-06
[DOCs]: Catching more broken URLs

serinko:
Ran a few more tests after the last deployment and have a few URLs to be fixed in order for the docs to load seamlessly from old sources.


documentation/dev-portal/src/clients/socks5-client.md
```diff
@@ -60,7 +60,7 @@ There are 2 pieces of software that work together to send SOCKS traffic through
 The `nym-socks5-client` allows you to do the following from your local machine:
 * Take a TCP data stream from a application that can send traffic via SOCKS5.
 * Chop up the TCP stream into multiple Sphinx packets, assigning sequence numbers to them, while leaving the TCP connection open for more data
-* Send the Sphinx packets through the mixnet to a [network requester](https://nymtech.net/operators/nodes/network-requester.md). Packets are shuffled and mixed as they transit the mixnet.
+* Send the Sphinx packets through the mixnet to a Network requester (a part of Nym Exit Gateway). Packets are shuffled and mixed as they transit the mixnet.
```
mfahampshire: either `network requester` or `Network Requester` imo

mfahampshire: I suppose the qu is whether we just start getting rid of the term entirely since its functionality is folded into the Exit Gateways anyway? Instead of saying 'NR as part of an EG' maybe we can just say 'run packets through the mixnet'?

mfahampshire (approved):

documentation/dev-portal/src/clients/socks5-client.md
mfahampshire: I prefer the idea of starting to remove the NR stuff entirely and we can just talk about EGs

serinko: definitely. I will do it like that. We will then redo the flow entirely elsewhere.

PR #4759, opened 2024-08-13
Make gateway latency check generic

octol:
Replace concrete gateway type with trait in latency check, so we can make use of it in the vpn client


neacsu (approved):

benedettadavico:
Testing Completion Date: [19/09/2024]

Testing Steps Performed:
Inited new nym-client with the `--latency-based-selection` flag and ensured it still works as normal. 
 
Validation:
✅ Test Passed

PR #4761, opened 2024-08-13
Serve OpenAPI for all endpoints

dynco-nym:
All endpoints should work as expected


common/cosmwasm-smart-contracts/mixnet-contract/src/gateway.rs
```diff
@@ -6,10 +6,11 @@ use cosmwasm_schema::cw_serde;
 use cosmwasm_std::{Addr, Coin};
 use std::cmp::Ordering;
 use std::fmt::Display;
+use utoipa::ToSchema;
 
 /// Information provided by the node operator during bonding that are used to allow other entities to use the services of this node.
 #[cw_serde]
-#[derive(PartialOrd)]
+#[derive(PartialOrd, ToSchema)]
```
jstuczyn: can you feature-lock utoipa in here? the problem with that type (and everything in `common/cosmwasm-smart-contracts`) is that it's pulled by our smart contracts which creates 2 problems:
- bloats the contract sizes
- (not sure if that's the case in this instance) possibly makes contracts incompatible by introducing float operations. for example some serde's features don't work correctly.

you can see an example of feature-locking utoipa here: https://github.com/nymtech/nym/blob/10ff165c187dd04180442fe126dceac2db39a644/common/exit-policy/src/policy/address_policy.rs#L238

jstuczyn: and I guess this comment will apply to everything in this crate

common/network-defaults/Cargo.toml
```diff
@@ -13,11 +13,12 @@ log = { workspace = true, optional = true }
 schemars = { workspace = true, features = ["preserve_order"], optional = true }
 serde = { workspace = true, features = ["derive"], optional = true }
 url = { workspace = true, optional = true }
+utoipa = { workspace = true }
 
 # please be extremely careful when adding new dependencies because this crate is imported by the ecash contract,
```
jstuczyn, line 18: do the same with utoipa plz : )

nym-api/nym-api-requests/src/models.rs
```diff
@@ -115,19 +116,45 @@ pub struct MixnodeStatusResponse {
...
+    pub rewarding_details: String,
+
+    // TODO dz copy from cosmwasm crate
+    /// Adjustments to the mixnode that are ought to happen during future epoch transitions.
+    pub pending_changes: String,
+}
+
+#[derive(ToSchema)]
+#[schema(as = String, title = Decimal)]
+pub struct DecimalSchema;
```
jstuczyn: do you need a separate `DecimalSchema`? iirc. for all intents and purposes it behaves like string when serialised

nym-api/src/network/handlers.rs
```diff
@@ -2,33 +2,59 @@
...
     extract::State(state): extract::State<AxumAppState>,
 ) -> axum::Json<NetworkDetails> {
     state.network_details().to_owned().into()
 }
 
+// it's used for schema generation so dead_code is fine
+#[allow(dead_code)]
+#[derive(ToSchema)]
+#[schema(title = "ContractVersion")]
+pub(crate) struct ContractVersionSchemaResponse {
```
jstuczyn, line 39: why do we have to re-define that struct? couldn't we have used `ContractInformation` directly with the aliases?

nym-api/src/nym_contract_cache/handlers.rs
```diff
@@ -17,59 +16,52 @@ use std::collections::HashSet;
...
+    )
+)]
 async fn get_mixnodes(
     extract::State(state): extract::State<AxumAppState>,
 ) -> axum::Json<Vec<MixNodeDetails>> {
     state.nym_contract_cache().mixnodes_filtered().await.into()
 }
 
-// TODO dz:see deprecated warnings below
+// TODO dz:is this warning still relevant?
```
jstuczyn, line 64: eh. kinda. it's complicated. we can't remove it because people might be relying on it. I've explicitly deprecated all the routes in swagger on my directory v3 branch

nym-api/src/support/http/mod.rs
```diff
@@ -173,27 +245,24 @@ pub(crate) async fn setup_routes(network_monitor: bool) -> anyhow::Result<Router
...
-        .merge(node_status_routes(network_monitor))
-        .merge(nym_network_routes())
-        .merge(status::handlers::api_status_routes())
-        .merge(nym_node_routes())
-        .merge(nym_node_routes_unstable())
-        // CORS layer needs to be "outside" of routes
         .layer(setup_cors());
 
     // needs to be after routes we want to log
     let router = router.layer(axum::middleware::from_fn(logger));
```
jstuczyn, line 271: nitpicking: just layer the logger middleware when you're creating the original `Router`

nym-api/src/network/handlers.rs
dynco-nym: > couldn't we have used ContractInformation directly with the aliases

what do you mean by this?

nym-api/nym-api-requests/src/models.rs
dynco-nym: didn't end up using this, removed

common/cosmwasm-smart-contracts/mixnet-contract/src/gateway.rs
dynco-nym: makes sense, I'll feature-guard it

nym-api/src/support/http/mod.rs
dynco-nym: this is intentional: logging needs to be outermost layer. In fact, I'm working on a design that will ensure logging and CORS layers are outermost and nobody can accidentally add a layer in the wrong place

nym-api/nym-api-requests/src/models.rs
```diff
@@ -115,19 +116,41 @@ pub struct MixnodeStatusResponse {
...
+    #[schema(value_type = String)]
     pub last_hour: Performance,
+    #[schema(value_type = String)]
     pub last_24h: Performance,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
+#[derive(ToSchema)]
+#[schema(title = "MixNodeDetails")]
+pub struct MixNodeDetailsSchema {
```
jstuczyn, line 131: nitpicking: this doesn't seem to be used anywhere, perhaps we should remove it then?
unless you wanted to use it for some schemar placeholder

nym-api/nym-api-requests/src/models.rs
```diff
@@ -338,45 +368,61 @@ pub struct MixnodeStatusReportResponse {
...
 pub struct GatewayUptimeHistoryResponse {
     pub identity: String,
     pub owner: String,
     pub history: Vec<HistoricalUptimeResponse>,
 }
 
-#[derive(Clone, Serialize, Deserialize, schemars::JsonSchema)]
+#[derive(ToSchema)]
+#[schema(title = "Coin")]
+pub struct CoinSchema {
```
jstuczyn, line 405: nitpicking: not sure if clippy/utoipa would even allow that (if not, plz do ignore this comment), but do you reckon it would perhaps make sense to **not** make this type public? 
I don't see any reasons for library consumers to ever use this type

nym-api/src/circulating_supply_api/handlers.rs
```diff
@@ -12,21 +11,25 @@ use nym_validator_client::nyxd::Coin;
...
 pub(crate) fn circulating_supply_routes() -> Router<AxumAppState> {
     Router::new()
+        .route("/", axum::routing::get(get_full_circulating_supply))
         .route(
-            &static_routes::v1::circulating_supply(),
-            axum::routing::get(get_full_circulating_supply),
-        )
-        .route(
-            &static_routes::v1::circulating_supply::circulating_supply_value(),
+            "/total-supply-value",
```
jstuczyn, line 16: just curious, why did you opt to go back to hardcoded routes rather than using the values you defined previously?

jstuczyn: (if it works, it works, no need to change it now)

nym-api/src/support/http/mod.rs
jstuczyn: aha! makes sense, good thinking!

jstuczyn (approved):

nym-api/src/circulating_supply_api/handlers.rs
dynco-nym: > why did you opt to go back to hardcoded routes rather than using the values you defined previously?

because of paths with parameters. They are defined in axum as 
- `/v1/api/path/:param`

while in utoipa annotations paths look like this
- `/v1/api/path/{param}`

which means I cannot reuse the predefined constants. If one is predefined as a constant, and I have to rewrite the other, then I don't see a point of having a constant at all.

nym-api/nym-api-requests/src/models.rs
dynco-nym: it's `utoipa`: I need to keep it `pub` (even `pub(crate)` doesn't work), otherwise [discovery](https://github.com/nymtech/nym/blob/dz-swagger-docs/nym-api/src/support/http/mod.rs#L170) doesn't work:

```rust
error[E0603]: struct `CoinSchema` is private
   --> nym-api/src/support/http/mod.rs:170:1
    |
170 | #[utoipauto(paths = "./nym-api/src")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | private struct
    | struct `CoinSchema` is not publicly re-exported
    |
note: the struct `CoinSchema` is defined here
   --> /nym/nym-api/nym-api-requests/src/models.rs:407:1
    |
407 | pub(crate) struct CoinSchema {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

nym-api/nym-api-requests/src/models.rs
dynco-nym: corrected, tnx

PR #4766, opened 2024-08-20
Avoid race on ip and registration structures

neacsu:
To avoid a state where the ip is being cleared out before the registration is also cleared out, couple the two structures under the same lock, since they are anyway very inter-dependent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6242,7 +6242,6 @@ name = "nym-wireguard-types"
 version = "0.1.0"
 dependencies = [
  "base64 0.21.7",
- "dashmap",
  "hmac",
  "log",
  "nym-config",
--- a/common/wireguard-types/Cargo.toml
+++ b/common/wireguard-types/Cargo.toml
@@ -12,7 +12,6 @@ license.workspace = true
 
 [dependencies]
 base64 = { workspace = true }
-dashmap = { workspace = true }
 log = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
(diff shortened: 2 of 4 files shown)


octol (approved):

tommyv1987:
**Testing Completion Date:** 16/09/2024

**Testing Steps Performed:**

1. - **Checked out** the release/2024.10-wedel branch containing the fix for the race condition on IP and registration structures
   - **Deployed** the on a controlled test environment to prevent interference

2. **Monitored Logs:**

   - **Enabled** debug logging to capture all events
   - **Monitored** logs in real-time to observe the handling of concurrent registration requests
   - **Checked** for any error messages, warnings, or indications of race conditions

3. **Verified Client Responses:**

   - Ensured that all clients received appropriate responses:
     - Successful registration with assigned IP and registration data
     - Appropriate error messages if no IPs were available or if other issues occurred
   - Confirmed that no clients were left in an inconsistent state (e.g., assigned an IP but not fully registered)

4. **Validated Normal Operation:**

    - **Conducted standard registration processes** with individual clients to confirm that regular functionality is unaffected via nym-vpn-cli
    - Ensured that authenticated clients could communicate over the network as expected

**Validation:**

- [X] Test Passed

**Notes**

- **No race conditions were observed** during the concurrent registration tests
- **Stale registrations were correctly identified and removed** after the timeout period, and associated IP addresses were freed
- **Logs did not contain any errors or panics** related to the registration process or IP management

PR #4775, opened 2024-08-21
Introduce axum feature flag

dynco-nym:


nym-api/src/support/config/mod.rs
```diff
@@ -246,9 +245,7 @@ pub struct Base {
     #[zeroize(skip)]
     pub local_validator: Url,
 
-    // TODO dz Andrew added this here ?!
-    /// Socket address this api will use for binding its http API.
-    /// default: `0.0.0.0:8080`
+    /// Socket address Axum will use for binding its HTTP API.
     #[zeroize(skip)]
     #[serde(default = "default_http_socket_addr")]
     pub bind_address: SocketAddr,
```
dynco-nym, line 252: @jstuczyn  I couldn't find this config value being used anywhere, could you confirm it isn't?

nym-api/src/support/config/template.rs
```diff
@@ -17,6 +15,9 @@ id = '{{ base.id }}'
 # Validator server to which the API will be getting information about the network.
 local_validator = '{{ base.local_validator }}'
 
+# Socket address this api will use for binding its http API.
```
jstuczyn, line 18: perhaps add a comment that this will only apply when `axum` feature flag is used

nym-api/src/support/cli/run.rs
```diff
@@ -62,24 +62,29 @@ pub(crate) async fn execute(args: Args) -> anyhow::Result<()> {
 
     config.validate()?;
 
-    let mut axum_shutdown = start_nym_api_tasks_v2(&config).await?;
-    let mut shutdown_handlers = start_nym_api_tasks(config).await?;
+    #[cfg(feature = "axum")]
```
jstuczyn, line 65: shouldn't it be the case that if we're using `axum` feature, then the rocket server does **not** startup? I think we just want one or the other, no?

durch (approved):

nym-api/src/support/cli/run.rs
dynco-nym: We might need both of them (as they are now) for testing purposes, to compare whether `axum` does the same ting as `rocket `

nym-api/src/support/config/template.rs
dynco-nym: done

jstuczyn (approved):

PR #4778, opened 2024-08-23
Enable dependabot version upgrades for root rust workspace

octol:


.github/dependabot.yml
```diff
@@ -14,3 +14,18 @@ updates:
         prefix: build
         prefix-development: chore
         include: scope
+  # Update the root workspace (only). For now we don't include
+  # the contracts workspcae.
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
```
dynco-nym, line 22: >When you set a weekly update schedule, by default, Dependabot checks for new versions on Monday at a random set time for the repository

I suggest setting [`schedule.time`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime) to something like 9:00, so it's not random

dynco-nym (approved):

PR #4801, opened 2024-08-27
Run cargo-autoinherit for a few new crates

octol:
- **Run cargo-autoinherit for a few new crates**
- **Sort crates list**


Cargo.toml
```diff
@@ -310,6 +313,7 @@ tungstenite = { version = "0.20.1", default-features = false }
 url = "2.4"
 utoipa = "4.2.0"
 utoipa-swagger-ui = "6.0.0"
+uuid = "*"
```
dynco-nym, line 316: why is this the only one with wildcard for a version?

octol: Good question, not sure. But I assume it was added for a good reason

octol: Also imho I'd prefer not to change it as part of this PR

jstuczyn (approved):

PR #4821, opened 2024-08-28
using legacy signing payload in CLI and verifying both variants in contract

jstuczyn:


benedettadavico (approved):

simonwicky (approved):

benedettadavico:
using a mixnet contract version from release `chomp`, bonding nodes from this branch is no longer breaking

PR #4825, opened 2024-08-29
Expose wireguard details on self described endpoint

neacsu:


octol (approved):

benedettadavico:
Testing Completion Date: [17/09/2024] 

Tests:
Wireguard details are now visible at the nym-node endpoint `/api/v1/gateway/client-interfaces` as well as on the nym-api self-described endpoint `/api/v1/gateways/described`, above the existing data displaying mixnet_websocket information. 

An example of what will be shown is: 
```
"wireguard": {
"port": 51822,
"public_key": "<some public key here>"
}
```
 
Validation:
✅ Test Passed

PR #4834, opened 2024-09-02
Eliminate cancel unsafe sig awaiting

neacsu:


gateway/src/node/client_handling/websocket/connection_handler/fresh.rs
```diff
@@ -720,111 +728,125 @@ where
...
-            };
+        let mut shutdown = self.shutdown.clone();
+        loop {
+            tokio::select! {
+                biased;
+                _ = shutdown.recv() => {
+                    trace!("received shutdown signal while performing initial authentication");
+                    return Err(InitialAuthenticationError::ReceivedShutdown);
+                },
+                msg = self.read_websocket_message() => {
```
pronebird, line 739: Nit, you might find it shorter to rewrite that with else-statement:

```rs
tokio::select! {
    biased;
    _ = shutdown.recv() => {
        // ..
    },
    Some(msg) = self.read_websocket_message() => {
        // ..
    },
    else => break,
}
```

gateway/src/node/client_handling/websocket/listener.rs
```diff
@@ -64,9 +65,9 @@ where
                                 active_clients_store.clone(),
                                 self.shared_state.clone(),
                                 remote_addr,
+                                shutdown,
                             );
-                            let shutdown = shutdown.clone().named(format!("ClientConnectionHandler_{remote_addr}"));
-                            tokio::spawn(async move { handle.start_handling(shutdown).await });
+                            tokio::spawn(async move { handle.start_handling().await });
```
pronebird: Nit: not certain but you may probably give it `handle.start_handling()` without await and the outer async block. Unless move semantics get in the way...

pronebird (approved):

gateway/src/node/client_handling/websocket/connection_handler/fresh.rs
neacsu: I like the current form a bit more because it's more explicit. In case a different select branch gets added with the match pattern, it seems to be it wouldn't be clear which one caused the else branch to get executed.

PR #4835, opened 2024-09-02
Update node_api_check.py & Documentation

serinko:
The CLI for operators self checking is now synced with the current API endpoints.


pwnfoo (approved):
LGTM

PR #4844, opened 2024-09-08
Fix test failure in ipr request size

octol:
Nightly build started failing due to a unit test using `now()`, changing the serialized size. Fixed to use a fixed date.


jstuczyn (changes requested):
as discussed this won't do as it will break in few months again anyway

octol:
Good catch by @jstuczyn, this is due to the timestamp using `now()`

jstuczyn (approved):

PR #4854, opened 2024-09-09
Node Status API background task

dynco-nym:
- code location up for discussion
- Configuration for one or more Nym APIs to use
- Background task to query for network information
- Store data in sqlite
- Migrations configured and working


.github/workflows/ci-build.yml
```diff
@@ -71,6 +71,19 @@ jobs:
           command: fmt
           args: --all -- --check
 
+      # due to sqlx dependency, prepare environment before building this package
+      - name: Pre-build steps for `node-status-api`
+        run: |
+          cargo install sqlx-cli
```
octol: heads up, let's add this to the runner image as a follow up to this PR so we don't have to build it on each run

nym-node-status-api/src/db/helpers.rs
```diff
@@ -0,0 +1 @@
+
```
octol: Empty file?

nym-node-status-api/src/error.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[error("Client connection error: {0}")]
+    ClientConnectionError(String),
+    #[error("DB error: {0}")]
+    DatabaseError(String),
+    #[error("Internal: {0}")]
+    Internal(String),
+}
+
+impl From<ExplorerApiError> for Error {
+    fn from(value: ExplorerApiError) -> Self {
```
octol: At first I was going to write that these can be auto-generated by `thiserror` by adding a `#[from]`. But then I saw you're using anyhow and this is a binary anyway that will not need to feed strongly typed errors anywhere.

So in that sense, if you want to you can consider just dropping this error enum entirely and just use anyhow everywhere, attaching context when needed.

That's probably a matter of taste, but something to consider

nym-node-status-api/src/logging.rs
```diff
@@ -0,0 +1,37 @@
...
+        .from_env_lossy()
+        // these crates are more granularly filtered
+        .add_directive(directive_checked("nym_bin_common=warn"))
+        .add_directive(directive_checked("nym_explorer_client=warn"))
+        .add_directive(directive_checked("nym_network_defaults=warn"))
+        .add_directive(directive_checked("nym_validator_client=warn"))
+        .add_directive(directive_checked("reqwest=error"))
+        .add_directive(directive_checked("rustls=error"))
+        .add_directive(directive_checked("hyper=error"))
+        .add_directive(directive_checked("sqlx=error"))
```
octol: This disables all warnings right? imho let's show at least info and up for all these crates. I tried running the app and I didn't see any of these crates logging anything anyway at INFO level when I removed all the directives so there is no spam happening that needs to be hidden

nym-node-status-api/launch_node_status_api.sh
```diff
@@ -0,0 +1,64 @@
...
+
+if [ "$clear_db" = true ] || [ "$init_only" = true ]; then
+    init_db
+fi
+
+if [ "$init_only" = true ]; then
+    exit 0
+fi
+
+cargo run --package nym-node-status-api -- --config-env-file ../envs/mainnet.env
```
octol: No need to pass in mainnet.env. Actually, if we're doing that then I would double check that mainnet.env is actually up to date with the `nym-network-defaults` crate that other code uses

(Side comment: I wish we had a test or something that confirmed that they are in sync)

nym-node-status-api/src/monitor/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+    })
+}
+
+async fn run(pool: &DbPool, network_details: &NymNetworkDetails) -> NodeStatusApiResult<()> {
+    let default_api_url = network_details
+        .endpoints
+        .first()
+        .expect("rust sdk mainnet default incorrectly configured")
+        .api_url
```
octol: you can call `.api_url()` here to make this call chain shorter

nym-node-status-api/src/monitor/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        .parse()
+        .expect("rust sdk mainnet default api_url not parseable");
+    let default_explorer_url = network_details.explorer_api.clone().map(|url| {
+        url.parse()
+            .expect("rust sdk mainnet default explorer url not parseable")
+    });
+
+    let default_explorer_url =
+        default_explorer_url.expect("explorer url missing in network config");
+    let explorer_client = ExplorerClient::new(default_explorer_url)?;
```
octol, line 69: I thought that explorer-api was going to be removed soonish? What is the reason of querying it? Just for completeness?

nym-node-status-api/src/monitor/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+    let api_client = NymApiClient::new(default_api_url);
+    let gateways = api_client.get_cached_described_gateways().await?;
+    tracing::debug!("Fetched {} gateways", gateways.len());
+    let skimmed_gateways = api_client.get_basic_gateways(None).await?;
+
+    let mixnodes = api_client.get_cached_mixnodes().await?;
+    tracing::debug!("Fetched {} mixnodes", mixnodes.len());
+
+    // TODO dz can we calculate blacklisted GWs from their performance?
+    // where do we get their performance?
```
octol, line 81: The performance is exposed in the skimmed endpoint I think

nym-node-status-api/src/db/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub(crate) struct Storage {
+    pool: DbPool,
+}
+
+impl Storage {
+    pub async fn init() -> Result<Self> {
+        let connection_url = read_env_var(DATABASE_URL_ENV_VAR)?;
+
+        let pool = sqlx::SqlitePool::connect(&connection_url)
```
octol: What does it connect to if you just run without setting `DATABASE_URL_ENV_VAR`?

I would add some logging here so that when you run it, it will log where it's storing stuff. I ran it locally and it wasn't showing anything.

Also I suggest using `disable_statement_logging()` here to not flood the log. That way you can remove that logging directive for sqlx, so that we don't hide other relevant logging for it.

nym-node-status-api/launch_node_status_api.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
octol, line 1: General comment, not necessarily something you need to address for this PR (up to you).

But do we even need a script? We can also let the Rust code do all this and not have to worry about bash. This would seem simpler imho

nym-node-status-api/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    logging::setup_tracing_logger();
+
+    let args = cli::Cli::parse();
+    // if dotenv file is present, load its values
+    // otherwise, default to mainnet
+    setup_env(args.config_env_file.as_ref());
+    tracing::trace!("{:?}", std::env::var("NETWORK_NAME"));
```
octol: Optional and subjective: I would place these at DEBUG instead of TRACE

nym-node-status-api/launch_node_status_api.sh
```diff
@@ -0,0 +1,64 @@
...
+# export DATABASE_URL as absolute path due to this
+# https://github.com/launchbadge/sqlx/issues/3099
+db_filename="nym-node-status-api.sqlite"
+script_abs_path=$(realpath "$0")
+package_dir=$(dirname "$script_abs_path")
+db_abs_path="$package_dir/data/$db_filename"
+dotenv_file="$package_dir/.env"
+echo "DATABASE_URL=sqlite://$db_abs_path" > "$dotenv_file"
+
+export RUST_LOG=trace
```
octol: Is it really practical to run at TRACE? Sounds like something you'd only set if you need to debug or investigate issues or similar

octol (approved):
Approved. Left some comments

nym-node-status-api/src/db/helpers.rs
dynco-nym: removed

nym-node-status-api/src/error.rs
dynco-nym: I started with that enum, but couldn't find a strong reason to keep expanding it. 

switching to anyhow

nym-node-status-api/src/logging.rs
dynco-nym: I'm against `info` because too low threshold can get spammy when you need to troubleshoot a running server 

for example (as you noted) `sqlx` uses `info` to log every query unless you suppress it, don't know how other crates behave and I currently have no integration tests to see real life logs

nym-node-status-api/src/monitor/mod.rs
dynco-nym: nice, done

nym-node-status-api/launch_node_status_api.sh
dynco-nym: build script now does this so the script is not mandatory

but I like it for convenience: as opposed to remembering the right combination of cargo flags: specifying package, env file, CLI argument for the binary etc. I don't run it in CI

nym-node-status-api/src/main.rs
dynco-nym: changed

nym-node-status-api/launch_node_status_api.sh
dynco-nym: I understand, removed CLI arg

nym-node-status-api/launch_node_status_api.sh
dynco-nym: changed it to `debug`, allowing override from environment

nym-node-status-api/src/monitor/mod.rs
dynco-nym: it's already being queried in HM, copying functionality from there. As far as I see, it's used for stats keeping

nym-node-status-api/src/db/mod.rs
dynco-nym: > What does it connect to if you just run without setting DATABASE_URL_ENV_VAR?

nothing, it fails during start up because it needs DB path

> I would add some logging here so that when you run it, it will log where it's storing stuff. I ran it locally and it wasn't showing anything.

it [does log](https://github.com/nymtech/nym/blob/97642507e21dc9d96ed5221247ed72d9991b3777/nym-node-status-api/src/main.rs#L35) env var after successfully reading it

dynco-nym: > Also I suggest using disable_statement_logging() here to not flood the log. That way you can remove that logging directive for sqlx, so that we don't hide other relevant logging for it.

done

nym-node-status-api/src/logging.rs
dynco-nym: changed to `warn`

PR #4856, opened 2024-09-09
Bugfix/client registration vol2

jstuczyn:


neacsu (approved):

common/gateway-storage/src/shared_keys.rs
```diff
@@ -40,9 +40,17 @@ impl SharedKeysManager {
         client_address_bs58: String,
         derived_aes128_ctr_blake3_hmac_keys_bs58: String,
     ) -> Result<i64, sqlx::Error> {
-        sqlx::query!("INSERT OR REPLACE INTO shared_keys(client_address_bs58, derived_aes128_ctr_blake3_hmac_keys_bs58) VALUES (?, ?)",
+        // https://stackoverflow.com/a/20310838
+        // we don't want to be using `INSERT OR REPLACE INTO` due to the foreign key on `available_bandwidth` if the entry already exists
```
neacsu, line 44: Just to double check, this is because of other table referencing `shared_keys`, right? We have the `INSERT OR REPLACE` command too https://github.com/nymtech/nym/blob/7e40207d4691610fef14d903d36421ffff3d60e1/common/gateway-storage/src/wireguard_peers.rs#L29 , but it's not with any foreign key, so it's probably safe _for now_. Maybe we should use the same approach there as well, in case things change in the future?

PR #4864, opened 2024-09-11
Disable push trigger and add missing paths in ci-build

octol:


.github/workflows/ci-build.yml
```diff
@@ -1,24 +1,25 @@
...
+  #    - 'sdk/rust/nym-sdk/**'
+  #    - 'service-providers/**'
+  #    - 'nym-api/**'
+  #    - 'nym-node/**'
+  #    - 'nym-outfox/**'
+  #    - 'nym-validator-rewarder/**'
+  #    - 'tools/**'
+  #    - 'Cargo.toml'
+  #    - 'Cargo.lock'
+  #    - '.github/workflows/ci-build.yml'
```
dynco-nym: why comment out and not delete?

octol: There is no good answer to that, let me delete that!

jstuczyn (approved):

PR #4868, opened 2024-09-11
Gateway database modifications for different modes

neacsu:
As gateway clients will not be solely from the mixnet, we need to split the table that handles shared keys from the client ids that are referenced from other tables. That way, the bandwidth table can be shared between different client types (entry mixnet, entry gateway, exit gateway), using the same `client_id` referencing.

### DIFF
--- a/common/gateway-storage/migrations/20240910120000_generic_client_id.sql
+++ b/common/gateway-storage/migrations/20240910120000_generic_client_id.sql
@@ -5,11 +5,11 @@
 
 CREATE TABLE clients (
    id              INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
-   client_type     TEXT NOT NULL CHECK(client_type IN ('entry mixnet', 'entry wireguard', 'exit wireguard'))
+   client_type     TEXT NOT NULL CHECK(client_type IN ('entry_mixnet', 'exit_mixnet', 'entry_wireguard', 'exit_wireguard'))
 );
 
 INSERT INTO clients (id, client_type)
-SELECT id, 'entry mixnet'
+SELECT id, 'entry_mixnet'
 FROM shared_keys;
 
 CREATE TABLE shared_keys_tmp (
--- a/common/gateway-storage/migrations/20240910120000_generic_client_id.sql
+++ b/common/gateway-storage/migrations/20240910120000_generic_client_id.sql
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+CREATE TABLE clients (
+   id              INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+   client_type     TEXT NOT NULL CHECK(client_type IN ('entry mixnet', 'entry wireguard', 'exit wireguard'))
+);
+
+INSERT INTO clients (id, client_type)
+SELECT id, 'entry mixnet'
+FROM shared_keys;
+
+CREATE TABLE shared_keys_tmp (
+   client_id                                INTEGER NOT NULL PRIMARY KEY REFERENCES clients(id),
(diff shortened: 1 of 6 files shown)


octol (approved):

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
```diff
@@ -0,0 +1,98 @@
...
+   expiration TIMESTAMP WITHOUT TIME ZONE
+);
+
+INSERT INTO available_bandwidth_tmp
+SELECT * FROM available_bandwidth;
+
+DROP TABLE available_bandwidth;
+ALTER TABLE available_bandwidth_tmp RENAME TO available_bandwidth;
+
+CREATE TABLE received_ticket_tmp (
```
jstuczyn, line 36: why are you recreating this table if nothing has changed in it? (the same applies to other tables too)

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
```diff
@@ -0,0 +1,98 @@
...
+   id              INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+   client_type     TEXT NOT NULL CHECK(client_type IN ('entry mixnet', 'entry wireguard', 'exit wireguard'))
+);
+
+INSERT INTO clients (id, client_type)
+SELECT id, 'entry mixnet'
+FROM shared_keys;
+
+CREATE TABLE shared_keys_tmp (
+   client_id                                INTEGER NOT NULL PRIMARY KEY REFERENCES clients(id),
```
jstuczyn, line 16: I'm not entirely sure whether this creates the relation you're looking for.
since `shared_keys` references `clients`, you will end up having multiple of those for the same underlying user (but in different mode).

how about something like this:
- the "master" `client` table that has the id, address and shared key
- bandwidth table with extra `client_type` referencing that master client table

so that the same entity would end up having different bandwidth metering based on its type

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+CREATE TABLE clients (
+   id              INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+   client_type     TEXT NOT NULL CHECK(client_type IN ('entry mixnet', 'entry wireguard', 'exit wireguard'))
```
jstuczyn: for completion sake we're missing exit mixnet. also, can you change the typing slightly? I feel like having no spaces in SQL might backfire : )

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
neacsu: Because `received_ticket`'s `ticket_id` references `received_ticket`'s `id`. All the references cascade because of one change. Things would have been much easier if the `ALTER TABLE MODIFY COLUMN` would have worked in sqlite, but since we have to do the create and rename "hack", this affects inter-table dependencies.

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
neacsu: I'm guessing you mean to remove the white spaces. Sure, no problem, I'll also add `exit_mixnet`

common/gateway-storage/migrations/20240910120000_generic_client_id.sql
neacsu: Yes, but what would the shared key and address be for the "master" client that is not of mixnet type then? This is exactly why I created the `clients` table.
Have multiple entries of in `clients` for the same underlying user doesn't seem like a problem to me, as each mode is essentially a different "client", the user doesn't have anything to do with it. The same user will actually have all 4 instances of client modes running at the same time.

PR #4876, opened 2024-09-14
Sync code with .env in build.rs

dynco-nym:
Keep dotenv file always up to date

```git
index 2be5300bd..e40e54a47 100644
--- a/envs/mainnet.env
+++ b/envs/mainnet.env
@@ -25,4 +25,4 @@ NYXD="https://rpc.nymtech.net"
-NYM_VPN_API="https://nymvpn.net/api/"
+NYM_VPN_API="https://nymvpn.com/api"
```

... and we already see its use case :)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4876)

### DIFF
--- a/common/network-defaults/build.rs
+++ b/common/network-defaults/build.rs
@@ -1,11 +1,15 @@
+use cargo_metadata::MetadataCommand;
 use regex::Regex;
-use std::{collections::HashMap, fs};
-
-const TARGET_ENV_FILE: &str = "../../envs/mainnet.env";
+use std::{collections::HashMap, fs, path::PathBuf};
 
 /// Sync variable values defined in code with .env file
 fn main() {
-    let file = include_str!("src/mainnet.rs");
+    let source_of_truth = include_str!("src/mainnet.rs");
+    let mut output_path = workspace_root();
+    output_path.push("envs");
+    output_path.push("mainnet.env");
+
+    println!("{}", output_path.display());
 
     let variables_to_track = [
         "NETWORK_NAME",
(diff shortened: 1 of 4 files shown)


common/network-defaults/build.rs
```diff
@@ -0,0 +1,64 @@
+use regex::Regex;
+use std::{collections::HashMap, fs};
+
+const TARGET_ENV_FILE: &str = "../../envs/mainnet.env";
```
octol: I wonder about this, do you think it's possible to make this relative to the repo root directory? Soon I would like to move the root Rust workspace to a subdirectory to avoid the workspace nesting, and this would then break

Also put in a check that we are writing to the directory we think we are. So that it doesn't try to write the file unless the directory path is correct

octol (changes requested):
I think this is a great idea!

Seeing this I realize that the mainnet default use for nym-vpi-api is wrong, it should be nymvpn.com not .net, so that's great that this code caught that :)

Put in some comments that I think it would be good to make it error if the directory is not correct (due e.g the crate being moved or built from somewhere else). For example what happens when you package this crate using `cargo package`? Another option here is move this code to a unit test that confirms that the file contents match the env file?

common/network-defaults/build.rs
dynco-nym: > make this relative to the repo root directory

done

> Also put in a check that we are writing to the directory we think we are. So that it doesn't try to write the file unless the directory path is correct

done

octol (approved):

PR #4885, opened 2024-09-16
Feature/updated gateway registration

jstuczyn:
This PR introduces support for aes256-gcm-siv shared keys between clients and gateways.

those changes should be fully backwards compatible. if they're not, there's a bug.

To test (I've briefly done it locally, but some more extensive tests would probably be needed):

From the top of my head, the following have to be tested:
I've done some very brief tests locally, such as new client -> old gateway followed by new client -> new gateway to see key being upgraded but by no means were they exhaustive.

- registration/authentication of a FRESH old client with a new gateway - should work as before
- registration/authentication of a FRESH new client with old gateway - should work as before
- registration/authentication of a FRESH new client with new gateway - should **derive** the new key
- authentication of an EXISTING (i.e. already has bunch of keys stored) old client with new gateway - should work as before
- authentication of an EXISTING (i.e. already has bunch of keys stored) new client with new gateway - should **UPGRADE** existing key

I might be stating the obvious here, but whenever I say "existing client", it also implies the associated gateway also has bunch of data stored on its side (since it needs to have the relevant legacy key on its end)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)


common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -448,40 +456,41 @@ impl<C, St> GatewayClient<C, St> {
...
+        let Some(shared_key) = self.shared_key.as_ref() else {
             return Err(GatewayClientError::NoSharedKeyAvailable);
-        }
+        };
+
+        assert!(shared_key.is_legacy());
+        let legacy_key = shared_key.unwrap_legacy();
+
+        let _ = legacy_key;
+        warn!("unimplemented: migration into aes256gcm-siv key!");
```
jstuczyn: this still remains to be implemented

octol (approved):
Looking good! But it's also thousands of lines of changes so ... 

> Reviewed 20 of 50 files at r1, 53 of 53 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885#-:-O78plSMEK-BMahhY6Xm:bf4uv36)* status: all files reviewed, 4 unresolved discussions (waiting on @jstuczyn)

___
*[`common/client-core/gateways-storage/src/backend/mem_backend.rs` line 115 at r2](https://reviewable.io/reviews/nymtech/nym/4885#-O78nHEM5zDGRjIliN5O:-O78nHEN7lc8J1IxJgSa:b-uv3fx9) ([raw file](https://github.com/nymtech/nym/blob/f5863b9668c611e1d805bd18f7f6d1d0841352c2/common/client-core/gateways-storage/src/backend/mem_backend.rs#L115)):*
> ```XML
>             // eh. that's nasty, but it's only ever used for ephemeral clients so should be fine for now...
>             details.shared_key = Arc::new(SharedGatewayKey::Current(
>                 SharedSymmetricKey::try_from_bytes(updated_key.as_bytes()).unwrap(),
> ```

If one would be nitpicky we should don't unwrap and instead error back up, but then again it looks like this one might be one of those that can in practive never fail?
___
*[`common/client-core/src/init/helpers.rs` line 336 at r2](https://reviewable.io/reviews/nymtech/nym/4885#-O78oB6eE2T8S0lEAGYK:-O78oB6eE2T8S0lEAGYL:b-9xgdre) ([raw file](https://github.com/nymtech/nym/blob/f5863b9668c611e1d805bd18f7f6d1d0841352c2/common/client-core/src/init/helpers.rs#L336)):*
> ```XML
>     // we can ignore the authentication result because we have **REGISTERED** a fresh client
>     // (we didn't have a prior key to upgrade/authenticate with)
>     assert!(!auth_response.requires_key_upgrade);
> ```

Just double checking so this will never actually happen?
___
*[`common/gateway-requests/src/registration/handshake/error.rs` line 10 at r2](https://reviewable.io/reviews/nymtech/nym/4885#-O78om64BdoAgObqDyZD:-O78om64BdoAgObqDyZE:b-5fbftc) ([raw file](https://github.com/nymtech/nym/blob/f5863b9668c611e1d805bd18f7f6d1d0841352c2/common/gateway-requests/src/registration/handshake/error.rs#L10)):*
> ```XML
> pub enum HandshakeError {
>     #[error("received key material of invalid length: {received}. Expected: {expected}")]
>     KeyMaterialOfInvalidSize { received: usize, expected: usize },
> ```

I like this, adding context to errors :)

jstuczyn:
*[`common/client-core/gateways-storage/src/backend/mem_backend.rs` line 115 at r2](https://reviewable.io/reviews/nymtech/nym/4885#-O78nHEM5zDGRjIliN5O:-O79k7JeBPpMfXaCIYE-:b-t7xnz4) ([raw file](https://github.com/nymtech/nym/blob/f5863b9668c611e1d805bd18f7f6d1d0841352c2/common/client-core/gateways-storage/src/backend/mem_backend.rs#L115)):*
<details><summary><i>Previously, octol (Jon Häggblad) wrote…</i></summary><blockquote>

If one would be nitpicky we should don't unwrap and instead error back up, but then again it looks like this one might be one of those that can in practive never fail?
</blockquote></details>

I really hate the existence of this one, but the interface behind saving keys, which provides everything with references, does not fully work with the "in memory storage" that keeps everything by value.

I also purposely don't expose any Clone/Copy methods on keys to prevent their misuse, so as a hack I convert it to and back from bytes. The reason I'm unwrapping is this can't possibly fail because it's just feeding its own bytes representation

jstuczyn:
*[`common/client-core/src/init/helpers.rs` line 336 at r2](https://reviewable.io/reviews/nymtech/nym/4885#-O78oB6eE2T8S0lEAGYK:-O79kmcf5DddVB4-1Tbh:br3ko1y) ([raw file](https://github.com/nymtech/nym/blob/f5863b9668c611e1d805bd18f7f6d1d0841352c2/common/client-core/src/init/helpers.rs#L336)):*
<details><summary><i>Previously, octol (Jon Häggblad) wrote…</i></summary><blockquote>

Just double checking so this will never actually happen?
</blockquote></details>

Yes, because for fresh registrations you will always have the up to date key and this code only ever registers the client. However, maybe rather than doing an assertion, I will instead return an error for when this inevitiably changes because of some accidental changes we're going to make in the future...

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)* status: 72 of 74 files reviewed, 5 unresolved discussions (waiting on @jstuczyn and @octol)

___
*[`common/client-core/gateways-storage/src/backend/mem_backend.rs` line 69 at r3](https://reviewable.io/reviews/nymtech/nym/4885#-O79sz8r6xgWrHKIetol:-O79sz8s348s-5f1ylwG:b-g4xpx2) ([raw file](https://github.com/nymtech/nym/blob/be7f00fe527d5dd096efef257997fb429f066e4f/common/client-core/gateways-storage/src/backend/mem_backend.rs#L69)):*
> ```XML
>     }
> 
>     async fn has_gateway_details(&self, gateway_id: &str) -> Result<bool, Self::StorageError> {
> ```

The function is infallible but returns a `Result`

jstuczyn:
*[`common/client-core/gateways-storage/src/backend/mem_backend.rs` line 69 at r3](https://reviewable.io/reviews/nymtech/nym/4885#-O79sz8r6xgWrHKIetol:-O79uAstBbwfRRBEsOUl:b-iy5asy) ([raw file](https://github.com/nymtech/nym/blob/be7f00fe527d5dd096efef257997fb429f066e4f/common/client-core/gateways-storage/src/backend/mem_backend.rs#L69)):*

The function is infallible but returns a `Result`
</blockquote></details>

it's because it's part of the trait method which can fail in other implementations

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)* status: 72 of 74 files reviewed, 4 unresolved discussions (waiting on @jstuczyn and @octol)

___
*[`common/client-core/gateways-storage/src/backend/mem_backend.rs` line 69 at r3](https://reviewable.io/reviews/nymtech/nym/4885#-O79sz8r6xgWrHKIetol:-O7A3_gHEUECQi4emNn3:b-gxztsp) ([raw file](https://github.com/nymtech/nym/blob/be7f00fe527d5dd096efef257997fb429f066e4f/common/client-core/gateways-storage/src/backend/mem_backend.rs#L69)):*
<details><summary><i>Previously, jstuczyn (Jędrzej Stuczyński) wrote…</i></summary><blockquote>

it's because it's part of the trait method which can fail in other implementations
</blockquote></details>

Fair enough.

octol (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)* status: all files reviewed, 2 unresolved discussions (waiting on @jstuczyn and @pronebird)

octol (approved):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)* status: all files reviewed, 1 unresolved discussion (waiting on @jstuczyn)

jstuczyn (commented):
> Reviewed 20 of 50 files at r1, 52 of 53 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/4885)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jstuczyn)

benedettadavico:
Testing Completion Date: [14/10/2024]

Testing Steps Performed:
For the following combinations I inited the client, ran the client, stopped the client, and ran the client again:
fresh client on new binary && gateway on old binary
fresh client on old binary && gateway on new binary
fresh client on new binary && gateway new binary  
existing old client on old binary && new gateway 

Validation:

- [x] ✅ Test Passed
- [ ] 🔴 Issues Found

Notes (if any):
not related, but nonetheless making a note here that when testing the nym-vpn-cli there were errors along the lines of `timeout waiting for connect response from exit gateway`

Attachments (if applicable):
N/A

PR #4901, opened 2024-09-19
Update cargo deny

octol:
Update to use latest cargo-deny.

- Regenerate deny.toml
- Backport old settings to deny.toml
- Explicitly allow GPL-3 only on our own specific crates
- Update deny.toml for latest changes
- Fix cargo-deny warnings for duplicate crates
- Update cargo-deny-action to v2

The security related changes about denying vulnerabilities and explicit ignores was not backported, since we don't run it anyway. Currently we're relying on dependabot for this

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4901)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -113,6 +113,12 @@ exceptions = [
     # Each entry is the crate and version constraint, and its specific allow
     # list
     #{ allow = ["Zlib"], crate = "adler32" },
+    { allow = ["GPL-3.0"], crate = "nym-api" },
+    { allow = ["GPL-3.0"], crate = "nym-gateway" },
+    { allow = ["GPL-3.0"], crate = "nym-mixnode" },
+    { allow = ["GPL-3.0"], crate = "nym-network-requester" },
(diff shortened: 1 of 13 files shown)


jstuczyn (approved):

deny.toml
```diff
@@ -57,37 +65,18 @@ feature-depth = 1
 # More documentation for the advisories section can be found here:
 # https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html
 [advisories]
-# The path where the advisory database is cloned/fetched into
-db-path = "~/.cargo/advisory-db"
+# The path where the advisory databases are cloned/fetched into
+#db-path = "$CARGO_HOME/advisory-dbs"
```
dynco-nym, line 69: is `db-path` intentionally commented out in the new revision?

octol: That's the default value. I'm not sure why it was intentionally set before, I don't see any reason why it has to be or that it makes a difference 

Also we don't check advisories in CI anymore (letting dependabot handle that)

PR #4936, opened 2024-09-26
bugfix: correctly paginate through 'search_tx' endpoint

jstuczyn:
when `results.append(&mut res.txs);` was called, `res.txs` was always empty thus it was impossible to return more than page size number of results

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4936)

### DIFF
--- a/common/client-libs/validator-client/src/nyxd/cosmwasm_client/client_traits/query_client.rs
+++ b/common/client-libs/validator-client/src/nyxd/cosmwasm_client/client_traits/query_client.rs
@@ -218,17 +218,19 @@ pub trait CosmWasmClient: TendermintRpcClient {
 
         loop {
             let mut res = self
-                .tx_search(query.clone(), false, page, 100, Order::Ascending)
+                .tx_search(query.clone(), false, page, per_page, Order::Ascending)
                 .await?;
 
-            results.append(&mut res.txs);
             // sanity check for if tendermint's maximum per_page was modified -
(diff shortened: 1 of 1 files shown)


common/client-libs/validator-client/src/nyxd/cosmwasm_client/client_traits/query_client.rs
```diff
@@ -218,17 +218,19 @@ pub trait CosmWasmClient: TendermintRpcClient {
...
                 .await?;
 
-            results.append(&mut res.txs);
             // sanity check for if tendermint's maximum per_page was modified -
             // we don't want to accidentally be stuck in an infinite loop
-            if res.total_count == 0 || res.txs.is_empty() {
+            let early_break = res.total_count == 0 || res.txs.is_empty();
+            results.append(&mut res.txs);
+
+            if early_break {
```
dynco-nym, line 229: Why instead of 
1. check
2. append
3. if (above check)

don't you just
1. append
2. check
```rust
results.append(&mut res.txs);

if res.total_count == 0 || res.txs.is_empty(); {
```

?

jstuczyn: because appending removes the data from the original vector and thus `res.txs.is_empty()` would always be true - which was the bug I was attempting to resolve : )

octol (approved):

PR #4938, opened 2024-09-27
Switch over the last set of jobs to arc runners

octol:
Switch over the remaining GH jobs using 16-core runners to self-hosted arc runners.

Since we can't currently use Docker on the ubuntu-20.04 runners, remove the matrix notification steps

Confirm that the deployment workflows work through manual testing
- [x] cd-docs
- [x] publish-sdk-npm

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4938)

### DIFF
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


Bykow (approved):
lgtm

PR #4941, opened 2024-09-27
Working HTTP server

dynco-nym:
```
./launch_node_status_api.sh
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4941)


nym-node-status-api/src/config.rs
```diff
@@ -0,0 +1,79 @@
...
+            anyhow::Error::from(e)
+        })
+    }
+
+    fn default_client_timeout() -> Duration {
+        Duration::from_secs(15)
+    }
+
+    fn default_http_port() -> u16 {
+        8000
```
octol, line 35: You can get these default values from the network-defaults crate I think?

octol (approved):

PR #4963, opened 2024-10-11
[Product Data] First step in gateway usage data collection

simonwicky:
## Overview 
This PR is the first step towards collecting data on gateway usage. It builds up on @neacsu's old code for what was then nym-connect.

## What it does
Exposes unique users count and connection time histogram on the `metrics/sessions` endpoint of the self-described API on entry-gateways. For the time being, data is collected by probing the `ActiveClientStore` every minute and extracting data from this. Data is stored internally and exposed on the next day, i.e. UTC day $d$ exposes data from day $d-1$.

## Future plans 
The [`statistics`](https://github.com/nymtech/nym/tree/simon/product_data/gateway/src/node/statistics) module will evolve as we add collection for product data and censorship resistance study. The collection will also eventually switch from probing to event-based for more accurate data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4963)


gateway/src/node/mod.rs
```diff
@@ -636,6 +660,13 @@ impl<St> Gateway<St> {
             shutdown.fork("mixnet_handling::Listener"),
         );
 
+        let shared_session_stats = self.session_stats.take().unwrap_or_default();
```
octol: Maybe just skip starting it if it has not been set? Or maybe it is assumed to always be running?

gateway/src/node/mod.rs
```diff
@@ -636,6 +660,13 @@ impl<St> Gateway<St> {
...
+        self.start_stats_collector(
+            active_clients_store.clone(),
+            shared_session_stats,
+            shutdown.fork("statistics::sessionCollector"),
```
octol: `camelCase` makes me think of Java and immediately get chills down my spine. Let's use `PascalCase` ;)
Also, isn't it called `StatisticsCollector` or `GatewayStatisticsCollector`?

octol (approved):

gateway/src/node/mod.rs
simonwicky: Sorry for the chills :)

Fixed in 65b83627fc31506ef8725adf0cd21632f68a55aa

gateway/src/node/mod.rs
simonwicky: I got a bit baited by that similar construction in [`mixing_stats`](https://github.com/nymtech/nym/blob/08aa0af56242bdc2fcf46f5e07b05557855f61e5/mixnode/src/node/mod.rs#L133).
If we have to create a default one here it won't be passed to the self described API so it's not really useful to start it.

Skipping start in 65b83627fc31506ef8725adf0cd21632f68a55aa

octol (approved):

mmsinclair (approved):

PR #4964, opened 2024-10-11
[Product Data] Second step in gateway usage data collection 

simonwicky:
## Overview 
This PR is the second step towards collecting data on gateway usage. 
It builds on top of #4963 

## What it does
Turn the data collection into an event-based one, which allows more accurate results and more flexibility for future data collection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4964)


common/statistics/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "nym-statistics-common"
+version = "1.1.1"
```
octol: Are we publishing this crate? Thinking about the version. Simplest is probably to just stick to 0.1.0

common/statistics/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition = "2021"
```
octol: Simplest is to stick to the workspace edition

common/statistics/src/events.rs
```diff
@@ -0,0 +1,37 @@
...
+        start_time: OffsetDateTime,
+        client: DestinationAddressBytes,
+    },
+    SessionStop {
+        stop_time: OffsetDateTime,
+        client: DestinationAddressBytes,
+    },
+}
+
+pub fn new_session_start_event(client: DestinationAddressBytes) -> StatsEvent {
```
octol: Add these two functions as constructors on `StatsEvent`?

gateway/src/node/client_handling/active_clients.rs
```diff
@@ -126,6 +130,9 @@ impl ActiveClientsStore {
     /// * `client`: address of the client for which to remove the handle.
     pub(crate) fn disconnect(&self, client: DestinationAddressBytes) {
         self.inner.remove(&client);
+        let _ = self
+            .stats_event_sender
+            .unbounded_send(events::new_session_stop_event(client));
```
octol: I know this was likely copying the pattern from surrounding code, but it was helpfully pointed out to me during another review that we should log errors for stuff like this, and thinking about it I kinda agree.
Imho let's log an error as a defensive action

gateway/src/node/client_handling/active_clients.rs
```diff
@@ -147,6 +154,9 @@ impl ActiveClientsStore {
         if self.inner.insert(client, entry).is_some() {
             panic!("inserted a duplicate remote client")
         }
+        let _ = self
+            .stats_event_sender
+            .unbounded_send(events::new_session_start_event(client));
```
octol: Same as comment above

gateway/src/node/mod.rs
```diff
@@ -623,6 +624,11 @@ impl<St> Gateway<St> {
                 return Err(GatewayError::InsufficientNodeBalance { account, balance });
             }
         }
+        let shared_session_stats = self.session_stats.take().unwrap_or_default();
+        let stats_event_sender = self.start_stats_collector(
+            shared_session_stats,
+            shutdown.fork("statistics::sessionCollector"),
```
octol: See comment in the other review about the name :)

gateway/src/node/statistics/mod.rs
```diff
@@ -1,4 +1,63 @@
...
+use tracing::trace;
+
+pub mod sessions;
+
+const STATISTICS_UPDATE_TIMER_INTERVAL: Duration = Duration::from_secs(3600); //update timer, no need to check everytime
+
+pub(crate) struct GatewayStatisticsCollector {
+    stats_event_rx: StatsEventReceiver,
+    session_stats: SessionStatsHandler,
+    //here go additionnal stats handler
```
octol: nit: spelling

gateway/src/node/statistics/mod.rs
```diff
@@ -1,4 +1,63 @@
...
+
+    async fn update_shared_state(&mut self, update_time: OffsetDateTime) {
+        self.session_stats.update_shared_state(update_time).await;
+        //here go additionnal stats handler update
+    }
+
+    pub async fn run(&mut self, mut shutdown: TaskClient) {
+        let mut update_interval = tokio::time::interval(STATISTICS_UPDATE_TIMER_INTERVAL);
+        while !shutdown.is_shutdown() {
+            tokio::select! {
```
octol, line 43: Consider handling the None case for `stats_event_rx.next()`, conveniently by having a `else => ` branch in the select

gateway/src/node/statistics/sessions.rs
```diff
@@ -0,0 +1,85 @@
...
+    //update shared state once a day has passed, with data from the previous day
+    pub(crate) async fn update_shared_state(&mut self, update_time: OffsetDateTime) {
+        let update_date = update_time.date();
+        if update_date != self.last_update_day {
+            let mut shared_state = self.shared_session_stats.write().await;
+            shared_state.update_time = self.last_update_day;
+            shared_state.unique_active_users = self.unique_users.len() as u32;
+            shared_state.session_started = self.sessions_started;
+            shared_state.session_durations = self.finished_sessions.clone();
+            drop(shared_state);
```
octol: nit: as a general thing I would recommend using scope over a explicit drop unless the drop is for a specific purpose

common/statistics/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "nym-statistics-common"
```
octol, line 5: The crate seems to be specifically about session stats (and not necessarily tied to gateway stats?). Probably a more precise name would be helpful, like `nym-session-statistics-common`, or `nym-gateway-statistics-common`, or `nym-gateway-statistics-events`, depending on the future plans for the crate. And the directory having the name matching (minus the `nym-` prefix)

octol (approved):

gateway/src/node/mod.rs
simonwicky: Fixed by the rebasing of this PR on top of the other one

simonwicky:
@octol 
Note from that PR about that comment https://github.com/nymtech/nym/pull/4963#discussion_r1797693450 :
The statistics collector here is made to allow more than just the Session collection module, so it is assumed to be always running. Hence the fact that it's not fixed here (unless you have another suggestion)

common/statistics/Cargo.toml
simonwicky: For now it is only session stats. In the future, I want to add more event types, that will be used by other components across the mono repo. Hence the name is kinda accurate for the end goal of this crate

gateway/src/node/statistics/mod.rs
simonwicky: I can't think of anything we want to do if that's the case. Sure we won't update the internal state, but we still need to update the shared object for the described API, so we can't stop.

common/statistics/Cargo.toml
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

common/statistics/Cargo.toml
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

common/statistics/src/events.rs
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

gateway/src/node/client_handling/active_clients.rs
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

gateway/src/node/client_handling/active_clients.rs
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

gateway/src/node/statistics/mod.rs
simonwicky: Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

gateway/src/node/statistics/sessions.rs
simonwicky: I need to drop the lock so I can reset the stats afterward, without retesting the condition. That being said, I didn't like that drop, so a scope it will be.

Fixed in adf198f6b8696b963643253dd0ee3aa6e745f13c

PR #4967, opened 2024-10-12
Import nym-vpn-api crates

octol:
Import `nym-vpn-api` crates from commit a414f4a2ee0.

Keep these crates in a separate workspace for now. The idea is to add them to the main workspace in time, but this appears to might require some changes to how sqlite is used. Alternatively these issues might go away once we upgrade sqlite in the main workspace.

Also we intend to rename some of these to something like `nym-credential-facade`, and the wasm lib should be incorporated in one of the existing crates in `common`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4967)


nym-vpn-api/vpn-api-lib-wasm/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+[dependencies]
+bs58 = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+time = { workspace = true, features = ["wasm-bindgen"] }
+thiserror.workspace = true
+zeroize = { workspace = true }
+
+# wasm-specific deps
+getrandom = { version = "0.2", features = ["js"] }
```
jstuczyn, line 23: since you're using the monorepo now, you might want to use workspace deps instead

nym-vpn-api/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+[workspace.package]
+authors = ["Nym Technologies SA"]
+repository = "https://github.com/nymtech/nym"
+homepage = "https://nymtech.net"
+documentation = "https://nymtech.net"
+edition = "2021"
+license = "GPL-3.0"
+
+[workspace.dependencies]
+async-trait = "0.1.80"
```
jstuczyn, line 27: remove those in favour of the main workspace?

jstuczyn (approved):

nym-vpn-api/vpn-api-lib-wasm/Cargo.toml
octol: Well these crates are not part of the workspace yet, but will be soon, hopefully after the sqlx upgrade is merged

nym-vpn-api/Cargo.toml
octol: Same as other comment, I agree we should switch to workspace deps asap after it's been embedded into it

PR #4970, opened 2024-10-14
feat: add clap and env vars through clap, add Dockerfile

farbanas:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4970)


nym-data-observatory/src/main.rs
```diff
@@ -8,29 +9,44 @@ mod db;
...
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     logging::setup_tracing_logger();
 
-    // if dotenv file is present, load its values
-    // otherwise, default to mainnet
-    setup_env(Some(".env"));
+    let args = Args::parse();
+
+    setup_env(args.env_file); // Defaults to mainnet if empty
```
dynco-nym, line 32: is this comment up to date?

if so, can you write that default in `Args` definition under `default_value`?
```rust
    /// Path to the environment file
    #[arg(short, long, default_value = None, env = "NYM_DATA_OBSERVATORY_ENV_FILE")]
    env_file: Option<String>,
```

nym-data-observatory/src/db/mod.rs
```diff
@@ -16,8 +16,8 @@ pub(crate) struct Storage {
 }
 
 impl Storage {
-    pub async fn init() -> Result<Self> {
-        let connection_url = std::env::var(DATABASE_URL_ENV_VAR).map_err(anyhow::Error::from)?;
+    pub async fn init(connection_url: Option<String>) -> Result<Self> {
+        let connection_url = connection_url.unwrap();
```
dynco-nym: can you get rid of `unwrap()`? you can use `.ok_or_else` with `anyhow!()`

nym-data-observatory/docker-compose.yml
```diff
@@ -2,12 +2,22 @@ services:
...
-    env_file:
-      - .env
+    environment:
+      POSTGRES_PASSWORD: password
     ports:
       - "5432:5432"
     volumes:
       - pgdata:/var/lib/postgresql/data
 
+  data-observatory:
```
dynco-nym, line 12: can you add postgres as a dependency of data-observatory? it needs postgres to run

nym-data-observatory/src/main.rs
farbanas: that's a comment that other files have because `setup_env` defaults to mainnet if `args.env_file` doesn't exist. I'm not sure if it makes sense to write it in the arg description, what do you think?

dynco-nym: didn't know that, just leave as is then

dynco-nym (approved):

PR #4974, opened 2024-10-15
[Product Data] Add session type based on ecash ticket received

simonwicky:
## What it does
Fire an `EcashTicket` event for the `GatewayStatisticsCollector`, when an Ecash ticket is being accepted. This allows to mark an active session as being a mixnet session or a vpn session.

It also changes the format of the related self-described data, to accommodate that new session type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4974)

### DIFF
--- a/gateway/src/node/statistics/sessions.rs
+++ b/gateway/src/node/statistics/sessions.rs
@@ -5,7 +5,7 @@ use nym_credentials_interface::TicketType;
 use nym_node_http_api::state::metrics::SharedSessionStats;
 use nym_sphinx::DestinationAddressBytes;
 use std::collections::{HashMap, HashSet};
-use time::{Date, OffsetDateTime};
+use time::{Date, Duration, OffsetDateTime};
 
 use nym_statistics_common::events::SessionEvent;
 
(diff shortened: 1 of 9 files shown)


gateway/src/node/statistics/sessions.rs
```diff
@@ -1,23 +1,92 @@
...
+            TicketType::V1MixnetEntry => Self::Mixnet,
+            TicketType::V1MixnetExit => Self::Mixnet,
+            TicketType::V1WireguardEntry => Self::Vpn,
+            TicketType::V1WireguardExit => Self::Vpn,
+        }
+    }
+}
+
+struct FinishedSession {
+    duration: u64, //in milliseconds,
```
octol: imho whenever a field or variable has a unit like seconds, ms, minutes, and is not strongly types, it's preferable to include it in the name like `duration_ms`.

That said, can we change it into a `Duration` instead?

nym-node/nym-node-http-api/src/state/metrics.rs
```diff
@@ -154,22 +154,30 @@ impl SharedSessionStats {
...
 
+type FinishedSessions = Vec<(u64, String)>;
+
 #[derive(Debug, Clone)]
 pub struct SessionStatsState {
     pub update_time: Date,
     pub unique_active_users: u32,
     pub session_started: u32,
-    pub session_durations: Vec<u64>,
+    pub sessions: FinishedSessions,
```
octol, line 164: So it's my understanding that these get purged nightly?

octol (approved):

nym-node/nym-node-http-api/src/state/metrics.rs
simonwicky: Finished sessions get purged nightly yes. Each night, finished sessions of the day get published on the self-described API and purged from the Collector state. Finished sessions that were already published get lost to history. 
Active sessions are not purged, they turn into Finished Sessions when clients disconnects

gateway/src/node/statistics/sessions.rs
simonwicky: Changed in eac43af

PR #4979, opened 2024-10-16
Expanding and Modularizing Metrics collection in Client

jmwample:
# Overview
This PR is a step towards more modular metrics collection in the client, including events based collection of abnormal network failures.

# What it does

Builds upon the existing `PacketStatisticsControl` infrastructure to allow a stats channel to be passed to reporting locations. However, the channel now allows a set of different metrics types to be reported over the channel. Each Modular Metrics type is responsible for serializing / marshalling and resetting it's own stats periodically when prompted by the `MetricsController`.

# Future Plans

Flesh out the collection points for API and Gateway metrics.

Expand on the ways that the `MetricsController` can report metrics. i.e. making it possible to anonymously share metrics with a status API endpoint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4979)


common/client-core/src/client/metrics/api_statistics.rs
```diff
@@ -0,0 +1,104 @@
...
+//!
+//! Metrics collected by the client while attempting to pull config from the API.
+
+use super::MetricsEvents;
+use std::collections::VecDeque;
+
+use nym_metrics::{inc, inc_by};
+
+#[derive(Default, Debug, Clone)]
+struct APIMetrics {
```
octol: nit: capitalization `ApiMetrics`

Also I would consider callig it `NymApiMetrics` to make it less ambiguous which API we're talking about

.... or now that I'm reading the PR a bit further, this might refer to some new metrics api service?

common/client-core/src/client/metrics.rs
```diff
@@ -0,0 +1,324 @@
...
+use futures::future::{FusedFuture, OptionFuture};
+use futures::FutureExt;
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use http_body_util::Full;
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use hyper::body::Bytes;
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use hyper::server::conn::http1;
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use hyper::service::service_fn;
```
octol: Looks like we can group some of these imports to reduce the number of cfg directives

common/client-core/src/client/metrics.rs
```diff
@@ -0,0 +1,324 @@
...
+// Time interval between reporting packet statistics
+const PACKET_REPORT_INTERVAL_SECS: u64 = 2;
+// Interval for taking snapshots of the packet statistics
+const SNAPSHOT_INTERVAL_MS: u64 = 500;
+
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub(crate) enum MetricsType {
+    PacketStatistics,
+    GatewayMetrics,
+    APIMetrics,
```
octol: nit: `ApiMetrics`

common/client-core/src/client/metrics.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    pub(crate) fn report(&self, event: MetricsEvents) {
+        if let Err(err) = self.stats_tx.send(event) {
+            log::error!("Failed to send stats event: {:?}", err);
+        }
+    }
+}
+
+pub(crate) trait MetricsObj: MetricsReporter + Send {
+    fn new() -> Self
```
octol: That's an interesting choice with `new()` as part of the trait. Looking at the code I don't think this is used though? `new()` looks like it's only called on the concrete types?

common/client-core/src/client/metrics.rs
```diff
@@ -0,0 +1,324 @@
...
+    fn marshall(&self) -> std::io::Result<String>;
+}
+
+/// Launches and manages metrics collection and reporting.
+///
+/// This is designed to be generic to allow for multiple types of metrics to be collected and
+/// reported.
+pub(crate) struct MetricsController {
+    /// Keep store the different types of metrics collectors
+    stats: HashMap<MetricsType, Box<dyn MetricsObj>>,
```
octol: It seems like the set of metrics types is closed, by using an enum for the types. But the set of collectors is open by using runtime dynamic dispatch.
It looks to me like you can consider simplifying by sticking to a closed set for the collectors too by wrapping in an enum

common/client-core/src/client/metrics.rs
```diff
@@ -0,0 +1,324 @@
...
+        stats.insert(
+            MetricsType::PacketStatistics,
+            Box::new(packet_statistics::PacketStatisticsControl::new()),
+        );
+
+        stats.insert(
+            MetricsType::GatewayMetrics,
+            Box::new(gateway_conn_statistics::GatewayMetricsControl::new()),
+        );
+        stats.insert(
```
octol: I would suggest to set these up statically instead of dynamically, for stronger compile time guarantees

common/client-core/src/client/metrics.rs
octol, line 1: Imho lets stick to a `mod.rs` module structure like the surrounding code

octol (commented):
Hey great work and nice to see more work on metrics, we need it! :)

Left some comments

I'm curious about the plan. If you want to extend this to be about more than just sphinx packets then this should probably be moved out of `nym-client-core`. I'd imagine you'd want to have the metrics aggregation and in particular the endpoint sitting not at the library layer but at the client (binary) crate level? Something like `nym-client`, or probably even more likely, the vpn client layer? And some of these types moved to it's own crate in `/common`

common/client-core/src/client/metrics.rs
jmwample: My initial thinking was that packages could own their own stats modules and register them with the central metrics controller that stores them and provides signals to periodically reports / reset epoch etc. But you are correct, that if they are all going to exist in a closed space here it makes sense not to take the dynamic dispatch hit. 

I will need to think more about this as I dig into the places where this needs to hook in.

jmwample:
I am revisiting this and working to hook the Nym API and gateway connection stats together. This is going to work closely with the reporting infrastructure in https://github.com/nymtech/nym/tree/simon/client_stats to the point where I may close this PR and make one against that branch if it makes more sense.

jmwample:
New stats combination PR here https://github.com/nymtech/nym/pull/5078

PR #5021, opened 2024-10-23
Fix/nym data observatory dockerfile

farbanas:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5021)


nym-data-observatory/src/main.rs
```diff
@@ -18,9 +18,25 @@ struct Args {
...
+    #[arg(short, long, default_value = None, env = "NYM_DATA_OBSERVATORY_CONNECTION_HOST")]
+    connection_host: String,
+
+    /// DB connection port
+    #[arg(short, long, default_value = None, env = "NYM_DATA_OBSERVATORY_CONNECTION_PORT")]
+    connection_port: String,
+
+    /// DB connection database name
+    #[arg(short, long, default_value = None, env = "NYM_DATA_OBSERVATORY_CONNECTION_DB")]
+    connection_db: String,
```
dynco-nym: I believe `short` defaults to the first letter of the field and will have runtime conflicts because all of these start with `c`

if so, you can set short to the initial of the 2nd word, e.g. 
- `short = 'd'` for `connection_db`
- `short = 'p'` for `connection_port`
etc

and try `cargo run`

dynco-nym (approved):

PR #5022, opened 2024-10-23
[Product Data] Introduce data persistence on gateways

simonwicky:
# What it does

This PR builds on top of #4974, not changing the behavior of the data collection, but persisting them in a sqlite database so they can be kept across restarts and crashes. 

It also leave the door open for other stats module to use that storage if needed.

# Points of Interest
- New [`gateway_stats_storage`](https://github.com/nymtech/nym/tree/simon/gateway_stats_persistence/common/gateway-stats-storage) crate
- [Config migration](https://github.com/nymtech/nym/blob/simon/gateway_stats_persistence/nym-node/src/config/old_configs/old_config_v4.rs) resulting from the added database.
- Resulting changes in the [`statistics`](https://github.com/nymtech/nym/tree/simon/gateway_stats_persistence/gateway/src/node/statistics) module to account the new storage system

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5022)

### DIFF
--- a/common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
+++ b/common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
@@ -1,6 +1,6 @@
 /*
- * Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
- * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: GPL-3.0-only
  */
 
 CREATE TABLE sessions_active
--- a/common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
+++ b/common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+CREATE TABLE sessions_active
+(
+    client_address  TEXT                        NOT NULL PRIMARY KEY UNIQUE,
+    start_time      TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+    typ             TEXT                        NOT NULL
+);
+
+CREATE TABLE sessions_finished
+(
+    id          INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    day         DATE    NOT NULL,
+    duration_ms INTEGER NOT NULL,
+    typ         TEXT    NOT NULL
+);
+
+CREATE TABLE sessions_unique_users
+(
+    day             DATE    NOT NULL,
+    client_address  TEXT    NOT NULL,
+    PRIMARY KEY (day, client_address)
+);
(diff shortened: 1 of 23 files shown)


nym-node/src/config/template.rs
```diff
@@ -185,6 +185,9 @@ announce_wss_port = {{#if entry_gateway.announce_wss_port }} {{ entry_gateway.an
 # derived shared keys, available client bandwidths and wireguard peers.
 clients_storage = '{{ entry_gateway.storage_paths.clients_storage }}'
 
+#Path to sqlite database containing all persistent stats data.
```
octol: nit: space before starting the text?

octol (approved):
Lgtm!

common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2021 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn: nit: wrong copyright year

common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
```
jstuczyn: given the rest of your code is under GPL-3.0, I guess that's wrong license too

gateway/src/config/persistence/paths.rs
```diff
@@ -39,6 +40,10 @@ pub struct GatewayPaths {
     #[serde(alias = "persistent_storage")]
     pub clients_storage: PathBuf,
 
+    /// Path to sqlite database containing all persistent stats data.
+    #[serde(alias = "persistent_stats_storage")]
```
jstuczyn: if it's a brand new field, what why do we need the alias? (it existed on `clients_storage` because we renamed it : ) )

nym-node/src/config/template.rs
```diff
@@ -237,6 +240,10 @@ upstream_exit_policy_url = '{{ exit_gateway.upstream_exit_policy_url }}'
 # derived shared keys, available client bandwidths and wireguard peers.
 clients_storage = '{{ exit_gateway.storage_paths.clients_storage }}'
 
+#Path to sqlite database containing all persistent stats data.
+stats_storage = '{{ exit_gateway.storage_paths.stats_storage }}'
```
jstuczyn, line 244: why does `nym-node` have 2 separate storages for exit and entry gateway? Just put it in one location and share it between "modes" since it's the same underlying gateway

nym-node/src/config/template.rs
simonwicky: Yes sir, fixed in https://github.com/nymtech/nym/pull/5022/commits/6486609040c2845c7785a029fad7c83f07ebf03a

common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
simonwicky: Changed in 6486609

common/gateway-stats-storage/migrations/20241017120000_create_initial_tables.sql
simonwicky: Fixed in 6486609

gateway/src/config/persistence/paths.rs
simonwicky: You're right, we don't need it
Removed in 6486609

nym-node/src/config/template.rs
simonwicky: Nothing to fix, following internal discussion

PR #5048, opened 2024-10-28
Fully functional network scores

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5048)


common/bin-common/src/lib.rs
```diff
@@ -10,3 +10,6 @@ pub mod completions;
 
 #[cfg(feature = "output_format")]
 pub mod output_format;
+
+#[cfg(feature = "models")]
+pub mod models;
```
jstuczyn: why did you put it in `bin-common`? it doesn't really feel like it belongs here. why would say `nym-api` or `nym-client` ever care about those models?

explorer-api/explorer-client/src/lib.rs
```diff
@@ -83,7 +83,9 @@ impl ExplorerClient {
         } else if response.status() == StatusCode::NOT_FOUND {
             Err(ExplorerApiError::NotFound)
         } else {
-            Err(ExplorerApiError::RequestFailure(response.text().await?))
+            let status = response.status();
+            let err_msg = format!("{}: {}", response.text().await?, status);
```
jstuczyn, line 87: sanity check: there's nothing relying on the content of that text, i.e. adding status prefix won't break any parsing?

nym-node-status-agent/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[dependencies]
+anyhow = { workspace = true}
+clap = { workspace = true, features = ["derive", "env"] }
+nym-bin-common = { path = "../common/bin-common", features = ["models"]}
+# release 0.2.4
+nym-gateway-probe = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "process"] }
+tokio-util = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
```
jstuczyn, line 25: you can get the tracing-subscriber via `basic-tracing` feature on `nym-bin-common`

nym-node-status-api/Cargo.toml
```diff
@@ -14,27 +14,31 @@ rust-version.workspace = true
...
 futures-util = { workspace = true }
 moka = { workspace = true, features = ["future"] }
-nym-bin-common = { path = "../common/bin-common" }
+nym-bin-common = { path = "../common/bin-common", features = ["models"]}
 nym-explorer-client = { path = "../explorer-api/explorer-client" }
-# TODO dz: ref before Nym API client changes. Update to latest develop once new Nym API is live
-nym-network-defaults = { git = "https://github.com/nymtech/nym", rev = "f86e08866" }
-nym-validator-client = { git = "https://github.com/nymtech/nym", rev = "f86e08866" }
+# TODO dz: before Nym API client breaking changes. Update to latest develop once new Nym API is live
+nym-network-defaults = { git = "https://github.com/nymtech/nym", branch = "pre-dir-v2-fork" }
```
jstuczyn, line 28: don't we have to update it to be compatible with magura asap?

nym-node-status-api/migrations/000_init.sql
```diff
@@ -98,3 +98,15 @@ CREATE TABLE
     FOREIGN KEY (mix_id) REFERENCES mixnodes (mix_id),
     UNIQUE (mix_id, date_utc) -- This constraint automatically creates an index
   );
+
+
+CREATE TABLE testruns
+(
+  id            INTEGER PRIMARY KEY AUTOINCREMENT,
+  gateway_id    INTEGER,
```
jstuczyn, line 106: what's `gateway_id` in this context? in your cli you're passing a `String`, so I assume this is not the identity key

nym-node-status-api/src/db/models.rs
```diff
@@ -298,3 +299,35 @@ pub(crate) mod gateway {
...
+    pub gateway_identity_key: String,
+    pub bonded: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct GatewayInfoDto {
+    pub id: i64,
+    pub gateway_identity_key: String,
+    pub self_described: Option<String>,
+    pub explorer_pretty_bond: Option<String>,
```
jstuczyn, line 333: why would we ever care about the "pretty_bond" from explorer? what does it have that we can't get from other sources?

nym-node-status-api/src/http/server.rs
```diff
@@ -22,6 +22,7 @@ pub(crate) async fn start_http_api(
 
     // TODO dz do we need this to be configurable?
     let bind_addr = format!("0.0.0.0:{}", http_port);
+    tracing::info!("Binding server to {}", bind_addr);
```
jstuczyn: nit:
```suggestion
    tracing::info!("binding server to {bind_addr}");
```

Cargo.toml
```diff
@@ -349,6 +352,10 @@ zeroize = "1.6.0"
 
 prometheus = { version = "0.13.0" }
 
+# -- internal -- #
+# release 0.2.4
+nym-gateway-probe = { git = "https://github.com/nymtech/nym-vpn-client", rev = "af3d103e" }
```
octol: Circular dependency, I wonder how that will play out :)

Should maybe the `nym-node-status-agent` should be part of the vpn client repo? Like the gateway probe

octol (approved):
I mostly skimmed. If it works, let's plow ahead

Cargo.toml
dynco-nym: good catch, fixed :)

nym-node-status-api/src/db/queries/testruns.rs
```diff
@@ -0,0 +1,156 @@
...
+use sqlx::{pool::PoolConnection, Sqlite};
+
+pub(crate) async fn get_testruns(conn: PoolConnection<Sqlite>) -> anyhow::Result<Vec<TestRunDto>> {
+    // TODO dz accept mut reference, repeat in all similar functions
+    let mut conn = conn;
+    let testruns = sqlx::query_as!(
+        TestRunDto,
+        r#"SELECT
+            id as "id!",
+            gateway_id as "gateway_id!",
```
jstuczyn: nit: do you have to expicitly tell sqlx that the field is not null? iirc it should be able to figure it out by itself

nym-node-status-api/src/db/queries/testruns.rs
```diff
@@ -0,0 +1,156 @@
...
+            ip_address as "ip_address!",
+            log as "log!"
+         FROM testruns
+         WHERE id = ?
+         ORDER BY timestamp_utc"#,
+        testrun_id
+    )
+    .fetch_one(&mut *conn)
+    .await
+    .map_err(|err| {
```
jstuczyn: fyi: `anyhow` has a useful `context()` method that does this very thing

nym-node-status-api/src/db/queries/testruns.rs
```diff
@@ -0,0 +1,156 @@
...
+    let now = now_utc().timestamp();
+    sqlx::query!(
+        "UPDATE gateways SET last_testrun_utc = ?, last_updated_utc = ? WHERE id = ?",
+        now,
+        now,
+        gateway_pk
+    )
+    .execute(&mut *conn)
+    .await
+    .map(drop)
```
jstuczyn, line 124: why do we have to explicitly drop it?

nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -1,7 +1,115 @@
...
+        "✅ Testrun row_id {} for gateway {} complete",
+        testrun.id,
+        testrun.gateway_id
+    );
+
+    Ok(StatusCode::CREATED)
+}
+
+fn get_result_from_log(log: &str) -> String {
+    let re = regex::Regex::new(r"\n\{\s").unwrap();
```
jstuczyn, line 109: I have a huge bias towards regexes, so take this with a grain of salt, but instead of passing a String body, can't we pass a proper struct and parse object properly with serde?

nym-node-status-api/src/monitor/mod.rs
```diff
@@ -97,14 +113,16 @@ async fn run(
...
 
-    // TODO dz can we calculate blacklisted GWs from their performance?
-    // where do we get their performance?
-    let gateways_blacklisted = api_client
-        .nym_api
-        .get_gateways_blacklisted()
-        .await
-        .map(|vec| vec.into_iter().collect::<HashSet<_>>())
-        .log_error("get_gateways_blacklisted")?;
+    let gateways_blacklisted = skimmed_gateways
```
jstuczyn, line 115: remind me again, why do we still care about blacklist?

nym-node-status-api/src/testruns/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub(crate) mod models;
+mod queue;
+pub(crate) use queue::{now_utc, now_utc_as_rfc3339};
+
+// TODO dz should be configurable
+const REFRESH_DELAY: Duration = Duration::from_secs(60 * 5);
+
+pub(crate) async fn spawn(pool: DbPool) {
+    let pool = pool.clone();
```
jstuczyn: nit: this close seems redundant as you're already passing by value

nym-node-status-api/src/testruns/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    tokio::spawn(async move {
+        // TODO dz delay for the first run, remove before merge
+        tokio::time::sleep(Duration::from_secs(90)).await;
```
jstuczyn: so let's remove before merge : )

nym-node-status-api/src/testruns/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    if pool.is_closed() {
+        tracing::debug!("DB pool closed, returning early");
+        return Ok(());
+    }
+
+    let mut conn = pool.acquire().await?;
+
+    let gateways = sqlx::query_as!(
+        GatewayIdentityDto,
+        r#"SELECT
```
jstuczyn, line 39: can we keep all sql contained within a single file (or maybe to a traint)? I think this makes it easier to manage rather than finding random sql in the middle of an async task

common/bin-common/src/lib.rs
dynco-nym: > why would say nym-api or nym-client ever care about those models?

if they don't, that's why I put it behind `models`

I needed a place from which both `node-status-api` and `node-status-agent` could pull a depending struct because i'd like to avoid having one of them depend on the other. 

Can you show me a better place to put it?

explorer-api/explorer-client/src/lib.rs
dynco-nym: I kept the same error Enum just in case someone was matching on it. As for text parsing, I surely hope we aren't doing that anywhere

nym-node-status-api/Cargo.toml
dynco-nym: I need to deploy this to be available for Nym VPN clients ASAP-er

nym-node-status-api/migrations/000_init.sql
dynco-nym: this ID is a PK for the gateways table

I mentioned in the code to avoid confusion

nym-node-status-api/src/db/models.rs
dynco-nym: maybe long term it can be removed: this was copied over from Harbour Master

nym-node-status-api/src/monitor/mod.rs
dynco-nym: It's part of reporing on the gateways page 

image


and part of overall network summary stats

nym-node-status-agent/Cargo.toml
dynco-nym: I'll look into it as an improvement, with other TODOs

nym-node-status-api/src/db/queries/testruns.rs
dynco-nym: not sure what you mean here, could you elaborate?

nym-node-status-api/src/db/queries/testruns.rs
dynco-nym: this isn't your straightforward destructor invocation. It's a mapping from 
```rust
Result<SqliteQueryResult, _>
````
to
```rust
Result<(), _>
```

because former is returned from `.execute`, but we don't need it

nym-node-status-api/src/http/api/testruns.rs
dynco-nym: this isn't the prettiest solution, but it's one of the things taken verbatim from HM

it can hopefully be improved upon

nym-node-status-api/src/testruns/mod.rs
dynco-nym: agree, will do

nym-node-status-api/src/db/queries/testruns.rs
dynco-nym: do you mean this?
```rust
.map_err(|err| {
    anyhow::anyhow!(
        "Couldn't retrieve testrun {}: {}",
        testrun_id,
        err.context()
    )
})
```

dynco-nym: you mean like this?
```rust
.fetch_one(&mut *conn)
.await
.context(format!("Couldn't retrieve testrun {testrun_id}"))
```

common/bin-common/src/lib.rs
jstuczyn: I'd say create another `common/<something>` crate, I don't want `bin-common` getting polluted because it's meant to be for common things shared by runnable binaries: like build information, setting up logging, etc.

nym-node-status-api/src/db/queries/testruns.rs
jstuczyn: as in do you have to be using 
```
  id as "id!",
  gateway_id as "gateway_id!",
  ...
```

as opposed to 
```
  id",
  gateway_id",
  ...
```

?

jstuczyn: but if it works, it works, that's good enough

nym-node-status-api/src/db/queries/testruns.rs
jstuczyn: yup

nym-node-status-api/src/db/queries/testruns.rs
jstuczyn: aha, makes sense, that's quite clever

nym-node-status-api/src/monitor/mod.rs
jstuczyn: gotcha. hopefully we'll be able to get away from it at some point...

nym-node-status-api/Cargo.toml
jstuczyn: yeah, totally, I'm just wondering whether it's going to be compatible, but I guess all "old" clients (so this one would apply) should work

common/bin-common/src/lib.rs
dynco-nym: done, moved to a lib called `common/models`

PR #5058, opened 2024-10-29
Use unstable explorer client

dynco-nym:
Clean up stale testruns & better logging
- use new `/unstable` endpoints on explorer for backwards compatibility
- log gw identity key
- better agent testrun logging
- log responses on server side
- change response code for agents 
- update sqlx data
- fix agent <-> probe gw bug

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5058)


nym-node-status-api/Cargo.toml
```diff
@@ -23,7 +23,8 @@ futures-util = { workspace = true }
 moka = { workspace = true, features = ["future"] }
 nym-bin-common = { path = "../common/bin-common", features = ["models"]}
 nym-common-models = { path = "../common/models" }
-nym-explorer-client = { path = "../explorer-api/explorer-client" }
+# nym-explorer-client = { path = "../explorer-api/explorer-client" }
+nym-explorer-client = { git = "https://github.com/nymtech/nym", branch = "release/2024.13-magura" }
```
octol: I assume these git dependencies into it's own repo is a temporary thing until we make it all work?

jstuczyn: oh yeah. since we're merging it into monorepo and we actually did update NS now, can we point to the local code instead? (I can resync develop with magura as I've done yesterday so that it contains the new endpoints we have just added)

nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -23,31 +24,32 @@ pub(crate) fn routes() -> Router<AppState> {
     Router::new()
         .route("/", axum::routing::get(request_testrun))
         .route("/:testrun_id", axum::routing::post(submit_testrun))
+        .layer(DefaultBodyLimit::max(1024 * 1024 * 5))
```
jstuczyn, line 27: any reason for explicitly adding it? were we accidentally ddosing ourselves? ; D

nym-node-status-api/src/testruns/mod.rs
```diff
@@ -74,3 +76,11 @@ async fn run(pool: &DbPool) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[instrument(level = "debug", skip_all)]
+async fn refresh_stale_testruns(pool: &DbPool, refresh_interval: Duration) -> anyhow::Result<()> {
+    let chrono_duration = chrono::Duration::from_std(refresh_interval)?;
```
jstuczyn, line 82: we're using `time` almost everywhere in the codebase rather than `chrono`, if it's an easy change, would it be possible if you migrated to that one instead? if not, then ignore this comment

jstuczyn (approved):

nym-node-status-api/src/http/api/testruns.rs
dynco-nym: this is request size limit

I was hitting it when probe log level was set to trace so I don't want the API to break if someone intentionally does that

and it's added ONLY here so other endpoints are protected by a [lower default (2MB)](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html)

nym-node-status-api/src/testruns/mod.rs
dynco-nym: `chrono` was already being used but I'll give it a shot.

nym-node-status-api/Cargo.toml
dynco-nym: done

PR #5069, opened 2024-10-31
chore: deprecated old nym-api client methods and replaced them when possible

jstuczyn:
this is to that the next time those methods are used outside the monorepo, the relevant calls flag up the CI via clippy

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5069)


octol (approved):
This will be most helpful

PR #5078, opened 2024-11-01
Statistics collection and reporting scaffolding

jmwample:
This PR attempts to integrate the statistics collection and reporting mechanisms we are putting together. The mechanisms are not fully done, this is meant to close the gaps on the implementation so we can work in parallel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5078)


simonwicky (approved):
LGTM

jmwample:
Rebased onto the changes in the upstream branch

I made things a little more static using an `enum` instead of boxed trait objects in `statistics_control.rs` to avoid dynamic dispatch since it isn't forced on us yet. There are some things that feel a bit repetitive at the moment, where we could likely combine / simplify.

For now I would like to merge this so I can start biting off smaller chunks.

PR #5086, opened 2024-11-04
Fix redirects

mfahampshire:
Fixing redirects to work with new base path ( s/o @fmtabbara :) )

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5086)


serinko (approved):
Tested! Ship it!

PR #5092, opened 2024-11-05
nyxd-scraper: add config to make pre-commit storage optional

pwnfoo:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5092)

### DIFF
--- a/common/nyxd-scraper/src/block_processor/mod.rs
+++ b/common/nyxd-scraper/src/block_processor/mod.rs
@@ -66,6 +66,7 @@ pub struct BlockProcessor {
     msg_modules: Vec<Box<dyn MsgModule + Send>>,
 }
 
+#[allow(clippy::too_many_arguments)]
 impl BlockProcessor {
     pub async fn new(
         pruning_options: PruningOptions,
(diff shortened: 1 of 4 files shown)


common/nyxd-scraper/src/block_processor/mod.rs
```diff
@@ -44,6 +44,7 @@ impl PendingSync {
 
 pub struct BlockProcessor {
     pruning_options: PruningOptions,
+    store_precommits: bool,
```
jstuczyn: perhaps define an intermediate `Config` struct with all the possible options, like storing precommits, pruning, etc?

nym-validator-rewarder/src/config/mod.rs
```diff
@@ -249,6 +251,9 @@ pub struct NyxdScraper {
     // if the value is missing, use `nothing` pruning as this was the past behaviour
     #[serde(default = "PruningOptions::nothing")]
     pub pruning: PruningOptions,
+
+    /// Specifies whether to store pre-commits within the database.
+    pub store_precommits: bool,
```
jstuczyn, line 257: you need to add `#[serde(default = true)]` so that old configs wouldn't fail to get deserialised

pwnfoo: Turns out it doesn't allow a `bool` to be default so had to hack around it

PR #5099, opened 2024-11-05
Add NYM_VPN_API to env files

pronebird:
Add missing NYM_VPN_API uri

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5099)

### DIFF
--- a/envs/qa.env
+++ b/envs/qa.env
@@ -21,3 +21,4 @@ REWARDING_VALIDATOR_ADDRESS=n1rfvpsynktze6wvn6ldskj8xgwfzzk5v6pnff39
 EXPLORER_API=https://qa-network-explorer.qa.nymte.ch/api
 NYXD=https://qa-validator.qa.nymte.ch
 NYM_API=https://qa-nym-api.qa.nymte.ch/api
+NYM_VPN_API=https://nym-vpn-api-git-deploy-qa-nyx-network-staging.vercel.app/api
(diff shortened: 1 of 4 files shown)


rokas-ambrazevicius (approved):

octol (dismissed):
Ah sorry but I'm not sure this is missing. If we are talking about the env file as a source for NymNetworkDetails, I would prefer if we went in the other direction. This env variable was temporarily added for mainnet but we are now using a discovery endpoint to dynamically set it. So having it in the env file is possibly confusing, and I think we should remove the mainnet one 

That said if you really want to add them for completeness sake then go ahead

pronebird:
> No this is not missing

We catch the following panic around time of starting account controller in nym_vpn_lib on iOS:

```
Caught a panic calling rust code: "called `Result::unwrap()` on an `Err` value: MissingApiUrl"
```

Setting this variable properly makes issue go away.

envs/qa.env
```diff
@@ -21,3 +21,4 @@ REWARDING_VALIDATOR_ADDRESS=n1rfvpsynktze6wvn6ldskj8xgwfzzk5v6pnff39
 EXPLORER_API=https://qa-network-explorer.qa.nymte.ch/api
 NYXD=https://qa-validator.qa.nymte.ch
 NYM_API=https://qa-nym-api.qa.nymte.ch/api
+NYM_VPN_API=https://nym-vpn-api-git-deploy-qa-nyx-network-staging.vercel.app/api/
```
dynco-nym, line 24: could you please add the same variable (with appropriate values) to 
- `envs/sandbox.env`
- `envs/canary.env`

octol (commented):
Retracting my request for changes, but see my comment why I think we shouldn't add these

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5099)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dynco-nym)

___
*[`envs/qa.env` line 24 at r1](https://reviewable.io/reviews/nymtech/nym/5099#gh-1830084985:-OB0W7oU-2PhFo9xUMrd:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/40cc90b1f2f1002600215b9686c182b3bdcb86cd/envs/qa.env#L24)):*
<details><summary><i>Previously, dynco-nym (Dinko Zdravac) wrote…</i></summary><blockquote>

could you please add the same variable (with appropriate values) to 
- `envs/sandbox.env`
- `envs/canary.env`
</blockquote></details>

Done.

envs/qa.env
pronebird: @dynco-nym bump

PR #5107, opened 2024-11-07
[Product Data] Client-side stats collection 

simonwicky:
# Why
The goal is to anonymously gather stats from nym-clients. These stats will be sent through the mixnet to a Nym run service provider that will gather them. 

# What this PR does
Sets the scene to send stats in a mixnet message to an address. The address can be set when the client is created. Current stats include some infos on sent packets along with platform information. If a receiving address is set, the client will send a mixnet packet every 5min to this address. Otherwise, nothing happens and the client runs as usual. 

To set an address, `BaseClientBuilder` and the sdk expose a `with_statistics_reporting` function.
For now, there is no address.

# Next steps

- Use address provided in .env file if there is one
- Design receiving client
- Publish address through VPN API so vpn client can grab it

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5107)


Cargo.toml
```diff
@@ -285,6 +285,7 @@ okapi = "0.7.0"
 once_cell = "1.20.2"
 opentelemetry = "0.19.0"
 opentelemetry-jaeger = "0.18.0"
+os_info = "3.8.2"
```
jstuczyn: why use `os_info` if we're already integrated into `sysinfo` that essentially gives the same info?

common/client-core/Cargo.toml
```diff
@@ -115,6 +116,7 @@ features = ["wasm-bindgen"]
 
 [dev-dependencies]
 tempfile = { workspace = true }
+pretty_env_logger = { workspace = true}
```
jstuczyn: why is the env_logger a dev dependency? 
any logging backends should ideally be pulled from `bin-common` and I can't see why it would be dev specific

common/client-core/src/client/base_client/mod.rs
```diff
@@ -586,11 +597,26 @@ where
...
+                let (stats_control, stats_reporter) = StatisticsControl::new(
+                    config.reporting_address,
+                    client_stats_id,
+                    config.reporting_type,
+                    input_sender.clone(),
+                );
+                stats_control.start_with_shutdown(shutdown.with_suffix("controller"));
+                stats_reporter
+            }
+            None => ClientStatsSender::sink(shutdown.with_suffix("sink")),
```
jstuczyn: what's the point of explicit `sink`. why not just not start anything?

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,174 @@
...
+};
+use nym_task::connections::TransmissionLane;
+
+use crate::{
+    client::inbound_messages::{InputMessage, InputMessageSender},
+    spawn_future,
+};
+
+/// Time interval between reporting statistics
+const STATS_REPORT_INTERVAL_SECS: u64 = 300;
```
jstuczyn, line 34: rather than having opaque local constants, why not put them explicitly in client's config (in one of `debug` sections) so that they could be adjusted more easily should we need it?

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,174 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: GPL-3.0-only
```
jstuczyn: isn't client code meant to be licensed under Apache?

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use crate::{
+    client::inbound_messages::{InputMessage, InputMessageSender},
+    spawn_future,
+};
+
+/// Time interval between reporting statistics
+const STATS_REPORT_INTERVAL_SECS: u64 = 300;
+/// Interval for taking snapshots of the statistics
+const SNAPSHOT_INTERVAL_MS: u64 = 500;
```
jstuczyn: nitpicking: you can just define an explicit const `Duration`

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub(crate) fn start_with_shutdown(mut self, task_client: nym_task::TaskClient) {
+        spawn_future(async move {
+            self.run_with_shutdown(task_client).await;
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    // use std::sync::Arc;
```
jstuczyn: if those tests are not run, why not just remove the dead code?

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,174 @@
...
+        loop {
+            tokio::select! {
+                stats_event = self.stats_rx.recv() => match stats_event {
+                        Some(stats_event) => self.stats.handle_event(stats_event),
+                        None => {
+                            log::trace!("StatisticsControl: shutting down due to closed stats channel");
+                            break;
+                        }
+                },
+                _ = snapshot_interval.tick() => {
```
jstuczyn, line 116: note that `snapshot_interval` and `report_interval` will be run immediately upon starting which I'm not sure is what you want given there will be nothing to report

common/statistics/src/report.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub(crate) gateway_conn_stats: GatewayStats,
+    pub(crate) nym_api_stats: NymApiStats,
+}
+
+impl TryFrom<ClientStatsReport> for Vec<u8> {
+    type Error = StatsError;
+
+    fn try_from(value: ClientStatsReport) -> Result<Self, Self::Error> {
+        let report_json = serde_json::to_string(&value)?;
+        Ok(report_json.as_bytes().to_vec())
```
jstuczyn: given you haven't implemented custom serialisation, it's safe to unwrap that result to simplify the typing

common/statistics/src/clients/gateway_conn_statistics.rs
```diff
@@ -0,0 +1,77 @@
...
+            GatewayStatsEvent::RealPacketSent(packet_size) => {
+                self.real_packets_sent += 1;
+                self.real_packets_sent_size += packet_size;
+                inc!("real_packets_sent");
+                inc_by!("real_packets_sent_size", packet_size);
+            }
+        }
+    }
+
+    fn summary(&self) -> (String, String) {
```
jstuczyn, line 36: can we use more strongly-typed return type? what's this tuple meant to represent? the second string is empty, how is anyone meant to know what it's meant to be?

common/statistics/src/report.rs
```diff
@@ -0,0 +1,42 @@
...
+use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
+
+/// Report object containing both data to be reported and client / device context. We take extra care not to overcapture context information.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ClientStatsReport {
+    pub(crate) last_update_time: OffsetDateTime,
+    pub(crate) client_id: String,
+    pub(crate) client_type: String,
+    pub(crate) os_information: os_info::Info,
```
jstuczyn: be extremely careful with it. as we found out, some operator, and I'd also assume users, do not want to report their OS information. if possible make it togglable with config (like we have in `nym-node`)

common/statistics/src/report.rs
```diff
@@ -0,0 +1,42 @@
...
+impl TryFrom<ClientStatsReport> for Vec<u8> {
+    type Error = StatsError;
+
+    fn try_from(value: ClientStatsReport) -> Result<Self, Self::Error> {
+        let report_json = serde_json::to_string(&value)?;
+        Ok(report_json.as_bytes().to_vec())
+    }
+}
+
+impl TryFrom<Vec<u8>> for ClientStatsReport {
```
jstuczyn: nitpicking: usually, for bytes conversion the associated methods take `&[u8]` as arguments since the bytes are hardly ever consumed and it makes them more versatile

common/statistics/src/report.rs
```diff
@@ -0,0 +1,42 @@
...
+    type Error = StatsError;
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        let report_str = String::from_utf8(value)
+            .map_err(|err| StatsError::ReportBytesDeserialization(err.to_string()))?;
+        Ok(serde_json::from_str(&report_str)?)
```
jstuczyn: rather than converting report to string and then deserialising string, you can use `serde_json::from_slice` directly

common/statistics/src/clients/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    nym_api_stats: nym_api_statistics::NymApiStatsControl,
+}
+
+impl ClientStatsController {
+    /// Creates a ClientStatsController given a client_id
+    pub fn new(client_id: String, client_type: String) -> Self {
+        ClientStatsController {
+            //Safety : 0 is always a valid number of seconds
+            #[allow(clippy::unwrap_used)]
+            last_update_time: OffsetDateTime::now_utc().replace_second(0).unwrap(), // allow a bigger anonymity set wrt
```
jstuczyn: will this really make the anonimity set bigger given it will have a pretty unique values for micro and nanoseconds?

common/statistics/src/clients/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Reset the metrics to their initial state.
+    ///
+    /// Used to periodically reset the metrics in accordance with periodic reporting strategy
+    pub fn reset(&mut self) {
+        self.nym_api_stats = Default::default();
+        self.gateway_conn_stats = Default::default();
+        //no periodic reset for packet stats
+
+        #[allow(clippy::unwrap_used)] //Safety : 0 is always a valid number of seconds
+        let now = OffsetDateTime::now_utc().replace_second(0).unwrap();
```
jstuczyn: given you're doing that replacement in multiple places, why not extract it to some function for easier updates?

common/statistics/src/clients/nym_api_statistics.rs
```diff
@@ -0,0 +1,77 @@
...
+                inc!("real_packets_sent");
+                inc_by!("real_packets_sent_size", packet_size);
+            }
+        }
+    }
+
+    fn summary(&self) -> (String, String) {
+        (
+            format!("packets sent: {}", self.real_packets_sent,),
+            String::new(),
```
jstuczyn: again, some strongly typed return type would definitely be useful here...

common/statistics/src/clients/packet_statistics.rs
```diff
@@ -1,48 +1,21 @@
...
-
-// Time interval between reporting packet statistics
-const PACKET_REPORT_INTERVAL_SECS: u64 = 2;
-// Interval for taking snapshots of the packet statistics
-const SNAPSHOT_INTERVAL_MS: u64 = 500;
 // When computing rates, we include snapshots that are up to this old. We set it to some odd number
 // a tad larger than an integer number of snapshot intervals, so that we don't have to worry about
 // threshold effects.
 // Also, set it larger than the packet report interval so that we don't miss notable singular events
 const RECORDING_WINDOW_MS: u64 = 2300;
```
jstuczyn, line 19: if we're going to be moving configuration values to config, can we move this one too?

common/statistics/src/gateways.rs
```diff
@@ -0,0 +1,89 @@
...
+impl GatewayStatsReporter {
+    /// Construct a new gateway statistics event reporter
+    pub fn new(stats_tx: tokio::sync::mpsc::UnboundedSender<GatewayStatsEvent>) -> Self {
+        Self { stats_tx }
+    }
+
+    /// Report a gateway statistivs event using the reporter
+    pub fn report(&self, event: GatewayStatsEvent) {
+        self.stats_tx.send(event).unwrap_or_else(|err| {
+            log::error!("Failed to report gateway stat event : {:?}", err);
```
jstuczyn: nitpicking: use `Display` impl of the error, i.e. `log::error!("failed to report gateway stat event: {err}");`

common/statistics/src/gateways.rs
```diff
@@ -0,0 +1,89 @@
...
+pub enum GatewayStatsEvent {
+    /// Events in the lifecycle of an established client tunnel
+    SessionStatsEvent(SessionEvent),
+}
+
+impl GatewayStatsEvent {
+    /// A new session between this gateway and the client remote has successfully opened
+    pub fn new_session_start(client: DestinationAddressBytes) -> GatewayStatsEvent {
+        GatewayStatsEvent::SessionStatsEvent(SessionEvent::SessionStart {
+            start_time: OffsetDateTime::now_utc(),
```
jstuczyn, line 41: in here you're not adjusting the event time to increase the anonymity set. shouldn't we do it everywhere?

Cargo.toml
simonwicky: Because I didn't see that there was already a crate for this, I'll change that

common/client-core/Cargo.toml
simonwicky: Pinging @jmwample for this one

common/client-core/src/client/base_client/mod.rs
simonwicky: Pinging @jmwample for this one. 
My guess is it's to avoid lots of error logging from event sending thread, so we don't mistake channel errors with no disabled collection

common/client-core/src/client/statistics_control.rs
simonwicky: Got it, I will integrate that along with the address distribution in .env file in the next PR

common/client-core/src/client/statistics_control.rs
simonwicky: I honestly don't know how we're licensing what, so you tell me

common/client-core/src/client/statistics_control.rs
simonwicky: Pinging @jmwample for this one

common/client-core/src/client/statistics_control.rs
simonwicky: Immediate report will at least be useful for the os_information so I see no problem for that, the snapshot stuff was already like that I think. it was just moved from elsewhere

common/statistics/src/clients/gateway_conn_statistics.rs
simonwicky: Pinging @jmwample for this one

common/statistics/src/clients/mod.rs
simonwicky: The idea is more to hide the exact moment the message was sent. Delays in the mixnet pretty much make the micro and nano second info useless, but the second info can still be leveraged by a GPA I reckon, hence this

common/statistics/src/clients/nym_api_statistics.rs
simonwicky: Pinging @jmwample for this one

common/statistics/src/clients/packet_statistics.rs
simonwicky: Yep, will do in the config PR next

common/statistics/src/gateways.rs
simonwicky: This one stays on gateway. When a session_end event is handled, we turn it into a duration, no need to replace timings here

common/statistics/src/report.rs
simonwicky: So, for now there is no collection anyway, as there is no service provider set up, so no address gets distributed. 
Once there will be, it will probably be up to the SDK using code to choose wether to report stuff or not

common/client-core/src/client/base_client/mod.rs
jmwample: This is because we use an event sender which is a (wrapped) channel. The sender gets shared all over to aggregate statistics events across subsystems. We either need some kind of channel to give to all of these places  which empties itself so that events aren't just added to a queue (this solution) or a way to give out a non-functioning channel. I avoided passing around an Option<sender> to reduce the places where we have to check if the stats sender is Some before sending.

I can play with make the inner channel an Option<Channel> and then on send check for Some and skip sending on None. That might be a cleaner solution.

common/client-core/Cargo.toml
jmwample: I was unaware of `bin-common` and used the env logger while testing the stats collection to ensure that threads launched, channels sent, and reports logged as expected. The test is removed now because we have swapped to a network reporter, so I can probably remove this unless we want to write stats reports to logs in place of or along with network reporting.

common/statistics/src/clients/gateway_conn_statistics.rs
jmwample: For now these are place-holders (and internal to the type) which remain as part of the skeleton from the `packet-statistics` blueprint. The statistics collected here will be expanded and the serialization will be more structured. The primary goal was to get the structures in place for aggregated stats collection from multiple sources.

common/client-core/src/client/statistics_control.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    pub(crate) fn create_and_start_with_shutdown(
+        reporting_config: Option<StatsReportingConfig>,
+        client_stats_id: String,
+        report_tx: InputMessageSender,
+        task_client: nym_task::TaskClient,
+    ) -> ClientStatsSender {
+        match reporting_config {
+            None => ClientStatsSender::new(None),
```
octol: If if read it correctly, it's conditionally started on the reporting config? We need to e.g always start this in the vpn since this is used to feed the UI

common/statistics/src/clients/packet_statistics.rs
octol: I notice that a set of variables that are deeply tied are split into different locations.
These values are judiciously set for how it affects the UI presentation in the vpn client. It feels like these now possibly start meaning different things now that they also used for other things?

Cargo.toml
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/2b6498da031f014035b64641294332ed5bac5eb3

common/client-core/src/client/statistics_control.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/2b6498da031f014035b64641294332ed5bac5eb3, along with other places I could think of

common/client-core/src/client/statistics_control.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/2b6498da031f014035b64641294332ed5bac5eb3

common/client-core/src/client/statistics_control.rs
simonwicky: Done in https://github.com/nymtech/nym/pull/5107/commits/2b6498da031f014035b64641294332ed5bac5eb3

common/statistics/src/clients/packet_statistics.rs
simonwicky: With the new collection module, the control is moved elsewhere. I made sure to keep that same small reporting interval to keep that module's behavior identical, and so other module can also profit from that

common/statistics/src/report.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/0b0f2c77cad0831e84bf88efe3dc94dcf9931faa

common/statistics/src/report.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/0b0f2c77cad0831e84bf88efe3dc94dcf9931faa

common/statistics/src/report.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/0b0f2c77cad0831e84bf88efe3dc94dcf9931faa

common/statistics/src/gateways.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/5107/commits/0b0f2c77cad0831e84bf88efe3dc94dcf9931faa

common/statistics/src/clients/mod.rs
simonwicky: Changed in https://github.com/nymtech/nym/pull/5107/commits/5b263148ffbb19b14d690537e2f622faebadb66a

octol (approved):
Thanks for incorporating the stats reporting we need for the UI :)

PR #5127, opened 2024-11-13
NS Agent auth with NS API

dynco-nym:
- NS Agent authenticates with key that was registered with NS API
- added flag to Agent to generate keypairs
- agent requests are signed by agent
- server-side requests are checked for authentication

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5127)


nym-node-status-agent/src/cli/generate_keypair.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub(crate) fn generate_key_pair(path: impl AsRef<Path>) -> anyhow::Result<()> {
+    let priv_key_path = path.as_ref();
+    let mut rng = rand::thread_rng();
+    let keypair = nym_crypto::asymmetric::identity::KeyPair::new(&mut rng);
+    info!("Generated keypair as Base58-encoded string");
+
+    let mut private_key_file = File::create(priv_key_path)?;
+    private_key_file.write_all(keypair.private_key().to_base58_string().as_bytes())?;
+
```
jstuczyn, line 12: nitpicking: if unused, each private key should be zeroized

nym-node-status-agent/src/cli/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+    RunProbe {
+        #[arg(short, long, env = "NODE_STATUS_AGENT_SERVER_ADDRESS")]
+        server_address: String,
+
+        #[arg(short = 'p', long, env = "NODE_STATUS_AGENT_SERVER_PORT")]
+        server_port: u16,
+
+        /// base58-encoded private key
+        #[arg(long, env = "NODE_STATUS_AGENT_AUTH_KEY")]
+        ns_api_auth_key: String,
```
jstuczyn, line 33: I'm not sure if passing PRIVATE key explicitly in cli is a good idea because it will linger in history and args forever. I think it'd be safer (if possible) to pass path to the file since you're saving it anyway

nym-node-status-api/launch_node_status_api.sh
```diff
@@ -7,6 +7,7 @@ export RUST_LOG=${RUST_LOG:-debug}
 export NYM_API_CLIENT_TIMEOUT=60
 export EXPLORER_CLIENT_TIMEOUT=60
 export NODE_STATUS_API_TESTRUN_REFRESH_INTERVAL=60
+export NODE_STATUS_API_AGENT_KEY_LIST="H4z8kx5Kkf5JMQHhxaW1MwYndjKCDHC7HsVhHTFfBZ4J,AKYZeKaJTzJrWtidk3BZz7wsm4GapaTY4GwQYVL43cmc"
```
jstuczyn: uhm... I hope we didn't just dox ourselves by committing qa api keys : D

nym-node-status-api/src/cli/mod.rs
```diff
@@ -56,19 +57,43 @@ pub(crate) struct Cli {
...
+}
+
+// We need a list of keys from clap. But if we define CLI argument as Vec<T>,
+// clap interprets that as type T which can be given as a CLI argument multiple
+// times (so all of them are stored in a Vec<T>). Thus we wrap Vec in a newtype
+// pattern to have a list of keys and make clap happy.
+#[derive(Debug, Clone)]
+struct KeyList(Vec<PublicKey>);
+
+fn parse_key_list(arg: &str) -> anyhow::Result<KeyList> {
```
jstuczyn: clap can do it automatically by using `value_delimiter`

nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -115,6 +133,23 @@ async fn submit_testrun(
     Ok(StatusCode::CREATED)
 }
 
+fn authenticate_agent(base58_pubkey: &str, state: &AppState) -> HttpResult<PublicKey> {
```
jstuczyn: I'm sorry to say, but that's an AWFUL authentication. you're just sending your **public** key. anyone can do it since it's public. and even if it wasn't, it's susceptible to replay attacks.

common/models/src/ns_api.rs
```diff
@@ -1,7 +1,15 @@
...
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct TestrunAssignment {
     pub testrun_id: i64,
     pub gateway_identity_key: String,
 }
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct SubmitResults {
+    pub message: String,
```
jstuczyn: what's the content of this message? can we guarantee it has something unique per request? otherwise we're susceptible to replay attacks.

common/models/src/ns_api.rs
```diff
@@ -1,7 +1,15 @@
...
+    pub signature: Signature,
+    pub public_key: PublicKey,
```
jstuczyn: shouldn't public key already be implicit given who has registered?

nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -28,10 +30,14 @@ pub(crate) fn routes() -> Router<AppState> {
...
 #[tracing::instrument(level = "debug", skip_all)]
-async fn request_testrun(State(state): State<AppState>) -> HttpResult<Json<TestrunAssignment>> {
-    // TODO dz log agent's key
+async fn request_testrun(
+    State(state): State<AppState>,
+    body: String,
+) -> HttpResult<Json<TestrunAssignment>> {
     // TODO dz log agent's network probe version
-    tracing::debug!("Agent requested testrun");
+    let agent_pubkey = authenticate_agent(&body, &state)?;
```
jstuczyn: as mentioned in the method itself, this authentication is insufficient.

nym-node-status-api/src/http/api/testruns.rs
jstuczyn: ideally, each signed message would contain a unique nonce (bonus points if it's controlled by the server).

why not, instead, just go for a simpler `Bearer token` approach?

nym-node-status-api/src/http/error.rs
```diff
@@ -15,6 +15,13 @@ impl HttpError {
         }
     }
 
+    pub(crate) fn unauthorized() -> Self {
+        Self {
+            message: serde_json::json!({"message": "Make sure your public key si registered with NS API"}).to_string(),
```
jstuczyn:
```suggestion
            message: serde_json::json!({"message": "Make sure your public key is registered with NS API"}).to_string(),
```

nym-node-status-agent/src/cli/run_probe.rs
```diff
@@ -0,0 +1,119 @@
...
+        .json(&results)
+        .send()
+        .await
+        .and_then(|response| response.error_for_status())?;
+
+    tracing::debug!("Submitted results: {})", res.status());
+    Ok(())
+}
+
+fn sign_message(key: PrivateKey, probe_outcome: String) -> SubmitResults {
```
jstuczyn: key is not zeroized here

jstuczyn (changes requested):

nym-node-status-agent/src/cli/generate_keypair.rs
dynco-nym: don't they implement [ZeroizeOnDrop](https://github.com/nymtech/nym/blob/develop/common/crypto/src/asymmetric/identity/mod.rs#L56), meaning they get zeroized a couple lines below?

nym-node-status-agent/src/cli/mod.rs
dynco-nym: it's not meant for passing via CLI: `clap` is just our convention because it handles loading from environment in a convenient way 

but for good measure, I removed the CLI option so now it's only loadable via env var

nym-node-status-api/launch_node_status_api.sh
dynco-nym: it's a keypair for testing only

nym-node-status-agent/src/cli/run_probe.rs
dynco-nym: currently, it's used to sign a message, extract a public key, then it's dropped (zeroized). 

what would you like to do differently?

common/models/src/ns_api.rs
dynco-nym: added `assigned_at_utc`: timestamp of when testrun was assigned. This is stored in the database for each testrun and checked upon submit

nym-node-status-api/src/http/api/testruns.rs
dynco-nym: each request now contains
- payload 
  - containing unique timestamp (serving as nonce) for replay protection
- signature 
  - of the payload with agent's private key

common/models/src/ns_api.rs
dynco-nym: if agent didn't sent his own public key, server would have to check the database to find agent's public key BEFORE even verifying the request authenticity. That way unauthorized requests could spam-lock our database. 

This way at least server ensures
-  message is signed and 
- key is registered 

before doing further checks

nym-node-status-agent/run.sh
```diff
@@ -1,25 +1,52 @@
...
+
+export RUST_LOG="debug"
+export NODE_STATUS_AGENT_SERVER_ADDRESS="http://127.0.0.1"
+export NODE_STATUS_AGENT_SERVER_PORT="8000"
 export NODE_STATUS_AGENT_PROBE_PATH="$crate_root/nym-gateway-probe"
 
+# each key is required for a separate agent when running in parallel: their
+# public counterparts need to be registered with NS API
+private_keys=("BjyC9SsHAZUzPRkQR4sPTvVrp4GgaquTh5YfSJksvvWT"
+    "4RqSKydrEuyGF8Xtwyauvja62SAjqxFPYQzW2neZdkQD"
```
jstuczyn: what are those keys? are those for some testnet? whatever they are, i dont think they should be hardcoded in a script...

nym-node-status-agent/Cargo.toml
```diff
@@ -16,12 +16,20 @@ readme.workspace = true
 
 [dependencies]
 anyhow = { workspace = true}
+bincode = { workspace = true }
+chrono = { workspace = true }
```
jstuczyn, line 20: plz let's not use `chrono`, it caused us enough headache, if possible use `time` like we do everywhere else

nym-node-status-agent/src/cli/generate_keypair.rs
jstuczyn: the keypair itself, sure, but youre then allocating a base58 string (which you then convert to bytes). that one will not get zeroized.

as a side note: why not write key bytes directly instead?

jstuczyn (approved):
🚢 it

nym-node-status-agent/run.sh
dynco-nym: exclusively for testing

nym-node-status-agent/Cargo.toml
dynco-nym: as discussed in private, not a change for this PR but I agree.

PR #5140, opened 2024-11-16
Respond to auth messages with same version

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5140)

### DIFF
--- a/common/authenticator-requests/src/traits.rs
+++ b/common/authenticator-requests/src/traits.rs
@@ -11,6 +11,30 @@ use nym_wireguard_types::PeerPublicKey;
 
 use crate::{v1, v2, v3, Error};
 
+#[derive(Copy, Clone, Debug)]
+pub enum AuthenticatorVersion {
+    V1,
+    V2,
+    V3,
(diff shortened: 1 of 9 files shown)


common/authenticator-requests/src/traits.rs
```diff
@@ -0,0 +1,269 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use std::net::IpAddr;
+
+use nym_credentials_interface::CredentialSpendingData;
+use nym_crypto::asymmetric::x25519::PrivateKey;
+use nym_sdk::mixnet::Recipient;
```
octol: Imho use the original type from crypto instead to avoid pulling in nym sdk

octol (approved):

common/authenticator-requests/src/traits.rs
neacsu: True, that's better. I've pushed the commit. Thanks.

PR #5145, opened 2024-11-18
Limit race probability

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5145)


octol (approved):

common/wireguard/src/peer_controller.rs
```diff
@@ -194,6 +194,10 @@ impl<St: Storage + Clone + 'static> PeerController<St> {
         );
         self.bw_storage_managers
             .insert(peer.public_key.clone(), bandwidth_storage_manager);
+        // try to immediately update the host information, to eliminate races
+        if let Ok(host_information) = self.wg_api.inner.read_interface_data() {
+            *self.host_information.write().await
```
jstuczyn: but will this actually write that host info, don't you need to assign it, i.e. `*self.host_information.write().await = host_information`?

neacsu: Ofc I need to 😅 Good catch, thanks

PR #5148, opened 2024-11-18
start session collection for exit gateways

simonwicky:
Apparently, exit gateways are also entry gateways so we need to start session stats for them as well

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5148)

### DIFF
--- a/nym-node/src/node/mod.rs
+++ b/nym-node/src/node/mod.rs
@@ -622,6 +622,7 @@ impl NymNode {
             self.exit_gateway.stats_storage.clone(),
         );
         exit_gateway.set_task_client(task_client);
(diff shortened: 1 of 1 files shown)


nym-node/src/node/mod.rs
```diff
@@ -622,6 +622,7 @@ impl NymNode {
             self.exit_gateway.stats_storage.clone(),
         );
         exit_gateway.set_task_client(task_client);
+        exit_gateway.set_session_stats(self.entry_gateway.sessions_stats.clone()); //Weird naming I'll give you that, but Andrew is gonna rework it anyway
```
jstuczyn, line 625: listen-one-job

jstuczyn (approved):

PR #5162, opened 2024-11-22
Add export_to_env to NymNetworkDetails

octol:
In `nym-vpn-core` we've started to read the network environment from a json file and then try to pass around `NymNetworkDetails` directly instead of relying on the exported environment.
However we still need to bridge with old code so we need to export the network details instance to the environment.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5162)

### DIFF
--- a/common/network-defaults/src/network.rs
+++ b/common/network-defaults/src/network.rs
@@ -157,6 +157,46 @@ impl NymNetworkDetails {
         }
     }
 
+    #[rustfmt::skip]
+    #[cfg(feature = "env")]
+    pub fn export_to_env(self) {
+        use crate::var_names;
+        use std::env::set_var;
+
+        fn set_optional_var(var_name: &str, value: Option<String>) {
+            if let Some(value) = value {
+                set_var(var_name, value);
+            }
+        }
+
+        set_var(var_names::NETWORK_NAME, self.network_name);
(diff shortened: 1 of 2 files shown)


common/network-defaults/src/network.rs
```diff
@@ -157,6 +157,46 @@ impl NymNetworkDetails {
...
+
+        set_var(var_names::MIX_DENOM, self.chain_details.mix_denom.base);
+        set_var(var_names::MIX_DENOM_DISPLAY, self.chain_details.mix_denom.display);
+
+        set_var(var_names::STAKE_DENOM, self.chain_details.stake_denom.base);
+        set_var(var_names::STAKE_DENOM_DISPLAY, self.chain_details.stake_denom.display);
+
+        set_var(var_names::DENOMS_EXPONENT, self.chain_details.mix_denom.display_exponent.to_string());
+
+        if let Some(e) = self.endpoints.first() {
```
jstuczyn, line 183: first or all of them?

jstuczyn (approved):

common/network-defaults/src/network.rs
octol: Nah just the first, like the existing setup

PR #5189, opened 2024-11-28
Fix/network tunnel script

tommyv1987:
- adjust forwarding rules were writing to multiple lines, only ensure 1 entry exists for ipv4 and ipv6 forwarding
- adjust the joke section to disclaim further debugging techniques
- remove all duplicate rules they may exist - iptables will need to be re-applied after
- remove duplicate code

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5189)


serinko (approved):
ship it!

PR #5190, opened 2024-11-28
Remove explorer dependency

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5190)


nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
+            .filter(|node| node.description.declared_role.entry)
+            .collect::<Vec<_>>();
+        tracing::debug!("Of those, {} gateways", gateways.len());
 
-    log_gw_in_explorer_not_api(explorer_gateways.as_slice(), gateways.as_slice());
+        let mixnodes = all_nodes
+            .iter()
+            .filter(|node| node.description.declared_role.mixnode)
+            .collect::<Vec<_>>();
+        tracing::debug!("Of those, {} mixnodes", mixnodes.len());
```
jstuczyn: technically you might end up with overlaps here because somebody (very soon) will be able to declare **both** entry and mixnode roles

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
+
+        let bonded_node_info = api_client
```
jstuczyn, line 131: this will **only** give you native nymnodes. so legacy gateways and legacy mixnodes will be missing in that response (but imo that's fine. if you didn't upgrade, we're not showing any of your details, you're just a "number"). just so you're aware.

nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
```diff
@@ -0,0 +1,55 @@
...
+                if matches!(err.kind(), ipinfo::IpErrorKind::RateLimitExceededError) {
+                    tracing::error!("Rate limit exceeded: {}", err);
+                }
+                anyhow::Error::from(err)
+            })
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct GatewayGeoData {
```
jstuczyn: nitpicking: would it make sense to just call it `NodeGeoData` given that we're locating all nym-nodes rather than just gateways?

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
+            .get_all_basic_nodes(None)
+            .await
+            .log_error("get_all_basic_nodes")?;
 
-    let mixnodes = api_client
-        .get_cached_mixnodes()
-        .await
-        .log_error("get_cached_mixnodes")?;
-    tracing::debug!("Fetched {} mixnodes", mixnodes.len());
+        let gateways_blacklisted = all_skimmed_nodes
```
jstuczyn, line 158: I know, I know, you're just replicating old apis, so don't act on this comment, I just have to express my feelings somewhere: it makes no sense to redo mixnode/gateway blacklists anymore. especially since we explicitly show performance and probe scores. @mmsinclair should just remove that column from harbourmaster

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
+            .nym_api
+            .get_mixnodes_detailed_unfiltered()
+            .await
+            .log_error("get_mixnodes_detailed_unfiltered")?;
+        let mixnodes_described = api_client
+            .nym_api
+            .get_mixnodes_described()
+            .await
+            .log_error("get_mixnodes_described")?;
+        let mixnodes_active = api_client
```
jstuczyn, line 182: I'm not sure this is an endpoint you want to be using. it will only give you **old legacy** mixnodes that happen to be in the rewarded set. I reckon you might get maybe like 10-20 of them

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
+
+        // Cached mixnodes don't include blacklisted nodes
+        // We need that to calculate the total locked tokens later
+        // TODO dz deprecated API, remove
+        let mixnodes = api_client
+            .nym_api
+            .get_mixnodes_detailed_unfiltered()
+            .await
+            .log_error("get_mixnodes_detailed_unfiltered")?;
+        let mixnodes_described = api_client
```
jstuczyn, line 177: I don't think you have to be making that query as you already grabbed the relevant info via
```

        let mixnodes = all_nodes
            .iter()
            .filter(|node| node.descri
```

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
-
-    // instead of counting blacklisted GWs returned from API cache, count from the active set
-    let count_gateways_blacklisted = gateways
-        .iter()
-        .filter(|gw| {
-            let gw_identity = gw.ed25519_identity_key().to_base58_string();
-            gateways_blacklisted.contains(&gw_identity)
-        })
-        .count();
+        // instead of counting blacklisted GWs returned from API cache, count from the active set
```
jstuczyn: active set? but you never queried for active set gateways. I'm a bit confused here

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -70,345 +85,342 @@ pub(crate) async fn spawn_in_background(
...
-                count: all_historical_mixnodes.cast_checked()?,
-                last_updated_utc: last_updated_utc.to_owned(),
+        let count_mixnodes_blacklisted = mixnodes.iter().filter(|elem| elem.blacklisted).count();
+
+        let recently_unbonded_gateways =
+            queries::ensure_gateways_still_bonded(&pool, &gateways).await?;
+        let recently_unbonded_mixnodes =
+            queries::ensure_mixnodes_still_bonded(&pool, &mixnodes).await?;
+
+        let count_bonded_mixnodes_reserve = 0; // TODO: NymAPI doesn't report the reserve set size
```
jstuczyn, line 251: because it doesn't exist anymore

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: I'm acting on it indeed, just not in this PR. Upcoming PR will remove those stats in favor of what we agreed was sensible

one thing at a time :wink:

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: rephrased comment to reflect that

nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
dynco-nym: renamed

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: correct, but returned type corresponds to what is currently being done with legacy nodes. Changing it would require a larger refactor (incoming)

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: outdated comment, removed

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: noted for upcoming refactor

jstuczyn (approved):

PR #5200, opened 2024-12-02
NS API: add mixnet scraper

pwnfoo:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5200)


dynco-nym (approved):
HTTP API already [exists](https://github.com/nymtech/nym/blob/develop/nym-node-status-api/nym-node-status-api/src/http/api/mixnodes.rs#L83) for this, approved

PR #5201, opened 2024-12-02
feature: hopefully final steps of the smoosh™️

jstuczyn:
this PR is a bit more extensive than I initially planned because of all the spaghetti that had to be untangled to make it all happen.

the general idea is that now it's possible to run your node simultanously in multiple modes. for example `mixnode` + `entry` or `entry` + `exit`, etc.

the modes do the following:
- `mixnode`: allows the node to handle **forward** sphinx packets
- `entry`: allows the node to accepts websocket connections from incoming clients and to handle **final hop** sphinx packets
- `exit`: allows the node to handle **final hop** sphinx packets as well as starts NR and IPR service providers

furthermore, if node runs with `wireguard` enabled, it will start local `authenticator` service provider which will also implicitly enable final hop packet handling regardless of the underlying mode.

there are also various of other smaller (and bigger) changes:
- upon receiving a forward packet, its delay is now clamped so that one could not dos the node by asking it to delay it for, for example, a year
- as mentioned above, node can now run in multiple modes at the same time,
- `/metrics/mixing` endpoint got deprecated (it still provides the same information, however) and might be removed in future release
- introduced `/metrics/packet-stats` endpoint that provides more extensive information about packets received/sent
- it is now possible to control whether the node should log its statistics to console
- the console logs are also updated with the current packet rates
- all nodes now run the "verloc" protocol and measure every other node on the network
- metrics got revamped and unified:
    - all running metrics are stored in a single `NymNodeMetrics` struct
    - there exists a `MetricsAggregator` that listens to any metrics events that might require additional processing
    - any new metrics event types can be easily added by registering a new handler via `register_handler`. the type must implement `MetricsHandler` trait and use unique variant of `MetricsEvent` enum. for example `GatewaySessionStatsHandler`. it is, however, possible to have opaque handlers, such as the already implemented `LegacyMixingStatsUpdater` and `MixnetMetricsCleaner`
- everything in `mixnode` directory has been removed because there was nothing really left there. the mixing socket listener was unified in `nym-node` and similarly `verloc` was also moved there
- `gateway` directory was similarly reduced in size. now it also creates appropriate tasks as opposed to the whole gateway process. eventually it might also be further stripped, but today is not that day.
- removed the generic parameter on the `GatewayStorage` to simplify all the generics down the stack. it wasn't used anyway

cli:
- added `--modes` argument to specify all node modes with a single command (or env variable). for example: `--modes="mixnode,entry"`. can't be used alongside `--modes`
- extended `--mode` argument to allow specifying it multiple times, for example: `--mode mixnode --mode entry`. can't be used alongside `--mode`

config changes:
- replaced `mode` with `modes` to allow setting the node to run with say, entry + mixnode roles simultanously
- added `maximum_forward_packet_delay` to `mixnet.debug` section
- moved `verloc` section from `mixnode.verloc` to top level
- removed `mixnode` section
- removed `entry_gateway` section to `gateway_tasks`
- moved `authenticator` section from `gateway_tasks` (`entry_gateway`) to `service_providers` (`exit_gateway`)
- renamed `exit_gateway` section to `service_providers`
- moved top level `authenticator` section to `service_providers` so that it'd live alongside NR and IPR
- added general `debug` section
- added `metrics` section

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5201)


Cargo.toml
```diff
@@ -149,15 +148,14 @@ members = [
...
     "tools/internal/testnet-manager",
-    "tools/internal/testnet-manager/dkg-bypass-contract",
+    "tools/internal/testnet-manager/dkg-bypass-contract", "common/verloc", "tools/internal/mixnet-connectivity-check",
 ]
 
 default-members = [
     "clients/native",
     "clients/socks5",
     "explorer-api",
     "gateway",
```
octol: should we remove `gateway` as well?

common/client-libs/mixnet-client/src/client.rs
```diff
@@ -61,6 +59,15 @@ pub struct Client {
     config: Config,
 }
 
+impl Clone for Client {
+    fn clone(&self) -> Self {
```
octol: The implementation seems to indicate it's not really a clone?

gateway/src/node/internal_service_providers.rs
```diff
@@ -0,0 +1,226 @@
...
+        GatewayError,
+    > {
+        let started_nr = self.network_requester.start_service_provider().await?;
+        let started_ipr = self.ip_router.start_service_provider().await?;
+
+        Ok((started_nr, started_ipr))
+    }
+}
+
+pub struct SPMessageRouterBuilder {
```
octol: The smallest of nits: capitalization. Use `SpMessageRouterBuilder` for consistency ;)

nym-node/src/node/http/mod.rs
```diff
@@ -1,40 +1,66 @@
...
+pub mod helpers;
+pub mod middleware;
+pub mod router;
+pub mod state;
+
+type InnerService = IntoMakeServiceWithConnectInfo<Router, SocketAddr>;
+type ConnectInfoExt = AddExtension<Router, ConnectInfo<SocketAddr>>;
+pub type ServeService = Serve<InnerService, ConnectInfoExt>;
+
+pub struct NymNodeHTTPServer {
```
octol: nit: `NymNodeHttpServer`

common/client-libs/mixnet-client/src/client.rs
jstuczyn: good catch. I added it for some intermediate code to pass things around. removed : )

gateway/src/node/internal_service_providers.rs
jstuczyn: shouldnt it be `SP` though for "Service Provider"?

benedettadavico:
Testing Completion Date: 14/01/2025

Testing Steps Performed:

Ran a series of tests for nodes focusing on the ability to handle multiple modes, network protocols, and configuration adjustments. Validated wireguard and metrics reporting.

Validation:

[x] Test Passed
Issues Found: None

PR #5235, opened 2024-12-09
Extend swagger docs

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5235)


nym-api/src/support/http/openapi.rs
```diff
@@ -13,85 +11,31 @@ use utoipauto::utoipauto;
...
+// cannot be autodiscovered because proc macros run before feature flags.
+// Tracking issue: https://github.com/ProbablyClem/utoipauto/issues/13
+
+#[utoipauto(paths = "./nym-api/src,
+    ./nym-api/nym-api-requests/src from nym-api-requests,
+    ./common/config/src from nym-config,
+    ./common/cosmwasm-smart-contracts/mixnet-contract/src from nym-mixnet-contract-common,
+    ./common/ticketbooks-merkle/src from nym-ticketbooks-merkle,
+    ./common/nym_offline_compact_ecash/src from nym_compact_ecash")]
 #[derive(OpenApi)]
```
dynco-nym: benefit of bumping `utoipauto`

common/nym_offline_compact_ecash/src/common_types.rs
```diff
@@ -7,15 +7,26 @@ use crate::helpers::{g1_tuple_to_bytes, recover_g1_tuple};
...
 pub struct Signature {
+    #[schema(value_type = G1ProjectiveSchema)]
     pub(crate) h: G1Projective,
+    #[schema(value_type = G1ProjectiveSchema)]
     pub(crate) s: G1Projective,
 }
 
+#[derive(ToSchema)]
+#[schema(title = "G1Projective")]
+pub struct G1ProjectiveSchema {
```
jstuczyn: I dont think we want that schema for the purposes of any api. if we're returning raw signature anywhere, they're using json serialiser and thus stringified, I think hex, encoding for G1 points

common/nym_offline_compact_ecash/src/proofs/proof_withdrawal.rs
```diff
@@ -32,11 +33,15 @@ pub struct WithdrawalReqWitness<'a> {
     pub private_attributes_openings: &'a Vec<Scalar>,
 }
 
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, ToSchema)]
 pub struct WithdrawalReqProof {
+    #[schema(value_type = [u64; 4])]
```
jstuczyn: I think the same applies to Scalars, please double check by trying to send some dummy values across

nym-api/nym-api-requests/src/ecash/models.rs
```diff
@@ -323,17 +340,28 @@ impl From<Vec<u8>> for SerialNumberWrapper {
...
-#[derive(Clone, Serialize, Deserialize, Debug, JsonSchema, PartialEq)]
+#[derive(Clone, Serialize, Deserialize, Debug, JsonSchema, PartialEq, ToSchema)]
 pub struct BatchRedeemTicketsBody {
     #[serde(with = "nym_serde_helpers::bs58")]
     #[schemars(with = "String")]
     pub digest: Vec<u8>,
     pub included_serial_numbers: Vec<SerialNumberWrapper>,
     pub proposal_id: u64,
     #[schemars(with = "String")]
+    #[schema(value_type = AccountIdSchema)]
```
jstuczyn: for the purposes of the api, it just returns a string

nym-api/src/node_status_api/models.rs
```diff
@@ -315,8 +316,13 @@ impl From<HistoricalUptime> for OldHistoricalUptimeResponse {
 
 // TODO rocket remove smurf name after eliminating `rocket`
 pub(crate) type AxumResult<T> = Result<T, AxumErrorResponse>;
+
+#[derive(ToSchema, ToResponse)]
```
jstuczyn: why do we need `ToSchema` on this type? it's never directly returned to the user. `status` is used to set correct HTTP status response and the underlying `message` is what reaches the user

jstuczyn (changes requested):

nym-api/src/node_status_api/models.rs
dynco-nym: Because it's listed as a returned type [here](https://github.com/nymtech/nym/blob/develop/nym-api/src/ecash/api_routes/partial_signing.rs#L45) and in many other places like that

dynco-nym: replaced with `String` like this

```rust
responses(
    // ...
    (status = 400, body = String),
)
```

common/nym_offline_compact_ecash/src/common_types.rs
dynco-nym: moved `ToSchema`  derives out of ecash crate, as discussed

common/nym_offline_compact_ecash/src/proofs/proof_withdrawal.rs
dynco-nym: [ditto](https://github.com/nymtech/nym/pull/5235#discussion_r1891517769)

nym-api/nym-api-requests/src/ecash/models.rs
```diff
@@ -272,19 +297,167 @@ pub struct AggregatedExpirationDateSignatureResponse {
...
     #[schemars(with = "PlaceholderJsonSchemaImpl")]
+    #[schema(value_type = openapi_schema::Signature)]
     pub signatures: Vec<AnnotatedCoinIndexSignature>,
 }
 
+/// duplicate types from `nym-compact-ecash`, but these do derive `ToSchema```
+pub mod openapi_schema {
+    #![allow(dead_code)]
+    use nym_compact_ecash::common_types::{G2Projective, Scalar};
+
```
dynco-nym, line 296: new code @jstuczyn

common/nym_offline_compact_ecash/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rand = { workspace = true }
 thiserror = { workspace = true }
 sha2 = "0.9"
 bs58 = { workspace = true }
+utoipa = { workspace = true }
```
jstuczyn: I thought you removed all utoipa stuff from the ecash library?

nym-api/nym-api-requests/src/ecash/models.rs
```diff
@@ -272,19 +297,167 @@ pub struct AggregatedExpirationDateSignatureResponse {
...
+        #[schema(value_type = String)]
+        joined_commitment: G1Projective,
+        #[schema(value_type = Vec<String>)]
+        private_attributes_commitments: Vec<G1Projective>,
+        zk_proof: WithdrawalReqProof,
+    }
+
+    #[derive(ToSchema)]
+    pub struct WithdrawalReqProof {
+        #[schema(value_type = [u64; 4], format = Binary)]
```
jstuczyn: I think `Scalar` is also using the same serialisation format as G1 and G2 points, i.e. it's hex-encoded. Please double check it!

nym-api/nym-api-requests/src/ecash/models.rs
```diff
@@ -424,8 +600,9 @@ impl IssuedTicketbooksForResponseBody {
 pub struct IssuedTicketbooksForResponse {
     pub body: IssuedTicketbooksForResponseBody,
 
-    /// Signature on the body    
+    /// Signature on the body
     #[schemars(with = "PlaceholderJsonSchemaImpl")]
```
jstuczyn: I guess now we might be able to remove all those `    #[schemars(with = "PlaceholderJsonSchemaImpl")]` I put everywhere? : )

nym-api/nym-api-requests/src/ecash/models.rs
dynco-nym: yep
```rust
impl fmt::Debug for Scalar {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let tmp = self.to_bytes();
        write!(f, "0x")?;
        for &b in tmp.iter().rev() {
            write!(f, "{:02x}", b)?;
        }
        Ok(())
    }
}
```

added `content_encoding = "Base16"`

common/nym_offline_compact_ecash/Cargo.toml
dynco-nym: removed

nym-api/nym-api-requests/src/ecash/models.rs
dynco-nym: done :heavy_check_mark:

nym-api/nym-api-requests/src/ecash/models.rs
jstuczyn: But that's a debug implementation which has nothing to do with serde. I was asking about how it's serialised

dynco-nym: there is no serde impl in the docs, but a chapter on [serialization](https://docs.rs/bls12_381/latest/bls12_381/notes/serialization/index.html) mentions that  `G_1` group uses `Scalar`s (elements of `F_p` )  for its coordinates

jstuczyn (approved):

PR #5247, opened 2024-12-11
Add control messages to GatewayTransciver

durch:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5247)

### DIFF
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -191,6 +191,8 @@ pub struct BaseClientBuilder<'a, C, S: MixnetClientStorage> {
 
     #[cfg(unix)]
     connection_fd_callback: Option<Arc<dyn Fn(RawFd) + Send + Sync>>,
+
+    forget_me: bool,
 }
 
 impl<'a, C, S> BaseClientBuilder<'a, C, S>
@@ -215,7 +217,14 @@ where
             setup_method: GatewaySetup::MustLoad { gateway_id: None },
             #[cfg(unix)]
             connection_fd_callback: None,
+            forget_me: false,
         }
+    }
+
+    #[must_use]
+    pub fn with_forget_me(mut self, forget_me: bool) -> Self {
+        self.forget_me = forget_me;
+        self
     }
 
     #[must_use]
@@ -643,6 +652,21 @@ where
         mix_tx
     }
 
+    fn start_mix_traffic_controller_with_forget_me(
+        gateway_transceiver: Box<dyn GatewayTransceiver + Send>,
+        shutdown: TaskClient,
+        forget_me: bool,
+    ) -> BatchMixMessageSender {
+        info!(
+            "Starting mix traffic controller with forget me: {}...",
+            forget_me
+        );
+        let (mix_traffic_controller, mix_tx) =
+            MixTrafficController::new_with_forget_me(gateway_transceiver, forget_me);
+        mix_traffic_controller.start_with_shutdown(shutdown);
+        mix_tx
+    }
+
     // TODO: rename it as it implies the data is persistent whilst one can use InMemBackend
     async fn setup_persistent_reply_storage(
         backend: S::ReplyStore,
@@ -820,10 +844,19 @@ where
         // that are to be sent to the mixnet. They are used by cover traffic stream and real
         // traffic stream.
         // The MixTrafficController then sends the actual traffic
-        let message_sender = Self::start_mix_traffic_controller(
-            gateway_transceiver,
-            shutdown.fork("mix_traffic_controller"),
-        );
+
+        let message_sender = if self.forget_me {
+            Self::start_mix_traffic_controller_with_forget_me(
+                gateway_transceiver,
+                shutdown.fork("mix_traffic_controller"),
+                self.forget_me,
+            )
+        } else {
+            Self::start_mix_traffic_controller(
+                gateway_transceiver,
+                shutdown.fork("mix_traffic_controller"),
+            )
+        };
 
         // Channels that the websocket listener can use to signal downstream to the real traffic
         // controller that connections are closed.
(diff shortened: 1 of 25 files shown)


common/client-core/src/client/base_client/mod.rs
```diff
@@ -622,6 +630,21 @@ where
         mix_tx
     }
 
+    fn start_mix_traffic_controller_with_forget_me(
```
jstuczyn: tiny nitpick: why define separate function for essentially the same thing? just have a `MixTrafficController::new()` that  takes `forget_me` flag which would be `false` by default

common/client-core/src/client/mix_traffic/mod.rs
```diff
@@ -111,7 +152,26 @@ impl MixTrafficController {
                 }
             }
             shutdown.recv_timeout().await;
+
+            if self.forget_me {
+                log::info!("Sending forget me request to the gateway");
+                match self
+                    .gateway_transceiver
+                    .send_client_request(ClientRequest::ForgetMe {
```
jstuczyn, line 127: have you checked it actually works? because this is received after the global shutdown has already been issued so it might be the case that the gateway client might not work anymore because some of the channels it was sending received messages on are already gone

common/client-core/src/client/mix_traffic/mod.rs
```diff
@@ -111,7 +152,26 @@ impl MixTrafficController {
...
+                        also_from_stats: true,
```
jstuczyn: should this be set to `true` every time?

common/client-core/src/client/mix_traffic/transceiver.rs
```diff
@@ -123,6 +138,20 @@ where
...
         self.gateway_client.ws_fd()
     }
+
+    async fn send_client_request(
+        &mut self,
+        message: ClientRequest,
+    ) -> Result<(), GatewayClientError> {
+        if let Some(shared_key) = self.gateway_client.shared_key() {
+            self.gateway_client
+                .send_websocket_message(message.encrypt(shared_key.unwrap_legacy())?)
```
jstuczyn: why are you attempting to unwrap the legacy shared key here? any new-ish client will be failing here. just use `message.encrypt(shared_key)` (because it will then correctly delegate to correct enum variant)

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
```diff
@@ -327,6 +327,13 @@ impl<R, S> AuthenticatedHandler<R, S> {
         }
     }
 
+    async fn handle_forget_me(
+        &mut self,
+        _also_from_stats: bool,
+    ) -> Result<ServerResponse, RequestHandlingError> {
+        unimplemented!()
```
jstuczyn: if we're leaving it unimplemented for now, maybe just log a `warn!` rather than crash the gateway itself : )

common/client-core/src/client/mix_traffic/mod.rs
durch: Not yet, similar thought occurred to me, I'll test it in `localnet` now that the whole thing is implemented

gateway/src/node/client_handling/websocket/connection_handler/authenticated.rs
durch: That was just a placeholder :)

common/client-core/src/client/mix_traffic/mod.rs
durch: No, I've elaborated that :)

common/client-core/src/client/mix_traffic/transceiver.rs
durch: Got tripped up by Arc here

nym-node/src/node/metrics/handler/client_sessions.rs
```diff
@@ -73,6 +73,14 @@ impl GatewaySessionStatsHandler {
         Ok(())
     }
 
+    async fn handle_session_delete(
+        &mut self,
+        client: DestinationAddressBytes,
+    ) -> Result<(), StatsStorageError> {
+        self.storage.delete_active_session(client).await?;
```
simonwicky, line 80: Can you delete the entry from `sessions_unique_users` as well? (For the correct day)

simonwicky (changes requested):
I looked at the session statistics part only. Can you make sure the unique users entry for that client gets deleted as well please?

nym-node/src/node/metrics/handler/client_sessions.rs
durch: I think we can assume that we won't be duplicating identities in the span of a few days, if we are we should fix that... Otherwise added

simonwicky: It's only a problem if we're deleting actual clients and that would mean we were able to generate the same keypair as somebody. It's fair to say it's not gonna happen, so we're good

simonwicky (commented):
LGTM

simonwicky (approved):

tommyv1987:
**Testing Completion Report: Add Control Messages to Gateway Transceiver**  
**Testing Completion Date:** 13/01/2025  

**Testing Steps Performed:**  


**Review and Testing: Forget Me Implementation**  
- Validated the encryption and delivery of `ForgetMe` control messages to the gateway 

**Testing: MixTrafficController Integration**  
- Verified that the `MixTrafficController` invokes `ForgetMe` logic correctly during shutdown  
- Tested behaviour for gateway transceiver failures while sending control messages  

**Testing: Gateway Storage Updates**  
- Confirmed successful deletion of client data (e.g., inbox messages, bandwidth allocations) from persistent storage  

**Validation:**  
- **Test Passed**  
- **Issues Found:** None

PR #5250, opened 2024-12-11
tweak fix

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5250)


serinko (approved):
As before, I would consider to keep convention `./<PATH_TO>/` but it's not the end of the world.

PR #5251, opened 2024-12-12
Add conversion unit tests for auth msg

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5251)


common/authenticator-requests/src/v4/conversion.rs
```diff
@@ -137,25 +140,51 @@ impl TryFrom<v4::response::AuthenticatorResponse> for v3::response::Authenticato
...
 }
 
+impl TryFrom<v3::response::AuthenticatorResponseData> for v4::response::AuthenticatorResponseData {
+    type Error = crate::Error;
+    fn try_from(
+        authenticator_response_data: v3::response::AuthenticatorResponseData,
+    ) -> Result<Self, Self::Error> {
+        match authenticator_response_data {
+            v3::response::AuthenticatorResponseData::PendingRegistration(_) => Err(
+                Self::Error::Conversion("mac hash breaking change".to_string()),
```
jstuczyn, line 158: what has actually changed in the mac? is there no way to backport the changes? I wonder if this might cause client breakage

jstuczyn (approved):

common/authenticator-requests/src/v4/conversion.rs
neacsu: The ipv6 was added to the payload. It's not possible to convert those messages without the private key. It's better to actually remove the conversion, because it forces the re-creation of the messages via the `new` function, which requires the private key and re-does the mac.

tommyv1987:
These tests strengthen the Authenticator service by:

- Supporting more request/response handling
- Improving internal consistency with better utilities and traits
- Simplifying unit testing through reusable constants and data structures

This should prevent any auth service mismatches in the future

PR #5255, opened 2024-12-12
Update TS bindings

durch:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5255)


mmsinclair (approved):
At some point in the future, we should format this - for now it is ok so that it stops creating noise in diffs

PR #5256, opened 2024-12-12
bugfix: make sure to apply gateway score filtering when choosing initial node

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5256)

### DIFF
--- a/sdk/lib/socks5-listener/src/lib.rs
+++ b/sdk/lib/socks5-listener/src/lib.rs
@@ -275,7 +275,7 @@ where
             specification: GatewaySelectionSpecification::UniformRemote {
                 must_use_tls: false,
             },
-            available_gateways: current_gateways(&mut rng, &nym_apis, None).await?,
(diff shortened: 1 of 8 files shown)


sdk/lib/socks5-listener/src/lib.rs
```diff
@@ -275,7 +275,7 @@ where
             specification: GatewaySelectionSpecification::UniformRemote {
                 must_use_tls: false,
             },
-            available_gateways: current_gateways(&mut rng, &nym_apis, None).await?,
+            available_gateways: current_gateways(&mut rng, &nym_apis, None, 50).await?,
```
octol, line 278: Hardcoding?

octol (approved):

sdk/lib/socks5-listener/src/lib.rs
jstuczyn: indeed, because socks5 didn't have proper config. but as discussed, it's going to get removed anyway, so that's fine : )

PR #5271, opened 2024-12-16
feature: nym topology revamp

jstuczyn:
this PR changes the internals of the `NymTopology` to blur the lines between explicit mixnodes and gateways so that what used to be considered a "mixnode" could be a valid egress point of the network.

`NymTopology` is no longer divided into `BTreeMap<MixLayer, Vec<mix::Node>>` and `Vec<gateway::Node>`. instead there's information about the current rewarded set (to support future VRF work) and a simple map of `HashMap<NodeId, RoutingNode>`. 

the new features are mostly controlled via 2 new flags/config values: 
- `use_extended_topology` that tells the client to retrieve **all** network nodes rather than the ones that got assigned "active" mixnode role (or support being a gateway)
- `ignore_egress_epoch_role` that tells the client it's  fine to construct egress packets to nodes that are **not** assigned entry or exit gateway role

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5271)


common/client-core/src/client/base_client/mod.rs
```diff
@@ -539,15 +538,15 @@ where
...
+                config_topology,
                 nym_api_urls,
                 user_agent,
             )),
             config::TopologyStructure::GeoAware(group_by) => {
-                Box::new(GeoAwareTopologyProvider::new(nym_api_urls, group_by))
+                warn!("using deprecated 'GeoAware' topology provider - this option will be removed very soon");
+
+                #[allow(deprecated)]
+                Box::new(crate::client::topology_control::GeoAwareTopologyProvider::new(nym_api_urls, group_by))
```
octol, line 549: I would consider actually just removing it, if it makes things simpler

octol (approved):

common/client-core/src/client/base_client/mod.rs
jstuczyn: I'd rather do it in stages (can't believe I'm the one saying this : D) because if I do it now, I'd need to write bunch of client config migrations which would further bloat this PR

benedettadavico:
The majority of the test cases have been preceded by https://github.com/nymtech/nym/pull/5388. 

Topology behaviour and regression is all passing and working as expected.

PR #5274, opened 2024-12-17
Nyx Chain Watcher

mmsinclair:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5274)


nym-validator-rewarder/src/config/mod.rs
```diff
@@ -129,7 +129,10 @@ impl Config {
             database_path: self.storage_paths.nyxd_scraper.clone(),
             pruning_options: self.nyxd_scraper.pruning,
             store_precommits: self.nyxd_scraper.store_precommits,
-            start_block_height: None,
+            start_block: StartingBlockOpts {
+                start_block_height: None,
+                use_best_effort_start_height: true,
```
pwnfoo, line 134: @jstuczyn hope this doesn't mess with the rewarder logic 👀

common/nyxd-scraper/src/scraper/mod.rs
```diff
@@ -226,32 +258,62 @@ impl NyxdScraper {
...
+                        Err(err) => {
+                            error!("failed to retrieve the block: {err}. stopping...");
+                            return Err(err);
+                        }
                     }
                 }
             }
+
+            // if we don't need to catch up, return early
+            if !must_catch_up {
```
jstuczyn, line 293: I keep forgetting, do we still need it?

nym-validator-rewarder/src/config/mod.rs
jstuczyn: I don't think it should because the rewarded has already been started so those settings wouldn't apply. we just need to make sure the config correctly deserialises

jstuczyn (approved):
note: I just looked at the scraper code changes not the chain watcher (lmk if you want me to review that)

tommyv1987:
Out of scope for Hu testing, internal tool. Will review at later date.

PR #5312, opened 2025-01-06
feature: periodically remove stale gateway messages

jstuczyn:
this PR introduces a simple task that removes gateway messages that haven't been retrieved in (by default) 24h.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5312)


gateway/src/node/stale_data_cleaner.rs
```diff
@@ -0,0 +1,59 @@
...
+                _ = self.task_client.recv() => {
+                    trace!("StaleMessagesCleaner: received shutdown");
+                }
+                _ = interval.tick() => {
+                    if let Err(err) = self.clean_up_stale_messages().await {
+                        warn!("failed to clean up stale messages: {err}");
+                    }
+                }
+            }
+        }
```
octol, line 53: Imho I like having a debug statement when the task finishes, to troubleshot tasks that gets stuck

Possibly we can make a cleaner solution for that with the cancellation token work in the future ...

octol (approved):

tommyv1987:
Test Report: Automation Script for Data Cleanup Validation
Test Objective Validate that the stale message cleanup mechanism in the database correctly removes records older than the configured threshold (24 hours).

Test Setup
1. Environment:
    * SQLite database
    * Bash script: used to insert data.

Steps Performed
1. Ran the insert_data.sh script to populate the database with test data:
    * Recent records inserted successfully.
2. Verified the database content post-insertion: sqlite3 clients.sqlite "SELECT * FROM message_store;"
3. 
    * Confirmed that all 20 records (10 recent + 10 stale) were present.
4. Allowed the system to run for 24 hours to trigger the cleanup mechanism.
5. Queried the database again after 24 hours: sqlite3 gateway_storage.db "SELECT * FROM message_store;"
6. 

Expected Result
* All stale records (older than 24 hours) should be removed.
* Recent records should remain in the database.
Actual Result
* Query after 24 hours showed only the 10 recent records.
* All 10 stale records were successfully removed.

Status: Pass

PR #5318, opened 2025-01-08
Bugfix/contract version assignment

jstuczyn:
this PR fixes updates to current nym-node version as well as introduces migration to fix the existing state of the mainnet contract

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5318)


contracts/mixnet/src/mixnet_contract_settings/storage.rs
```diff
@@ -34,9 +34,13 @@ impl NymNodeVersionHistory<'_> {
     }
 
     fn next_id(&self, storage: &mut dyn Storage) -> Result<u32, MixnetContractError> {
-        let next = self.id_counter.may_load(storage)?.unwrap_or_default();
```
jstuczyn, line 37: this was the bug. it was always returning "0"

PR #5325, opened 2025-01-09
feature: `CancellationToken`-based shutdowns

jstuczyn:
this PR introduces scaffolding for using `CancellationToken` and `TaskTracker` for our graceful shutdowns rather than the existing `TaskClient` and `TaskManager`. 

it focues on creating the new API and migrating **some** of the existing tasks inside `NymNode`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5325)

### DIFF
--- a/common/task/src/cancellation.rs
+++ b/common/task/src/cancellation.rs
@@ -50,7 +50,7 @@ impl Deref for ShutdownToken {
 
 impl ShutdownToken {
     const MAX_NAME_LENGTH: usize = 128;
-    const OVERFLOW_NAME: &'static str = "reached maximum TaskClient children name depth";
+    const OVERFLOW_NAME: &'static str = "reached maximum ShutdownToken children name depth";
 
     pub fn new(name: impl Into<String>) -> Self {
         ShutdownToken {
@@ -183,6 +183,14 @@ impl ShutdownManager {
     pub fn with_shutdown_duration(mut self, duration: Duration) -> Self {
         self.max_shutdown_duration = duration;
         self
+    }
+
+    pub fn child_token<S: Into<String>>(&self, child_suffix: S) -> ShutdownToken {
+        self.root_token.child_token(child_suffix)
+    }
+
+    pub fn clone_token<S: Into<String>>(&self, child_suffix: S) -> ShutdownToken {
+        self.root_token.clone_with_suffix(child_suffix)
     }
 
     pub async fn wait_for_shutdown(&self) {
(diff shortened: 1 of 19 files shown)


common/task/src/cancellation.rs
```diff
@@ -0,0 +1,358 @@
...
+#[cfg(not(target_arch = "wasm32"))]
+use tokio::signal::unix::{signal, SignalKind};
+
+pub const DEFAULT_MAX_SHUTDOWN_DURATION: Duration = Duration::from_secs(5);
+
+pub fn token_name(name: &Option<String>) -> String {
+    name.clone().unwrap_or_else(|| "unknown".to_string())
+}
+
+// pending name
```
octol: Still pending?

common/task/src/cancellation.rs
```diff
@@ -0,0 +1,358 @@
...
+            format!("unknown-{suffix}")
+        };
+
+        child.name = Some(child_name);
+        child
+    }
+
+    // expose the method with the old name for easier migration
+    #[must_use]
+    pub fn fork<S: Into<String>>(&self, child_suffix: S) -> Self {
```
octol, line 108: I think we should add a note that these are deprecated and should be removed later

common/task/src/cancellation.rs
```diff
@@ -0,0 +1,358 @@
...
+impl Deref for ShutdownManager {
+    type Target = TaskTracker;
+
+    fn deref(&self) -> &Self::Target {
+        &self.tracker
+    }
+}
+
+impl ShutdownManager {
+    pub fn new(root_token: impl Into<String>) -> Self {
```
octol: Name rather than token?

nym-node/src/node/mod.rs
```diff
@@ -447,6 +448,10 @@ impl NymNode {
             wireguard: Some(wireguard_data),
             config,
             accepted_operator_terms_and_conditions: false,
+            shutdown_manager: ShutdownManager::new("NymNode")
```
octol, line 451: Is there a reason why this one PascalCase while the others are kebab-case?

octol (approved):
I wonder what the impact is when now switching to not-dropguard behaviour by default? I know in the vpn client we seem to hit the drop guards for the gateway-client

jstuczyn:
> I wonder what the impact is when now switching to not-dropguard behaviour by default? I know in the vpn client we seem to hit the drop guards for the gateway-client

i think it would have to be dealt with on case-by-case basis, because with those changes on-drop behaviour is explicit

nym-node/src/node/mod.rs
jstuczyn: I'm not sure I've been particularly consistent, but I think with recent changes I attempted to use PascalCase for the top level name and kebab-case or snake_case for any child components

PR #5344, opened 2025-01-14
Feature/legacy explorer message

windy-ux:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5344)


fmtabbara (changes requested):
Instead of adding the <ReleaseAlert /> component on every page, juts add it inside the DesktopNav and MobileNav components, above the children.

windy-ux:
> Instead of adding the component on every page, juts add it inside the DesktopNav and MobileNav components, above the children.

Done

windy-ux:
Will make new PR for it

PR #5354, opened 2025-01-15
Max/tcp proxy bin sdk readme

mfahampshire:
* move tcp proxy standalone binaries to bin/
* add stub readme to rust sdk

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5354)


octol (commented):
What do you think about having this as a separate crate, sitting side-by-side with the nym-sdk?

octol (approved):

mfahampshire:
> What do you think about having this as a separate crate, sitting side-by-side with the nym-sdk?

I think it totally could. Considering it consumed so much of the SDK / was sort of a 'front' to the SDK I could move it but I also thought it made sense to include as something to import from it. 

I think it really depends on how we want to organise the SDK, I'm happy to roll with whatever we decide :)

PR #5355, opened 2025-01-16
DNS resolver configuration for internal HTTP client lookups

jmwample:
The resolver itself is the set combination of the google, cloudflare, and quad9 endpoints supporting DoH, DoT, and for google DoH3 as well.

This resolver implements a fallback mechanism where, should the DNS-over-X resolution fail, a followup resolution will be done using the hosts configured default (e.g. `/etc/resolve.conf` on linux).

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5355)


common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|ip_addr| SocketAddr::new(ip_addr, 0))
+    }
+}
+
+/// Create a new resolver with a custom DoT based configuration. The options are overridden to look
+/// up for both IPv4 and IPv6 addresses to work with "happy eyeballs" algorithm.
+fn new_resolver() -> Result<TokioAsyncResolver, HickoryDnsSystemConfError> {
+    let mut name_servers = NameServerConfigGroup::google_tls();
```
jstuczyn, line 115: if it's not too much of a hassle (but it might be, so you might want to ignore this comment), would it be possible to allow user to specify which name servers could be used via a `debug` section in their config files? (so nym-node, all flavours of clients, etc). essentially just start off by having an extra argument to this constructor

common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,153 @@
...
+    // name_servers.merge(NameServerConfigGroup::google_h3());
+    name_servers.merge(NameServerConfigGroup::quad9_tls());
+    name_servers.merge(NameServerConfigGroup::quad9_https());
+    name_servers.merge(NameServerConfigGroup::cloudflare_tls());
+    name_servers.merge(NameServerConfigGroup::cloudflare_https());
+
+    let config = ResolverConfig::from_parts(None, Vec::new(), name_servers);
+
+    let mut opts = ResolverOpts::default();
+    opts.ip_strategy = LookupIpStrategy::Ipv4thenIpv6;
```
jstuczyn: shouldn't it actually be the other way around, i.e. prefer ipv6 and use ipv4 as fallback?

common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,153 @@
...
+    // Would like to enable this when 0.25 stabilizes
+    // opts.server_ordering_strategy = ServerOrderingStrategy::RoundRobin;
+
+    Ok(TokioAsyncResolver::tokio(config, opts))
+}
+
+/// Create a new resolver with the default configuration, which reads from `/etc/resolve.conf`. The
```
jstuczyn: you're saying it reads from `/etc/resolve.conf`. obviously this file doesn't exist on all operating systems, so how does it work on, for example, windows?

common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,153 @@
...
+            Ok(addrs)
+        })
+    }
+}
+
+impl Iterator for SocketAddrs {
+    type Item = SocketAddr;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|ip_addr| SocketAddr::new(ip_addr, 0))
```
jstuczyn, line 88: is it safe to use port value of 0 here?

common/http-api-client/src/lib.rs
```diff
@@ -90,7 +91,8 @@ impl ClientBuilder {
                 url: url.into_url()?,
                 timeout: None,
                 custom_user_agent: false,
-                reqwest_client_builder: reqwest::ClientBuilder::new(),
+                reqwest_client_builder: reqwest::ClientBuilder::new()
+                    .dns_resolver(Arc::new(HickoryDnsResolver::default())),
```
jstuczyn: this will not build in `wasm` : (

common/http-api-client/src/dns.rs
jmwample: This is a function of the `hickory_resolver` library. Their implementation is [generic across unix and windows](https://docs.rs/hickory-resolver/latest/src/hickory_resolver/system_conf/mod.rs.html). The `/etc/resolv.conf` is there to say "for example". This comment, and a solid portion of the code for this resolver, came from [the resolver that reqwest implements using hickory](https://docs.rs/reqwest/latest/src/reqwest/dns/hickory.rs.html). The reqwest resolver that uses hickory is not externally configurable in the way I needed so I extended it.  I will update the comment too be more generic for platform target though so as not to be misleading.

common/http-api-client/src/dns.rs
jmwample: This is an easy change to make, hickory [supports basically all different orderings](https://docs.rs/hickory-resolver/latest/hickory_resolver/config/enum.LookupIpStrategy.html). I set this while I was troubleshooting an (ultimately unrelated) issue since I was on a network with only IPv4. `reqwest` sets to `Ipv4andIpv6` so maybe that is what we would want as well.

common/http-api-client/src/dns.rs
jmwample: This was my original goal, but having thought about this while working on the DNS part I think it would be best to leave that for a separate effort. The required extra config is non-trivial when you allow DoH and DoT and the broad scope of places that we are using the API library means to me that we needs some more planning to allow custom DNS from config correctly without creating a giant headache.

common/http-api-client/src/dns.rs
jmwample: I am not sure what the implications are actually. This comes from the `reqwest` implementation, but I will double check that this is safe to do in our context. To start though, the data is coming from the responses to DNS requests, so there isn't really any port to give them here (which SocketAddr requires). I am guessing this is implemented so that you can call things like `tokio::net::TcpStream::connect(...)` on the result.

I will investigate further.

common/http-api-client/src/lib.rs
jmwample: hickory (`reqwest`s choice of alternative resolver supporting DoH & DoT) does not seem to support wasm for now. So this pr only affects non wasm architectures. `wasm32` targets should successfully compile now.

octol:
It's strange to see that CI is not being triggered, I don't know what the issue is. Not seeing any failed triggered builds in the action tab either

common/http-api-client/src/dns.rs
jmwample: Okay so returning an `IntoIter<SockAddr>` is just part of interface that the `reqwest`  uses for dns resolution. This somehow has to do with reqwest internally using `hyper` / `tower` I think. So as long as the `HickoryDnsResolver` is used in conjunction with `reqwest` the ports are managed independently, and the `SockAddr`s with unset ports conforms to expected behavior.

To gain some separation from this I changed the public function to return the same iterator that hickory itself returns which is an `IntoIterator<IpAddr>`. So you have to go out of your way now by using the `reqwest` dns interface to get an iterator of `SockAddr`s (which is the expected behavior for `reqwest`), so shooting youself in the foot with unset ports should be pretty hard.

common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,186 @@
...
+    Ok(TokioAsyncResolver::tokio(config, opts))
+}
+
+impl fmt::Display for HickoryDnsSystemConfError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("error reading DNS system conf for hickory-dns")
+    }
+}
+
+impl std::error::Error for HickoryDnsSystemConfError {
```
octol: Is there a reason why Error is implemented manually instead of just `#[derive(thiserror::Error)]`? Same with `impl fmt::Display`

common/http-api-client/src/dns.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|ip_addr| SocketAddr::new(ip_addr, 0))
+    }
+}
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+impl HickoryDnsResolver {
+    pub async fn resolve_str(&self, name: &str) -> Result<LookupIp, BoxError> {
```
octol: Question, what about just returning the concrete error type?

common/http-api-client/src/dns.rs
jmwample: Done, this was implemented manually in the reqwest hickory resolver, so I was just working from their example. I also didn't know we imported thiserror. Easy change :+1:

common/http-api-client/src/dns.rs
jmwample: Done

octol (approved):

jstuczyn (approved):

PR #5361, opened 2025-01-16
Bind to [::] on nym-node for both IP versions

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5361)


jstuczyn (approved):

benedettadavico:
Test Report: IPv4 Configuration and Migration Testing
Testing Completion Date: 23/01/2025

Testing Steps:
- Initiated and ran a nym-node with version 1.3.1 on an IPv4-only machine, then updated it to the new 'hu', 1.4.0 version.
- Initiated and ran a new nym-node with 'hu', 1.4.0 always on a machine with only IPv4.
- Initiated and ran a new nym-node with 'hu', 1.4.0 on a machine with both IPv4 and IPv6 for regression testing.

Results:
- No functional issues during version updates.
- Logged message on all versions: "no registered client for destination: ff02::2"

Status: Pass

PR #5364, opened 2025-01-17
Max/tssdk docs maintenance

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5364)


windy-ux (commented):
You could create component and import it on the sites so the text and all it needs would be specified on one place.

mfahampshire:
> You could create component and import it on the sites so the text and all it needs would be specified on one place.

Since this is just hopefully for a few days I thought I'd just do a quick copy paste everywhere instead of messing with a component

serinko (approved):

PR #5381, opened 2025-01-22
MixnetClient can send ClientRequests

durch:
Ended up adding another channel to the `MixnetClient`, so that it can signal to the `MixTrafficController` that can send a `ClientRequest` via its `GatewayTransciver`.

Had to add a sleep to the `disconnect` call as the socket would close before the channel message got processed, briefly considered adding an `ack` channel so that we can signal back when the message got sent, but opted for simplicity for now. I've also bumped the default timeout to 10 secs.

Looking at this code, I'm not sure if channels between MixnetClient and MixTrafficController are needed, since messages are processed serially, we might be able to pass an `Arc<Mutex<MixTrafficController>>` to MixnetClient and use it directly, remove a level of indirection from the code, nothing actionable now just shower thoughts... 

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5381)

### DIFF
--- a/sdk/rust/nym-sdk/src/mixnet/native_client.rs
+++ b/sdk/rust/nym-sdk/src/mixnet/native_client.rs
@@ -5,12 +5,15 @@ use async_trait::async_trait;
 use futures::{ready, Stream, StreamExt};
 use log::error;
 use nym_client_core::client::base_client::GatewayConnection;
+use nym_client_core::client::mix_traffic::ClientRequestSender;
 use nym_client_core::client::{
     base_client::{ClientInput, ClientOutput, ClientState},
     inbound_messages::InputMessage,
     received_buffer::ReconstructedMessagesReceiver,
 };
+use nym_client_core::ForgetMe;
 use nym_crypto::asymmetric::identity;
+use nym_gateway_requests::ClientRequest;
 use nym_sphinx::addressing::clients::Recipient;
 use nym_sphinx::{params::PacketType, receiver::ReconstructedMessage};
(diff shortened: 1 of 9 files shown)


sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -215,11 +215,12 @@ impl MixnetClient {
     /// client.
     pub async fn disconnect(mut self) {
         if self.forget_me.any() {
-            log::info!("Sending forget me request: {:?}", self.forget_me);
+            log::debug!("Sending forget me request: {:?}", self.forget_me);
             match self.send_forget_me().await {
                 Ok(_) => (),
                 Err(e) => error!("Failed to send forget me request: {}", e),
             };
+            tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
jstuczyn, line 223: this seems like a very fragile solution if we depend on extra 2s sleep : / 
I wonder, would it make sense to maybe add some `Arc<Notify>` on the `forget_me` message so we could wait (with some timeout) until it gets sent?

jstuczyn (approved):

sdk/rust/nym-sdk/src/mixnet/native_client.rs
durch: You're absolutely right, great call on the notify, will look into it quickly

durch: Unfortunatley a no go because of serialization.

PR #5386, opened 2025-01-23
Use secure DNS for websocket connection establishment

jmwample:
Adjust the websocket connection establishment (using `tokio_tungstenite`) to use a custom DNS resolver that tries DoH and DoT.

instead of relying on the `tokio_tungstenite` to open the `tokio::net::TcpStream` ([which it does here](https://github.com/snapview/tokio-tungstenite/blob/aafb2f9e036162f7bffa002cfea502376a690724/src/connect.rs#L91)) which would kick in the default network resolver we do the resolution and open the `TcpStream` ourselves before handing off to the websocket libary, 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5386)

### DIFF
--- a/common/client-core/src/error.rs
+++ b/common/client-core/src/error.rs
@@ -36,6 +36,9 @@ pub enum ClientCoreError {
     #[error("no gateway with id: {0}")]
     NoGatewayWithId(String),
 
+    #[error("Invalid URL: {0}")]
+    InvalidURL(String),
+
     #[error("no gateways on network")]
     NoGatewaysOnNetwork,
 
--- a/common/client-core/src/error.rs
+++ b/common/client-core/src/error.rs
@@ -37,7 +37,7 @@ pub enum ClientCoreError {
     NoGatewayWithId(String),
 
     #[error("Invalid URL: {0}")]
-    InvalidURL(String),
+    InvalidUrl(String),
 
     #[error("no gateways on network")]
     NoGatewaysOnNetwork,
--- a/common/client-core/src/error.rs
+++ b/common/client-core/src/error.rs
@@ -39,6 +39,7 @@ pub enum ClientCoreError {
     #[error("Invalid URL: {0}")]
     InvalidUrl(String),
 
+    #[cfg(not(target_arch = "wasm32"))]
     #[error("resolution failed: {0}")]
     ResolutionFailed(#[from] nym_http_api_client::HickoryDnsError),
 
--- a/common/client-core/src/error.rs
+++ b/common/client-core/src/error.rs
@@ -39,6 +39,9 @@ pub enum ClientCoreError {
     #[error("Invalid URL: {0}")]
     InvalidUrl(String),
 
+    #[error("resolution failed: {0}")]
+    ResolutionFailed(#[from] nym_http_api_client::HickoryDnsError),
+
     #[error("no gateways on network")]
     NoGatewaysOnNetwork,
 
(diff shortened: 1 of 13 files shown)


common/client-core/src/error.rs
```diff
@@ -36,6 +36,9 @@ pub enum ClientCoreError {
     #[error("no gateway with id: {0}")]
     NoGatewayWithId(String),
 
+    #[error("Invalid URL: {0}")]
+    InvalidURL(String),
```
octol: Sorry for being picky about it, but that capitalization makes my right eye twitch ;)
Convention in Rust is InvalidUrl

jmwample: I agree. However, this was taken as is from the GatewayClient - https://github.com/nymtech/nym/blob/9d014742776d880ad5f467868fed226428a80c4c/common/client-libs/gateway-client/src/error.rs#L45-L48

So there is precedent in this case! I can change both though if we want lol.

jmwample: Done

common/client-libs/gateway-client/src/client/websockets.rs
```diff
@@ -0,0 +1,60 @@
...
+        Host::Domain(domain) => {
+            // Do a DNS lookup for the domain using our custom DNS resolver
+            resolver
+                .resolve_str(domain)
+                .await
+                .map_err(|_| {
+                    // failed to resolve
+                    GatewayClientError::NetworkConnectionFailed {
+                        address: endpoint.to_owned(),
+                        source: UrlError::NoPathOrQuery.into(),
```
octol: we are not forwarding the source error from resolve_str?

common/client-core/src/init/websockets.rs
```diff
@@ -0,0 +1,50 @@
...
+                })?
+                .into_iter()
+                .map(|a| SocketAddr::new(a, port))
+                .collect()
+        }
+    };
+
+    let stream = TcpStream::connect(&sock_addrs[..]).await?;
+
+    tokio_tungstenite::client_async_tls(endpoint, stream)
```
jstuczyn, line 41: but we can't use `tls` unconditionally. as a matter of fact most of our gateways do **not** support it

jmwample: I am pretty sure this is how  tokio_tungstnite is meant to be used for hosts that maybe support tls. The function naming is misleading, [in the docs](https://docs.rs/tokio-tungstenite/latest/tokio_tungstenite/fn.client_async_tls.html) it says:

```
Creates a WebSocket handshake from a request and a stream, upgrading the stream to TLS if required.
```
 
Also if you look at the control flow from `tokio_tungstanite::connect_async()` (which we were using previous to these changes) inside of the library it opens a `tokio::TcpStream` and then calls (effectively) `client_async_tls` [here](https://github.com/snapview/tokio-tungstenite/blob/aafb2f9e036162f7bffa002cfea502376a690724/src/connect.rs#L97).

It would definitely be good to double check that this works the way we want it to, I am not sure what the easiest way to do that would be.

common/client-libs/gateway-client/src/client/websockets.rs
jmwample: This is how a resolution error would be passed on to us from `tokio_tungstenite` before this change. If we want to capture and forward the error we can. I can add a specific error enum item `ResolutionFailed` or something.

jmwample (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5386)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @aniampio, @jstuczyn, and @octol)

common/client-libs/gateway-client/src/client/websockets.rs
jmwample: Error forwarding added.

octol (approved):
Assuming it works in wasm, looks good!

jmwample:
Wasm is intentionally and explicitly (via compiler directives) excluded from these changes. There is no change to the wasm behavior.

PR #5400, opened 2025-01-29
Yana/pending staking events

yanok87:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5400)


fmtabbara (approved):

explorer-nextjs/src/app/(pages)/account/[address]/page.tsx
```diff
@@ -1,81 +1,45 @@
...
-import ExplorerButtonGroup from "@/components/toggleButton/ToggleButton";
+import { fetchNodes } from "@/app/api";
+import type { NodeData } from "@/app/api/types";
 import { Box, Typography } from "@mui/material";
 import Grid from "@mui/material/Grid2";
+import { AccountBalancesCard } from "../../../../components/accountPageComponents/AccountBalancesCard";
+import { AccountInfoCard } from "../../../../components/accountPageComponents/AccountInfoCard";
+import { ContentLayout } from "../../../../components/contentLayout/ContentLayout";
+import SectionHeading from "../../../../components/headings/SectionHeading";
+import ExplorerButtonGroup from "../../../../components/toggleButton/ToggleButton";
```
Comment (fmtabbara) [inline explorer-nextjs/src/app/(pages)/account/[address]/page.tsx:10]: can you use the `@` import alias for imports. It's a bit neater.

explorer-nextjs/src/components/toggleButton/ToggleButton.tsx
```diff
@@ -12,17 +18,64 @@ type Options = [Option, Option];
...
+  const handleRefetch = useCallback(() => {
+    queryClient.invalidateQueries();
+  }, [queryClient]);
+
+  useEffect(() => {
+    if (!hasEpochStarted) return;
+    handleRefetch();
+    const interval = setInterval(handleRefetch, 30000);
+    return () => clearInterval(interval);
+  }, [hasEpochStarted, handleRefetch]);
```
fmtabbara: this logic cant live in this component. what if want to use the component somewhere else in the app and have it do something different? It should have an onToggle prop or the logic should live on the page it needs to work on

explorer-nextjs/src/components/toggleButton/ToggleButton.tsx
```diff
@@ -12,17 +18,64 @@ type Options = [Option, Option];
...
+      if (!data?.dateTime) return;
+      const oneHourLater = subSeconds(new Date(data.dateTime), 30).getTime();
+      const now = Date.now();
+      setHasEpochStarted(now >= oneHourLater);
+    };
+
+    checkEpochStatus();
+    const interval = setInterval(checkEpochStatus, 60000);
+    return () => clearInterval(interval);
+  });
```
fmtabbara: can you explain this as I'm not fully clear on what is going on or why we need this?

explorer-nextjs/src/components/toggleButton/ToggleButton.tsx
```diff
@@ -12,17 +18,64 @@ type Options = [Option, Option];
...
+  useEffect(() => {
+    const checkEpochStatus = () => {
+      if (!data?.dateTime) return;
+      const oneHourLater = subSeconds(new Date(data.dateTime), 30).getTime();
+      const now = Date.now();
+      setHasEpochStarted(now >= oneHourLater);
+    };
+
+    checkEpochStatus();
+    const interval = setInterval(checkEpochStatus, 60000);
```
fmtabbara: if this block of logic is used in other parts of the app the it would be worth making a custom hook. But before that let's figure our what it's doing and why it's needed.

fmtabbara (changes requested):

explorer-nextjs/src/app/(pages)/account/[address]/page.tsx
Comment (yanok87) [reply explorer-nextjs/src/app/(pages)/account/[address]/page.tsx:10]: for some reason I've started having import red errors with "@/...", have refactored to this to get rid of the errors

explorer-nextjs/src/components/toggleButton/ToggleButton.tsx
yanok87: ok, will find a better component to hang the epoch listening functionality on it

explorer-nextjs/src/components/toggleButton/ToggleButton.tsx
yanok87: when the epoch time changes, there's a delay of the new data being returned from the fetch request, because there is a cache on the contract data. So, even if it's past the current epoch time, the fetch requests still return the last epoch time, resulting into bad UX. For example,  it looks buggy if it's written next epoch starts at 18:26pm and it's already 18:29pm. I've tested and had up to 5min delay... It's always different, but always a delay.
Therefore, I was advised to show "Waiting for the next epoch to start.." until we get new data back.
Also, if the new epoch has started, I'm resetting all the queries and fetching new data across the app. Lot's of queries get updated even if the contract epoch hasn't returned new epoch time yet, because not all endpoints are updated simultaneously with the epoch change. Some update sooner, some later.

fmtabbara: why don't we check the current time against most recently fetched epoch time, if the currenttime > epochtime show "getting new epoch time" (or something like that). If we check the epoch every 30 or 60 secs eventually it's going to show correctly?

The frontend shouldnt resolving issues in epoch timing. If there's an issue with caching on the BE it should be fixed there.

explorer-nextjs/src/app/(pages)/account/[address]/page.tsx
Comment (fmtabbara) [reply explorer-nextjs/src/app/(pages)/account/[address]/page.tsx:10]: what are the errors?

explorer-nextjs/src/providers/EpochProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+      );
+
+      setEpochStatus("active");
+      return newRefetchInterval;
+    },
+  });
+
+  useEffect(() => {
+    const refreshQueries = async () => {
+      await QueryClient.invalidateQueries({
```
yanok87, line 86: @fmtabbara this fixes the infinite loop :)

PR #5435, opened 2025-02-07
Remove all recv_with_delay and add shutdown condition to loops in client-core

octol:
Inside client-core we want to prepare the ground for moving a behaviour
close to what we have in the vpn client.

Remove all the recv_with_delay since we want to just stop

Add shutdown condition to all select loops to guard against the shutdown
listener being polled inside the select blocks.

Remove unwraps when sending on unbounded channels in case the receiver exits before the sender

Move `TaskClient` to be a member field so make it easier to wrap log errors in a shutdown check.

Update all fork names to use underscore consistently, since the task separator is hyphen

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5435)

### DIFF
--- a/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
+++ b/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
@@ -224,10 +224,7 @@ where
         }
     }
 
-    async fn on_message(
-        &mut self,
-        next_message: StreamMessage,
-    ) {
+    async fn on_message(&mut self, next_message: StreamMessage) {
         trace!("created new message");
 
         let (next_message, fragment_id, packet_size) = match next_message {
--- a/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
+++ b/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
@@ -198,7 +198,9 @@ where
         // queues and client load rather than the required delay. So realistically we can treat
         // whatever is about to happen as negligible additional delay.
         trace!("{} is about to get sent to the mixnet", frag_id);
-        self.sent_notifier.unbounded_send(frag_id).unwrap();
+        if let Err(err) = self.sent_notifier.unbounded_send(frag_id) {
+            error!("Failed to notify about sent message: {err}");
+        }
     }
 
     fn loop_cover_message_size(&mut self) -> PacketSize {
--- a/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
+++ b/common/client-core/src/client/real_messages_control/real_traffic_stream.rs
@@ -219,7 +219,11 @@ where
         }
     }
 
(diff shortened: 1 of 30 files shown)


common/statistics/src/clients/mod.rs
```diff
@@ -36,9 +36,9 @@ impl ClientStatsSender {
     /// Report a statistics event using the sender.
     pub fn report(&self, event: ClientStatsEvents) {
         if let Some(tx) = &self.stats_tx {
-            if let Err(err) = tx.send(event) {
-                log::error!("Failed to send stats event: {:?}", err);
-            }
+            tx.send(event)
+                .inspect_err(|err| log::debug!("Failed to send stats event: {err}"))
```
jstuczyn: nit-question: any reason for downgrading it from `error` to `debug`?

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -271,7 +277,9 @@ where
         };
 
         if let Err(err) = self.mix_tx.send(vec![next_message]).await {
-            log::error!("Failed to send: {err}");
+            if !task_client.is_shutdown_poll() {
```
jstuczyn: would it perhaps be possible to move `TaskClient` to `Self` and call it that way instead? I think that's quite a good pattern to only log those errors if we're not in shutdown

jstuczyn (approved):

common/statistics/src/clients/mod.rs
octol: Mostly because it proved difficult to pass a TaskClient all the way down here and sending stats didn't seem critical (compared to other tasks)

But I'll see if I can maybe change that by moving the TaskClient from the run call arguments to the task structs instead

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
octol: The problem is that we end up having two select branches with `&mut self`. (With `CancellationToken` this will not be a problem)

octol: Well I guess we can fork a clone for use in the select loop specifically

jstuczyn (approved):
lgtm, let's just fix that contract issue and rebase the branch first

octol:
Notes for testing: this has potential impact for all binaries. Things to look for is things like the `nym-node` crashing with `UnexpectedHalt` error

benedettadavico:
Testing Completion Date: [28/02/2025]

Testing Steps Performed:
Validated that all binaries including nym-node, nym-client, nym-network-requester, and nym-socks5-client are behaving well without indicating the presence of any unexpected errors or crashes

Validation:
[x] Test Passed
[ ] Test Failed

Notes:
Also tested the vpn client, and everything is behaving well

Attachments:
N/A

PR #5440, opened 2025-02-10
Seedable clients

durch:
### `Adds DerivationMaterial and accompanying methods to builders`

`DerivationMaterial` encapsulates parameters for deterministic key derivation using HKDF (SHA-512).

It consists of:

- **Master Key (`master_key`)**: The base secret.
- **Index (`index`)**: Ensures unique derivations.
- **Salt (`salt`)**: Adds additional uniqueness.

Use the `derive_secret()` method to generate a 32-byte secret. To prepare for a new derivation, call the `next()` method which increments the index. **It is the caller's responsibility to track and persist the derivation index if keys need to be rederived.**

#### Example

```rust
let master_key = [0u8; 32]; // your secret master key
let salt = "unique-salt-value".to_string();
let material = DerivationMaterial::new(master_key, 0, salt.as_bytes());

// Derive a secret
let secret = material.derive_secret().expect("Failed to derive secret");

// Prepare for the next derivation
let next_material = material.next();
```
---

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5440)

### DIFF
--- a/common/crypto/src/asymmetric/identity/mod.rs
+++ b/common/crypto/src/asymmetric/identity/mod.rs
@@ -82,13 +82,13 @@ impl KeyPair {
         }
     }
 
-    pub fn from_secret(secret: SecretKey) -> Self {
+    pub fn from_secret(secret: SecretKey, index: u32) -> Self {
         let ed25519_signing_key = SigningKey::from(secret);
 
         KeyPair {
             private_key: PrivateKey(ed25519_signing_key.to_bytes()),
             public_key: PublicKey(ed25519_signing_key.verifying_key()),
-            index: fake_index(secret.as_slice()),
+            index,
         }
     }
 
--- a/common/crypto/src/asymmetric/identity/mod.rs
+++ b/common/crypto/src/asymmetric/identity/mod.rs
@@ -76,7 +76,7 @@ impl KeyPair {
         KeyPair {
             private_key: PrivateKey(ed25519_signing_key.to_bytes()),
             public_key: PublicKey(ed25519_signing_key.verifying_key()),
-            index: index,
+            index,
         }
     }
 
--- a/common/crypto/src/asymmetric/identity/mod.rs
+++ b/common/crypto/src/asymmetric/identity/mod.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub use ed25519_dalek::SignatureError;
-use ed25519_dalek::{Signer, SigningKey};
+use ed25519_dalek::{SecretKey, Signer, SigningKey};
 pub use ed25519_dalek::{Verifier, PUBLIC_KEY_LENGTH, SECRET_KEY_LENGTH, SIGNATURE_LENGTH};
 use nym_pemstore::traits::{PemStorableKey, PemStorableKeyPair};
 use std::fmt::{self, Debug, Display, Formatter};
@@ -18,7 +18,7 @@ pub mod serde_helpers;
 use nym_sphinx_types::{DestinationAddressBytes, DESTINATION_ADDRESS_LENGTH};
 
 #[cfg(feature = "rand")]
-use rand::{CryptoRng, RngCore};
+use rand::{CryptoRng, Rng, RngCore};
 #[cfg(feature = "serde")]
 use serde::de::Error as SerdeError;
 #[cfg(feature = "serde")]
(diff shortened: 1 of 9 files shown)


common/crypto/src/asymmetric/identity/mod.rs
```diff
@@ -62,16 +62,31 @@ pub struct KeyPair {
...
     public_key: PublicKey,
+
+    #[zeroize(skip)]
+    index: u32,
 }
 
 impl KeyPair {
     #[cfg(feature = "rand")]
     pub fn new<R: RngCore + CryptoRng>(rng: &mut R) -> Self {
+        let index = rng.gen();
```
jstuczyn, line 75: I'm not sure I understand the point of having a random index for the random keys

common/crypto/src/asymmetric/identity/mod.rs
```diff
@@ -87,15 +102,31 @@ impl KeyPair {
         Ok(KeyPair {
             private_key: PrivateKey::from_bytes(priv_bytes)?,
             public_key: PublicKey::from_bytes(pub_bytes)?,
+            index: fake_index(pub_bytes),
```
jstuczyn, line 107: but the resultant `index` here will be different when you call `KeyPair::from_bytes` and `KeyPair::from_secret` even if you use the same underlying keypair - is that intended?

common/crypto/src/hkdf.rs
```diff
@@ -8,6 +8,9 @@ use hkdf::{
     },
     Hkdf,
 };
+use sha2::Sha512;
```
jstuczyn: nit: if we're using a new hkdf somewhere, we might as well go with the blake3 digest because it's supposedly better than sha

simonwicky (approved):

common/crypto/src/hkdf.rs
```diff
@@ -28,3 +31,64 @@ where
...
+/// let salt = "unique-salt-value".to_string();
+/// let material = DerivationMaterial::new(master_key, 0, salt);
+///
+/// // Derive a secret
+/// let secret = material.derive_secret().expect("Failed to derive secret");
+///
+/// // Prepare for the next derivation
+/// let next_material = material.next();
+/// ```
+pub struct DerivationMaterial {
```
jstuczyn, line 62: this guy should probably be `ZeroizeOnDrop`

common/crypto/src/hkdf.rs
```diff
@@ -28,3 +31,64 @@ where
...
+    master_key: [u8; 32],
+    index: u32,
+    salt: String,
```
jstuczyn: why not use proper bytes for the salt rather than a `String`?

common/crypto/src/hkdf.rs
```diff
@@ -28,3 +31,64 @@ where
...
+}
+
+impl DerivationMaterial {
+    /// Derives a 32-byte seed from a master seed and an index using HKDF (with SHA-512).
+    ///
+    /// The `salt` and the use of the index (as info) bind this derivation to an application/client.
+    pub fn derive_secret(&self) -> Result<[u8; 32], hkdf::InvalidLength> {
```
jstuczyn, line 80: also it seems like we might be reinventing the wheel here a bit, aren't there some libraries for this purpose out there? something like bip32 maybe?

durch: I like the bip32 idea, it should be lighter then hkdf if nothing else

common/crypto/src/asymmetric/identity/mod.rs
durch: Yeah, this is not a derived key, we just need something that is not None or a constant, in case something somewhere inspects the keys and behaves differently if it can figure out they're derived

common/crypto/src/asymmetric/identity/mod.rs
durch: Its just to have something, that is not None or constant

durch: I'll document the intent, otherwise I'll forget :D

common/crypto/src/hkdf.rs
durch: blake3::Hasher can't be natively used, misses the CryptoProxy trait, so I'll keep it as is

PR #5450, opened 2025-02-11
Condense core API functionalities and enable gzip decompression for reqwest payloads

jmwample:
This PR is intended to take out variables from our usage of HTTP requests in the `nym-http-api-client`. 

To do this the PR:
* Adds the `ApiClientCore` trait with the minimal feature required to send a request
* Turns all request sending into trait extension automatically implemented for any type that implements `ApiClientCore`

This has the benefit of consistent expected behavior for all clients using `nym-http-api-client`, including features added going forward. FOR EXAMPLE this pr adds a default header `"accept-encoding: gzip;q=1.0, *;q=0.5"` which indicates that compression is preferred whenever available. Other features to keep in mind here are things like configurable retries, domain fallbacks, etc.

Note: the `Apiclient` interface could be simplified, but that would require refactoring our downstream usages of the API. For now this isn't necessary as `ApiClient` is implemented automatically so it costs nothing to have it this way. It just allows divergent usage in downstream crates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5450)

### DIFF
--- a/common/http-api-client/src/lib.rs
+++ b/common/http-api-client/src/lib.rs
@@ -1,18 +1,154 @@
 // Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+//! Nym HTTP API Client
+//!
+//! Centralizes and implements the core API client functionality. This crate provides custom,
+//! configurable middleware for a re-usable HTTP client that takes advantage of connection pooling
+//! and other benefits provided by the [`reqwest`] `Client`.
+//!
+//! ## Making GET requests
+//!
+//! Create an HTTP `Client` and use it to make a GET request.
+//!
+//! ```rust
+//! # use url::Url;
+//! # use nym_http_api_client::{ApiClient, NO_PARAMS, HttpClientError};
+//!
+//! # type Err = HttpClientError<String>;
+//! # async fn run() -> Result<(), Err> {
+//! let url: Url = "https://nymvpn.com".parse()?;
+//! let client = nym_http_api_client::Client::new(url, None);
+//!
+//! // Send a get request to the `/v1/status` path with no query parameters.
+//! let resp = client.send_get_request(&["v1", "status"], NO_PARAMS).await?;
+//! let body = resp.text().await?;
+//!
+//! println!("body = {body:?}");
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! ## JSON
+//!
+//! There are also json helper methods that assist in executing requests that send or receive json.
+//! It can take any value that can be serialized into JSON.
+//!
+//! ```rust
+//! # use std::collections::HashMap;
+//! # use std::time::Duration;
+//! use nym_http_api_client::{ApiClient, HttpClientError, NO_PARAMS};
+//!
+//! # use serde::{Serialize, Deserialize};
+//! #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+//! pub struct ApiHealthResponse {
+//!     pub status: ApiStatus,
+//!     pub uptime: u64,
+//! }
+//!
+//! #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+//! pub enum ApiStatus {
+//!     Up,
+//! }
+//!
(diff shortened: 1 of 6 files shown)


tommyv1987 (approved):

tommyv1987 (approved):

common/http-api-client/src/lib.rs
```diff
@@ -96,8 +244,17 @@ impl ClientBuilder {
...
+            let reqwest_client_builder = {
+                let r = reqwest::ClientBuilder::new()
+                    .dns_resolver(Arc::new(HickoryDnsResolver::default()));
+
+                // Note this is extra as the `gzip` feature for `reqwest` crate should be enabled which
+                // `"Enable[s] auto gzip decompression by checking the Content-Encoding response header."`
+                //
+                // I am going to leave it here anyways so that gzip decompression is attempted even if
+                // that feature is removed.
+                r.gzip(true)
```
jstuczyn, line 256: does that mean the client will always be sending compressed requests? if so, should it maybe be a builder method instead (that'd be enabled by default). though then again, i dont see any reason where you wouldn't want to use it

common/http-api-client/src/lib.rs
```diff
@@ -215,95 +397,120 @@ impl Client {
...
-        V: AsRef<str>,
+        S: AsRef<str>,
     {
-        let url = sanitize_url(&self.base_url, path, params);
-        self.reqwest_client.post(url).json(json_body)
+        // Use a stand-in url to extract the path and queries from the provided endpoint string
+        // which could potentially fail.
+        //
+        // This parse cannot fail
+        let mut standin_url: Url = "http://example.com".parse().unwrap();
```
jstuczyn, line 431: why can't we use the provided `endpoint` instead? (if we have to change the generic parameters slightly, I think that's fine). the standin here is quite odd to be honest

common/http-api-client/src/lib.rs
```diff
@@ -215,95 +397,120 @@ impl Client {
...
         K: AsRef<str>,
         V: AsRef<str>,
-        E: Display,
     {
         let url = sanitize_url(&self.base_url, path, params);
 
         let mut request = self.reqwest_client.request(method.clone(), url);
 
+        // Indicate that compressed responses are preferred, but if not supported other encodings are fine.
+        // TODO: Down the road we can be more selective about adding this, but it's inclusion here guarantees
```
jstuczyn, line 499: definitely something we should do down the road : )

jstuczyn (commented):
I love all the docs you added, I wish we had time to do it everywhere ourselves : (

octol (approved):

common/http-api-client/src/lib.rs
jmwample: > does that mean the client will always be sending compressed requests?

I don't believe that this impacts outgoing request construction in any way. My interpretation of the [docs](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.gzip) is that this ensures that HTTP responses with `"gzip"` in the `Content-Encoding` header (indicating a compressed payload) will be automatically decompressed by the `reqwest` crate for us.

common/http-api-client/src/lib.rs
jmwample: The `url` crate wont parse things as a `URL` if they are missing a base or host. The stand-in url is there so that those pieces are guaranteed to be present. In this function the endpoint param is ONLY path and/or query parameters that we will add to the `base_url` for this Client in `create_request`. However because I wanted a default implementation for this there is no `base_url` available -- hence the stand-in.

aniampio:
Are we gonna manage to squeeze this into this release before the code freezes?

benedettadavico:
_Testing Completion Date:_ [24/02/2025]

_Testing Steps Performed:_
Measured nym-api response times before and after updating the API client using curl
Checked if the client automatically decompressed Gzip responses

_Validation:_
 [x] Test Passed
 [ ] Test Failed

_Notes:_
Response times improved significantly for most endpoints
Minor response time increase observed for `/unstable/nym-nodes/gateways/skimmed` for example

_Attachments:_
N/A

PR #5451, opened 2025-02-11
Add a middleware layer to the nym api allowing for data compression

jmwample:
~~The changes in this PR are based on theory and documentation for axum and tower-http. I have not actually been able to test that this work with an `accept-encoding` header to compress data in a response yet.~~

After Testing in a minimal example, this does work as expected. Routes still default to plain encoding, however if a client indicates support for a compressed encoding using the `Accept-Encoding` header then the served response will be compressed with the appropriate `Content-Encoding` header. ([Proof-of-Concept](https://gist.github.com/jmwample/c15a983e804fc338fee3d1b037d216b0))

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5451)


nym-api/src/network/handlers.rs
```diff
@@ -6,16 +6,29 @@ use crate::support::http::state::AppState;
...
 use utoipa::ToSchema;
 
 pub(crate) fn nym_network_routes() -> Router<AppState> {
+    // This should compress responses unless:
+    //
+    // * They’re gRPC, which has its own protocol specific compression scheme.
+    // * It’s an image as determined by the content-type starting with image/.
+    // * They’re Server-Sent Events (SSE) as determined by the content-type being text/event-stream.
+    // * The response is less than 32 bytes.
+    let compression_layer: CompressionLayer = CompressionLayer::new()
```
jstuczyn: given you're creating this compression layer in 4 different spots, would it perhaps make sense to create a helper method for it instead? So that if we decide to change one of the settings, we wouldn't accidentally miss one of the instances.
Perhaps you could put it into `http-api-common/middleware` so that other APIs could benefit from that too

jstuczyn (approved):

octol (approved):

benedettadavico:
_Testing Completion Date:_ [24/02/2025]

_Testing Steps Performed:_
Sent requests with and without Accept-Encoding: gzip to observe response behaviour
Verified if responses included content-encoding: gzip when compression was requested
Checked response file output to confirm actual compression occurred

_Validation:_
 [x] Test Passed
 [ ] Test Failed

_Notes:_
The server now correctly compresses responses when requested, improving efficiency
Compression leads to lower bandwidth usage and faster response times for clients
No impact on API behaviour when Accept-Encoding: gzip is omitted, ensuring backward compatibility

_Attachments:_
N/A

PR #5455, opened 2025-02-12
Move all workflows on ubuntu-20 to ubuntu-22

octol:
Ubuntu 20.04 runners are scheduled to be removed from set of github hosted runners. Now is as good a time as any to finally take the plunge to switch away from ubuntu-20.04 to the shiny oh so modern ubuntu-22.04 runners

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5455)

### DIFF
--- a/.github/workflows/cd-docs.yml
+++ b/.github/workflows/cd-docs.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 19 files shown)


jstuczyn (approved):

octol:
Fails are due to new rustc released today. Merging as is

PR #5456, opened 2025-02-12
Provide Interval context with node descriptor endpoints

jmwample:
Add current interval context information to existing enpoints using `build_skimmed_nodes_response` under the hood. This allows clients checking for a refresh to send the `epoch_uid` as a query parameter when fetching updates so the server can tell it that there have been no changes, instead of sending duplicate data over and over.

The changes in this PR should be backwards compatible and never interfere with existing clients. The additions to the `NodeParams` are optional, so there is no error if a client does not send them. Old clients will not send `epoch_uid` by accident so they cannot accidentally clear their set of known nodes.

In the response the status field is optional so if it is missing (e.g. if a new client talks to an old server) the connection still works. For old clients speaking to new servers, the json parsing will simply ignore extra information not included in the objects json spec.

Open to feedback on the approach here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5456)

### DIFF
--- a/nym-api/src/nym_nodes/handlers/unstable/mod.rs
+++ b/nym-api/src/nym_nodes/handlers/unstable/mod.rs
@@ -114,7 +114,7 @@ impl From<NodesParamsWithRole> for NodesParams {
             no_legacy: params.no_legacy,
             page: params.page,
             per_page: params.per_page,
-            epoch_uid: params.epoch_uid,   
+            epoch_uid: params.epoch_uid,
         }
     }
 }
--- a/nym-api/src/nym_nodes/handlers/unstable/mod.rs
+++ b/nym-api/src/nym_nodes/handlers/unstable/mod.rs
@@ -85,6 +85,11 @@ struct NodesParamsWithRole {
     no_legacy: Option<bool>,
     page: Option<u32>,
     per_page: Option<u32>,
+
+    // Identifier for the current epoch of the topology state. When sent by a client we can check if
+    // the client already knows about the latest topology state, allowing a `no-updates` response
+    // instead of wasting bandwidth serving an unchanged topology.
+    epoch_uid: Option<u32>,
 }
 
 #[derive(Debug, Deserialize, utoipa::IntoParams)]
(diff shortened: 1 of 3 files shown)


nym-api/src/nym_nodes/handlers/unstable/mod.rs
```diff
@@ -85,6 +85,11 @@ struct NodesParamsWithRole {
     no_legacy: Option<bool>,
     page: Option<u32>,
     per_page: Option<u32>,
+
+    // Identifier for the current epoch of the topology state. When sent by a client we can check if
+    // the client already knows about the latest topology state, allowing a `no-updates` response
+    // instead of wasting bandwidth serving an unchanged topology.
+    epoch_uid: Option<u32>,
```
jstuczyn: just a nit: in all other places in our code base, we use `epoch_id` rather than `epoch_uid`

nym-api/src/nym_nodes/handlers/unstable/skimmed.rs
```diff
@@ -124,6 +124,21 @@ where
     // (ideally it'd be tied directly to the NI iterator, but I couldn't defeat the compiler)
     let describe_cache = state.describe_nodes_cache_data().await?;
 
+    let maybe_interval = state
+        .nym_contract_cache()
+        .current_interval()
+        .await
+        .to_owned();
+
+    // 4.0 If the client indicates that they already know about the current topology send empty response
```
jstuczyn, line 133: just thinking aloud here: I guess in the longer term, there'd need to be some even further identifier to distinguish between updates mid-epoch, e.g. some node changed its ip address or port, etc. because ideally the client should receive those as soon as possible rather than having to wait until next epoch

jstuczyn (approved):

aniampio:
Are we gonna manage to squeeze this one for the release?

benedettadavico:
Testing Completion Date: [25/02/2025]

Testing Steps Performed:
Made a request to the `/api/v1/unstable/nym-nodes/semi-skimmed` endpoint:
- without entering an epoch_id
- with entering the current epoch_id
- with entering an old epoch_id
- with entering a future epoch_id
All results returned the same data, as expected

Validation:
[x] Test Passed
[ ] Test Failed

Notes:
The `"status": ` field is always `null` at the moment and will be fixed in a future release alongside some other changes

Attachments:
N/A

PR #5479, opened 2025-02-17
feature: upgrade cosmwasm to 2.2

jstuczyn:
this PR updates our contracts (and all relevant code) to use the most recent version of cosmwasm, i.e. 2.2. 

to use it, however, we first have to complete the chain upgrade

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5479)


octol (approved):
Seems good to me, nice that we can bump the toolchain version

.github/workflows/publish-nym-contracts.yml
```diff
@@ -2,19 +2,18 @@ name: publish-nym-contracts
...
   workflow_dispatch:
   release:
-    types: [created]
+    types: [ created ]
 
 jobs:
   build:
     if: ${{ (startsWith(github.ref, 'refs/tags/nym-contracts-') && github.event_name == 'release') || github.event_name == 'workflow_dispatch' }}
-    runs-on: [self-hosted, custom-ubuntu-20.04]
+    runs-on: [ self-hosted, custom-ubuntu-20.04 ]
```
octol: This is a bit unexpected though, should be maybe use the GH hosted or ARC ubuntu runners here instead?

jstuczyn: good question - I don't know. for now I'd rather not change what's not needed : )

PR #5494, opened 2025-02-20
Connection fd callback before actual connection

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5494)

### DIFF
--- a/common/client-libs/gateway-client/src/client/websockets.rs
+++ b/common/client-libs/gateway-client/src/client/websockets.rs
@@ -1,6 +1,11 @@
 use crate::error::GatewayClientError;
 
 use nym_http_api_client::HickoryDnsResolver;
+#[cfg(unix)]
+use std::{
+    os::fd::{AsRawFd, RawFd},
+    sync::Arc,
+};
 use tokio::net::TcpStream;
 use tokio_tungstenite::{MaybeTlsStream, WebSocketStream};
 use tungstenite::handshake::client::Response;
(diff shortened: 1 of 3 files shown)


common/client-libs/gateway-client/src/client/websockets.rs
```diff
@@ -37,14 +45,41 @@ pub(crate) async fn connect_async(
...
+        }
+        .map_err(|err| GatewayClientError::NetworkConnectionFailed {
             address: endpoint.to_owned(),
-            source: error.into(),
+            source: err.into(),
+        })?;
+
+        #[cfg(unix)]
+        if let Some(callback) = connection_fd_callback.as_ref() {
+            callback.as_ref()(socket.as_raw_fd());
```
jstuczyn, line 64: sanity check: do we want to call the callback on **every** possible socket (because we can resolve to multiple socket addresses) or just the one we succeeded? I don't know what that callback is doing so it's very likely it's intended

neacsu: We only want to call the callback on the **one** socket that is used in the end, but we can't know which one that would be until we call connect. And the callback is designed to be called before actually connecting, otherwise we wouldn't have infiltrated it until this point. The main reason for the callback is to have a custom way (depending on platform) of bypassing obstacles that would prevent the `connect` function to work (firewall, mobile sandboxes etc.)

neacsu: In the less probable case that we do go through more then one socket for some reason, there shouldn't be a problem to have called the function on the previous `fd`s, as they would get dropped and freed anyway in the next iteration.

jstuczyn (approved):

common/client-libs/gateway-client/src/client/websockets.rs
jstuczyn: gotcha. I didn't know if attempting to call the guy twice (even with different args) would have some weird side effects

pronebird: That looks sane. We need to punch the hole for socket before the call to connect() to make sure that it succeeds. On android that prevents network loop and on unix desktops it ensures that the right network interface is used when establishing the connection.

pronebird (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5494)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @jmwample, @jstuczyn, and @octol)

PR #5495, opened 2025-02-20
Fix stats bug & remove HM caching

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5495)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/http/state.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/state.rs
@@ -57,6 +57,8 @@ static MIXSTATS_LIST_KEY: &str = "mixstats";
 static SUMMARY_HISTORY_LIST_KEY: &str = "summary-history";
 static SESSION_STATS_LIST_KEY: &str = "session-stats";
 
+const MIXNODE_STATS_HISTORY_DAYS: usize = 30;
+
 #[derive(Debug, Clone)]
 pub(crate) struct HttpCache {
     gateways: Cache<String, Arc<RwLock<Vec<Gateway>>>>,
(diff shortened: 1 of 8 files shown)


nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -285,24 +200,27 @@ impl HttpCache {
...
                 let read_lock = guard.read().await;
-                read_lock.to_vec()
+                let mut stats = read_lock.to_vec();
+
+                stats.truncate(MIXNODE_STATS_HISTORY_DAYS + offset);
+                stats.into_iter().skip(offset).collect()
             }
             None => {
-                let new_node_stats = crate::db::queries::get_daily_stats(db, offset)
+                let mut new_node_stats = crate::db::queries::get_daily_stats(db, 0)
```
jstuczyn: just to confirm, you meant to use 0 offset here?

nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -285,24 +200,27 @@ impl HttpCache {
...
-                self.upsert_mixnode_stats(mixnode_stats.clone()).await;
-                mixnode_stats
+                    .unwrap_or_default()
+                    .into_iter()
+                    .rev()
+                    .collect::<Vec<_>>();
+                // cache result without offset
+                self.upsert_mixnode_stats(new_node_stats.clone()).await;
+
+                new_node_stats.truncate(MIXNODE_STATS_HISTORY_DAYS + offset);
```
jstuczyn: why are you truncating it here rather than just getting correct number from db in the first place?

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/http/state.rs
dynco-nym: because of how caching works: imagine I truncated result directly on DB side

then truncated value is cached

next request comes in with a different offset: how do I give them data they're requesting if my cache has incomplete data?

nym-node-status-api/nym-node-status-api/src/http/state.rs
dynco-nym: yes, but I'll remove it since truncating happens on retrieved data

PR #5496, opened 2025-02-20
feature: allow nym-nodes to understand future version of sphinx packets

jstuczyn:
this PR updates the used version of the `sphinx-packet` dependency and allows our nym-nodes to understand the current version of sphinx packets as well as the future ones including fixed crypto. 

it also includes additional prometheus metric gauge to count the number of packets with given version.

## Testing remarks
you need to check the following:
- whether prometheus metrics are correctly updated based on packets of particular versions
- being able to send and receive packets generated from client from an older version (forward compatibility)
- being to send and receive packets generated by the client from this branch by a nym-node from old version (backwards compatibility)
- being able to send and receive packets by the client from this branch **BUT WITH DISABLED LEGACY MODE** (future compatibility). to enable it you have to remove `with_version(...)` argument in the packet builder here: https://github.com/nymtech/nym/blob/f3022b44fbe4e6acd15a10d26ddb145bd9066680/common/nymsphinx/types/src/lib.rs#L95

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5496)


simonwicky (approved):
Review limited to Outfox changes

benedettadavico:
Testing Completion Date: [3/03/2025]

Testing Steps Performed:
Baseline check to confirm existing behaviour still works.
Tested forward compatibility, making sure old clients work with the new node.
Tested backwards compatibility, so the new client works with old nodes.
Ensured everything runs smoothly when both sides are upgraded, still supporting old packet formats.
Tested future compatibility of sphinx v2.

Validation:
[x] Test Passed
[ ] Test Failed

Notes:
The next release will be the most important one to test/see a surge in sphinx_v2 packets

Attachments:
n/a

PR #5514, opened 2025-02-24
Fix total_stake on SQL update

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5514)


pwnfoo (approved):
LGTM! Nice catch

PR #5521, opened 2025-02-26
Fix/pull from harbor

farbanas:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5521)


Bykow (approved):
As mentioned, careful is running the action OUTSIDE of our infra, as these pulls require being in corpo vpn range.

PR #5523, opened 2025-02-26
Shared instance for DNS AsyncResolver

jmwample:
Make the `HickoryDnsResolver` use a shared instance by default to limit fd use. Now the multiple `nym_http_api_client::Client`s that get built should take advantage of a shared connection pool for DNS lookups.

If for some reason a client does need an independent AsyncResolver this can still be done with the added `thread_resolver`  function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5523)


jstuczyn (approved):
before merging, make sure to retarget and rebase it against `release/2025.4-dorina`

PR #5526, opened 2025-02-27
feature: disallow routing mix packets to nodes not present in the topology

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5526)


nym-node/src/config/gateway_tasks.rs
```diff
@@ -56,12 +60,14 @@ pub struct Debug {
 
 impl Debug {
     const DEFAULT_MESSAGE_RETRIEVAL_LIMIT: i64 = 100;
+    pub const DEFAULT_MINIMUM_MIX_PERFORMANCE: u8 = 60;
```
octol: Where does 60 come from?

nym-node/src/node/mixnet/packet_forwarding/global.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
```
octol: Not GPL?

octol (approved):

benedettadavico:
Testing Completion Date: [17/03/2025]

Testing Steps Performed:
Verified that mixnodes only route packets to nodes present in the active topology. 
No stability or performance issues were observed.

Validation:
[x] Test Passed
[] Test Failed

Notes:
When the NM runs, the logs will show an increase in received [mix] even if the node is not in the active topology

Attachments:
n/a

PR #5536, opened 2025-02-28
Clean stale partially received buffers

octol:
Clean up old `ReconstructionBuffer`s that have not received fragments in a long time. The motivation is that we want to allow the client to cap the number of retransmissions done, meaning we can't assume that these will eventually arrive

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5536)

### DIFF
--- a/common/nymsphinx/chunking/src/reconstruction.rs
+++ b/common/nymsphinx/chunking/src/reconstruction.rs
@@ -33,6 +33,7 @@ struct ReconstructionBuffer {
     /// everything in order the whole time, allowing for O(1) insertions and O(n) reconstruction.
     fragments: Vec<Option<Fragment>>,
 
+    /// The timestamp of the last received fragment. Used for cleaning up stale buffers.
     last_fragment_timestamp: Instant,
 }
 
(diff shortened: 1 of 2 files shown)


common/nymsphinx/chunking/src/reconstruction.rs
```diff
@@ -296,6 +307,25 @@ impl MessageReconstructor {
     pub fn recover_fragment(&self, fragment_data: Vec<u8>) -> Result<Fragment, ChunkingError> {
         Fragment::try_from_bytes(&fragment_data)
     }
+
+    pub fn cleanup_stale_buffers(&mut self) {
+        trace!("Cleaning up stale buffers");
+        let now = Instant::now();
+        self.reconstructed_sets.retain(|_, set_buf| {
```
jstuczyn, line 314: I wonder, and I know I'm contradicting myself, but should we be calling `cleanup_stale_buffers` on **every** fragment? reason being is that cost of `.retain` is non-negligible as it has to iterate through whole thing every time. though maybe if we don't expect a lot of entries there it's fine. 

perhaps we could keep internal counter of fragments received and call `cleanup_stale_buffers` if either `received % 1000 == 0 `( or whatever constant) or `last_cleaned_up < threshold`?

octol: I think that makes sense, let me fix that

jstuczyn (approved):

PR #5547, opened 2025-03-03
Handle disconnect in IPR

octol:
Handle disconnect in the IPR, removing clients that requests to

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5547)

### DIFF
--- a/service-providers/ip-packet-router/src/clients/connected_clients.rs
+++ b/service-providers/ip-packet-router/src/clients/connected_clients.rs
@@ -59,6 +59,24 @@ impl ConnectedClients {
             .any(|client| client.client_id == *client_id)
     }
 
+    pub(crate) fn disconnect_client(&mut self, client_id: &ConnectedClientId) {
+        if let Some(ips) = self.lookup_ip_from_client_id(client_id) {
+            log::info!("Disconnect client that requested to do so: {ips}");
+            self.disconnect_client_handle(ips);
+        }
+    }
+
+    fn disconnect_client_handle(&mut self, ips: IpPair) {
+        self.clients_ipv4_mapping.remove(&ips.ipv4);
+        self.clients_ipv6_mapping.remove(&ips.ipv6);
+        self.tun_listener_connected_client_tx
+            .send(ConnectedClientEvent::Disconnect(DisconnectEvent(ips)))
+            .inspect_err(|err| {
+                log::error!("Failed to send disconnect event: {err}");
+            })
+            .ok();
+    }
+
     pub(crate) fn lookup_ip_from_client_id(&self, client_id: &ConnectedClientId) -> Option<IpPair> {
         self.clients_ipv4_mapping
             .iter()
@@ -164,14 +182,7 @@ impl ConnectedClients {
     ) {
         for (ips, _) in &stopped_clients {
             log::info!("Disconnect stopped client: {ips}");
-            self.clients_ipv4_mapping.remove(&ips.ipv4);
-            self.clients_ipv6_mapping.remove(&ips.ipv6);
-            self.tun_listener_connected_client_tx
-                .send(ConnectedClientEvent::Disconnect(DisconnectEvent(*ips)))
-                .inspect_err(|err| {
-                    log::error!("Failed to send disconnect event: {err}");
-                })
-                .ok();
+            self.disconnect_client_handle(*ips);
         }
     }
 
@@ -181,14 +192,7 @@ impl ConnectedClients {
     ) {
         for (ips, _) in &inactive_clients {
             log::info!("Disconnect inactive client: {ips}");
-            self.clients_ipv4_mapping.remove(&ips.ipv4);
-            self.clients_ipv6_mapping.remove(&ips.ipv6);
-            self.tun_listener_connected_client_tx
-                .send(ConnectedClientEvent::Disconnect(DisconnectEvent(*ips)))
-                .inspect_err(|err| {
-                    log::error!("Failed to send disconnect event: {err}");
-                })
-                .ok();
+            self.disconnect_client_handle(*ips);
         }
     }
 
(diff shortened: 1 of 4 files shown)


neacsu (commented):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5547)* status: all files reviewed, 2 unresolved discussions (waiting on @jstuczyn)

___
*[`service-providers/ip-packet-router/src/clients/connected_clients.rs` line 64 at r1](https://reviewable.io/reviews/nymtech/nym/5547#-OK_V1Vc7FcGzrK42ROM:-OK_V1Vc7FcGzrK42RON:b-byfgiy) ([raw file](https://github.com/nymtech/nym/blob/ef3edc056593bce0e77e3959966b9d65090f81ad/service-providers/ip-packet-router/src/clients/connected_clients.rs#L64)):*
> ```XML
>     pub(crate) fn disconnect_client(&mut self, client_id: &ConnectedClientId) {
>         if let Some(ips) = self.lookup_ip_from_client_id(client_id) {
>             log::info!("Disconnect client that requested to do so: {ips}");
> ```

This looks like it could be to verbose, if every IPR client will send disconnect.
___
*[`service-providers/ip-packet-router/src/mixnet_listener.rs` line 317 at r1](https://reviewable.io/reviews/nymtech/nym/5547#-OK_UThTEdPx5YiNVPF6:-OK_UThTEdPx5YiNVPF7:b-xg9wh3) ([raw file](https://github.com/nymtech/nym/blob/ef3edc056593bce0e77e3959966b9d65090f81ad/service-providers/ip-packet-router/src/mixnet_listener.rs#L317)):*
> ```XML
>                     request_id,
>                     reply: DisconnectResponse::Failure(
>                         crate::messages::response::DisconnectFailureReason::ClientNotConnected,
> ```

Maybe this is already how we do it on the other disconnects as well, but isn't this some attack vector, in which you can find out if a client is connected or not? Or are there additional measures that prevent this, on a different layer?

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -292,9 +292,46 @@ impl MixnetListener {
         }))
     }
 
-    fn on_disconnect_request(&self, _disconnect_request: DisconnectRequest) -> PacketHandleResult {
-        log::info!("Received disconnect request: not implemented, dropping");
-        Ok(None)
+    async fn on_disconnect_request(
```
jstuczyn: but this doesn't need to by async from what I can see?

service-providers/ip-packet-router/src/clients/connected_clients.rs
```diff
@@ -59,23 +59,42 @@ impl ConnectedClients {
...
-    //}
-    //
-    //fn lookup_client(&self, client_id: &ConnectedClientId) -> Option<&ConnectedClient> {
-    //    self.clients_ipv4_mapping
-    //        .values()
-    //        .find(|connected_client| connected_client.client_id == *client_id)
-    //}
+    pub(crate) fn disconnect_client(&mut self, client_id: &ConnectedClientId) {
+        if let Some(ips) = self.lookup_ip_from_client_id(client_id) {
+            log::info!("Disconnect client that requested to do so: {ips}");
```
jstuczyn: I agree with Bogdan, it could be quite verbose here

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -292,9 +292,46 @@ impl MixnetListener {
...
+        &mut self,
+        disconnect_request: DisconnectRequest,
+    ) -> PacketHandleResult {
+        log::info!(
```
jstuczyn, line 300: sanity check, for a client with known address you have to pass some signature/authentication check to get here, right?

jstuczyn (approved):

service-providers/ip-packet-router/src/mixnet_listener.rs
octol: disconnect requests should be signed yes, but let me double check that

octol: Yes disconnect requests are signed when it's an explicit reply_to. For surbs the tag is secret

service-providers/ip-packet-router/src/mixnet_listener.rs
octol: fixed

service-providers/ip-packet-router/src/clients/connected_clients.rs
octol: fixed

service-providers/ip-packet-router/src/mixnet_listener.rs
octol: done

service-providers/ip-packet-router/src/clients/connected_clients.rs
octol: fixed

octol (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5547)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @jstuczyn and @neacsu)

___
*[`service-providers/ip-packet-router/src/mixnet_listener.rs` line 317 at r1](https://reviewable.io/reviews/nymtech/nym/5547#-OK_UThTEdPx5YiNVPF6:-OKaPhpt2dz5mLnQYOjS:b-1j26f0) ([raw file](https://github.com/nymtech/nym/blob/ef3edc056593bce0e77e3959966b9d65090f81ad/service-providers/ip-packet-router/src/mixnet_listener.rs#L317)):*

Maybe this is already how we do it on the other disconnects as well, but isn't this some attack vector, in which you can find out if a client is connected or not? Or are there additional measures that prevent this, on a different layer?
</blockquote></details>

For v4 the disconnect request is signed by the client identity key. For v5 it's not signed, since sender tags are deemed secret

PR #5548, opened 2025-03-03
Explorer V2

fmtabbara:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5548)


explorer-v2/src/app/i18n/locales/en/alpha.json
mmsinclair, line 1: Are all of these translations needed? It seems to be all of them from the website?

mmsinclair (changes requested):
@yanok87 @fmtabbara it would be much easier to know what is a server-rendered component if all files had either `"use server"` or `"use client"` in them

I'm thinking it might be best to enforce this in `*.tsx` and `*.jsx` files

mmsinclair (approved):

PR #5569, opened 2025-03-05
Add /v3/nym-nodes

dynco-nym:
- returns extended node info from local DB
- endpoint caching
- add bond_info & self_described to DB nym_nodes
- update mixnode & gateway bond status on data refresh
- add `active` column to DB nym_nodes
- use only active & bonded nodes in scraping/testrun tasks

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5569)


nym-node-status-api/nym-node-status-agent/run.sh
```diff
@@ -3,7 +3,7 @@
 set -eu
 export ENVIRONMENT=${ENVIRONMENT:-"mainnet"}
 
-probe_git_ref="nym-vpn-core-v1.3.2"
+probe_git_ref="nym-vpn-core-v1.4.0"
```
jstuczyn, line 6: sanity check, since it's been a while since you created this PR (sorry!) - is this ref still valid?

nym-node-status-api/nym-node-status-api/migrations/005_node_self_described.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE nym_nodes ADD COLUMN self_described TEXT;
+ALTER TABLE nym_nodes ADD COLUMN bond_info TEXT;
```
jstuczyn, line 2: nit: Mark recently showed me that you can actually use proper json in sqlite. assuming it works and is supported by sqlx - would it be more fitting here?

nym-node-status-api/nym-node-status-api/migrations/005_node_self_described.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE nym_nodes ADD COLUMN active INTEGER CHECK (active in (0, 1)) NOT NULL DEFAULT 0;
```
jstuczyn: does it make sense to store active status in the db? it will be changing hourly and is not really that "persistent"

nym-node-status-api/nym-node-status-api/src/db/models.rs
```diff
@@ -433,13 +450,34 @@ pub(crate) struct NymNodeInsertRecord {
...
+        bond_info: Option<&NymNodeDetails>,
+        self_described: Option<&NymNodeDescription>,
+    ) -> anyhow::Result<Self> {
         let now = OffsetDateTime::now_utc().to_string();
 
+        // if a node doesn't expose its self-described endpoint, it can't route traffic
+        // https://nym.com/docs/operators/nodes/nym-node/bonding
+        // same if it's not bonded in the mixnet smart contract
+        // https://nym.com/docs/operators/tokenomics/mixnet-rewards#rewarded-set-selection
+        let active = self_described.is_some() && bond_info.is_some();
```
jstuczyn: just so we're clear, this `active` is not the same as "active" we've been using everywhere else. 
what we refer to as "active node" means that node that is in the active part of the rewarded set in the mixmining contract

nym-node-status-api/nym-node-status-api/src/db/queries/scraper.rs
```diff
@@ -7,45 +7,56 @@ use crate::{
 };
 use anyhow::Result;
 use chrono::Utc;
+use nym_validator_client::nym_api::SkimmedNode;
 
 pub(crate) async fn get_nodes_for_scraping(pool: &DbPool) -> Result<Vec<ScraperNodeInfo>> {
```
jstuczyn, line 12: remind me, where is this being used for? because "SkimmedNodes" are already pre-scraped (if by scraping you mean querying their self-described API)

nym-node-status-api/nym-node-status-api/src/db/queries/scraper.rs
```diff
@@ -7,45 +7,56 @@ use crate::{
...
-                    .map(|ip| ip.to_string())
-                    .collect::<Vec<_>>(),
-                http_api_port: node.mix_port.into(),
-            })
-        });
+            }
+        })
+    })?;
+
+    skimmed_nodes.for_each(|node| {
```
jstuczyn, line 32: didn't we want to remove legacy nodes in this iteration? (or was it the next one)

nym-node-status-api/nym-node-status-api/src/http/models.rs
```diff
@@ -45,6 +47,23 @@ pub struct Mixnode {
...
+#[derive(Clone, Debug, utoipa::ToSchema, Deserialize, Serialize)]
+pub(crate) struct ExtendedNymNode {
+    pub(crate) node_id: NodeId,
+    pub(crate) identity_key: String,
+    pub(crate) uptime: f64,
+    #[schema(value_type = String)]
+    pub(crate) total_stake: Decimal,
+    pub(crate) original_pledge: u128,
+    pub(crate) bonding_address: String,
+    pub(crate) bonded: bool,
```
jstuczyn, line 59: under what conditions would you have node details for node that's not bonded?

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
```diff
@@ -308,17 +326,45 @@ impl Monitor {
         }
     }
 
+    fn prepare_nym_node_data(
+        &self,
+        skimmed_nodes: Vec<SkimmedNode>,
+        bonded_node_info: &HashMap<NodeId, NymNodeDetails>,
+        described_nodes: &HashMap<NodeId, NymNodeDescription>,
```
jstuczyn, line 337: this looks like a lot of data duplication because skimmednode is derived out of bond info and description

nym-node-status-api/nym-node-status-api/src/db/queries/scraper.rs
dynco-nym: it's used for 
- collecting packets mixed from `/stats`
- scraping `/v1/api/description`

nym-node-status-api/nym-node-status-agent/run.sh
dynco-nym: will update before merge, but this is just for local development anyway

nym-node-status-api/nym-node-status-api/migrations/005_node_self_described.sql
dynco-nym: All data is persistent in the DB. Other stats are also refreshed hourly so they're still in DB like routing score, testruns etc.

nym-node-status-api/nym-node-status-api/src/db/models.rs
dynco-nym: we collect stats by quering nodes for packets mixed every hour or so

if I query only nodes mixing in this hour, I might miss a node that was mixing packets in previous hour, but is not mixing currently

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
dynco-nym: this description isn't the same as description returned by `get_all_described_nodes()`

```
pub struct NodeDescriptionResponse {
    pub moniker: Option<String>,
    pub website: Option<String>,
    pub security_contact: Option<String>,
    pub details: Option<String>,
}
```

nym-node-status-api/nym-node-status-api/src/http/models.rs
dynco-nym: details from when it was bonded? what are you suggesting?

nym-node-status-api/nym-node-status-api/src/db/queries/scraper.rs
dynco-nym: leaving for now

nym-node-status-api/nym-node-status-api/src/db/models.rs
jstuczyn: so could you just use a different name for it? because "active" has a different meaning in the context of nym nodes

nym-node-status-api/nym-node-status-api/src/http/models.rs
jstuczyn: that most of those fields make absolutely no sense if node is no longer bonded, like pledge, stake, rewarding details, uptime. etc. so I wonder if we should just purge if it it's not bonded (because after it rebonds, it will have a brand new node_id anyway)

nym-node-status-api/nym-node-status-api/src/monitor/mod.rs
jstuczyn: gotcha, my bad!

nym-node-status-api/nym-node-status-api/src/db/models.rs
dynco-nym: removed that field, I'll filter based on `self_described` / `bonded`  individually

nym-node-status-api/nym-node-status-api/src/http/models.rs
dynco-nym: leaving `bonded` field because it's required on the API, but I changed insert logic to delete  unbonded nodes from the DB

nym-node-status-api/nym-node-status-api/migrations/005_node_self_described.sql
dynco-nym: I'm aware of [this](https://docs.rs/sqlx/latest/sqlx/types/struct.Json.html), which requires `sqlx::types::Json` wrappers in Rust code, but it doesn't offer any utility over `serde` (which is also supported by `sqlx`)

`sqlx::types::Json`:
```rust 
    books: sqlx::types::Json<Book>
```

serde:
```
    self_described as "self_described: serde_json::Value",
    bond_info as "bond_info: serde_json::Value"
```

jstuczyn (approved):

PR #5577, opened 2025-03-06
Server Side internal DoT/DoH opt out

jmwample:
Allow server side (nodes and nym-api) to use nym-http-api-client without  DoH/DoT resolver.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5577)

### DIFF
--- a/common/client-libs/validator-client/src/coconut/mod.rs
+++ b/common/client-libs/validator-client/src/coconut/mod.rs
@@ -8,7 +8,6 @@ use nym_coconut_dkg_common::types::{EpochId, NodeIndex};
 use nym_coconut_dkg_common::verification_key::ContractVKShare;
 use nym_compact_ecash::error::CompactEcashError;
 use nym_compact_ecash::{Base58, VerificationKeyAuth};
-use nym_http_api_client::HttpClientError;
 use std::fmt::{Display, Formatter};
 use thiserror::Error;
 use url::Url;
@@ -72,12 +71,6 @@ pub enum EcashApiError {
         #[from]
         source: crate::ValidatorClientError,
     },
-
-    #[error("error building or using HTTP client: {source}")]
-    ClientError {
-        #[from]
-        source: HttpClientError,
-    },
 }
 
 impl TryFrom<ContractVKShare> for EcashApiClient {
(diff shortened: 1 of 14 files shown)


common/client-libs/validator-client/src/coconut/mod.rs
```diff
@@ -83,8 +90,19 @@ impl TryFrom<ContractVKShare> for EcashApiClient {
 
         let url_address = Url::parse(&share.announce_address)?;
 
+        // Note: we expect this to be used server-side (gateways / nodes / APIs) not client-side
```
jstuczyn: that one is tricky because clients will use it when attempting to get credentials locally, independently, but it kinda blurs the line of whether this dns is really needed in that instance and whether it really applies to the vpn

common/http-api-client/src/lib.rs
```diff
@@ -233,22 +235,36 @@ impl ClientBuilder {
...
             // TODO: or should we maybe default to https?
             Self::new(alt)
+        } else {
+            Ok(Self::new_with_url(url.into_url()?))
+        }
+    }
+
+    /// Constructs a new http `ClientBuilder` from a valid url.
+    pub fn new_with_url(url: Url) -> Self {
+        if !url.scheme().starts_with("http") {
```
jstuczyn, line 259: I'm trying to understand the logic here, if the url does not start with `http://` or `https://` you want to force `https` on it? how can we ensure it actually supports ssl?

jstuczyn (commented):
also forgot to mention, you probably also want to look into node status api... sorry...

common/http-api-client/src/lib.rs
jmwample: This was an attempt to reproduce the logic from the url parsing in the `&str` based constructor but you are correct that it doesn't make much sense. I have changed it now to just warn if the scheme for this api client url is not http/https. No error or panic for handling, but if someone misuses this and then wonders why things are broken the logs will have a hint.

jmwample (commented):
I have added changes for `nym-node-status-api`

common/client-libs/validator-client/src/coconut/mod.rs
jmwample: Maybe the simplest solution for now is to revert this one. We have covered several longer lived clients with the other changes and I can't think of a clean solution for disentangling client vs non-client usage of this code right now.

jstuczyn (approved):
just add that comment for ecash client and we're good to go

PR #5584, opened 2025-03-08
Always purge sender tags on startup

octol:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5584)


octol:
Changing to draft since we don't want to merge this

PR #5617, opened 2025-03-13
Add workflow

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5617)

### DIFF
--- a/.github/workflows/ci-check-ns-api-version.yml
+++ b/.github/workflows/ci-check-ns-api-version.yml
@@ -1,0 +1,46 @@
+name: ci-check-ns-api-version
+
+on:
+  pull_request:
+    paths:
+      - "nym-node-status-api/**"
+
+      # TODO dz parse Cargo.toml version
+      # TODO dz check if there is a harbor image with that tag already
+      # TODO dz if there is, return non-zero
+
+env:
+  WORKING_DIRECTORY: "nym-node-status-api/nym-node-status-api"
+
+jobs:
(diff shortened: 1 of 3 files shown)


farbanas (approved):

.github/workflows/ci-check-ns-api-version.yml
```diff
@@ -0,0 +1,57 @@
+name: ci-check-ns-api-version
```
octol, line 1: I'm happy to see that this naming convention is used :)

.github/workflows/ci-check-ns-api-version.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: arc-ubuntu-22.04-dind
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Get version from cargo.toml
+        uses: mikefarah/yq@v4.45.1
+        id: get_version
+        with:
+          cmd: yq -oy '.package.version' ${{ env.WORKING_DIRECTORY }}/Cargo.toml
```
octol, line 22: You can use `cargo metadata` for stuff like this as well I think

octol (approved):

.github/workflows/ci-check-ns-api-version.yml
dynco-nym: currently runner image doesn't have cargo

PR #5618, opened 2025-03-13
Lower bandwidth used for topology refresh

jmwample:
Prevents re-pulling node descriptors that are already known while using the `TopologyRefresher`

**What does this do**
- Adds an endpoint to the Nym API that allows pulling batches of node descriptors by `node_id`
- Creates a `TopologyProvider` that caches topologies and only goes to network if the known topology is older than the configured cache ttl.
- When refreshing the Topology we compare the node IDs in the cache to the node IDs in the refreshed layer assignments. 
  -  If the epoch has not changed no descriptors are pulled
  - if the epoch has changed layer assignments are pulled and updated in the cached topology
  - Only descriptors for node IDs in the new layer assignments that are NOT in the known topology cache are pulled. 
- Adds a `Performance` field to the stored `RoutingNode` in the Cached topology. This way the cache contains all node descriptors and queries simply filter based on the stored performance, as opposed to filtering before storing and downloading more descriptors.

**What does this not do**
- Each created `MixnetClient` still uses an independent `TopologyProvider` and cached topology. This PR does not create a shared topology provider that can be used for every new connection.
- The first refresh pulls the full topology. This PR does not pre-cache or add any topology serialization deserialization that would prevent the initial pull.
- This PR does not change the stage at which the initial topology is pulled. As is the `TopologyRefresher` is created as a `MixnetCient` is starting up. This means that the initial topology pull doesn't happen until we have already started the process of connecting to the mixnet.
- This does not effect the Topology information used by the VPN client. That is a different API with different data and happens in the VPN client repo.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5618)


common/client-core/src/client/topology_control/smart_api_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+    EpochRewardedSet, NymTopology, RoutingNode, TopologyProvider,
+};
+use nym_validator_client::UserAgent;
+use rand::{prelude::SliceRandom, thread_rng};
+use url::Url;
+
+/// Topology Provider build around a cached piecewise provider that uses the Nym API to
+/// fetch changes and node details.
+#[derive(Clone)]
+pub struct NymApiTopologyProvider {
```
octol, line 23: What about using something like `NymApiSmartTopologyProvider`? Having two types with the same name was a bit confusing initially when reading the the code

common/topology/src/lib.rs
```diff
@@ -546,4 +546,19 @@ impl NymTopology {
...
+    }
+
+    pub fn all_node_ids(&self) -> impl Iterator<Item = &NodeId> {
+        self.node_details.keys()
+    }
+
+    pub fn gateways(&self) -> impl Iterator<Item = &RoutingNode> {
+        self.node_details.values().filter(|n| {
+            self.rewarded_set.entry_gateways.contains(&n.node_id)
+                || self.rewarded_set.exit_gateways.contains(&n.node_id)
```
octol, line 561: Would be good to confirm with @jstuczyn that this is correct?

octol (approved):
Great work! Would be good to have @jstuczyn to quickly skim it as well though

common/topology/src/lib.rs
jstuczyn: no, it's not. but it's easy to miss because we've been going back and forth about this for quite a while. the current consensus is that you can use **any** node that supports gateway mode regardless if its in the active set or not

nym-api/src/network_monitor/monitor/preparer.rs
```diff
@@ -231,6 +231,8 @@ impl PacketPreparer {
                     mixnet_entry: false,
                     mixnet_exit: false,
                 },
+                // We have no information about performance in legacy node formats
+                performance: Percent::hundred(),
```
jstuczyn, line 235: at this point I guess we should probably use performance of zero if nodes are still outdated...

nym-api/src/nym_nodes/handlers/unstable/skimmed.rs
```diff
@@ -326,7 +327,47 @@ pub(super) async fn nodes_basic_all(
...
+    path = "batch",
+    context_path = "/v1/unstable/nym-nodes/skimmed",
+    responses(
+        (status = 200, body = PaginatedCachedNodesResponseSchema)
+    )
+)]
+pub(super) async fn nodes_basic_batch(
+    state: State<AppState>,
+    Query(query_params): Query<NodesParamsWithRole>,
+    Json(ids): Json<Vec<u32>>,
```
jstuczyn, line 359: there should be some `limit` on the number of nodes you can request (because _certain_ people might complain it could lead to DoS)

jstuczyn (commented):
few notes:
- > Only descriptors for node IDs in the new layer assignments that are NOT in the known topology cache are pulled.

sometimes you might have to pull information about nodes more often than that. regardless if they're known or not because they might have changed their internal configuration (like started using different mix port)

- you have to be pulling all gateways, not just the ones in the rewarded set, as clients are allowed to use any of them at any time

PR #5619, opened 2025-03-14
Feature/paginated ticketbooks challenge

jstuczyn:
this PR makes changes to how ticketbook rewarding works and in particular the challenge portion of the exchange. rather than requesting merkle proof alongside all binary data of the associated ticketbooks in a single `/issued-ticketbooks-challenge` query (that could have potentially lead to ddos of nym-apis as it made them load all of the data into the memory at once) the query got split into `/issued-ticketbooks-challenge-commitment` for obtaining the merkle proof and `/issued-ticketbooks-data` for obtaining the binary data. note that the latter query might have to be called multiple times with different arguments due to the enforced limit on the maximum number of values returned at once.

I've also added some helper queries to nym-api for getting more insight into pure numbers of issued ticketbooks.

Moreover, some of the logic regarding the rewarding has changed so some extensive testing would be needed before deploying it in the wild.

nym-api routes:

added:
- `/issued-ticketbooks-count`
- `/issued-ticketbooks-for-count/:expiration_date`
- `/issued-ticketbooks-on-count/:issuance_date`
- `/issued-ticketbooks-challenge-commitment`
- `/issued-ticketbooks-data`

removed:
- `/issued-ticketbooks-challenge`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5619)


benedettadavico:
Testing Completion Date: [28/03/2025]

Testing Steps Performed:
New issue ticketbook related endpoints work -  tickets are issued and consumed as before. 
Redeeming tickets is happening as normal on the gateway side.

Validation:
✅ Test Passed

Notes (if any):
Chuckles will not require the rewarder binary to be updated; as there will be additional changes to rewarding, this will be extensively tested in the future.

PR #5632, opened 2025-03-17
Rework IPR codec to extract out timer and implement AsyncWrite

octol:
The goal is to be able to handle backpressure in the vpn client. For that we need to extract out the timer from the codec since we want to move control over that to the mixnet processor. Once the timer is removed we can reformulate the mixnet sender as a `Sink` wrapped in a `FramedWrite`

- Extract out timer from IPR codec
- Implement AsyncWrite on mixnet client sender
- Add functions to wait for lanes to clear on `LaneQueueLenghts`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5632)

### DIFF
--- a/service-providers/ip-packet-router/src/clients/connected_client_handler.rs
+++ b/service-providers/ip-packet-router/src/clients/connected_client_handler.rs
@@ -109,12 +109,12 @@ impl ConnectedClientHandler {
         (forward_from_tun_tx, close_tx, handle)
     }
 
-    async fn handle_packet(&mut self, packet: Vec<u8>) -> Result<()> {
+    async fn handle_packet(&mut self, packet: IprPacket) -> Result<()> {
         self.activity_timeout.reset();
         self.payload_topup_interval.reset();
 
         self.mixnet_ip_packet_sink
-            .send(IprPacket::from(packet))
+            .send(packet)
             .await
             .map_err(|source| IpPacketRouterError::FailedToEncodeMixnetMessage { source })
     }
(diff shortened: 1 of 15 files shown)


service-providers/ip-packet-router/src/clients/connected_client_handler.rs
```diff
@@ -141,16 +130,19 @@ impl ConnectedClientHandler {
...
-                packets = self.encoder.buffer_timeout() => match packets {
-                    Some(packets) => {
-                        if let Err(err) = self.handle_buffer_timeout(packets).await {
-                            log::error!("client handler: failed to handle buffer timeout: {err}");
-                        }
-                    },
-                    None => log::trace!("no packets to send"),
+                _ = self.payload_topup_interval.tick() => {
+                    // Send an empty packet to trigger the buffer timeout
+                    if let Err(err) = self.handle_packet(Vec::new()).await {
```
jstuczyn: are we sure this won't result in client actually sending empty messages over the mixnet (and getting those warnings with "empty payload" on the other side?)

service-providers/ip-packet-router/src/clients/connected_client_handler.rs
```diff
@@ -141,16 +130,19 @@ impl ConnectedClientHandler {
...
+                        log::error!("client handler: failed to handle packet: {err}");
+                    }
                 },
                 packet = self.forward_from_tun_rx.recv() => match packet {
                     Some(packet) => {
+                        // I don't think this should ever happen, so log this so we are aware of it
+                        if packet.is_empty() {
+                            log::warn!("client handler: received empty packet");
```
jstuczyn: oh yeah, this warning : )

jstuczyn (approved):

service-providers/ip-packet-router/src/clients/connected_client_handler.rs
octol: Oops, that should be a `IprPacket::Flush` and not an empty packet

PR #5642, opened 2025-03-18
Add max_retransmissions flag on each message

octol:
Add an optional `max_retransmissions` field on `InputMessage`, to be able to selectively disable or turn down the number of retransmissions done for specific packets

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5642)


common/client-core/src/client/real_messages_control/message_handler.rs
```diff
@@ -367,12 +374,12 @@ where
     pub(crate) async fn try_send_reply_chunks(
         &mut self,
         target: AnonymousSenderTag,
-        fragments: Vec<(TransmissionLane, Fragment)>,
+        fragments: Vec<(TransmissionLane, (Fragment, MaxRetransmissions))>,
```
jstuczyn: nit: that tuple looks quite disgusting tbh. can we perhaps wrap it in some other structure or at least the second part of it(since it's used elsewhere)? idk `FragmentWithRetransmissions`. maybe it should even be 
```rust
struct Foo {
fragments: Vec<Fragment>,
max_retransmissions: MaxRetransmissions
}
```

because realistically you'd want all fragments of your underlying message to have the same retransmitting behaviour

common/client-core/src/client/real_messages_control/message_handler.rs
```diff
@@ -495,11 +517,15 @@ where
             reply_surbs,
         ));
 
+        // When sending SURBs we want to retransmit
+        let max_retransmissions = None;
```
jstuczyn, line 530: and to confirm, this also applies to surb requests, et al.?

octol: yes correct, that's the intention. Everything surb related should stick to the normal retransmit behaviour

common/client-core/src/client/real_messages_control/message_handler.rs
octol: Let me sort that out

octol: Done!

jstuczyn (approved):

PR #5652, opened 2025-03-20
Update wallet to include Interval Operator Cost and Profit Margin

tommyv1987:
- Created a new modal for Interval Operator Cost and Profit Margin changes
- Uses existing update node operating cost functionality for old mixnodes (If it works it works)
- Prompts a warning that a user "Can send the request as many times as they want but it will only update the last message when the interval changes"

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5652)


octol:
There is a CI fail for ci-lint-typescript.yml due to missing RUSTUP_PERMIT_COPY_RENAME in the workflow that I just fixed, rebase on develop

jstuczyn (approved):

PR #5660, opened 2025-03-25
Remove Google public DNS

pronebird:
We had a lot of complaints about Google, so we're removing this in VPN client, nym should follow to avoid issues with firewall.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5660)


jmwample (approved):
lgtm :+1:

lets double check whether this PR needs targeted against dev or against the latest release branch.

jmwample:
Associated PR in the VPN client for tunnel trafic for reference:  https://github.com/nymtech/nym-vpn-client/pull/2558

PR #5664, opened 2025-03-28
Minor fixes involving key cloning and hashing

jmwample:
Adds fixes for several minor key related issues:
- copying a private keys to perform a DH
  - perform DH against the reference to the static key to prevent memory copy of secret key material
- hashing an x25519 public key without first converting to a canonical representation
  - use derived `Hash` implementation which internally [converts to a canonical representation](https://github.com/dalek-cryptography/curve25519-dalek/blob/fbf1fb5339b22eaf9925e520c90f1821f79ef5db/curve25519-dalek/src/montgomery.rs#L103) before hashing

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5664)

### DIFF
--- a/common/authenticator-requests/src/v5/registration.rs
+++ b/common/authenticator-requests/src/v5/registration.rs
@@ -168,13 +168,12 @@ impl GatewayClient {
     // Client should perform this step when generating its payload, using its own WG PK
     #[cfg(feature = "verify")]
     pub fn verify(&self, gateway_key: &PrivateKey, nonce: u64) -> Result<(), Error> {
-        let client_pubkey = PublicKey::from(*self.pub_key.deref());
-
-        let dh = gateway_key.diffie_hellman(&client_pubkey);
+        // use gateways key as a ref to an x25519_dalek key
+        let dh = (gateway_key.as_ref()).diffie_hellman(&self.pub_key);
 
         // TODO: change that to use our nym_crypto::hmac module instead
         #[allow(clippy::expect_used)]
-        let mut mac = HmacSha256::new_from_slice(&dh[..])
+        let mut mac = HmacSha256::new_from_slice(dh.as_bytes())
             .expect("x25519 shared secret is always 32 bytes long");
 
         mac.update(self.pub_key.as_bytes());
--- a/common/authenticator-requests/src/v5/registration.rs
+++ b/common/authenticator-requests/src/v5/registration.rs
@@ -15,7 +15,7 @@ use std::{fmt, ops::Deref, str::FromStr};
 #[cfg(feature = "verify")]
 use hmac::{Hmac, Mac};
 #[cfg(feature = "verify")]
-use nym_crypto::asymmetric::encryption::PrivateKey;
+use nym_crypto::asymmetric::encryption::{PrivateKey, PublicKey};
 #[cfg(feature = "verify")]
 use sha2::Sha256;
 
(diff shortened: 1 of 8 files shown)


common/authenticator-requests/src/v5/registration.rs
```diff
@@ -143,24 +143,22 @@ impl GatewayClient {
         private_ips: IpPair,
         nonce: u64,
     ) -> Self {
-        // convert from 1.0 x25519-dalek private key into 2.0 x25519-dalek
-        #[allow(clippy::expect_used)]
-        let static_secret = x25519_dalek::StaticSecret::from(local_secret.to_bytes());
-        let local_public: x25519_dalek::PublicKey = (&static_secret).into();
+        let local_public = PublicKey::from(local_secret);
```
jstuczyn, line 146: while you're there, mind fixing that conversion everywhere? we're using `dalek` 2.0 everywhere now so it's quite reduntant

common/authenticator-requests/src/v5/registration.rs
```diff
@@ -143,24 +143,22 @@ impl GatewayClient {
...
+        let mut mac = HmacSha256::new_from_slice(&dh[..])
             .expect("x25519 shared secret is always 32 bytes long");
 
         mac.update(local_public.as_bytes());
         mac.update(private_ips.to_string().as_bytes());
         mac.update(&nonce.to_le_bytes());
 
         GatewayClient {
-            pub_key: PeerPublicKey::new(local_public),
+            pub_key: PeerPublicKey::new(local_public.into()),
```
jstuczyn, line 161: I think that `into()` might be reduntant? though if clippy is happy, I'm happy

common/authenticator-requests/src/v5/registration.rs
```diff
@@ -170,15 +168,13 @@ impl GatewayClient {
     // Client should perform this step when generating its payload, using its own WG PK
     #[cfg(feature = "verify")]
     pub fn verify(&self, gateway_key: &PrivateKey, nonce: u64) -> Result<(), Error> {
-        // convert from 1.0 x25519-dalek private key into 2.0 x25519-dalek
-        #[allow(clippy::expect_used)]
-        let static_secret = x25519_dalek::StaticSecret::from(gateway_key.to_bytes());
+        let client_pubkey = PublicKey::from(*self.pub_key.deref());
 
-        let dh = static_secret.diffie_hellman(&self.pub_key);
+        let dh = gateway_key.diffie_hellman(&client_pubkey);
```
jstuczyn:
```suggestion
        let dh = gateway_key.diffie_hellman(&self.pub_key)
```

common/wireguard-types/src/public_key.rs
```diff
@@ -6,13 +6,12 @@ use base64::engine::general_purpose;
...
 use serde::Serialize;
 use std::fmt;
-use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 use std::str::FromStr;
 
 use x25519_dalek::PublicKey;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
```
jstuczyn, line 14: good catch!

jstuczyn (approved):
I'll pre-approve it so it wouldn't block you later as my comments are relatively minor, but please do take a look at them before merging : )

common/authenticator-requests/src/v5/registration.rs
jmwample: `gateway_key` is of type `nym_crypto::asymmetric::encryption::PrivateKey` and the self.pub_key istype `nym_wireguard_types::PeerPublicKey`. Those two don't have a direct way to do a dh against each other despite both wrapping dalek keys. So the option is converting or using a reference. Or adding more interop code to those types.

```
// convert public key to appropriate type
let client_pubkey = PublicKey::from(*self.pub_key.deref());
let dh = gateway_key.diffie_hellman(&client_pubkey);


// use gateways key as a ref to an x25519_dalek key
let dh = (gateway_key.as_ref()).diffie_hellman(&self.pub_key)

```

So I can change to using the second option there, but i think they are effectively pretty similar.

common/authenticator-requests/src/v5/registration.rs
jmwample: The `into()` is not redundant, again this is the difference between `nym_crypto::asymmetric::encryption::PublicKey` and  `nym_wireguard_types::PeerPublicKey` which don't have direct conversion, though they both wrap `x25519_dalek` things. so the conversion here is `nym_crypto -> x25519_dalek -> nym_wireguard_types` .

jmwample:
Added an equivalent change for `common/crypto/src/asymmetric/identity/` to use the derived `Hash` implementation. This is an edwards point ( as compared to a MontgomeryPoint for x25519 pubkeys) so  I don't believe that there are multiple representations for the key. However,  there isn't a reason to keep our own matching `Hash` impl ([default impl here](https://github.com/dalek-cryptography/curve25519-dalek/blob/fbf1fb5339b22eaf9925e520c90f1821f79ef5db/ed25519-dalek/src/verifying.rs#L78)).

PR #5670, opened 2025-04-01
Add contains ticketbook data db query

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5670)

### DIFF
--- a/common/credential-storage/src/backends/sqlite.rs
+++ b/common/credential-storage/src/backends/sqlite.rs
@@ -95,6 +95,20 @@ impl SqliteEcashTicketbookManager {
         Ok(())
     }
 
+    pub(crate) async fn contains_ticketbook_data(&self, data: &[u8]) -> Result<bool, sqlx::Error> {
+        let exists = sqlx::query(
+            r#"
+                SELECT * FROM ecash_ticketbook WHERE ticketbook_data = ?
+            "#,
+        )
+        .bind(data)
+        .fetch_optional(&self.connection_pool)
+        .await?
+        .is_some();
+
+        Ok(exists)
+    }
+
     pub(crate) async fn get_ticketbooks_info(
         &self,
     ) -> Result<Vec<BasicTicketbookInformation>, sqlx::Error> {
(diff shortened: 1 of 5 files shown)


jstuczyn:
what's the purpose of this query? because checking for binary data is quite bad for checking for reuse due to the coconu's magic "randomisation" property

neacsu:
> what's the purpose of this query? because checking for binary data is quite bad for checking for reuse due to the coconu's magic "randomisation" property

Due to intermediate errors, we might end up into a state where we import a credential without marking it as done on the vpn-api side. As a temporary fix we check for that error which is a false positive and just ignore it: https://github.com/nymtech/nym-vpn-client/pull/2596

It would be better to be able to check that we already imported it and move ahead from it https://github.com/nymtech/nym-vpn-client/pull/2598 . This is via a retry mechanism that the account controller uses to manage zk-nyms.

This shouldn't touch the randomisation property because vpn-api doesn't do any randomisation between two different requests from the client, it just tries to return the same credential again. On a second attempt (or third etc), we might eventually be able to mark it as done on server side as well. If the same re-randomised credential gets stored twice, that would be a logic error and we'd get hit with double spending, but imo the database structure should allow for it

common/credential-storage/src/backends/sqlite.rs
```diff
@@ -95,6 +95,20 @@ impl SqliteEcashTicketbookManager {
         Ok(())
     }
 
+    pub(crate) async fn contains_ticketbook_data(&self, data: &[u8]) -> Result<bool, sqlx::Error> {
+        let exists = sqlx::query(
+            r#"
+                SELECT * FROM ecash_ticketbook WHERE ticketbook_data = ?
```
jstuczyn: optimisation: rather than retrieving data, allocating it and then finally comparing it, why not do everything inside the db via `EXISTS` operator?

jstuczyn (approved):

common/credential-storage/src/backends/sqlite.rs
neacsu: Done

PR #5673, opened 2025-04-01
Add /account/{address}

dynco-nym:
- meant to replace `https://explorer.nymtech.net/api/v1/tmp/unstable/account/{address}` for Explorer v2

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5673)


nym-api/build.rs
```diff
@@ -1,13 +1,20 @@
 use sqlx::{Connection, FromRow, SqliteConnection};
-use std::env;
+use std::{env, fs::Permissions, os::unix::fs::PermissionsExt};
```
jstuczyn: this fails windows build

nym-api/build.rs
```diff
@@ -62,3 +69,26 @@ async fn main() {
...
+async fn write_db_path_to_file(out_dir: &str, db_filename: &str) -> anyhow::Result<()> {
+    let mut file = File::create("settings.sql").await?;
+    let settings = ".mode columns
+.headers on";
+    file.write_all(settings.as_bytes()).await?;
+
+    let mut file = File::create("enter_db.sh").await?;
+    let contents = format!(
+        "#!/bin/bash\n\
+        sqlite3 -init settings.sql {}/{}",
```
jstuczyn, line 90: can we guarantee `sqlite3` exists on every machine? why not rely on sqlx directly?

nym-api/build.rs
jstuczyn, line 1: why are we changing the build.rs to begin with?

nym-api/src/unstable_routes/models.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde::{Deserialize, Serialize};
+use utoipa::schema;
+
+#[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema, utoipa::ToResponse)]
+pub struct NyxAccountDelegationDetails {
+    pub node_id: NodeId,
+    #[schema(value_type = String)]
+    pub delegated: Coin,
+    pub height: u64,
+    #[schema(value_type = String)]
```
jstuczyn: `value_type` is not a string, because a `None` will get serialised to `null` rather than `""`. either set it to `Option<String>` or remove the field completely. alternatively change it to `boolean` to indicate whether it's a legacy vested delegation that somebody didn't bother to migrate (and that we should display a big ❗ next to)

nym-api/src/unstable_routes/models.rs
```diff
@@ -0,0 +1,57 @@
...
+use cosmwasm_std::{Addr, Coin};
+use nym_topology::NodeId;
+use serde::{Deserialize, Serialize};
+use utoipa::schema;
+
+#[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema, utoipa::ToResponse)]
+pub struct NyxAccountDelegationDetails {
+    pub node_id: NodeId,
+    #[schema(value_type = String)]
+    pub delegated: Coin,
```
jstuczyn, line 20: similarly `Coin` is not a `String` either

nym-api/src/unstable_routes/models.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn: nit: 2025

nym-api/src/unstable_routes/models.rs
jstuczyn: (this comment applies to all following occurences of it)

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn: nit: 2025

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+use utoipa::ToSchema;
+
+pub(crate) fn routes() -> Router<AppState> {
+    Router::new().route("/:address", get(address))
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, ToSchema, utoipa::IntoParams)]
+pub struct AddressQueryParam {
+    #[serde(default)]
+    pub address: String,
```
jstuczyn: could we perhaps change it to `AccountId` so that value validation would occur during deserialisation?

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+        .await
+        .into_iter()
+        .map(|node| (node.node_id(), node))
+        .collect::<HashMap<_, _>>();
+
+    let state_client = state.nyxd_client;
+    let mut total_value = 0u128;
+
+    // 1. get balances of chain tokens
+    let balances = state_client
```
jstuczyn: if we're only concerned about `unym`, query for only `unym` balance rather than getting `all_balances`. this way you'd also avoid having to mess with `Vec` and adding up amounts to `total_value`

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+    Path(AddressQueryParam { address }): Path<AddressQueryParam>,
+    State(state): State<AppState>,
+) -> AxumResult<Json<NyxAccountDetails>> {
+    let account_id = AccountId::from_str(&address).map_err(|err| {
+        error!("{err}");
+        AxumErrorResponse::not_found(&address)
+    })?;
+
+    let node_cache = state.nym_contract_cache();
+    let nym_nodes = node_cache
```
jstuczyn: 1. first query for account balance before attempting to retrieve and clone internal caches. if account doesn't exist, there's no point in doing any other operation

jstuczyn: 2. do we really need to clone the full cache of all nym-nodes? there's quite a lot of them and the cost overhead might not be negligible. perhaps `all_cached_nym_nodes` would have worked better? also do note it does not include **all** nodes. if somebody has an ancient delegation towards a legacy mixnode, it wouldn't come up here

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+                .rewarding_details
+                .determine_delegation_reward(delegation)
+            {
+                Ok(delegation_reward) => {
+                    rewards_map.insert(
+                        node_id,
+                        NyxAccountDelegationRewardDetails {
+                            node_id: delegation.node_id,
+                            rewards: decimal_to_coin(delegation_reward, "unym"),
+                            amount_staked: nym_node_details.original_pledge().clone(),
```
jstuczyn: just so we're clear, this is just pledge of the operator, not the full stake

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+                        "Couldn't determine delegations for {} on node {}: {}",
+                        &account_id, node_id, err
+                    )
+                }
+            }
+        }
+    }
+
+    // 4. make the map of rewards into a vec and sum the rewards and delegations
+    let accumulated_rewards: Vec<NyxAccountDelegationRewardDetails> =
```
jstuczyn: I don't understand. why are you creating the map only to immediately convert it into vec?

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+    let mut total_delegations = 0u128;
+    for r in &accumulated_rewards {
+        claimable_rewards += r.rewards.amount.u128();
+        total_delegations += r.amount_staked.amount.u128();
+        total_value += r.rewards.amount.u128();
+        total_value += r.amount_staked.amount.u128();
+    }
+
+    // 5. get vesting account details (if any): none because everyone has already fully vested
+    let vesting_account: Option<NymVestingAccount> = None;
```
jstuczyn: well.........
yes. they have fully vested, but not everyone has actually migrated the tokens to liquid. (and we didn't have  time to purge it ourselves)

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    if let Some(vesting_account) = vesting_account.clone() {
```
jstuczyn: why the clone

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+        total_value += vesting_account.locked.amount.u128();
+        total_value += vesting_account.spendable.amount.u128();
+    }
+
+    // 6. get operator rewards: is this an operator?
+    let mut operator_rewards = 0;
+
+    for (_, node_details) in nym_nodes.iter() {
```
jstuczyn:
```suggestion
    for node_details in nym_nodes.values() {
```

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+        if address == node_details.bond_information.owner.as_str() {
+            let pending_operator_reward = node_details.pending_operator_reward().amount.u128();
+
+            // add operator rewards
+            operator_rewards += pending_operator_reward;
+
+            // add to totals
+            total_value += pending_operator_reward;
```
jstuczyn: `break` at this point, because there's no point in continuing to iterate the map as there can only be a single bond

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+    }
+
+    // 7. convert totals
+    let claimable_rewards = Coin::new(claimable_rewards, "unym");
```
jstuczyn: I don't really like the hardcoded `unym` everywhere. what if some testnet uses different denom? can you just pull it from one of many `Coin`s you got it at the start?

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,207 @@
...
+    get,
+    path = "/{address}",
+    context_path = "/v1/unstable/account",
+    responses(
+        (status = 200, body = NyxAccountDetails)
+    ),
+    params(AddressQueryParam)
+)]
+#[instrument(level = "info", skip_all, fields(address=address))]
+async fn address(
```
jstuczyn: this requires some `AccountCache` behind it, like we have for `ChainStatusCache`. this is a huge potential source of DoS because there's a lot of expensive queries and operations for one account

nym-api/src/unstable_routes/models.rs
dynco-nym: fixed

nym-api/src/unstable_routes/account.rs
dynco-nym: unfortunately not, because `AccountId` is a foreign type which doesn't implement `utoipa::ToSchema`

I could do the whole "newtype pattern with `#[schema(value_type = AccountIdSchema)]`" thing but I don't see an advantage over current approach in that case

nym-api/src/unstable_routes/account.rs
dynco-nym: done

nym-api/src/unstable_routes/models.rs
dynco-nym: used newtype pattern to correctly derive `ToSchema` for Coin and its fields

nym-api/src/unstable_routes/account.rs
dynco-nym: fixed

nym-api/src/unstable_routes/account.rs
dynco-nym: replaced with `all_cached_nym_nodes`

it should be read from the same list, except this doesn't clone the whole vec?

nym-api/src/unstable_routes/account.rs
dynco-nym: copy/paste from old explorer code, fixed

nym-api/src/unstable_routes/account.rs
dynco-nym: good suggestion, reading it from `NymNetworkDetails`

nym-api/build.rs
dynco-nym: this generates a git-ignored bash file that allows you to easily shell into a running sqlite DB of nym-api

It's in the `build.rs` because sqlite location changes, depending on target dir which on debug builds have a new name every time you `cargo clean` it

changed: 
- only happens on unix and **not** in CI
- even if function fails, it won't unwrap the whole script

it's very useful while debugging so I wanted to share it. If you feel strongly against, I'll remove it

nym-api/build.rs
dynco-nym: see [other reply](https://github.com/nymtech/nym/pull/5673#discussion_r2044546718)

nym-api/build.rs
dynco-nym: see [other reply](https://github.com/nymtech/nym/pull/5673#discussion_r2044546718)

nym-api/src/unstable_routes/account.rs
dynco-nym: removing this field

nym-api/build.rs
jstuczyn: I'm not against it - I was just confused by it. if you feel it's useful for debugging, I don't see any reasons to remove it as long as it doesn't fail the build : )

nym-api/src/unstable_routes/account.rs
jstuczyn: but for the purposes of schema, it's just a string, so couldn't we just put `#[schema(value_type = String)]` there instead like we're doing already with so many types?

nym-api/src/unstable_routes/account.rs
jstuczyn: yup. exactly

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,199 @@
...
+        AxumErrorResponse::not_found(&address)
+    })?;
+    let base_denom = &state.network_details().network.chain_details.mix_denom.base;
+
+    let state_client = &state.nyxd_client;
+    let mut total_value = 0u128;
+
+    // 1. get balances of chain tokens
+    let balance = state_client
+        .get_address_balance(&account_id, base_denom)
```
jstuczyn: I actually don't remember from the top of my head if, when account doesnt exist, whether it will return an `Err` or `Ok(None)` - can you double check it with some dummy account?

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,199 @@
...
+        .collect::<HashSet<_>>();
+
+    let mut operator_rewards = 0;
+    let nym_nodes = state
+        .nym_contract_cache()
+        .all_cached_nym_nodes()
+        .await
+        .ok_or_else(AxumErrorResponse::service_unavailable)?
+        .iter()
+        .filter_map(|node_details| {
```
jstuczyn: I thought I already left a comment about it - there's no point in iterating the whole cache, just do a `.find()` because there can be at most one node that belongs to particular address

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,199 @@
...
+            // is this an operator of this node?
+            if account_id.to_string() == node_details.bond_information.owner.as_str() {
+                let pending_operator_reward = node_details.pending_operator_reward().amount.u128();
+
+                // add operator rewards
```
jstuczyn: I think that's something to discuss against requirements. do we want here tokens that could be claimed (i.e. pending rewards) or tokens that are actually associated with the account (`.operator` field - which includes original pledge **and** pending rewards)

jstuczyn: (sorry, I meant to add this comment for `total_value += pending_operator_reward;`)

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    // 3. get the current reward for each active delegation
+    // calculate rewards from nodes this delegator delegated to
+    let mut claimable_rewards = 0u128;
+
+    let mut accumulated_rewards = Vec::new();
+    for delegation in og_delegations.iter() {
+        let node_id = &delegation.node_id;
+
+        if let Some(nym_node_details) = nym_nodes.get(node_id) {
```
jstuczyn: ah. right. that's why you need all nym nodes in a map

nym-api/src/unstable_routes/account.rs
```diff
@@ -0,0 +1,199 @@
...
+    get,
+    path = "/{address}",
+    context_path = "/v1/unstable/account",
+    responses(
+        (status = 200, body = NyxAccountDetails)
+    ),
+    params(AddressQueryParam)
+)]
+#[instrument(level = "info", skip_all, fields(address=address))]
+async fn address(
```
jstuczyn: if possible, can you perhaps split this function into some smaller chunks, say one for retrieving delegation rewards, one for operator stuff, etc. i think it would make it slightly easier to read

nym-api/src/unstable_routes/account.rs
dynco-nym: what do you mean by account not existing?

if the address is invalid, it will fail validation before reaching this step

nym-api/src/unstable_routes/account.rs
dynco-nym: currently, unclaimed rewards _are_ displayed in totals so this keeps that behaviour

nym-api/src/unstable_routes/account.rs
dynco-nym: as you [noticed](https://github.com/nymtech/nym/pull/5673#discussion_r2044979383), I have to go through nodes at least once

nym-api/src/unstable_routes/account.rs
dynco-nym: done

nym-api/src/unstable_routes/account.rs
jstuczyn: by not existing, I mean it not having on-chain representation. there's a difference between 0 balance and no balance (i.e. not existing). for example if you generate a brand new mnemonic, **without sending any transactions**, that account wouldn't exist and if, for example, you queried for its sequence number, it wouldn't return anything

dynco-nym: tested with a fresh address and it handles it fine, filling all fields and returning 200
```
n19ha49y8rwyz2xemf92q7kre80elahr2ykrexm4
```

nym-api/src/unstable_routes/account/cache.rs
```diff
@@ -0,0 +1,115 @@
...
+#[derive(Clone)]
+pub(crate) struct AddressInfoCache {
+    inner: Cache<String, Arc<RwLock<NyxAccountDetails>>>,
+}
+
+impl AddressInfoCache {
+    pub(crate) fn new() -> Self {
+        // TODO dz values should come from config
+        let cache_ttl = Duration::from_secs(60);
+        let max_capacity = 1000;
```
dynco-nym: @jstuczyn sensible defaults?

nym-api/src/unstable_routes/account.rs
dynco-nym: added

dynco-nym: [added](https://github.com/nymtech/nym/pull/5673#discussion_r2060104264)

nym-api/src/support/http/state.rs
```diff
@@ -292,4 +295,12 @@ impl AppState {
             .await
             .ok_or_else(AxumErrorResponse::internal)
     }
+
+    pub(crate) async fn get_address_info(
+        self,
+        account_id: nym_validator_client::nyxd::AccountId,
+    ) -> Result<NyxAccountDetails, AxumErrorResponse> {
+        let address_info_cache = self.address_info_cache.clone();
```
jstuczyn: I'm not sure I understand the logic here, we're cloning the address cache and consuming `self`, only to call a method on address cache with `self` as the argument, i.e. changing `state.foo(cache)` to `cache.bar(state)`. why not just do everything here without that clone?

nym-api/src/unstable_routes/account/data_collector.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    pub(crate) async fn get_address_balance(
+        &mut self,
+    ) -> AxumResult<nym_validator_client::nyxd::Coin> {
+        let balance = self
+            .app_state
+            .nyxd_client
+            .get_address_balance(&self.account_id, &self.base_denom)
+            .await?
+            .unwrap_or_else(|| nym_validator_client::nyxd::Coin::new(0u128, &self.base_denom));
```
jstuczyn, line 53: so what was the result of your test, i.e. if the account didn't exist did it return an `Err` or `Ok(None)`?
essentially, would this function return an error so no further queries would be invoked?

jstuczyn (approved):
i think I added few nits and questions, but lgtm

nym-api/src/unstable_routes/account/data_collector.rs
dynco-nym: as we discussed in private: this returns `Ok(Some)` in all cases

there is [get_account](https://github.com/nymtech/nym/blob/develop/common/client-libs/validator-client/src/nyxd/cosmwasm_client/client_traits/query_client.rs#L90), which fails with `Err - not found`

but that's an overkill for this. To reduce number of queries, we fetch balance with `get_address_balance`, assume account with 0 unym is freshly created, has no delegtions and isn't an operator and we return early in that case

nym-api/src/support/http/state.rs
dynco-nym: fixed

nym-api/src/unstable_routes/account.rs
dynco-nym: as agreed, ignoring vesting info but leaving here in case it's required in the future

PR #5686, opened 2025-04-04
Expand /v3/nym-nodes with geodata

dynco-nym:
- includes node description and geodata
- expanded scope of included geodata

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5686)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
+++ b/nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
@@ -63,7 +63,6 @@ impl IpInfoClient {
     }
 }
 
-// TODO dz: are fields other than location used?
 #[derive(Debug, Clone, Serialize)]
 pub(crate) struct ExplorerPrettyBond {
     pub(crate) identity_key: String,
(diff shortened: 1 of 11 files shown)


pwnfoo (approved):

pwnfoo:
probably needs a version bump too for deployment?

nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
```diff
@@ -91,6 +106,19 @@ pub(crate) struct LocationResponse {
     pub(crate) two_letter_iso_country_code: String,
     #[serde(deserialize_with = "deserialize_loc")]
     pub(crate) loc: Coordinates,
+    // TODO dz consider making them optional?
```
jstuczyn: nit: would it make sense to just wrap all the geodata in separate struct and put it in an `Option` as you suggested yourself?

jstuczyn: especially since it's the case you will either have everything or nothing at all, i.e. it won't be the case you will have ip but not timezone

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/monitor/geodata.rs
dynco-nym: unfortunately, that is exactly the case: some fields may be missing, depending on the IP

PR #5687, opened 2025-04-04
Tauri V2 - Wallet Migration

tommyv1987:
- The core of the lifting was done via the migrate command
- A lot of API's changed
- Some help from SDK (Andrew) this is always welcomed
- Improved styling
- Pipelines are fixed for all platforms
- **TauriV2 is here**

A lot of pieces were refactored ☝️ 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5687)


nym-wallet/package.json
```diff
@@ -35,7 +36,11 @@
     "@nymproject/react": "^1.0.0",
     "@nymproject/types": "^1.0.0",
     "@storybook/react": "^6.5.15",
-    "@tauri-apps/api": "^1.2.0",
+    "@tauri-apps/api": "^2.4.0",
+    "@tauri-apps/cli": "^2.4.0",
```
doums: `@tauri-apps/cli` should be defined as dev-dependency only

nym-wallet/src-tauri/Cargo.toml
```diff
@@ -1,25 +1,29 @@
...
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [build-dependencies]
-tauri-build = { version = "=1.2.1", features = [] }
+tauri-build = { version = "2.1.1", features = [] }
 
-tauri-codegen = "=1.2.1"
-tauri-macros = "=1.2.1"
+tauri-codegen = "=2.1.1"
+tauri-macros = "2.1.1"
```
doums: Are these 2 deps still needed?

nym-wallet/src-tauri/capabilities/main-capability.json
```diff
@@ -0,0 +1,31 @@
...
+    "permissions": [
+        "core:default",
+        "core:path:default",
+        "core:event:default",
+        "core:window:default",
+        "core:app:default",
+        "core:resources:default",
+        "core:menu:default",
+        "core:tray:default",
+        "shell:allow-open",
```
doums: Suggestion: for opener it is encouraged to switch on https://tauri.app/plugin/opener/ (and drop shell)

nym-wallet/src-tauri/src/operations/app/link.rs
```diff
@@ -0,0 +1,22 @@
...
+#[tauri::command]
+pub async fn open_url(url: String) -> Result<(), String> {
+    println!("Opening URL: {}", url);
+
+    let status = match std::env::consts::OS {
+        "macos" => Command::new("open").arg(&url).status(),
+        "windows" => Command::new("cmd").args(["/c", "start", &url]).status(),
+        "linux" => Command::new("xdg-open").arg(&url).status(),
+        os => return Err(format!("Unsupported OS: {}", os)),
+    };
```
doums: yea instead of this manual implem you should use the opener plugin https://tauri.app/plugin/opener/

nym-wallet/src/components/Send/SendInputModal.tsx
```diff
@@ -52,28 +65,124 @@ export const SendInputModal = ({
...
+  const [errorFee, setErrorFee] = useState<string | undefined>();
+  const [noAccount, setNoAccount] = useState(false);
+
+  useEffect(() => {
+    if (!balance || balance === '0' || parseFloat(balance) === 0) {
+      setNoAccount(true);
+    } else {
+      setNoAccount(false);
+    }
+  }, [balance]);
```
doums: With such simple state/props conditions, it is better to avoid unnecessary `useEffect`:
```
// component root scope
const noAccount = !balance || balance === '0' || parseFloat(balance) === 0;
```

nym-wallet/src/components/TauriLinkWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  const handleClick = async (event: React.MouseEvent<HTMLAnchorElement>) => {
+    if (onClick) {
+      onClick(event);
+    }
+
+    if (href && (href.startsWith('http://') || href.startsWith('https://'))) {
+      event.preventDefault();
+      console.log('Opening link in browser:', href);
+      await openInBrowser(href);
```
doums: suggestion: you can avoid custom implem when using plugin-opener

```
import { openUrl } from '@tauri-apps/plugin-opener';
```

nym-wallet/src/theme/NymWalletTheme.tsx
```diff
@@ -1,17 +1,38 @@
...
 import { getDesignTokens } from './theme';
 import '@assets/fonts/non-variable/fonts.css';
 
+// Component that applies the Inter font for better typography
+const FontLoader = () => {
+  useEffect(() => {
+    // Add Inter font for modern typography
+    const interFontLink = document.createElement('link');
+    interFontLink.rel = 'stylesheet';
+    interFontLink.href = 'https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700&display=swap';
```
doums, line 21: This should be setup before entering react app, ie. in the JS entry point
Also note that when running in development react hooks run twice for stress purposes.
At least, you should protect it against re-runs

```
import [...]

// outside react tree
let init = false;

const FontLoader = () => {
  useEffect(() => {
    if (init) { return; }
    init = true;
    // logic
```

doums:
In the `Cargo.toml` I'm not sure if this feature is still needed
```
custom-protocol = ["tauri/custom-protocol"]
```

nym-wallet/src-tauri/Cargo.toml
tommyv1987: Removed, you're right they're not needed 👍

nym-wallet/package.json
tommyv1987: Fixed 👍

nym-wallet/src/components/TauriLinkWrapper.tsx
tommyv1987: Replaced this with the `plugin-opener` :)

nym-wallet/src/theme/NymWalletTheme.tsx
tommyv1987: Noted and changed

nym-wallet/src/components/Send/SendInputModal.tsx
tommyv1987: Nice one, changed this as well

nym-wallet/src-tauri/src/operations/app/link.rs
tommyv1987: Now using the plugin

nym-wallet/src-tauri/capabilities/main-capability.json
tommyv1987: Dropped this too

fmtabbara:
Generally looks OK. A couple things that stand out from a frontend perspective are

1. There are quite a lot of inline styles now - I think this is OK for now and necessary to get the new theme in quickly. If/when there's time in the future we should update the global theme in order to reduce the number of inline styles.

2. The vesting feature is still there - Can this be removed yet?

nym-wallet/src/theme/NymWalletTheme.tsx
fmtabbara: I'm not sure that this is the correct way to set the fonts up. If it's working leave it like this but we may want revisit - https://mui.com/material-ui/react-typography

tommyv1987:
> Generally looks OK. A couple things that stand out from a frontend perspective are
> 
> 1. There are quite a lot of inline styles now - I think this is OK for now and necessary to get the new theme in quickly. If/when there's time in the future we should update the global theme in order to reduce the number of inline styles.
> 2. The vesting feature is still there - Can this be removed yet?

Unfortunately we can't remove vesting just yet @fmtabbara, otherwise this would've been the first thing to remove :) 

As for styling let's look at this after as well

tommyv1987:
@fmtabbara I haven't had the chance to update the storybook components, but that too, can come in the future

doums (approved):

benedettadavico (approved):

PR #4360, opened 2024-01-30
Feature Noise XKpsk3 integration

simonwicky:
# Description
 This is the base of stacked PRs to integrate XKpsk3 Noise connection between nym nodes

This PR concerns self-described API and Nym-API changes.

The stack of PR is as follow :
#4360 (this one) -> [#4361](https://github.com/nymtech/nym/pull/4361) -> #4373 -> [#4362](https://github.com/nymtech/nym/pull/4362)

# Checklist:

- [ ] added a changelog entry to `CHANGELOG.md`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/4360)

PR #5692, opened 2025-04-07
[Feature] Noise XKpsk3 integration (2025 version)

simonwicky:
# Description
Noise PR https://github.com/nymtech/nym/pull/4360 is dead, long live Noise PR #5692.

No stacked PRs this time, commits description are quite explicit (and it's less of a mess than last time)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5692)

### DIFF
--- a/common/client-libs/mixnet-client/src/client.rs
+++ b/common/client-libs/mixnet-client/src/client.rs
@@ -3,6 +3,8 @@
 
 use dashmap::DashMap;
 use futures::StreamExt;
+use nym_noise::config::NoiseConfig;
+use nym_noise::upgrade_noise_initiator;
 use nym_sphinx::forwarding::packet::MixPacket;
 use nym_sphinx::framing::codec::NymCodec;
 use nym_sphinx::framing::packet::FramedNymPacket;
@@ -52,6 +54,7 @@ pub trait SendWithoutResponse {
 
 pub struct Client {
     active_connections: ActiveConnections,
+    noise_config: NoiseConfig,
     connections_count: Arc<AtomicUsize>,
     config: Config,
 }
@@ -97,6 +100,7 @@ impl ConnectionSender {
 
 struct ManagedConnection {
     address: SocketAddr,
+    noise_config: NoiseConfig,
     message_receiver: ReceiverStream<FramedNymPacket>,
     connection_timeout: Duration,
     current_reconnection: Arc<AtomicU32>,
@@ -105,12 +109,14 @@ struct ManagedConnection {
 impl ManagedConnection {
     fn new(
         address: SocketAddr,
+        noise_config: NoiseConfig,
         message_receiver: mpsc::Receiver<FramedNymPacket>,
         connection_timeout: Duration,
         current_reconnection: Arc<AtomicU32>,
     ) -> Self {
         ManagedConnection {
             address,
+            noise_config,
             message_receiver: ReceiverStream::new(message_receiver),
             connection_timeout,
             current_reconnection,
@@ -125,9 +131,21 @@ impl ManagedConnection {
             Ok(stream_res) => match stream_res {
                 Ok(stream) => {
                     debug!("Managed to establish connection to {}", self.address);
-                    // if we managed to connect, reset the reconnection count (whatever it might have been)
+
+                    let noise_stream =
+                        match upgrade_noise_initiator(stream, &self.noise_config).await {
+                            Ok(noise_stream) => noise_stream,
+                            Err(err) => {
+                                error!("Failed to perform Noise handshake with {address} - {err}");
+                                // we failed to finish the noise handshake - increase reconnection attempt
+                                self.current_reconnection.fetch_add(1, Ordering::SeqCst);
+                                return;
+                            }
+                        };
+                    // if we managed to connect AND do the noise handshake, reset the reconnection count (whatever it might have been)
                     self.current_reconnection.store(0, Ordering::Release);
-                    Framed::new(stream, NymCodec)
+                    debug!("Noise initiator handshake completed for {:?}", address);
+                    Framed::new(noise_stream, NymCodec)
                 }
                 Err(err) => {
                     debug!("failed to establish connection to {address} (err: {err})",);
@@ -160,9 +178,14 @@ impl ManagedConnection {
 }
 
 impl Client {
-    pub fn new(config: Config, connections_count: Arc<AtomicUsize>) -> Client {
+    pub fn new(
+        config: Config,
+        noise_config: NoiseConfig,
+        connections_count: Arc<AtomicUsize>,
+    ) -> Client {
         Client {
             active_connections: Default::default(),
+            noise_config,
             connections_count,
             config,
         }
@@ -217,6 +240,7 @@ impl Client {
         let initial_connection_timeout = self.config.initial_connection_timeout;
 
         let connections_count = self.connections_count.clone();
+        let noise_config = self.noise_config.clone();
         tokio::spawn(async move {
             // before executing the manager, wait for what was specified, if anything
             if let Some(backoff) = backoff {
@@ -227,6 +251,7 @@ impl Client {
             connections_count.fetch_add(1, Ordering::SeqCst);
             ManagedConnection::new(
                 address,
+                noise_config,
                 receiver,
                 initial_connection_timeout,
                 current_reconnection_attempt,
@@ -291,8 +316,12 @@ impl SendWithoutResponse for Client {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use nym_crypto::asymmetric::x25519;
+    use nym_noise::config::NoiseNetworkView;
+    use rand::rngs::OsRng;
 
     fn dummy_client() -> Client {
+        let mut rng = OsRng; //for test only, so we don't care if rng source isn't crypto grade
         Client::new(
             Config {
                 initial_reconnection_backoff: Duration::from_millis(10_000),
@@ -300,6 +329,11 @@ mod tests {
                 initial_connection_timeout: Duration::from_millis(1_500),
                 maximum_connection_buffer_size: 128,
             },
+            NoiseConfig::new(
+                Arc::new(x25519::KeyPair::new(&mut rng)),
+                NoiseNetworkView::new_empty(),
+                Duration::from_millis(1_500),
+            ),
             Default::default(),
         )
     }
--- a/common/nymnoise/src/config.rs
+++ b/common/nymnoise/src/config.rs
@@ -1,0 +1,171 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{
+    collections::HashMap,
+    net::{IpAddr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
+
+use arc_swap::ArcSwap;
+use nym_crypto::asymmetric::x25519;
+use nym_noise_keys::{NoiseVersion, VersionedNoiseKey};
+use snow::params::NoiseParams;
+
+use strum::{EnumIter, FromRepr};
+
+#[derive(Default, Debug, Clone, Copy, EnumIter, FromRepr, Eq, PartialEq)]
+#[repr(u8)]
+#[non_exhaustive]
+pub enum NoisePattern {
+    #[default]
+    XKpsk3 = 1,
+    IKpsk2 = 2,
+}
+
+impl NoisePattern {
+    pub(crate) const fn as_str(&self) -> &'static str {
+        match self {
+            Self::XKpsk3 => "Noise_XKpsk3_25519_AESGCM_SHA256",
+            Self::IKpsk2 => "Noise_IKpsk2_25519_ChaChaPoly_BLAKE2s", //Wireguard handshake (not exactly though)
+        }
+    }
+
+    // SAFETY: we have tests to ensure that hardcoded pattern are correct
+    #[allow(clippy::unwrap_used)]
+    pub(crate) fn psk_position(&self) -> u8 {
+        //automatic parsing, works for correct pattern, more convenient
+        match self.as_str().find("psk") {
+            Some(n) => {
+                let psk_index = n + 3;
+                let psk_char = self.as_str().chars().nth(psk_index).unwrap();
+                psk_char.to_string().parse().unwrap()
+            }
+            None => 0,
+        }
+    }
+
+    // SAFETY : we have tests to ensure that hardcoded pattern are correct
+    #[allow(clippy::unwrap_used)]
+    pub(crate) fn as_noise_params(&self) -> NoiseParams {
+        self.as_str().parse().unwrap()
+    }
+}
+
+#[derive(Debug, Default)]
+struct SocketAddrToKey {
+    inner: ArcSwap<HashMap<SocketAddr, VersionedNoiseKey>>,
+}
+
+// SW NOTE : Only for phased upgrade. To remove once we decide all nodes have to support Noise
+#[derive(Debug, Default)]
+struct IpAddrToVersion {
+    inner: ArcSwap<HashMap<IpAddr, NoiseVersion>>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct NoiseNetworkView {
+    keys: Arc<SocketAddrToKey>,
+    support: Arc<IpAddrToVersion>,
+}
+
+impl NoiseNetworkView {
+    pub fn new_empty() -> Self {
+        NoiseNetworkView {
+            keys: Default::default(),
+            support: Default::default(),
+        }
+    }
+
+    pub fn swap_view(&self, new: HashMap<SocketAddr, VersionedNoiseKey>) {
+        let noise_support = new
+            .iter()
+            .map(|(s_addr, key)| (s_addr.ip(), key.supported_version))
+            .collect::<HashMap<_, _>>();
+        self.keys.inner.store(Arc::new(new));
+        self.support.inner.store(Arc::new(noise_support));
+    }
+}
+
+#[derive(Clone)]
+pub struct NoiseConfig {
+    network: NoiseNetworkView,
+
+    pub(crate) local_key: Arc<x25519::KeyPair>,
+    pub(crate) pattern: NoisePattern,
+    pub(crate) timeout: Duration,
+
+    pub(crate) unsafe_disabled: bool, // allows for nodes to not attempt to do a noise handshake, VERY UNSAFE, FOR DEBUG PURPOSE ONLY
+}
+
+impl NoiseConfig {
+    pub fn new(
+        noise_key: Arc<x25519::KeyPair>,
+        network: NoiseNetworkView,
+        timeout: Duration,
+    ) -> Self {
+        NoiseConfig {
+            network,
+            local_key: noise_key,
+            pattern: Default::default(),
+            timeout,
+            unsafe_disabled: false,
+        }
+    }
+
+    #[must_use]
+    pub fn with_noise_pattern(mut self, pattern: NoisePattern) -> Self {
+        self.pattern = pattern;
+        self
+    }
+
+    #[must_use]
+    pub fn with_unsafe_disabled(mut self, disabled: bool) -> Self {
+        self.unsafe_disabled = disabled;
+        self
+    }
+
+    pub(crate) fn get_noise_key(&self, s_address: &SocketAddr) -> Option<VersionedNoiseKey> {
+        self.network.keys.inner.load().get(s_address).copied()
+    }
+
+    // Only for phased update
+    //SW This can lead to some troubles if two nodes shares the same IP and one support Noise but not the other. This in only for the progressive update though and there is no workaround
+    pub(crate) fn get_noise_support(&self, ip_addr: IpAddr) -> Option<NoiseVersion> {
+        let plain_ip_support = self.network.support.inner.load().get(&ip_addr).copied();
+
+        // SW default bind address being [::]:1789, it can happen that a responder sees the ipv6-mapped address of the initiator, this check for that
+        let canonical_ip = &ip_addr.to_canonical();
+        let canonical_ip_support = self.network.support.inner.load().get(canonical_ip).copied();
+        plain_ip_support.or(canonical_ip_support)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use snow::params::NoiseParams;
+
+    use super::NoisePattern;
+    use std::str::FromStr;
+    use strum::IntoEnumIterator;
+
+    // The goal of these is to make sure every NoisePatterns are correct and unwrap can be used on them
+
+    #[test]
+    fn noise_patterns_are_valid() {
+        for pattern in NoisePattern::iter() {
+            assert!(NoiseParams::from_str(pattern.as_str()).is_ok())
+        }
+    }
+
+    #[test]
+    fn noise_patterns_psk_position_is_valid() {
+        for pattern in NoisePattern::iter() {
+            match pattern {
+                NoisePattern::XKpsk3 => assert_eq!(pattern.psk_position(), 3),
+                NoisePattern::IKpsk2 => assert_eq!(pattern.psk_position(), 2),
+            }
+        }
+    }
+}
--- a/common/nymnoise/src/lib.rs
+++ b/common/nymnoise/src/lib.rs
@@ -1,0 +1,118 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use nym_noise_keys::NoiseVersion;
+use snow::error::Prerequisite;
+use snow::Error;
+use tokio::net::TcpStream;
+use tracing::{error, warn};
+
+pub mod config;
+pub mod connection;
+pub mod error;
+pub mod stream;
+
+use crate::config::NoiseConfig;
+use crate::connection::Connection;
+use crate::error::NoiseError;
+use crate::stream::NoiseStreamBuilder;
+
+const NOISE_PSK_PREFIX: &[u8] = b"NYMTECH_NOISE_dQw4w9WgXcQ";
+
+pub const LATEST_NOISE_VERSION: NoiseVersion = NoiseVersion::V1;
+
+// TODO: this should be behind some trait because presumably, depending on the version,
+// other arguments would be needed
+mod psk_gen {
+    use crate::error::NoiseError;
+    use crate::stream::Psk;
+    use crate::NOISE_PSK_PREFIX;
+    use nym_crypto::asymmetric::x25519;
+    use nym_noise_keys::NoiseVersion;
+    use sha2::{Digest, Sha256};
+
+    pub(crate) fn generate_psk(
+        responder_pub_key: x25519::PublicKey,
+        version: NoiseVersion,
+    ) -> Result<Psk, NoiseError> {
+        match version {
+            NoiseVersion::V1 => Ok(generate_psk_v1(responder_pub_key)),
+            NoiseVersion::Unknown(noise_version) => {
+                Err(NoiseError::PskGenerationFailure { noise_version })
+            }
+        }
+    }
+
+    fn generate_psk_v1(responder_pub_key: x25519::PublicKey) -> [u8; 32] {
+        let mut hasher = Sha256::new();
+        hasher.update(NOISE_PSK_PREFIX);
+        hasher.update(responder_pub_key.to_bytes());
+        hasher.finalize().into()
+    }
+}
+
+pub async fn upgrade_noise_initiator(
+    conn: TcpStream,
+    config: &NoiseConfig,
+) -> Result<Connection<TcpStream>, NoiseError> {
+    if config.unsafe_disabled {
+        warn!("Noise is disabled in the config. Not attempting any handshake");
+        return Ok(Connection::Raw(conn));
+    }
+
+    //Get init material
+    let responder_addr = conn.peer_addr().map_err(|err| {
+        error!("Unable to extract peer address from connection - {err}");
+        Error::Prereq(Prerequisite::RemotePublicKey)
+    })?;
+
+    let Some(key) = config.get_noise_key(&responder_addr) else {
+        warn!("{responder_addr} can't speak Noise yet, falling back to TCP");
+        return Ok(Connection::Raw(conn));
+    };
+
+    let handshake_version = match key.supported_version {
+        NoiseVersion::V1 => NoiseVersion::V1,
+
+        // We're talking to a more recent node, but we can't adapt. Let's try to do our best and if it fails, it fails.
+        // If that node sees we're older, it will try to adapt too.
+        NoiseVersion::Unknown(version) => {
+            warn!("{responder_addr} is announcing an v{version} version of Noise that we don't know how to parse, we will attempt to downgrade to our current highest supported version");
+            LATEST_NOISE_VERSION
+        }
+    };
+
+    NoiseStreamBuilder::new(conn)
+        .perform_initiator_handshake(config, handshake_version, key.x25519_pubkey)
+        .await
+        .map(|stream| Connection::Noise(Box::new(stream)))
+}
+pub async fn upgrade_noise_responder(
+    conn: TcpStream,
+    config: &NoiseConfig,
+) -> Result<Connection<TcpStream>, NoiseError> {
+    if config.unsafe_disabled {
+        warn!("Noise is disabled in the config. Not attempting any handshake");
+        return Ok(Connection::Raw(conn));
+    }
+
+    //Get init material
+    let initiator_addr = match conn.peer_addr() {
+        Ok(addr) => addr,
+        Err(err) => {
+            error!("Unable to extract peer address from connection - {err}");
+            return Err(Error::Prereq(Prerequisite::RemotePublicKey).into());
+        }
+    };
+
+    // if responder doesn't announce noise support, we fallback to tcp
+    if config.get_noise_support(initiator_addr.ip()).is_none() {
+        warn!("{initiator_addr} can't speak Noise yet, falling back to TCP",);
+        return Ok(Connection::Raw(conn));
+    };
+
+    NoiseStreamBuilder::new(conn)
+        .perform_responder_handshake(config)
+        .await
+        .map(|stream| Connection::Noise(Box::new(stream)))
+}
--- a/nym-node/src/node/mod.rs
+++ b/nym-node/src/node/mod.rs
@@ -50,6 +50,8 @@ use nym_network_requester::{
 use nym_node_metrics::events::MetricEventsSender;
 use nym_node_metrics::NymNodeMetrics;
 use nym_node_requests::api::v1::node::models::{AnnouncePorts, NodeDescription};
+use nym_noise::config::{NoiseConfig, NoiseNetworkView};
+use nym_noise_keys::VersionedNoiseKey;
 use nym_sphinx_acknowledgements::AckKey;
 use nym_sphinx_addressing::Recipient;
 use nym_task::{ShutdownManager, ShutdownToken, TaskClient};
@@ -1031,6 +1033,7 @@ impl NymNode {
         active_clients_store: &ActiveClientsStore,
         replay_protection_bloomfilter: ReplayProtectionBloomfilters,
         routing_filter: F,
+        noise_config: NoiseConfig,
         shutdown: ShutdownToken,
     ) -> Result<(MixForwardingSender, ActiveConnections), NymNodeError>
     where
@@ -1054,6 +1057,7 @@ impl NymNode {
         );
         let mixnet_client = nym_mixnet_client::Client::new(
             mixnet_client_config,
+            noise_config.clone(),
             self.metrics
                 .network
                 .active_egress_mixnet_connections_counter(),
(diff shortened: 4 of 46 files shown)


timkuijsten:
Is there any rationale of the choice for XKpsk3? I.e. how to do DoS mitigation? Also I'm curious how this relates to the announcement of planning to use [McEliece](https://mailarchive.ietf.org/arch/msg/cfrg/O_TCtOactOG3X3UPqM3_7DomaNY/).

aniampio:
@timkuijsten Thanks for the question! We decided to use XKpsk3 because it provides the best forward secrecy, authentication, replay protection, and identity-hiding guarantees for client-server settings where the client cannot be identified based on its source IP address. Also, as PQ-security is on our roadmap, we opted for the variant with PSK, as the pre-shared key can be later used to inject Post-Quantum safety into the protocol. We also decided to use the XKpsk3 between nodes, as we are considering supporting private gateways in the future, so the identity of the initiator may not be obvious from the source IP address (entry-mixnode). Then, for the rest of the connections (mix node to mix node, and mix node to exit) we also use the same pattern to keep the usage across the network uniform. 

In terms of DoS protection, deploying Noise at the transport layer helps shield the Sphinx layer, as each message must first be validated by Noise before any Sphinx processing occurs. However, Noise itself can still be vulnerable to DoS attacks. WireGuard mitigates this risk using a cookie-based mechanism. We could adopt a similar approach, or explore alternative strategies — all of which are currently under consideration.

nym-api/src/nym_nodes/handlers/unstable/semi_skimmed.rs
```diff
@@ -1,27 +1,132 @@
...
+) -> Vec<SemiSkimmedNode>
+where
+    NI: Iterator<Item = &'a NymNodeDescription> + 'a,
+{
+    let mut nodes = Vec::new();
+    for nym_node in nym_nodes_subset {
+        let node_id = nym_node.node_id;
+
+        let role: NodeRole = rewarded_set.role(node_id).into();
+
```
jstuczyn: you copied it from `skimmed` response but you dropped filtering for inactive nodes, what gives?

common/nymnoise/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    // SAFETY: if unwraps failed, it means hardcoded patterns were wrong
+    #[allow(clippy::unwrap_used)]
+    pub(crate) fn psk_position(&self) -> u8 {
+        //automatic parsing, works for correct pattern, more convenient
+        match self.as_str().find("psk") {
+            Some(n) => {
+                let psk_index = n + 3;
+                let psk_char = self.as_str().chars().nth(psk_index).unwrap();
```
jstuczyn, line 42: thinking about it, given we're using hardcoded patterns, couldn't we also hardcode that `psk_position` to avoid `unwrap` alltogether?

common/nymnoise/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    pub(crate) fn get_noise_key(&self, s_address: &SocketAddr) -> Option<VersionedNoiseKey> {
+        self.network.keys.inner.load().get(s_address).copied()
+    }
+
+    // Only for phased update
+    //SW This can lead to some troubles if two nodes shares the same IP and one support Noise but not the other. This in only for the progressive update though and there is no workaround
+    pub(crate) fn get_noise_support(&self, ip_addr: IpAddr) -> Option<NoiseVersion> {
+        self.network
```
jstuczyn: can you break that getter a bit into few lines? It's quite unreadable : )

common/nymnoise/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+            .inner
+            .load()
+            .get(&ip_addr)
+            .copied()
+            .or_else(|| {
+                self.network
+                    .support
+                    .inner
+                    .load()
+                    .get(&ip_addr.to_canonical()) // SW default bind address being [::]:1789, it can happen that a responder sees the ipv6-mapped address of the initiator, this check for that
```
jstuczyn: why would that matter?

jstuczyn: also, wouldn't it be better if we always and **only** used canonical address?

common/nymnoise/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+        self.unsafe_disabled = disabled;
+        self
+    }
+
+    pub(crate) fn get_noise_key(&self, s_address: &SocketAddr) -> Option<VersionedNoiseKey> {
+        self.network.keys.inner.load().get(s_address).copied()
+    }
+
+    // Only for phased update
+    //SW This can lead to some troubles if two nodes shares the same IP and one support Noise but not the other. This in only for the progressive update though and there is no workaround
```
jstuczyn, line 134: i hope this doesn't introduce any problems/vulnerabilities, because somebody will attempt to use that.

couldn't we simply say that if we have connection from ip A.B.C.D and it doesn't use noise, all connections from that ip are not allowed to use noise (and vice versa)

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+use snow::{error::Prerequisite, Builder, Error};
+use tokio::net::TcpStream;
+use tracing::*;
+
+pub mod config;
+pub mod connection;
+pub mod error;
+pub mod stream;
+
+const NOISE_PSK_PREFIX: &[u8] = b"NYMTECH_NOISE_dQw4w9WgXcQ";
```
jstuczyn, line 20: *sigh*. also, is it fine if the key is hardcoded?

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+    remote_pub_key: &x25519::PublicKey,
+) -> Result<Connection, NoiseError> {
+    trace!("Perform Noise Handshake, initiator side");
+
+    let secret = [
+        NOISE_PSK_PREFIX.to_vec(),
+        remote_pub_key.to_bytes().to_vec(),
+    ]
+    .concat();
+    let secret_hash = Sha256::digest(secret);
```
jstuczyn: given this is computed in two places, I'd recomend putting it in a separate function to avoid error if/once we decide to update it, e.g.
```rust
fn produce_secret_hash(pub_key: x25519::PublicKey) -> [u8; 32] {
    let mut hasher = Sha256::new();
    hasher.update(NOISE_PSK_PREFIX);
    hasher.update(pub_key.to_bytes());
    hasher.finalize().into()
}
```

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let handshake = Builder::new(config.pattern.as_str().parse()?)
```
jstuczyn: if we have to parse values from the config (which seem to be fallible!) why not just store `NoiseParams` directly in the config to avoid this issue?

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+        .local_private_key(&config.local_key.private_key().to_bytes())
```
jstuczyn: do we really have to attach private key bytes here? if so, wrap them in `Zeroize` because without it those private bytes will live in memory, e.g. `&Zeroizing::new(config.local_key.private_key().to_bytes())`

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let secret = [
+        NOISE_PSK_PREFIX.to_vec(),
+        config.local_key.public_key().to_bytes().to_vec(),
+    ]
+    .concat();
+    let secret_hash = Sha256::digest(secret);
+
+    let handshake = Builder::new(config.pattern.as_str().parse()?)
+        .local_private_key(&config.local_key.private_key().to_bytes())
```
jstuczyn: sanity check: do we not have to attach remote public key here like with the initiator?

common/nymnoise/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+        .psk(config.pattern.psk_position(), &secret_hash)
+        .build_responder()?;
+
+    let noise_stream = NoiseStream::new(conn, handshake);
+
+    Ok(Connection::Noise(noise_stream.perform_handshake().await?))
```
jstuczyn: I'm confused. you're creating `NoiseStream` with `handshake` information only to call `perform_handshake` which immediately calls `self.handshake = None;`. so what was the point in creating it at all?

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+            dec_buffer: VecDeque::with_capacity(MAXMSGLEN),
+        }
+    }
+
+    pub(crate) async fn perform_handshake(mut self) -> Result<Self, NoiseError> {
+        //Check if we are in the correct state
+        let Some(mut handshake) = self.handshake else {
+            return Err(NoiseError::IncorrectStateError);
+        };
+        self.handshake = None;
```
jstuczyn: why is this immediatlely reset

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+        Ok(self)
+    }
+
+    async fn send_handshake_msg(
+        &mut self,
+        handshake: &mut HandshakeState,
+    ) -> Result<(), NoiseError> {
+        let mut buf = BytesMut::zeroed(MAXMSGLEN + TAGLEN);
+        let len = handshake.write_message(&[], &mut buf)?;
+        buf.truncate(len);
```
jstuczyn: i'm confused. you're creating a buffer of `MAXMSGLEN + TAGLEN` zeroes and then calling truncate to... `MAXMSGLEN + TAGLEN` again. why? also why are we sending so many zeroes during noise handshake

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    async fn recv_handshake_msg(
+        &mut self,
+        handshake: &mut HandshakeState,
+    ) -> Result<(), NoiseError> {
+        match self.inner_stream.next().await {
+            Some(Ok(msg)) => {
+                let mut buf = vec![0u8; MAXMSGLEN];
+                handshake.read_message(&msg, &mut buf)?;
```
jstuczyn: shouldn't we retrieve number of bytes read to correctly truncate the buf?

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let projected_self = self.project();
+
+        match projected_self.inner_stream.poll_next(cx) {
+            Poll::Pending => {
+                //no new data, waking is already scheduled.
```
jstuczyn: > waking is already scheduled.

where?

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            Poll::Ready(Some(Err(err))) => return Poll::Ready(Err(err)),
+
+            //Stream is done, return Ok with nothing in buf
+            Poll::Ready(None) => return Poll::Ready(Ok(())),
+        }
+
+        //check and return what we can
+        let read_len = min(buf.remaining(), projected_self.dec_buffer.len());
```
jstuczyn: but if `buf.remaining()` gives you 0 because it's full, it means `read_len > 0` will fail, `Poll::Pending` will be returned and the task will be starved because it won't ever be woken up

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+const TAGLEN: usize = 16;
+
+/// Wrapper around a TcpStream
+#[pin_project]
+pub struct NoiseStream {
+    #[pin]
+    inner_stream: Framed<TcpStream, LengthDelimitedCodec>,
+    handshake: Option<HandshakeState>,
+    noise: Option<TransportState>,
+    dec_buffer: VecDeque<u8>,
```
jstuczyn: look at `Bytes` and `BytesMut`. it might be more appropriate (and efficient) than `VecDeque` here

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,191 @@
...
+                let mut noise_buf = BytesMut::zeroed(MAXMSGLEN + TAGLEN);
+
+                let Ok(len) = (match projected_self.noise {
+                    Some(transport_state) => transport_state.write_message(buf, &mut noise_buf),
+                    None => return Poll::Ready(Err(io::ErrorKind::Other.into())),
+                }) else {
+                    return Poll::Ready(Err(io::ErrorKind::InvalidInput.into()));
+                };
+                noise_buf.truncate(len);
+                match projected_self.inner_stream.start_send(noise_buf.into()) {
```
jstuczyn: but `start_send` does not imply the data has been sent

nym-node/src/node/mod.rs
```diff
@@ -700,7 +702,10 @@ impl NymNode {
         let host_details = sign_host_details(
             &self.config,
             self.x25519_sphinx_keys.public_key(),
-            self.x25519_noise_keys.public_key(),
+            &VersionedNoiseKey {
+                version: nym_noise::LATEST_NOISE_VERSION,
```
jstuczyn: this comment would be more appropriate in a different spot, but I just thought about it after seeing it yet again. I think it would be useful if the key serialised on disk also contained this version information because presumably `v5` key wouldn't be compatible with `v1`

nym-node/src/node/shared_network.rs
```diff
@@ -53,10 +54,10 @@ impl NodesQuerier {
         res
     }
 
-    async fn current_nymnodes(&mut self) -> Result<Vec<SkimmedNode>, ValidatorClientError> {
+    async fn current_nymnodes(&mut self) -> Result<Vec<SemiSkimmedNode>, ValidatorClientError> {
         let res = self
             .client
-            .get_all_basic_nodes()
+            .get_all_expanded_nodes()
```
jstuczyn, line 72: there's a tiny issue. currently nym api will return unimplemented on that endpoint : (

nym-node/src/node/shared_network.rs
```diff
@@ -263,6 +272,22 @@ impl NetworkRefresher {
...
         self.routing_filter.pending.clear().await;
 
+        //update noise Noise Nodes
+        let noise_nodes = nodes
+            .iter()
+            .filter(|n| n.x25519_noise_versioned_key.is_some())
+            .flat_map(|n| {
+                n.basic.ip_addresses.iter().map(|ip_addr| {
+                    (
+                        SocketAddr::new(*ip_addr, n.basic.mix_port),
```
jstuczyn, line 326: wait, hold on a second here. i'm not sure this is the port you want to be using. this is the port the remote node is **listening** on. when it tries to establish connection to **your** node, it will be different

nym-api/src/nym_nodes/handlers/unstable/semi_skimmed.rs
simonwicky: Since (at least for the moment) we need this endpoint to get the Noise keys, we also need it for inactive nodes (tests and stuff) so I did not why we would need it only for active ones. 
That being said, I can add the filtering back

common/nymnoise/src/config.rs
simonwicky: When adding/modifying patterns, I don't want to have that info in two places, that can lead to mismatch if not maintained properly. 

The `snow` library is parsing the pattern string to get its scheme so it's not really far from that.

Given that, I feel like this is still the better option no?

common/nymnoise/src/config.rs
simonwicky: Agreed. The formatter is being silly here. I'll see what I can do

common/nymnoise/src/config.rs
simonwicky: It does matter because I have seen a case where the responder sees the IPv6-mapped version of the initiator's IP, and it needs to canonicalize it to find it. 

As for only using the canonical version, using solely the IP for this is kinda shaky, so let's put all the chances on our side no? We're more likely to miss a Noise supporting node than to think a node is supporting it when it doesn't

common/nymnoise/src/config.rs
simonwicky: They might exploit and block you if you don't update. If you do, they can't. 

We could but then it means somebody can force a downgrade on your up-to-date node. Better to cause problems on older than newer nodes no?

common/nymnoise/src/lib.rs
simonwicky: Internal notes from Cryspen says it is (I know it's weird)
TLDR : a noise PSK protocol is always better than its non-psk variant, even if the psk is public and static

common/nymnoise/src/lib.rs
simonwicky: We do, it's a `snow` lib method. I'll add the zeroize stuff though

common/nymnoise/src/lib.rs
simonwicky: We do not. We have an XK scheme, the responder doesn't know the initiator's public key so there is no remote public key here

nym-node/src/node/mod.rs
simonwicky: Noise in any form requires an encryption (x25519) key. As long as we're using Noise, we can keep the same key. 
The Noise versions are here so we can change the scheme and/or the PSK (for PQ encryption for example). In any case, we can still use the same key, so we don't need that info on disk

nym-node/src/node/shared_network.rs
simonwicky: Well let's sure to update the nym-api before the nodes. 
This endpoint is implemented in that PR

nym-node/src/node/shared_network.rs
simonwicky: This is exactly the info I want. The port info is used by initiator, which is sending to your node. 
The responder will only use the IP part that is constructed here, where I'm removing the port info : https://github.com/nymtech/nym/blob/fba6c26ebf92115a4ec1ff9a178938d671615289/common/nymnoise/src/config.rs#L70

common/nymnoise/src/lib.rs
simonwicky: Creating `NoiseParams` is fallible so then creating the config would be fallible. 
The pattern string will be correct, so in could almost be an `unwrap`
In that spirit, let's group the potential errors

common/nymnoise/src/lib.rs
simonwicky: In the spirit of `nymnoise` being a library, it might the case that we want to do something else in between

simonwicky: Also, I don't really like having an `async fn new` that goes ahead and do network stuff

common/nymnoise/src/stream.rs
simonwicky: `snow` uses a `HandshakeState` that gets turned into a `TransportState`. If we start the handshake and fail, we cannot reuse the `HandshakeState`, hence it being reset

common/nymnoise/src/stream.rs
simonwicky: ~~The message can't be bigger than `MAXMSGLEN`, the inner_stream is using `LengthDelimitedCodec` and `read_message` is from `snow` and does the job. 
I don't think we need to do anything else~~
Edit : allocation size reduced in ea233c78e2e6016939f2662762bd0e0397e18541
See one of the comments below as for why

common/nymnoise/src/stream.rs
simonwicky: If `inner_stream.poll_next` returns `Poll::Pending`, the waking is already scheduled (by the `poll_next` call).

timkuijsten:
> In terms of DoS protection, deploying Noise at the transport layer helps shield the Sphinx layer, as each message must first be validated by Noise before any Sphinx processing occurs. However, Noise itself can still be vulnerable to DoS attacks. WireGuard mitigates this risk using a cookie-based mechanism. We could adopt a similar approach, or explore alternative strategies — all of which are currently under consideration.

I guess XK + cookie and making sure there is no responder-side state for unauthenticated connections (like rosenpass did) would put you in a good position. Are there already any public discussions, designs or transcripts?

common/nymnoise/src/config.rs
simonwicky: Fixed in f60a3177044849c6c78973a276cc59b957b01e94

common/nymnoise/src/lib.rs
simonwicky: Fixed in f60a3177044849c6c78973a276cc59b957b01e94

common/nymnoise/src/lib.rs
simonwicky: Fixed in f60a3177044849c6c78973a276cc59b957b01e94

common/nymnoise/src/stream.rs
simonwicky: Nice catch, I had in mind that `buf.remaining()` couldn't be 0.
Fixed in fc16dc6dd8a36ee4081293d26ee626e1909853bc along with a fix to the `Poll::Ready(None)` arm

common/nymnoise/src/stream.rs
simonwicky: Fixed in fc16dc6dd8a36ee4081293d26ee626e1909853bc

common/nymnoise/src/stream.rs
simonwicky: Added flushing afterwards in fc16dc6dd8a36ee4081293d26ee626e1909853bc

nym-api/src/nym_nodes/handlers/unstable/semi_skimmed.rs
jstuczyn: but it's not designed to be a noise-exlusive endpoint : )

common/nymnoise/src/config.rs
jstuczyn: > When adding/modifying patterns, I don't want to have that info in two places, that can lead to mismatch if not maintained properly.

that's why we write unit tests : )

jstuczyn: you can derive `EnumIter` on the guy and use it in tests to make sure every variant is always covered and returns expected value

common/nymnoise/src/lib.rs
jstuczyn: > Creating NoiseParams is fallible so then creating the config would be fallible.

sure, but we create config once, at startup. if that fails, node doesn't launch which is preferable imo

common/nymnoise/src/lib.rs
jstuczyn: actually, what'd be even better, because I see this is an option, is to use `as_bytes()` that doesn't make any allocation of the copy of the private key bytes : )

common/nymnoise/src/lib.rs
jstuczyn: sure, but it means building that whole `Handshake` is pointless, no?

common/nymnoise/src/stream.rs
jstuczyn: oh. Oh. i've just realised you've done a weird workaround to not use `.take()`
for readibility, can  you do 
```rust
        let Some(mut handshake) = self.handshake.take() else {
            return Err(NoiseError::IncorrectStateError);
        };
```

jstuczyn: also, you perhaps want to slap some `timeout` on that handshake. we don't want to be stuck here forever

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+use std::io;
+use std::pin::Pin;
+use std::task::Poll;
+use tokio::{
+    io::{AsyncRead, AsyncWrite, ReadBuf},
+    net::TcpStream,
+};
+use tokio_util::codec::{Framed, LengthDelimitedCodec};
+
+const MAXMSGLEN: usize = 65535;
```
jstuczyn: I've just noticed that both of those are defined inside `snow` instead. perhaps use those imports directly instead?

common/nymnoise/src/config.rs
jstuczyn: I'm not sure I understand. wouldn't using canonical address help us in reducing chances of accidentally having duplicates of the same underlying node?

common/nymnoise/src/stream.rs
jstuczyn: true

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+
+        let pending = match projected_self.inner_stream.poll_next(cx) {
+            Poll::Pending => {
+                //no new data, a return value of Poll::Pending means the waking is already scheduled
+                //Nothing new to decrypt, only check if we can return something from dec_storage, happens after
+                true
+            }
+
+            Poll::Ready(Some(Ok(noise_msg))) => {
+                //We have a new moise msg
```
jstuczyn:
```suggestion
                // We have a new noise msg
```

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+                let mut dec_msg = vec![0u8; MAXMSGLEN];
```
jstuczyn: I wonder, do we always have to allocate so much data per each message? i don't think that was the noise design

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+            None => Err(NoiseError::HandshakeError),
+        }
+    }
+
+    pub fn peer_addr(&self) -> Result<std::net::SocketAddr, io::Error> {
+        self.inner_stream.get_ref().peer_addr()
+    }
+}
+
+impl AsyncRead for NoiseStream {
```
jstuczyn: rather than re-implementing the wheel, wouldn't https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html work better?

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+                //Nothing new to decrypt, only check if we can return something from dec_storage, happens after
+                true
+            }
+
+            Poll::Ready(Some(Ok(noise_msg))) => {
+                //We have a new moise msg
+                let mut dec_msg = vec![0u8; MAXMSGLEN];
+                let len = match projected_self.noise {
+                    Some(transport_state) => {
+                        match transport_state.read_message(&noise_msg, &mut dec_msg) {
```
jstuczyn: why not just return those bytes immediately here

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+    ) -> Poll<Result<usize, std::io::Error>> {
+        let mut projected_self = self.project();
+
+        match projected_self.inner_stream.as_mut().poll_ready(cx) {
+            Poll::Pending => Poll::Pending,
+
+            Poll::Ready(Err(err)) => Poll::Ready(Err(err)),
+
+            Poll::Ready(Ok(())) => {
+                let mut noise_buf = BytesMut::zeroed(MAXMSGLEN + TAGLEN);
```
jstuczyn: why not reuse existing  buf or at least use its length? it's seems very wasteful to always allocate 60k for each tiny write...

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,202 @@
...
+        if pending {
+            //If we end up here, it means the previous poll_next was pending as well, hence waking is already scheduled
+            Poll::Pending
+        } else {
+            Poll::Ready(Ok(()))
+        }
+    }
+}
+
+impl AsyncWrite for NoiseStream {
```
jstuczyn: would https://docs.rs/tokio-util/latest/tokio_util/io/struct.SinkWriter.html work instead of implementing `AsyncWrite`? actually, why do we even need `AsyncWrite + AsyncRead`?

common/nymnoise/src/stream.rs
jstuczyn: this won't work as you wouldn't be able to flush it before sending is finished. you'd need some sort of state machine to keep track of it. i think. perhaps use one of tokio-util's adapters?

common/nymnoise/src/lib.rs
simonwicky: Tell that to the `snow` library author. I have to build it at some point, I could build it inside the `new` call though

common/nymnoise/src/stream.rs
simonwicky: My bad. I didn't know `take` existed when I wrote this lol.
There used to be a timeout, I need to find out what happened to it

common/nymnoise/src/stream.rs
simonwicky: The `write_message` call isn't sending anything. It's encrypting stuff. It's doing the handshake, so it's not encrypting any content (it's AEAD, but without the AD part).
I'm then truncating the buffer to the actual size of the message that will be sent, returned by `write_message`

common/nymnoise/src/stream.rs
simonwicky: Say what now ?!

common/nymnoise/src/stream.rs
simonwicky: Edit : Reading the code (from `snow` too) again, indeed allocating `noise_message.len() - TAGLEN` will suffice, fixed in ea233c78e2e6016939f2662762bd0e0397e18541

common/nymnoise/src/stream.rs
simonwicky: Maybe we have things in the buffer that needs to be returned before that. 
A call to this `poll_read` might end up with us having data we have decrypted but can't return yet. 
Hence we need to check it first before returning that part

common/nymnoise/src/stream.rs
simonwicky: In theory the writes can be up to 60k bytes though

common/nymnoise/src/stream.rs
simonwicky: Without Noise, we're wrapping the TcpStream into a `tokio_util::codec::Framed` with the sphinx packet codec. 
To code a one-for-one replacement, NoiseStream has to be `AsyncWrite + AsyncRead` in a way or another

simonwicky:
@timkuijsten There are no public discussions or designs yet no.
Using cookies is a good lead, although it's just used to confirm the IP address so it can then be limited with other means.

timkuijsten:
> Using cookies is a good lead, although it's just used to confirm the IP address so it can then be limited with other means.

But nym uses TCP right? Isn't TCP's three-way-handshake enough for IP ownership confirmation? (unlike WireGuard which needs the cookie because it uses UDP).

simonwicky:
@timkuijsten For the moment it does indeed. That's a fair point, I'll need to think about it.

common/nymnoise/src/stream.rs
simonwicky: It's in a private module though, we can't import it from there

nym-api/src/nym_nodes/handlers/unstable/semi_skimmed.rs
simonwicky: Added the option for later in d4d6e427777a51ec35edf942eea97faa0cf7413e
I didn't implement the `active` version of the endpoint though

common/nymnoise/src/config.rs
simonwicky: I added tests to ensure validity of all the noise pattern we hardcode in 090026d729cb993b020f220872edd9e1cb582d70
Let's keep the hardcoded position in the tests though

common/nymnoise/src/lib.rs
simonwicky: I added tests to make sure our hardcoded pattern can be safely unwrap when turned into a `NoiseParams`
090026d729cb993b020f220872edd9e1cb582d70

common/nymnoise/src/lib.rs
simonwicky: Added `as_bytes`  in 090026d729cb993b020f220872edd9e1cb582d70

common/nymnoise/src/lib.rs
simonwicky: I bundled the handshake object creation in the `Noisestream::new` call in 090026d729cb993b020f220872edd9e1cb582d70

common/nymnoise/src/stream.rs
simonwicky: `take` added in 090026d729cb993b020f220872edd9e1cb582d70. 
Timeout is WIP

common/nymnoise/src/stream.rs
simonwicky: fixed in 090026d729cb993b020f220872edd9e1cb582d70

common/nymnoise/src/stream.rs
simonwicky: Hold up, Tokio is doing exactly that in their SinkWriter implementation to go from a Sink to an AsyncWrite : 
https://docs.rs/tokio-util/latest/src/tokio_util/io/sink_writer.rs.html#104

common/nymnoise/src/stream.rs
simonwicky: Sure StreamReader can turn our underlying Stream into an AsyncRead, but it won't handle the encryption layer happening in the middle :(

common/nymnoise/src/stream.rs
simonwicky: SinkWriter doesn't work for the same reason StreamReader doesn't. We need to handle the encryption in the middle

common/nymnoise/src/stream.rs
simonwicky: Additional note : `write_message` requires the `buf` to be long enough, otherwise it returns an error. We're in the handshake here, so I will make it smaller

common/nymnoise/src/stream.rs
simonwicky: Edit : Reading the code (from `snow` too) again, indeed allocating `buf.len() + TAGLEN` will suffice, fixed in ea233c78e2e6016939f2662762bd0e0397e18541

common/nymnoise/src/stream.rs
simonwicky: Since allocation behavior changed, this isn't needed anymore

common/nymnoise/src/stream.rs
simonwicky: Smaller allocation in ea233c78e2e6016939f2662762bd0e0397e18541

common/nymnoise/src/stream.rs
simonwicky: Reusing the timeout duration from the connection in ea233c78e2e6016939f2662762bd0e0397e18541

simonwicky: Alternatively, we can also use a fixed one just like `WEBSOCKET_HANDSHAKE_TIMEOUT`. (I saw that just now)

common/client-libs/mixnet-client/src/client.rs
```diff
@@ -132,9 +138,21 @@ impl ManagedConnection {
...
                     debug!("Managed to establish connection to {}", self.address);
-                    // if we managed to connect, reset the reconnection count (whatever it might have been)
+
+                    let noise_stream =
+                        match upgrade_noise_initiator(stream, &self.noise_config).await {
+                            Ok(noise_stream) => noise_stream,
+                            Err(err) => {
+                                error!("Failed to perform Noise handshake with {address} - {err}");
+                                // we failed to finish the noise handshake - increase reconnection attempt
+                                self.current_reconnection.fetch_add(1, Ordering::SeqCst);
```
jstuczyn, line 141: wouldn't we end up in a constant reconnection loop if receiver doesn't support noise?

common/nymnoise/src/config.rs
jstuczyn: but why not also use constant for the `psk_position` itself xD (I guess I really don't like unwraps that could be removed)

common/nymnoise/src/stream.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub(crate) async fn perform_handshake(mut self) -> Result<Self, NoiseError> {
+        //Check if we are in the correct state
+        let Some(mut handshake) = self.handshake.take() else {
+            return Err(NoiseError::IncorrectStateError);
+        };
+
+        while !handshake.is_handshake_finished() {
+            if handshake.is_my_turn() {
+                self.send_handshake_msg(&mut handshake).await?;
```
jstuczyn: looking slightly deeper into it. i'm a bit confused: why are you having a method on `self` that takes `handshake` argument that originally was part of self but you temporarily removed it?

common/nymnoise/src/config.rs
simonwicky: I reckon having the automated construction somewhere and the constants somewhere else is a good thing to ensure validity. 
I'm team automated construction in the actual code and constants in the tests, if the tests pass, those `unwrap`s cannot fail. 
Alternatively, I can remove them and make the `fn` fallible, since it's already used in a fallible function. But again, if the tests pass, they can't fail

common/nymnoise/src/stream.rs
simonwicky: We still need the `inner_stream` from `self` to send the encrypted handshake message.
I'm taking the `handshake` out from `self` to prevent duplicate calls to this `fn` from messing with the first one.

common/client-libs/mixnet-client/src/client.rs
simonwicky: The error message is a bit misleading tbh, my bad. 
If the receiver doesn't support Noise, there won't be any handshake, and we will just return the `TcpStream` wrapped in a `Connection` for compatibility

jstuczyn (approved):

PR #5696, opened 2025-04-08
Make mix hops optional for Mixnet Client

jmwample:
As is the route selection for packets, reply SURBs, and acknowledgements are selected deep withing the `MixnetClient` machinery. This makes custom route selection (i.e. for authenticating / registering wiregaurd mode vpn clients) is not supported.

This PR adds configuration toggle that allows `MixnetClient` to be built where packets are sent direct through entry to egress gateway nodes -- no mix hops. 

The intended usage in the VPN is something like this:
```rust
    let mut debug_config = nym_client_core::config::DebugConfig::default();
    if two_hop_mode {
        // If operating in two hop mode, we disable mix hops for the mixnet connection.
        debug_config.traffic.disable_mix_hops = true;
    }

    //...

    let builder = MixnetClientBuilder::new_with_storage(storage)
        .debug_config(debug_config)
        .build()
        .map_err(MixnetError::FailedToBuildMixnetClient)?
        .connect_to_mixnet()
        .await
        .expect("connection failed");
```

Notes:
* for now SURB messages are still constructed such that they will be routed though mix hops, as the packet format / parsing currently does not allow for variable route lengths.
* This does not add the ability for fully custom route selection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5696)

### DIFF
--- a/common/client-core/config-types/src/lib.rs
+++ b/common/client-core/config-types/src/lib.rs
@@ -414,6 +414,12 @@ pub struct Traffic {
     pub use_legacy_sphinx_format: bool,
 
     pub packet_type: PacketType,
+
+    /// Indicates whether to mix hops or not. If mix hops are enabled, traffic
+    /// will be routed as usual, to the entry gateway, through three mix nodes, egressing
+    /// through the exit gateway. If mix hops are disabled, traffic will be routed directly
+    /// from the entry gateway to the exit gateway, bypassing the mix nodes.
+    pub disable_mix_hops: bool,
 }
 
 impl Traffic {
@@ -444,6 +450,7 @@ impl Default for Traffic {
             // we should use the legacy format until sufficient number of nodes understand the
             // improved encoding
             use_legacy_sphinx_format: true,
+            disable_mix_hops: false,
         }
     }
 }
--- a/common/nymsphinx/src/preparer/mod.rs
+++ b/common/nymsphinx/src/preparer/mod.rs
@@ -3,7 +3,6 @@
 
 use crate::message::{NymMessage, ACK_OVERHEAD, OUTFOX_ACK_OVERHEAD};
 use crate::NymPayloadBuilder;
-use log::debug;
 use nym_crypto::asymmetric::x25519;
 use nym_crypto::Digest;
 use nym_sphinx_acknowledgements::surb_ack::SurbAck;
@@ -19,6 +18,7 @@ use nym_sphinx_types::{Delay, NymPacket};
 use nym_topology::{NymRouteProvider, NymTopologyError};
 use rand::{CryptoRng, Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
+use tracing::*;
 
 use nym_sphinx_chunking::monitoring;
 use std::time::Duration;
@@ -52,6 +52,10 @@ pub trait FragmentPreparer {
     type Rng: CryptoRng + Rng;
 
     fn use_legacy_sphinx_format(&self) -> bool;
+    fn mix_hops_disabled(&self) -> bool {
+        // Unless otherwise configured, mix hops are enabled
+        false
+    }
 
     fn deterministic_route_selection(&self) -> bool;
     fn rng(&mut self) -> &mut Self::Rng;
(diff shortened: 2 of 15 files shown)


common/topology/src/lib.rs
```diff
@@ -176,6 +176,18 @@ impl NymRouteProvider {
             .random_route_to_egress(rng, egress_identity, self.ignore_egress_epoch_roles)
     }
 
+    /// Tries to create a route to the egress point, such that it goes through mixnode on layer 1,
```
jstuczyn: I don't think this comment is accurate as you're not actually going through any mixnodes. just straight into egress

common/client-core/config-types/src/lib.rs
```diff
@@ -711,6 +711,12 @@ pub struct DebugConfig {
 
     /// Defines all configuration options related to the forget me flag.
     pub forget_me: ForgetMe,
+
+    /// Indicates whether to mix hops or not. If mix hops are enabled, traffic
+    /// will be routed as usual, to the entry gateway, through three mix nodes, egressing
+    /// through the exit gateway. If mix hops are disabled, traffic will be routed directly
+    /// from the entry gateway to the exit gateway, bypassing the mix nodes.
+    pub disable_mix_hops: bool,
```
jstuczyn: I think it'd make more sense if it was part of `Traffic` section rather than being a generic setting. 
this would also give us explicit lack of control over cover traffic which would always send full packets

common/nymsphinx/src/preparer/mod.rs
```diff
@@ -59,28 +58,38 @@ pub trait FragmentPreparer {
...
-            use_legacy_sphinx_format,
-            recipient,
-            ack_key,
-            fragment_id.to_bytes(),
-            ack_delay,
-            topology,
-            packet_type,
-        )
-    }
+    ) -> Result<SurbAck, NymTopologyError>;
```
jstuczyn: I don't quite understand - why did you remove default impl from the trait?

jstuczyn: (the same comment applies to other methods too)

common/nymsphinx/src/preparer/mod.rs
```diff
@@ -362,108 +282,183 @@ where
...
+        } else {
+            // generate pseudorandom route for the packet
+            topology.random_route_to_egress(self.rng(), destination)?
+        };
+
+        let destination = packet_recipient.as_sphinx_destination();
+
+        // including set of delays
+        let delays =
+            nym_sphinx_routing::generate_hop_delays(self.average_packet_delay(), route.len());
```
jstuczyn: this is a side note but makes me wonder whether there's a bug in delay generation. it's because we're generating 1 delay here, but, when egress receives the packet and unwraps the header, the delay information wouldn't be there. the whole content would be immediately forwarded to the recipient... 🤔

jstuczyn (approved):

common/nymsphinx/src/preparer/mod.rs
jmwample: The trait implementation doesn't have access to the fields of the type implementing it so I moved to the implementation to the type. However, now looking at some other related PRs it looks like maybe you would prefer just adding another accessor method to the trait so that the impl can stay in the trait itself?

In general I am confused by the use of the trait here in the first place. We are not using the `FragmentPreparer`  trait anywhere else so is the point just to allow some overriding in a mock version in a foreign crate?  (i.e. `node-tester-utils/srs/tester.rs`)

common/topology/src/lib.rs
jmwample: Fixed

common/client-core/config-types/src/lib.rs
jmwample: Added to the `Traffic` config object

common/nymsphinx/src/preparer/mod.rs
jstuczyn: I think it might have been some legacy leftover - feel free to try to remove the trait. if it works without it, great!

jmwample: I just added another accessor method for the path of least resistance here (and least code review)

common/nymsphinx/src/preparer/mod.rs
jmwample: I don't know the answer to this, is this worth opening an issue? Not sure this is directly related to this PR

PR #5698, opened 2025-04-09
Feature/updated sphinx payload keys

jstuczyn:
This PR uses the new version of the sphinx-packet that allows us to derive payload encryption keys from the seed rather than having to attach the keys themselves. what it means in practice is that each reply surb is now ~60% smaller.

However, while this sounds amazing, currently all of those functionalities are **DISABLED** by default. it is because it required nodes to actually understand the new scheme. it shouldn't be enabled until sufficient number of nodes, particularly mixnodes, has upgraded. I reckon it will take few releases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5698)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -317,7 +317,7 @@ serde_with = "3.9.0"
 serde_yaml = "0.9.25"
 sha2 = "0.10.8"
 si-scale = "0.2.3"
-sphinx-packet = "=0.5.0"
+sphinx-packet = "0.5.0"
 sqlx = "0.7.4"
 strum = "0.26"
 strum_macros = "0.26"
@@ -406,6 +406,10 @@ wasm-bindgen = "0.2.99"
 wasm-bindgen-futures = "0.4.49"
 wasmtimer = "0.4.1"
 web-sys = "0.3.76"
+
+
+[patch.crates-io]
+sphinx-packet = { path = "../sphinx" }
 
 # Profile settings for individual crates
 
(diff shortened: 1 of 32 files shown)


Cargo.toml
```diff
@@ -407,6 +407,11 @@ wasm-bindgen-futures = "0.4.49"
 wasmtimer = "0.4.1"
 web-sys = "0.3.76"
 
+
+# for local development:
+#[patch.crates-io]
+#sphinx-packet = { path = "../sphinx" }
+
```
simonwicky, line 414: Did you mean for this to stay there for future dev work? Or is it just a miss?

common/nymsphinx/anonymous-replies/Cargo.toml
```diff
@@ -19,6 +19,7 @@ nym-sphinx-params = { path = "../params" }
 nym-sphinx-routing = { path = "../routing" }
 nym-sphinx-types = { path = "../types" }
 nym-topology = { path = "../../topology" }
+log = "0.4.27"
```
simonwicky: No `workspace = true` here ?

simonwicky (approved):

common/nymsphinx/anonymous-replies/Cargo.toml
jstuczyn: huh. good catch. my IDE probably decided it knows better and it imported it. if anything it should be using `tracing`...

Cargo.toml
jstuczyn: left it commented out, similarly to what we're doing in the vpn core repo, so that it's easy to swap the used crate (I didn't remember the correct syntax from the top of my head : P )

PR #5706, opened 2025-04-11
Yana/explorer caching

yanok87:
Add moniker to Magic Search

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5706)

### DIFF
--- a/explorer-v2/src/app/api/urls.ts
+++ b/explorer-v2/src/app/api/urls.ts
@@ -2,8 +2,7 @@ export const CURRENT_EPOCH =
   "https://validator.nymtech.net/api/v1/epoch/current";
 export const CURRENT_EPOCH_REWARDS =
   "https://validator.nymtech.net/api/v1/epoch/reward_params";
-export const NYM_NODE_BONDED =
-  "https://validator.nymtech.net/api/v1/nym-nodes/bonded";
+
 export const NYM_ACCOUNT_ADDRESS =
   "https://explorer.nymtech.net/api/v1/tmp/unstable/account";
 export const NYM_PRICES_API = "https://api.nym.spectredao.net/api/v1/nym-price";
@@ -18,3 +17,8 @@ export const DATA_OBSERVATORY_BALANCES_URL =
   "https://api.nym.spectredao.net/api/v1/balances";
 export const OBSERVATORY_GATEWAYS_URL =
   "https://mainnet-node-status-api.nymtech.cc/v2/gateways";
+
+export const NS_API_MIXNODES_STATS =
+  process.env.NEXT_PUBLIC_NS_API_MIXNODES_STATS;
+
+export const NS_API_NODES = process.env.NEXT_PUBLIC_NS_API_NODES;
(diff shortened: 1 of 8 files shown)


explorer-v2/src/app/api/urls.ts
```diff
@@ -18,3 +17,8 @@ export const DATA_OBSERVATORY_BALANCES_URL =
   "https://api.nym.spectredao.net/api/v1/balances";
 export const OBSERVATORY_GATEWAYS_URL =
   "https://mainnet-node-status-api.nymtech.cc/v2/gateways";
+
```
windy-ux, line 20: I think that this is not actually needed to be exported as this variable is accessible everywhere across nextjs through that 'process.env' variable
Might be cleaner to just import it from there on the place you need.
const { NS_API_MIXNODES_STATS } = process.env

explorer-v2/src/components/copyToClipboard/CopyToClipboard.tsx
```diff
@@ -33,16 +34,19 @@ const CopyToClipboard = ({
 
   if (hasCopied) {
     return (
-      <Typography sx={{ color: "pine.950" }} variant="h6" color="textSecondary">
+      <Typography
+        sx={{ color: isDarkMode ? "base.white" : "pine.950" }}
```
windy-ux, line 38: Would be better to use some generic color definition referring to its usage instead of writing color directly

explorer-v2/src/components/landingPageComponents/NoiseCard.tsx
```diff
@@ -124,11 +135,22 @@ export const NoiseCard = () => {
...
+          open={tooltipOpen}
+          onClose={handleTooltipClose}
           onClick={(e) => e.stopPropagation()}
+          enterNextDelay={300}
+          leaveDelay={200}
         >
-          <Typography variant="h4" sx={{ color: "#8482FD", cursor: "pointer" }}>
+          <Typography
+            variant="h4"
+            sx={{ color: "#8482FD", cursor: "pointer" }}
```
windy-ux, line 148: Same for here, using general usage of that color rather then staticaly written.

explorer-v2/src/components/search/NodeAndAddressSearch.tsx
```diff
@@ -33,64 +50,164 @@ const NodeAndAddressSearch = () => {
...
             setIsLoading(false); // Stop loading
 
             return;
           }
         } else {
           setErrorText(
-            "It seems that this node or account does not exist. Please enter a complete Node ID or an existing Nym wallet address.",
+            "No node found with the provided Name, Node ID or Identity Key. Please check your input and try again."
           );
           setIsLoading(false); // Stop loading
```
windy-ux, line 63: In the future it would be good to change loading logic into using suspense https://relay.dev/docs/guided-tour/rendering/loading-states/

windy-ux (approved):

explorer-v2/src/components/landingPageComponents/NoiseCard.tsx
yanok87: this one is not a standard color to add to the theme

explorer-v2/src/components/copyToClipboard/CopyToClipboard.tsx
yanok87: yes, it's just the colors for h6 vary through the designs..

explorer-v2/src/app/api/urls.ts
yanok87: yes, you're right! But I want to have all endpoints in one file, otherwise it's not practical to search them through the app.. Has costed me now one forgotten endpoint for Dinko to replace

PR #5728, opened 2025-04-22
Eliminate duplicate node_ids from endpoint

dynco-nym:
- sort returned nodes on `/nym-nodes` so that their order is deterministic
  - otherwise, it could potentially return the same node in different pages of pagination, if requesting this endpoint several times in succession
- store gateway descriptions to nym_nodes table so the info can be retrieved from nym-nodes endpoint as well
- improve swagger type definitions
- log every HTTP request

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5728)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6269,6 +6269,7 @@ dependencies = [
  "nym-contracts-common",
  "nym-crypto",
  "nym-http-api-client",
+ "nym-http-api-common",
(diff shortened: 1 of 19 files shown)


pwnfoo (approved):
LGTM 🛳️

pwnfoo:
ns-api probably needs a version bump?

PR #5742, opened 2025-05-06
[Feature] `RememberMe` is the new don't `ForgetMe`

simonwicky:
# Description

Sessions stats collected from gateway proved themselves to be incredibly noisy. In an effort to reduce the noise, `ForgetMe` was introduced at the end of sessions to prompt gateways to forget a specific session. 
This flag helped reduce the amount of noise, but alone it's not enough (e.g. `ForgetMe` can't be sent by non-sdk client)

Therefore we're turning the whole thing and around and introducing a `RememberMe` flag, to prompt the gateways to actually remember said session.

## What this PR does
- Disable the `stats` part of the `ForgetMe` flag. It's still there for compatibility, but it now has no effect
- Introduce a `RememberMe` message, that gets sent at the end of the connection by SDK clients. It include a session type to tag said session as well
- Change collection logic on gateways : The new default is to **NOT** record any session data, unless a `RememberMe` message was received. In that case, the session will get its type and be recorded. The data being served on the self-described API doesn't change

## What this PR doesn't do
- This PR doesn't allow native and socks5 client to use that `RememberMe` flag. It's impossible to send something right before stopping the process, since parts needed to do this might have already shut down on their own. This is already the case of the `ForgetMe` flag

### Once this is merged

The vpn-client code should use the `with_remember_me` option when building a MixnetClient, once it's gonna use a nym-core version that contains this PR
https://github.com/nymtech/nym-vpn-client/pull/2737

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5742)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6255,7 +6255,7 @@ dependencies = [
 
 [[package]]
 name = "nym-node-status-api"
-version = "2.3.0"
+version = "2.3.1"
 dependencies = [
  "ammonia",
  "anyhow",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5176,6 +5176,7 @@ dependencies = [
  "nym-pemstore",
  "nym-sphinx-addressing",
  "nym-sphinx-params",
+ "nym-statistics-common",
  "serde",
  "thiserror 2.0.12",
  "url",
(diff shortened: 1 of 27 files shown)


durch:
We keep adding complexity to address design issues. I don't mind merging this in now, but lets please schedule proper work around this, I'd really like to understand the data collection requirements, and then figure out how to get the data, I'm pretty sure that we can land on a simpler solution but it will require some design

durch (approved):
https://github.com/nymtech/nym/pull/5742#issuecomment-2866324348

PR #5745, opened 2025-05-07
feat: nym-api bincode + yaml support

jstuczyn:
this PR introduces support for nym-api to return responses as either bincode or yaml (on top of existing, default, json). furthermore, initial client support for nodes-related queries is introduced. further endpoints can be enhanced by simply pushing `("output", "bincode")` params.

here is the sample response size difference when using json vs bincode (with and without gzip enabled on top of it):
| endpoint                                            | json    | json + gzip | bincode | bincode + gzip |
| --------------------------------------------------- | ------- | ----------- | ------- | -------------- |
| `/v1/nym-nodes/described`                           | 882,684 | 204,265     | 437,204 | 192,752        |
| `/v1/unstable/nym-nodes/skimmed/entry-gateways/all` | 116,268 | 27,991      | 39,755  | 25,589         |
| `/v1/unstable/nym-nodes/skimmed/mixnodes/active`    | 44,646  | 11,371      | 14,021  | 10,235         |
| `/v1/ecash/aggregated-coin-indices-signatures`      | 11,669  | 6,037       | 4,952   | 4,980          |
| `/v1/network/chain-status`                          | 1,523   | 828         | 1,004   | 585            |
| `/v1/nym-nodes/rewarded-set`                        | 1,221   | 616         | 679     | 529            |

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5745)

### DIFF
--- a/common/http-api-client/Cargo.toml
+++ b/common/http-api-client/Cargo.toml
@@ -12,6 +12,7 @@ license.workspace = true
 
 [dependencies]
 async-trait = { workspace = true }
+bincode = { workspace = true, optional = true }
 reqwest = { workspace = true, features = ["json", "gzip"] }
 http.workspace = true
 url = { workspace = true }
@@ -26,7 +27,7 @@ bytes = { workspace = true }
 encoding_rs = { workspace = true }
 mime = { workspace = true }
 
-
+nym-http-api-common = { path = "../http-api-common", default-features = false, optional = true }
 nym-bin-common = { path = "../bin-common" }
 
 [target."cfg(not(target_arch = \"wasm32\"))".dependencies]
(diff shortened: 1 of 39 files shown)


common/http-api-client/Cargo.toml
```diff
@@ -39,3 +40,6 @@ features = ["tokio"]
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt", "macros"] }
+
+[features]
+bincode = ["dep:bincode", "nym-http-api-common", "nym-http-api-common/bincode"]
```
octol, line 45: why not consistently use `dep:`?

octol (approved):
very nice! :)

common/http-api-client/Cargo.toml
jstuczyn: wanted to only add the prefix for the external dependency that was the source of the name clash, but I have no strong feelings about it

PR #5755, opened 2025-05-12
feat: expires header for `/active` nym-api responses

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5755)


common/http-api-common/src/response/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            }
+            #[cfg(feature = "bincode")]
+            FormattedResponse::Bincode(inner) => {
+                FormattedResponse::Bincode(inner.with_header(name, value))
+            }
+        }
+    }
+
+    #[must_use]
+    pub fn with_expires_header(self, expiration: OffsetDateTime) -> FormattedResponse<T> {
```
octol, line 91: maybe also a variant of this that takes a time delta?

octol (approved):

PR #5756, opened 2025-05-12
Raw route submissions

durch:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5756)

### DIFF
--- a/nym-network-monitor/src/accounting.rs
+++ b/nym-network-monitor/src/accounting.rs
@@ -486,11 +486,8 @@ async fn submit_accounting_routes_to_db(client: Arc<Client>) -> anyhow::Result<(
 }
 
 pub async fn submit_metrics(database_url: Option<&String>) -> anyhow::Result<()> {
-    match submit_metrics_to_db(database_url).await {
-        Ok(_) => (),
-        Err(e) => {
-            error!("Error submitting metrics to db: {}", e);
-        }
+    if let Err(e) = submit_metrics_to_db(database_url).await {
+        error!("Error submitting metrics to db: {}", e);
     }
 
     if let Some(private_key) = PRIVATE_KEY.get() {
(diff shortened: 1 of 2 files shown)


nym-network-monitor/src/accounting.rs
```diff
@@ -384,17 +424,73 @@ async fn submit_gateway_stats_to_db(client: Arc<Client>) -> anyhow::Result<()> {
...
+    writer.finish().await?;
+
     Ok(())
 }
 
-pub async fn submit_metrics(client: Option<Arc<Client>>) -> anyhow::Result<()> {
-    if let Some(client) = client {
-        submit_metrics_to_db(client).await?;
+pub async fn submit_metrics(database_url: Option<&String>) -> anyhow::Result<()> {
+    match submit_metrics_to_db(database_url).await {
```
jstuczyn: nit: (because i reckon clippy will complain about it), you could have instead used the `if Err(err) = submit_metrics_to_db(...)` pattern

nym-network-monitor/src/accounting.rs
```diff
@@ -384,17 +424,73 @@ async fn submit_gateway_stats_to_db(client: Arc<Client>) -> anyhow::Result<()> {
...
+        Ok(Some((client, handle)))
+    } else {
+        Ok(None)
+    }
+}
+pub async fn submit_metrics_to_db(database_url: Option<&String>) -> anyhow::Result<()> {
+    if let Some((client, handle)) = db_connection(database_url).await? {
+        let client = Arc::new(client);
+        let client2 = Arc::clone(&client);
+        let client3 = Arc::clone(&client);
```
jstuczyn, line 446: why are you cloning all those pointers? I'm not sure you have to (unless you wanted to execute all of those submissions in parallel)

jstuczyn (approved):

nym-network-monitor/src/accounting.rs
durch: Since all of them take an Arc its possible I wanted to execute them in parallel, maybe DB was complaining, interesting

durch: Arc is implicitly moved always, hence the cloning, ie just passing the same Arc will not work

PR #5759, opened 2025-05-14
Add node_bonded field to delegations

dynco-nym:
- clarifies whether the delegation is to a bonded or unbonded node
- include delegations to unbonded nodes in the returned list

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5759)

### DIFF
--- a/nym-api/src/unstable_routes/account/data_collector.rs
+++ b/nym-api/src/unstable_routes/account/data_collector.rs
@@ -87,10 +87,10 @@ impl AddressDataCollector {
                 if let Some(delegation) = og_delegations.get(&node_details.node_id()) {
                     node_delegation_info.push(AddressDelegationInfo {
                         details: delegation.clone(),
-                        node_reward_info: NodeBondStatus::Bonded(
-                            node_details.rewarding_details.to_owned(),
-                            node_details.is_unbonding(),
-                        ),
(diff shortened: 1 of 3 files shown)


nym-api/src/unstable_routes/account/data_collector.rs
```diff
@@ -171,24 +181,33 @@ impl AddressDataCollector {
...
+
+    pub(crate) fn is_node_bonded(&self) -> bool {
+        matches!(self.node_reward_info, NodeBondStatus::Bonded(_, _))
     }
 }
 
-type RewardAndBondInfo = (NodeRewarding, bool);
+pub(crate) enum NodeBondStatus {
+    /// bool: is node in the process of unbonding
+    Bonded(NodeRewarding, bool),
```
jstuczyn: if you have to document what a random bool does, why not just use proper field names instead, i.e.
```rust
pub(crate) enum NodeBondStatus {
  Bonded {
    rewarding_info: NodeRewarding,
    unbonding: bool,
  },
  Unbonded,
}
```

jstuczyn (approved):

nym-api/src/unstable_routes/account/data_collector.rs
dynco-nym: implemented

PR #5763, opened 2025-05-15
NMV2 metrics fanout and corrections

durch:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5763)

 Summary

  This PR enhances the Network Monitor with a "metrics fanout" capability, allowing monitoring metrics to be submitted to multiple API endpoints concurrently. It also introduces a new database schema for storing detailed
  route performance data to improve network reliability tracking and analysis.

  Key Features

  - Metrics Fanout System: Submit monitoring data to multiple `nym-api` instances simultaneously
  - Route Performance Tracking: Store route metrics in a new database table

  Technical Changes

  - Added support for multiple `NYM_API` endpoints via the `nym_apis` CLI parameter and `NYM_APIS` environment variable
  - Created new routes table schema in the database to store detailed route metrics:
    - Records node IDs for each layer (layer1, layer2, layer3, gateway)
    - Tracks success/failure status for each packet route
    - Includes created_at timestamps, would be better if these were received at, but out of scope for now
    - Added appropriate indexes for efficient querying
  - Implemented parallel submission of route data to all configured API endpoints using `FuturesUnordered`
  - Added support for efficient binary data copying to the database for metrics storage
  - Enhanced error handling and logging for metrics submission

  Testing Notes

  - The metrics fanout functionality has been tested against multiple API endpoints to ensure proper data submission
  - Route performance data can be verified in the database by querying the new routes table
  - The implementation includes proper error handling to ensure metrics submission continues even if one endpoint fails
  - Performance impact is minimal as metrics submission happens in parallel using asynchronous requests

  Deployment Considerations

  - To use this feature, configure multiple API endpoints using either:
    - Command-line arguments: --nym-apis <API_URL1> <API_URL2> ...
    - Environment variable: NYM_APIS=<API_URL1>,<API_URL2>,...
  - Database schema will need to be updated using the new migration (20250513104800_routes_table.sql)
  - No changes to existing functionality or backwards compatibility issues
  - For optimal performance, API endpoints should be geographically distributed

  This enhancement significantly improves the robustness of the Nym Network Monitor by allowing redundant metric submission to multiple endpoints, ensuring that network performance data is properly captured and available for
  analysis even if individual API endpoints experience issues. The detailed route tracking will also provide valuable insights into network performance patterns and help identify potential bottlenecks.


nym-api/migrations/20250513104800_routes_table.sql
```diff
@@ -0,0 +1,17 @@
+-- Add routes table for storing route metrics data
+CREATE TABLE IF NOT EXISTS routes (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    layer1 INTEGER NOT NULL,   -- NodeId of layer 1 mixnode
+    layer2 INTEGER NOT NULL,   -- NodeId of layer 2 mixnode
+    layer3 INTEGER NOT NULL,   -- NodeId of layer 3 mixnode
+    gw INTEGER NOT NULL,       -- NodeId of gateway
+    success BOOLEAN NOT NULL,  -- Whether the packet was delivered successfully
+    timestamp INTEGER NOT NULL DEFAULT (unixepoch()) -- When the measurement was taken
```
jstuczyn, line 9: nit: if possible, perhaps store it with explicit type for easier queries, i.e. `TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP` so that it could be coerced into `OffsetDateTime` with sqlx

nym-api/src/node_status_api/handlers/without_monitor.rs
```diff
@@ -87,18 +138,21 @@ pub(crate) async fn submit_gateway_monitoring_results(
...
-        .submit_gateway_statuses_v2(message.results())
-        .await
-    {
-        Ok(_) => Ok(()),
-        Err(err) => {
-            error!("failed to submit gateway monitoring results: {err}");
-            Err(AxumErrorResponse::internal_msg(
-                "failed to submit gateway monitoring results",
-            ))
+    match message.results() {
```
jstuczyn, line 141: I guess this is out of scope for this PR, but just wondering, does it still make sense to have separate results for gateways and mixnodes?

nym-api/src/support/storage/manager.rs
```diff
@@ -51,6 +52,25 @@ impl AvgGatewayReliability {
...
+#[derive(Debug, Default, Clone)]
+struct NodeCorrectionIntervalState {
+    pos_samples: u32,
+    neg_samples: u32,
+    fail_seq: u32,
+}
+
+// Output struct for the calculated corrected reliability for an interval
+#[derive(Debug, serde::Serialize)] // Add utoipa::ToSchema if this struct is directly exposed via API
+pub struct CorrectedNodeIntervalReliability {
```
jstuczyn, line 65: should the correction be done by nym  api or rather NM? (ideally, in stage 2, NM would be submitting corrected results to the contract)

nym-api/src/support/storage/manager.rs
```diff
@@ -51,6 +52,25 @@ impl AvgGatewayReliability {
...
+    pub identity: Option<String>, // Base58 public key
+    pub reliability: f64,
+    pub pos_samples_in_interval: u32,
+    pub neg_samples_in_interval: u32,
+    pub final_fail_seq_in_interval: u32,
+}
+
 // all SQL goes here
 impl StorageManager {
     pub(super) async fn get_all_avg_mix_reliability_in_last_24hr(
```
jstuczyn, line 76: another nit: imo `StorageManager` should be concerned with just SQL queries and the logic for calculating all this stuff should go up one (or maybe even two) layers of abstractions. either `NymApiStorage` or maybe even above it

nym-api/src/support/storage/manager.rs
```diff
@@ -1329,6 +1386,146 @@ impl StorageManager {
...
+
+        Ok(db_routes
+            .into_iter()
+            .map(|r| {
+                (
+                    r.layer1 as NodeId,
+                    r.layer2 as NodeId,
+                    r.layer3 as NodeId,
+                    r.gw as NodeId,
+                    r.success.unwrap_or_default(),
```
jstuczyn, line 1439: why would success be missing? can't we set it to `NOT NULL` and then avoid this whole iterator?

nym-api/src/support/storage/manager.rs
```diff
@@ -1329,6 +1386,146 @@ impl StorageManager {
...
+                for &node_id in &path_node_ids {
+                    let state = node_states.entry(node_id).or_default();
+                    state.fail_seq += 1;
+                    current_path_node_ids_for_blame.push(node_id);
+                }
+
+                let mut guilty_nodes_in_path: Vec<NodeId> = Vec::new();
+                for &node_id in &current_path_node_ids_for_blame {
+                    if let Some(state_after_update) = node_states.get(&node_id) {
+                        if state_after_update.fail_seq > 2 {
```
jstuczyn, line 1477: can you start distributing the blame before you accumulated all results?

nym-api/src/support/storage/manager.rs
```diff
@@ -1329,6 +1386,146 @@ impl StorageManager {
...
+            let total_samples = state.pos_samples + state.neg_samples;
+            let reliability = if total_samples == 0 {
+                0.0 // Default for no samples in this interval. Consider Option<f64> or filtering.
+            } else {
+                state.pos_samples as f64 / total_samples as f64
+            };
+
+            // Attempt to fetch identity, first as mixnode, then as gateway if not found.
+            // This assumes get_mixnode_identity_key and get_gateway_identity_key return Result<Option<String>, sqlx::Error>
+            let mut identity: Option<String> =
```
jstuczyn, line 1511: do we even care about identity key at this point?

nym-network-monitor/src/main.rs
```diff
@@ -172,6 +172,18 @@ async fn nym_topology_from_env() -> anyhow::Result<NymTopology> {
     let mut topology = NymTopology::new_empty(rewarded_set);
     topology.add_skimmed_nodes(&nodes);
 
+    let node_ids = topology.node_details().keys().cloned().collect::<Vec<_>>();
```
jstuczyn: that's the only **NOT NIT** I've noticed: you DON'T WANT to be doing this. because `node_details` will also give you **GATEWAYS** that can't operate as mixnodes

nym-network-monitor/src/accounting.rs
```diff
@@ -491,49 +494,100 @@ pub async fn submit_metrics(database_url: Option<&String>) -> anyhow::Result<()>
...
-                    PRIVATE_KEY.get().expect("We've set this!"),
+        if let Some(nym_api_urls) = NYM_API_URLS.get() {
+            info!("Submitting metrics to {} nym apis", nym_api_urls.len());
+            for nym_api_url in nym_api_urls {
+                info!("Submitting metrics to {}", nym_api_url);
+                let node_stats = monitor_mixnode_results().await?;
+                let gateway_stats = monitor_gateway_results().await?;
+                let client = reqwest::Client::new();
+
+                let node_submit_url =
```
jstuczyn, line 505: rather than using hardcoded strings, can't we just construct a `NymApiClient` and call methods on that guy instead?

jstuczyn (changes requested):
almost entirely nits apart from one thing:
`let node_ids = topology.node_details().keys().cloned().collect::<Vec<_>>();`
^ this will not work the way you want as it also retrieves gateways (which also have to be tested, but not as mixnodes : D )

nym-api/src/support/storage/manager.rs
durch: The thing is that corrections need a lot of results to be useful, and NMs submit every 5 or 10 minutes, or similar cadence. So I've pushed the correction as close to the actual usage of the results as possible otherwise it would not be doing much

nym-api/src/support/storage/manager.rs
durch: I don't know, just being paranoid, if we set it to NOT NULL then inserts might start failing quietly if we have a problem somewhere else, better be explicit then sorry :)

nym-api/src/support/storage/manager.rs
durch: A lot of this is legacy, I think we just clean it all up once we get to the performance contract

nym-network-monitor/src/main.rs
durch: Nice one, will look into chaining a filter in here, hopefully that info is in there

PR #5777, opened 2025-05-16
feat: key rotation

jstuczyn:
that one was definitely another ecach-size adventure. you can find detailed description on our [confluence page](https://nymtech.atlassian.net/wiki/spaces/KB/pages/139853839/Sphinx+key+rotation)

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5777)


durch (approved):
This is all very well thought out and executed. I've got to conceptual comments, or weak points, but nothing that we can address, more something to be vigilant about:

1. The config upgrade process, particularly upgrade_sphinx_key, involves network calls and multiple file operations. A failure mid-process could leave the node in a partially upgraded or broken state.
2. The KeyRotationController's  error handling for failed key/filter operations (that don't poison a mutex) primarily relies on logging.  Depending on the failure, this might lead to the node operating with inconsistent keys/filters until the next successful cycle.

jstuczyn:
>     1. The config upgrade process, particularly upgrade_sphinx_key, involves network calls and multiple file operations. A failure mid-process could leave the node in a partially upgraded or broken state.

I've slightly adjusted order of operations so that if the network call fails, nothing else gets done. As for IO failures - I'm not sure we can do much about them there. But it shouldn't be unrecoverable. Old keys are not deleted until new ones have been saved. And you won't receive "upgraded" config unless everything has succeeeded. Worst case scenario, you might have to repeat the process

> 
>     2. The KeyRotationController's  error handling for failed key/filter operations (that don't poison a mutex) primarily relies on logging.  Depending on the failure, this might lead to the node operating with inconsistent keys/filters until the next successful cycle.

That's a tricky one that I'm not sure if it has any good solution. Because if an operation fails and we miss it, we're already in an incocistent state and the only thing we could do is wait to retry...

PR #5778, opened 2025-05-16
Clippy

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5778)

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,3 +1,5 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
 allow-panic-in-tests = true
+large-error-threshold = 196
+enum-variant-size-threshold = 1000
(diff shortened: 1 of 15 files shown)


clippy.toml
```diff
@@ -1,3 +1,5 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
-allow-panic-in-tests = true
\ No newline at end of file
+allow-panic-in-tests = true
+large-error-threshold = 196
```
jstuczyn, line 4: this seems like just kicking the can down the road -> could we not fix it "properly" instead? which errors is clippy complaining about? could we box something?

jstuczyn:
superseded by https://github.com/nymtech/nym/pull/5802

PR #5785, opened 2025-05-20
swap a decode into a fromrow to please future postgres feature

simonwicky:
This PR change a sqlx derive from `Decode` to `FromRow` because a future PR will bring the `postgres` feature and it doesn't like that `Decode` there

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5785)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6306,7 +6306,7 @@ dependencies = [
 
 [[package]]
 name = "nym-node-status-api"
-version = "2.3.1"
+version = "2.3.2"
 dependencies = [
  "ammonia",
  "anyhow",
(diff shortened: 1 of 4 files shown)


dynco-nym:
Please add `"nym-http-api-client",` to a list of crates [here](https://github.com/nymtech/nym/blob/simon/nsapi_pgsql_preparation/nym-node-status-api/nym-node-status-api/src/logging.rs#L27)

dynco-nym:
also, build fails due to missing feature, try `cd nym-node-status-api/nym-node-status-api && cargo build`

dynco-nym (approved):

PR #5786, opened 2025-05-20
[DOCs/operators]: Rewards calculator quick tweak

serinko:
This hard-coded intermittent solution will by default list a value corresponding with an onchain data rather than 0, while letting the operator to change these values in case they changed since the docs build. 

This hardcoded way is a quick tweak which will be changed within a week for a scraper filling the values on book build based on real chain data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5786)

### DIFF
--- a/documentation/docs/components/operators/interactive/calculators/reward-calculator.jsx
+++ b/documentation/docs/components/operators/interactive/calculators/reward-calculator.jsx
@@ -1,11 +1,11 @@
 import React, { useState } from 'react'
 
 export default function RewardsCalculator() {
-  const [a, setA] = useState(0)
+  const [a, setA] = useState(5241)
   const [b, setB] = useState(0)
   const [c, setC] = useState(0)
   const [d, setD] = useState(0)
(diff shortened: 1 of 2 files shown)


documentation/docs/components/operators/interactive/calculators/reward-calculator.jsx
```diff
@@ -1,11 +1,11 @@
 import React, { useState } from 'react'
 
 export default function RewardsCalculator() {
-  const [a, setA] = useState(0)
+  const [a, setA] = useState(5241)
```
mfahampshire, line 4: Is this meant to be modifiable in the calculator still, since we're going to hardcode it?

mfahampshire (approved):
Approved with comment

documentation/docs/components/operators/interactive/calculators/reward-calculator.jsx
serinko: yes, they can

PR #5789, opened 2025-05-20
HTTP Client Retries, Fallbacks, and Redirects

jmwample:
Implements changes used for NET-353 NET-354 NET-355 NET-356

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5789)


common/http-api-client/src/fronted.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl ClientBuilder {
+    /// Enable and configure request tunneling for API requests.
+    #[cfg(feature = "tunneling")]
+    pub fn with_fronting(mut self, policy: FrontPolicy) -> Self {
+        let front = Front::new(policy);
+
+        // Check if any of the supplied urls even support fronting
+        if self.urls.iter().filter(|url| url.has_front()).count() == 0 {
```
jstuczyn: nit: rather than going through the whole list of urls and counting the total number of results, you could use the `.any()` method to short-circuit it after you've found the first url with fronting support

common/http-api-client/src/fronted.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{ApiClientCore, Url, NO_PARAMS};
+
+    #[tokio::test]
+    async fn nym_api_works() {
+        let url1 = Url::new(
+            "https://validator.global.ssl.fastly.net",
```
jstuczyn, line 89: is this url stable, i.e. would this test start failing in few weeks/months?

common/http-api-client/src/lib.rs
```diff
@@ -194,7 +198,7 @@ pub enum HttpClientError<E: Display = String> {
     },
 
     #[error("failed to deserialize received response: {source}")]
-    ResponseDeserialisationFailure { source: serde_json::Error },
+    ResponseDeserializationFailure { source: serde_json::Error },
```
jstuczyn: you really don't like UK spelling, do you? 😄

common/http-api-client/src/lib.rs
```diff
@@ -247,15 +251,121 @@ impl HttpClientError {
...
+        let req = self.create_request(method, path, params, json_body);
+        self.send(req).await
+    }
+}
+
 /// A `ClientBuilder` can be used to create a [`Client`] with custom configuration applied consistently
 /// and state tracked across subsequent requests.
 pub struct ClientBuilder {
-    url: Url,
+    urls: Vec<Url>,
```
jstuczyn, line 357: given this was changed to a vec of urls, it might be useful to add additional method to push additional url (rather than overwriting all of them at once)

jstuczyn (approved):

common/http-api-client/src/fronted.rs
jmwample: The more stable ones haven't been set up yet, and they will be included in the `common/network-defaults` so I can make sure to update this when those are added in a separate PR.

common/http-api-client/src/lib.rs
jmwample: they show up as spelling mistakes sometimes and I change them out of habit :rofl:

common/http-api-client/src/fronted.rs
jmwample: This has been added. I also added a method to get a mutable reference to the urls for a `Client` in case that is useful at some point.

common/http-api-client/src/lib.rs
jmwample: This has been added. I also added a method to get a mutable reference to the urls for a `Client` in case that is useful at some point.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5789)* status: 0 of 13 files reviewed, 5 unresolved discussions

___
*[`common/http-api-client/src/url.rs` line 151 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORokbkABazfs5xQ6f_f:-ORokbkABazfs5xQ6f_g:bde0k37) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L151)):*
> ```XML
> }
> 
> impl Url {
> ```

Such innocent object as `Url` seems to be having so much state. I'd suggest to break it on few structures. The mapping itself does not need to carry `Url`s as it's merely host to host dictionary, i.e:

```rs
struct DomainMapping {
  original_domain: String,
  fronting_domains: Vec<String>
}
```

For better separation of concern we can move the atomic counter into individual object that can be shared across various instances of HTTP client, i.e:

```rs
struct DomainFrontStrategy {
  mapping: DomainMapping,
  current: AtomicUsize
}
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5789)* status: 0 of 13 files reviewed, 6 unresolved discussions

___
*[`common/http-api-client/src/url.rs` line 146 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORooHUT8OMIbUuyee1i:-ORooHUT8OMIbUuyee1j:b-19njx5) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L146)):*
> ```XML
>             url,
>             fronts: None,
>             current_front: Arc::new(AtomicUsize::new(0)),
> ```

This counter is not shared across multiple HTTP clients so that means we'll be stepping on the same banana peel in each HTTP client we instantiate.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5789)* status: 0 of 13 files reviewed, 7 unresolved discussions

___
*[`common/http-api-client/src/url.rs` line 229 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORook90BGb1eBqI3lqH:-ORook90BGb1eBqI3lqI:bknykug) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L229)):*
> ```XML
>     }
> 
>     /// Return the scheme of this URL, lower-cased, as an ASCII string without the ‘:’ delimiter.
> ```

All methods below repeat already existing functionality. Composition is a better choice.

jmwample (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5789)* status: 0 of 13 files reviewed, 7 unresolved discussions

___
*[`common/http-api-client/src/url.rs` line 146 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORooHUT8OMIbUuyee1i:-ORqPOBg1yOdzN1QBPly:b-j3yty6) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L146)):*

This counter is not shared across multiple HTTP clients so that means we'll be stepping on the same banana peel in each HTTP client we instantiate.
</blockquote></details>

Do you think a design that maintains some kind of client state that can be exported / shared across multiple clients would work better here? So downstream users could store url, user-agent etc, and a shared client state that they are somehow given?

At that point shouldn't they just keep / share the HTTP client itself?
___
*[`common/http-api-client/src/url.rs` line 151 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORokbkABazfs5xQ6f_f:-ORqQsZ83JaERoUgDFvd:b-w9b33h) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L151)):*

Such innocent object as `Url` seems to be having so much state. I'd suggest to break it on few structures. The mapping itself does not need to carry `Url`s as it's merely host to host dictionary, i.e:

```rs
struct DomainMapping {
  original_domain: String,
  fronting_domains: Vec<String>
}
```

For better separation of concern we can move the atomic counter into individual object that can be shared across various instances of HTTP client, i.e:

```rs
struct DomainFrontStrategy {
  mapping: DomainMapping,
  current: AtomicUsize
}
```
</blockquote></details>

I think this is a good separation of state and config and allows for simpler support of serialized representations (e.g. json).

However, I think storing them as strings rather than `URL`s might be unfortunate as we want to operate on them as URLs which would then mean calling `parse` often. So then maybe something like this makes sense?

```rust
pub struct DomainMapping {
  pub original_domain: String,
  pub fronting_domains: Vec<String>.
}

struct DomainMappingInner {
  original_domain: Url,
  fronting_domains: Vec<Url>.
}

pub struct DomainFrontStrategy {
  mapping: DomainMappingInner,
  current: AtomicUsize
}
```
___
*[`common/http-api-client/src/url.rs` line 229 at r2](https://reviewable.io/reviews/nymtech/nym/5789#-ORook90BGb1eBqI3lqH:-ORqQB6Y6etwXN2nrond:b-w2lmcu) ([raw file](https://github.com/nymtech/nym/blob/80dcb4723aafd0e5c1ac527bad045479bd19eb66/common/http-api-client/src/url.rs#L229)):*

All methods below repeat already existing functionality. Composition is a better choice.
</blockquote></details>

I am not sure what you mean by this. Is this not effectively the composition pattern? It acts as a URL (which doesn't have any trait behavior) It externally provides the same key functions as the base `Url`. Would a better pattern here be to define our own trait for Url behavior and then implement it for both types?

PR #5792, opened 2025-05-21
add notice re sdks

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5792)


serinko (approved):
There is a bit of indentation on some of the callouts. I would jusgt make sure that it's not printing it with it, unless intended. 

for future, I would consider to make it a component and import it across the pages, but it's not a must. 

LGTM!

PR #5795, opened 2025-05-24
Update to sqlx 0.8.6

pronebird:
Sqlx updates have been painful and we should strive keeping it fresh to reduce amount of efforts needed for upgrade. But also due to [RUSTSEC-2024-0363](https://rustsec.org/advisories/RUSTSEC-2024-0363.html) which may or may not be affecting us, but still, we should consider upgrading. In part I am looking to update sqlx because we want fresh copy of it in `nym-vpn-client` but also because I hold a belief that perhaps one of those updates will fix some of the issues with sql databases that we witnessed on Windows.

Please inspect and test this upgrade carefully because there have been a few critical changes in sqlx:

1. `COUNT` queries return `i64` instead of `i32` which seems sensible
2. `query_as` no longer understands `COALESCE` and fails to compile when it spots `nullable` field
3. Conversations between `Option<OffsetDateTime>` and `Option<chrono::DateTime>` aren't available, so I had to switch from `chrono::NaiveDateTime` to `OffsetDateTime` in one occurrence to circumvent that. However the code looks nicer to due to present `From` traits between `SystemTime` and `OffsetDateTime`. I assume there is no difference functionally. Going forward, my suggestion would be to replace `time` with `chrono` which should reduce some frictions in sqlx as it seems like it decides to use `OffsetDateTime` for `WireguardPeer` struct despite `sqlx::types::chrono::NaiveDateTime` being defined for `last_handshake`
4. `fetch_one` no longer returns `Option<T>` but `T`

Connected issue: JIRA-VPN-3456

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5795)


common/gateway-storage/src/models.rs
```diff
@@ -113,7 +115,7 @@ pub struct WireguardPeer {
     pub preshared_key: Option<String>,
     pub protocol_version: Option<i64>,
     pub endpoint: Option<String>,
-    pub last_handshake: Option<sqlx::types::chrono::NaiveDateTime>,
```
neacsu, line 116: I think @simonwicky was also looking into changing this type, not sure where it got to

neacsu (approved):

common/gateway-storage/src/models.rs
simonwicky: I had correct code to remove the `chrono` type in favor of `time` because it was messing with my statistics-API. 
I didn't commit anything because I couldn't make sure it would work for existing gateways. 
I found a solution to my issue this morning, so that particular bit isn't problematic for me anymore.

simonwicky:
> Going forward, my suggestion would be to replace `time` with `chrono`

IIRC, we use `time` almost everywhere, `chrono` might be the easier to replace. 
From my statistics API adventures, I also read that while sqlx 0.7 preferred `chrono` over `time` when both features were enabled, it was the opposite in sqlx 0.8. (although I didn't test it). It might then be worth to replace `chrono` for `time`

pronebird:
> > Going forward, my suggestion would be to replace `time` with `chrono`
> 
> IIRC, we use `time` almost everywhere, `chrono` might be the easier to replace. From my statistics API adventures, I also read that while sqlx 0.7 preferred `chrono` over `time` when both features were enabled, it was the opposite in sqlx 0.8. (although I didn't test it). It might then be worth to replace `chrono` for `time`

I see the same behaviour hence changed chrono to time although I prefer chrono mostly because I am used to it.

jstuczyn:
> Going forward, my suggestion would be to replace `time` with `chrono`

I'm not a fan of this - we're using `time` almost **everywhere** apart from a single place. if anything we should purge `chrono` in favour of `time` imo

nym-node-status-api/nym-node-status-api/src/db/queries/packet_stats.rs
```diff
@@ -61,7 +61,8 @@ pub(crate) async fn get_raw_node_stats(
         // if no packets are found, it's fine to assume 0 because that's also
         // SQL default value if none provided
         ScrapeNodeKind::LegacyMixnode { mix_id } => {
-            sqlx::query_as!(
+            // FIXME: we must find a way to switch back to query_as!
+            sqlx::query_as_unchecked!(
```
jstuczyn: what's preventing us from using `query_as!` (or `query_as`) ?

pronebird: I think that it doesn't like that `packets_*` fields can be NULL and it does not understand `COALESCE` that turns `NULL` into `0`

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5795)* status: 0 of 113 files reviewed, 2 unresolved discussions (waiting on @dynco-nym, @jstuczyn, and @neacsu)

nym-node-status-api/nym-node-status-api/src/db/queries/packet_stats.rs
dynco-nym: why does `query_as` work with `COALESCE` in `nym-node-status-api/nym-node-status-api/src/db/queries/mixnodes.rs` ?

dynco-nym: btw I'm fine with this being replaced with `query_as_unchecked` since the underlying query is unchanged (so we know it works)

dynco-nym (approved):
if someone can vouch for `common/gateway-storage/src/models.rs` date change, the rest looks good to me

nym-node-status-api/nym-node-status-api/src/db/queries/packet_stats.rs
pronebird: > why does `query_as` work with `COALESCE` in `nym-node-status-api/nym-node-status-api/src/db/queries/mixnodes.rs` ?

I am not sure however I can try reverting it tomorrow.

pronebird: This is the error that I see with `query_as!`

```
error: no built in mapping found for type NULL of column #1 ("packets_received!"); a type override may be required, see documentation for details
  --> nym-node-status-api/nym-node-status-api/src/db/queries/packet_stats.rs:65:13
   |
65 | /             sqlx::query_as!(
66 | |                 NodeStats,
67 | |                 r#"
68 | |                 SELECT
...  |
77 | |                 mix_id
78 | |             )
   | |_____________^
   |
   = note: this error originates in the macro `$crate::sqlx_macros::expand_query` which comes from the expansion of the macro `sqlx::query_as` (in Nightly builds, run with -Z macro-backtrace for more info)

```

I am having hard time to understand why it works for`nym-node-status-api/nym-node-status-api/src/db/queries/mixnodes.rs` but not here. One way to go about it is to force override the type in order to continue using `query_as!`

pronebird: Fixed with type-override because for some reason `COALESCE` would still yield `Option<i64>`

jstuczyn (approved):

pronebird:
@benedettadavico can you check the milestone for this please?

PR #5796, opened 2025-05-24
Close sqlite pool before moving or reopening databases

pronebird:
Our sqlite db recovery and backup strategies don't work on Windows because the OS prevents moving the already open file. This PR attempts to address this and in principle it attempts to close all connections to the database and ensure that sqlx relinquishes the access to db file(s).

Short summary of what this PR introduces:

1. Close sqlx connection pools in order to close/release the db file
2. Poll until there are no more db file descriptors left open. See: https://github.com/launchbadge/sqlx/issues/3217

JIRA-VPN-1471

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5796)

### DIFF
--- a/common/client-core/src/client/base_client/non_wasm_helpers.rs
+++ b/common/client-core/src/client/base_client/non_wasm_helpers.rs
@@ -73,7 +73,7 @@ async fn archive_corrupted_database<P: AsRef<Path>>(db_path: P) -> io::Result<()
     let renamed = db_path.with_extension(new_extension);
 
     tokio::fs::rename(db_path, &renamed).await.inspect_err(|_| {
-        tracing::error!(
+        error!(
             "Failed to rename corrupt database file: {} to {}",
             db_path.display(),
             renamed.display()
--- a/common/client-core/src/client/base_client/non_wasm_helpers.rs
+++ b/common/client-core/src/client/base_client/non_wasm_helpers.rs
@@ -1,20 +1,22 @@
 // Copyright 2022-2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::client::replies::reply_storage::{
-    fs_backend, CombinedReplyStorage, ReplyStorageBackend,
+use crate::{
+    client::replies::reply_storage::{fs_backend, CombinedReplyStorage, ReplyStorageBackend},
+    config,
+    config::Config,
+    error::ClientCoreError,
 };
-use crate::config;
-use crate::config::Config;
-use crate::error::ClientCoreError;
+#[cfg(windows)]
+use log::debug;
 use log::{error, info, trace};
 use nym_bandwidth_controller::BandwidthController;
 use nym_client_core_gateways_storage::OnDiskGatewaysDetails;
 use nym_credential_storage::storage::Storage as CredentialStorage;
-use nym_validator_client::nyxd;
-use nym_validator_client::QueryHttpRpcNyxdClient;
-use std::path::Path;
-use std::{fs, io};
+use nym_validator_client::{nyxd, QueryHttpRpcNyxdClient};
+#[cfg(windows)]
+use std::time::Duration;
+use std::{io, path::Path};
 use time::OffsetDateTime;
 use url::Url;
 
@@ -22,11 +24,14 @@ async fn setup_fresh_backend<P: AsRef<Path>>(
     db_path: P,
     surb_config: &config::ReplySurbs,
 ) -> Result<fs_backend::Backend, ClientCoreError> {
-    info!("creating fresh surb database");
+    info!(
+        "creating fresh surb database: {}",
+        db_path.as_ref().display()
+    );
     let mut storage_backend = match fs_backend::Backend::init(db_path).await {
         Ok(backend) => backend,
         Err(err) => {
-            error!("failed to setup persistent storage backend for our reply needs: {err}");
+            error!("setup_fresh_backend: Failed to setup persistent storage backend for our reply needs: {err}");
             return Err(ClientCoreError::SurbStorageError {
                 source: Box::new(err),
             });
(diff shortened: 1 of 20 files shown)


sqlx-pool-guard/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[dependencies]
+tokio = { workspace = true, features = [
+    "rt-multi-thread",
+    "macros",
+    "time",
+    "fs",
+] }
+sqlx = { workspace = true, features = ["runtime-tokio-rustls", "sqlite"] }
+log.workspace = true
```
dynco-nym: why not use tracing? it allows for structured logging and is[ compatible](https://docs.rs/tracing/latest/tracing/#emitting-log-records) with binaries using simple `log` crate

common/client-core/src/client/base_client/non_wasm_helpers.rs
```diff
@@ -87,13 +89,12 @@ pub async fn setup_fs_reply_surb_backend<P: AsRef<Path>>(
...
     if db_path.exists() {
-        info!("loading existing surb database");
+        info!("Loading existing surb database");
         match fs_backend::Backend::try_load(db_path, surb_config.fresh_sender_tags).await {
             Ok(backend) => Ok(backend),
             Err(err) => {
-                error!("failed to setup persistent storage backend for our reply needs: {err}. We're going to create a fresh database instead. This behaviour might change in the future");
-
-                archive_corrupted_database(db_path)?;
+                error!("setup_fs_reply_surb_backend: Failed to setup persistent storage backend for our reply needs: {err}. We're going to create a fresh database instead. This behaviour might change in the future");
```
dynco-nym, line 96: if you don't want to instrument the function, event macro [supports](https://docs.rs/tracing/latest/tracing/#configuring-attributes) setting `target` and `name` attributes in a structured way

dynco-nym (commented):
looks good provided you tested this

sqlx-pool-guard/Cargo.toml
pronebird: I don't mind. I use tracing everywhere but it seems that a lot of code here uses `log` so I just stuck to it. Should I be changing this?

common/client-core/src/client/base_client/non_wasm_helpers.rs
pronebird: I needed to distinguish because there are two functions that used to output the same message :)

pronebird:
> looks good provided you tested this

It's tested. Also there is a test that you can run.

sqlx-pool-guard/Cargo.toml
pronebird: Fixed to use tracing.

common/client-core/src/client/base_client/non_wasm_helpers.rs
pronebird: Errrm so I am not looking to change this entire crate from log to tracing.

jstuczyn (approved):

sqlx-pool-guard/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+impl Deref for SqlitePoolGuard {
+    type Target = sqlx::SqlitePool;
+
+    fn deref(&self) -> &Self::Target {
+        &self.connection_pool
+    }
+}
```
neacsu, line 124: Isn't this enough to do `&self.connection_pool` instead of `&*self.connection_pool` in the execution functions above?

pronebird: > Isn't this enough to do `&self.connection_pool` instead of `&*self.connection_pool` in the execution functions above?

No it doesn't seem to work.

```
error[E0277]: the trait bound `&SqlitePoolGuard: Executor<'_>` is not satisfied
   --> common/credential-storage/src/backends/sqlite.rs:37:18
    |
37  |         .execute(&self.connection_pool)
    |          ------- ^^^^^^^^^^^^^^^^^^^^^ the trait `Executor<'_>` is not implemented for `&SqlitePoolGuard`
    |          |
    |          required by a bound introduced by this call
    |
note: required by a bound in `Query::<'q, DB, A>::execute`
   --> /Users/pronebird/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sqlx-core-0.7.4/src/query.rs:155:12
    |
151 |     pub async fn execute<'e, 'c: 'e, E>(self, executor: E) -> Result<DB::QueryResult, Error>
    |                  ------- required by a bound in this associated function
...
155 |         E: Executor<'c, Database = DB>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Query::<'q, DB, A>::execute`
help: consider dereferencing here
    |
37  |         .execute(&*self.connection_pool)
    |                   +
```

If I just deref then the following error is emitted:

```
error[E0277]: the trait bound `Pool<Sqlite>: Executor<'_>` is not satisfied
   --> common/credential-storage/src/backends/sqlite.rs:37:18
    |
37  |         .execute(*self.connection_pool)
    |          ------- ^^^^^^^^^^^^^^^^^^^^^ the trait `Executor<'_>` is not implemented for `Pool<Sqlite>`
    |          |
    |          required by a bound introduced by this call
    |
note: required by a bound in `Query::<'q, DB, A>::execute`
   --> /Users/pronebird/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sqlx-core-0.7.4/src/query.rs:155:12
    |
151 |     pub async fn execute<'e, 'c: 'e, E>(self, executor: E) -> Result<DB::QueryResult, Error>
    |                  ------- required by a bound in this associated function
...
155 |         E: Executor<'c, Database = DB>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Query::<'q, DB, A>::execute`
help: consider borrowing here
    |
37  |         .execute(&*self.connection_pool)
```

It's also possible to implement `AsRef` but then `.as_ref()` needs to be used to obtain the underlying executor and `&self.connection_pool` wouldn't work:


```rs
impl AsRef<sqlx::SqlitePool> for SqlitePoolGuard {
    fn as_ref(&self) -> &sqlx::SqlitePool {
        &self.connection_pool
    }
}
```

neacsu: Ah right, I confounded it with `AsRef`. Strange that, implementing that, the `&x` approach wouldn't work

pronebird: Further thoughts: we could implement `SqliteExecutor` on `SqlitePoolGuard` and delegate all work to `self.connection_pool`. This would make this type fully transparent. However, I don't think I have capacity to do that right now. Note that eventually this type should go away once the problem is fixed in sqlx.

jstuczyn:
seems you might have to do some additional feature-locking because `mio` is pulled into wasm build (from my experience, in 99.9% cases it's due to one of tokio's features)

pronebird:
> seems you might have to do some additional feature-locking because `mio` is pulled into wasm build (from my experience, in 99.9% cases it's due to one of tokio's features)

Fixed 👍

pronebird:
We'll conduct a bit of testing in nym-vpn-client before merging this. Thank you for reviewing this work.

PR #5811, opened 2025-05-29
Replace chrono with time in NS API

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5811)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
@@ -157,7 +157,7 @@ async fn submit_testrun(
     let created_at = unix_timestamp_to_utc_rfc3339(assigned_testrun.created_utc);
     let last_assigned = assigned_testrun
         .last_assigned_utc
-        .map(|timestamp| unix_timestamp_to_utc_rfc3339(timestamp))
+        .map(unix_timestamp_to_utc_rfc3339)
         .unwrap_or_else(|| String::from("never"));
     tracing::info!(
         "✅ Testrun row_id {} for gateway {} complete (last assigned {}, created at {})",
--- a/nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
@@ -1,6 +1,6 @@
 use crate::db::models::TestRunStatus;
 use crate::db::queries;
-use crate::testruns::now_utc;
+use crate::utils::{now_utc, unix_timestamp_to_utc_rfc3339};
 use crate::{
     db,
     http::{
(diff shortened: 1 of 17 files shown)


nym-node-status-api/nym-node-status-api/src/utils.rs
```diff
@@ -41,6 +42,22 @@ mod test {
     }
 }
 
+pub(crate) fn now_utc() -> time::UtcDateTime {
+    time::UtcDateTime::now()
+}
+
+pub(crate) fn unix_timestamp_to_utc_rfc3339(unix_timestamp: i64) -> String {
+    let timestamp = time::UtcDateTime::UNIX_EPOCH + time::Duration::seconds(unix_timestamp);
```
jstuczyn, line 50: use time's method for converting from unix timestamp, i.e. `OffsetDateTime::from_unix_timestamp(...)` rather than manually adding seconds to the `UNIX_EPOCH`

nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -188,12 +186,18 @@ fn authenticate(request: &impl VerifiableRequest, state: &AppState) -> HttpResul
...
+
 fn is_fresh(request_time: &i64) -> HttpResult<()> {
     // if a request took longer than N minutes to reach NS API, something is very wrong
-    let freshness_cutoff = chrono::Duration::minutes(1);
-    let cutoff_timestamp = (now_utc() - freshness_cutoff).timestamp();
-    if *request_time < cutoff_timestamp {
-        tracing::warn!("Request older than {}s, rejecting", cutoff_timestamp);
+    let request_time = time::UtcDateTime::from_unix_timestamp(*request_time).map_err(|e| {
+        warn!("Failed to parse request time: {e}");
+        HttpError::unauthorized()
```
jstuczyn, line 195: nit: if server somehow fails to parse the current time, is it really an `unauthorized` error?

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/utils.rs
jstuczyn: `UtcDateTime` has the same method

nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
dynco-nym: in this specific case: yes, because that means user is not authorized to submit that testrun 

(it can only happen if client code has been tampered with and it's not an authentic request)

nym-node-status-api/nym-node-status-api/src/utils.rs
dynco-nym: tried that initially but that function is fallible as opposed to this approach

PR #5818, opened 2025-06-02
chore: adjust heuristic for wireguard peer activity

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5818)

### DIFF
--- a/common/wireguard/src/peer_controller.rs
+++ b/common/wireguard/src/peer_controller.rs
@@ -281,11 +281,11 @@ impl PeerController {
         for (peer_key, peer) in new_host.peers.iter() {
             match old_host.peers.get(peer_key) {
                 // only consider pre-existing peers for the purposes of bandwidth accounting,
-                // do that the value would always be increasing.
+                // so that the value would always be increasing.
                 Some(prior) => {
                     // 1. determine bandwidth changes
                     let delta_rx = peer.rx_bytes.saturating_sub(prior.rx_bytes);
-                    let delta_tx = prior.tx_bytes.saturating_sub(prior.tx_bytes);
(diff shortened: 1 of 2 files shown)


common/wireguard/src/peer_controller.rs
```diff
@@ -279,26 +279,51 @@ impl PeerController {
...
 
         for (peer_key, peer) in new_host.peers.iter() {
-            // only consider pre-existing peers,
-            // so that the value would always be increasing
-            if let Some(prior) = old_host.peers.get(peer_key) {
-                let delta_rx = peer.rx_bytes.saturating_sub(prior.rx_bytes);
-                let delta_tx = prior.tx_bytes.saturating_sub(prior.tx_bytes);
+            match old_host.peers.get(peer_key) {
+                // only consider pre-existing peers for the purposes of bandwidth accounting,
+                // do that the value would always be increasing.
```
durch: Typo, `*so`

durch (changes requested):
`delta_tx` calculation needs to be fixed, otherwise good to go

common/wireguard/src/peer_controller.rs
```diff
@@ -279,26 +279,51 @@ impl PeerController {
...
+                Some(prior) => {
+                    // 1. determine bandwidth changes
+                    let delta_rx = peer.rx_bytes.saturating_sub(prior.rx_bytes);
+                    let delta_tx = prior.tx_bytes.saturating_sub(prior.tx_bytes);
```
durch: This will always be 0

jstuczyn: well, that explains why the old metrics were always showing 0 for that : D

PR #5825, opened 2025-06-04
[Stats API] Infallible network view

simonwicky:
Current `develop` API client doesn't work with currently deployed nym API

This PR downgrades the clients to `cheddar` for compatibility and make building the network view infallible, so that any future incompatibility doesn't prevent the stats API to start

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5825)

### DIFF
--- a/nym-statistics-api/Cargo.toml
+++ b/nym-statistics-api/Cargo.toml
@@ -39,7 +39,6 @@ utoipauto.workspace = true
 
 #internal
 nym-bin-common = { path = "../common/bin-common" }
-nym-http-api-client = { path = "../common/http-api-client" }
 nym-http-api-common = { path = "../common/http-api-common", features = [
     "middleware",
 ] }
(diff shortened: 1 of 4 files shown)


nym-statistics-api/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "nym-statistics-api"
-version = "0.1.0"
+version = "0.1.1"
```
dynco-nym, line 6: interestingly, this tag already exists? did it get created automatically in some github workflow?

dynco-nym (approved):
don't forget to fix the tag :)

nym-statistics-api/Cargo.toml
simonwicky: I triggered the workflow manually from that branch, to make sure it was working, so the harbor image with this tag is actually this one, so all good on that front

PR #5829, opened 2025-06-05
Feature/node status dvpn directory

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5829)


nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -158,17 +167,129 @@ impl HttpCache {
...
+                    *guard = new_gateway_list;
+                    v.clone()
+                } else {
+                    Arc::new(RwLock::new(new_gateway_list))
+                }
+            })
+            .await
+    }
+
+    pub async fn get_dvpn_gateway_list(&self, db: &DbPool) -> Vec<DVpnGateway> {
```
dynco-nym: @mmsinclair

nym-node-status-api/nym-node-status-api/src/db/queries/gateways.rs
```diff
@@ -111,7 +111,8 @@ pub(crate) async fn get_all_gateways(pool: &DbPool) -> anyhow::Result<Vec<Gatewa
             error!("Conversion from DTO failed: {e}. Invalidly stored data?");
             e
         })?;
-    tracing::trace!("Fetched {} gateways from DB", items.len());
+    // TODO dz revert to trace
+    tracing::info!("Fetched {} gateways from DB", items.len());
```
jstuczyn: so revert to trace? : )

nym-node-status-api/nym-node-status-api/src/http/error.rs
```diff
@@ -16,6 +16,14 @@ impl HttpError {
         }
     }
 
+    pub(crate) fn invalid_country_code() -> Self {
+        Self {
+            message: serde_json::json!({"message": "Only two letter country code is allowed"})
```
jstuczyn: do we have to wrap it in json? (if we do, that's fine, just asking what might be expecting that)

nym-node-status-api/nym-node-status-api/src/mixnet_scraper/helpers.rs
```diff
@@ -206,7 +206,7 @@ pub async fn update_daily_stats(
     let date_utc = format!(
```
jstuczyn, line 206: nit: you could just use `format_description!` instead (https://docs.rs/time/latest/time/format_description/index.html)

nym-node-status-api/nym-node-status-api/src/testruns/mod.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) async fn spawn(pool: DbPool, refresh_interval: Duration) {
     });
 }
 
-#[instrument(level = "debug", name = "testrun_queue", skip_all)]
+#[instrument(level = "info", name = "testrun_queue", skip_all)]
```
jstuczyn, line 27: just to check, we want to leave this at `info`? it wouldn't cause too much noise?

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
```diff
@@ -0,0 +1,95 @@
...
+#[instrument(level = tracing::Level::INFO, skip(state), fields(two_letter_country_code = two_letter_country_code))]
+pub async fn get_gateways_by_country(
+    Path(TwoLetterCountryCodeParam {
+        two_letter_country_code,
+    }): Path<TwoLetterCountryCodeParam>,
+    state: State<AppState>,
+) -> HttpResult<Json<Vec<DVpnGateway>>> {
+    let country_filter = two_letter_country_code.to_lowercase();
+    match two_letter_country_code.len() {
+        2 => Ok(Json(
```
jstuczyn: isn't this match redundant given your param validation?

jstuczyn: if you want to be super type safe, you could use `celes` that does all the parsing, deserialisation, etc. we're already using it in couple places

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
```diff
@@ -0,0 +1,95 @@
...
+            state
+                .cache()
+                .get_dvpn_gateway_list(state.db_pool(), &MIN_SUPPORTED_VERSION)
```
jstuczyn: I recognise this might be difficult to change now, but eventually we should do filtering on the retrieval level. because as it works right now it will get **all** gateways from the db (assuming cache was empty) and then filter from the whole list

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
```diff
@@ -0,0 +1,95 @@
...
+#[instrument(level = tracing::Level::INFO, skip(state))]
+pub async fn get_gateway_countries(state: State<AppState>) -> HttpResult<Json<Vec<String>>> {
+    Ok(Json(
+        state
+            .cache()
+            .get_dvpn_gateway_list(state.db_pool(), &MIN_SUPPORTED_VERSION)
+            .await
+            .into_iter()
+            .map(|gw| gw.location.two_letter_iso_country_code.to_string())
+            // dedup relies on iterator being sorted by country, but we already do that
```
jstuczyn, line 82: another nit for the future. rather than iterating through the list of all gateways, couldn't we just query the db for unique country codes?

nym-node-status-api/nym-node-status-api/src/http/models.rs
```diff
@@ -23,6 +32,223 @@ pub struct Gateway {
...
+    pub ip_packet_router: Option<IpPacketRouterDetails>,
+    pub authenticator: Option<AuthenticatorDetails>,
+    pub location: Location,
+    pub last_probe: Option<DirectoryGwProbe>,
+    pub ip_addresses: Vec<String>,
+    pub mix_port: u16,
+    pub role: NodeRole,
+    pub entry: Option<BasicEntryInformation>,
+    // The performance data here originates from the nym-api, and is effectively mixnet performance
+    // at the time of writing this
```
jstuczyn, line 65: in that case couldn't we store it as a slightly more type safe integer/float?

nym-node-status-api/nym-node-status-api/src/http/models.rs
```diff
@@ -23,6 +32,223 @@ pub struct Gateway {
...
+#[derive(Debug, Clone, Deserialize, Serialize, ToSchema)]
+pub struct Location {
+    pub two_letter_iso_country_code: String,
+    pub latitude: f64,
+    pub longitude: f64,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, ToSchema)]
+pub struct DVpnGateway {
+    pub identity_key: String,
```
jstuczyn, line 53: assuming it wouldn't cause too many issues down the callstack, could we use a stronger type here, i.e. `ed25519::PublicKey` instead?

nym-node-status-api/nym-node-status-api/src/http/models.rs
```diff
@@ -23,6 +32,223 @@ pub struct Gateway {
...
+    pub name: String,
+    pub ip_packet_router: Option<IpPacketRouterDetails>,
+    pub authenticator: Option<AuthenticatorDetails>,
+    pub location: Location,
+    pub last_probe: Option<DirectoryGwProbe>,
+    pub ip_addresses: Vec<String>,
```
jstuczyn: `Vec<IpAddr>`?

nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -158,17 +168,166 @@ impl HttpCache {
...
+
+                let gateways = self.get_gateway_list(db).await;
+
+                let started_with = gateways.len();
+                let Ok(skimmed_nodes) = crate::db::queries::get_described_bonded_nym_nodes(db)
+                    .await
+                    .map(|records| {
+                        records
+                            .into_iter()
+                            .filter_map(|dto| SkimmedNode::try_from(dto).ok())
```
jstuczyn: we're kinda silently ignoring nodes that are seemingly invalid. at the very least we should log them to know about it

nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -158,17 +168,166 @@ impl HttpCache {
...
+                        if let Ok(gw_version) = Version::parse(gw_version) {
+                            &gw_version >= min_node_version
+                        } else {
+                            warn!("Failed to parse GW version {}", gw_version);
+                            false
+                        }
+                    })
+                    .filter(|gw| {
+                        // gateways must have a country
+                        if gw.location.two_letter_iso_country_code.len() == 2 {
```
jstuczyn, line 283: nit: just use stronger typing via `celes` and we can avoid this filtering : )

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/testruns/mod.rs
dynco-nym: I like instruments to be at a high level, adds call context to further logs

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
dynco-nym: > isn't this match redundant given your param validation?

validation you're referring to happens in swagger UI. If someone uses `curl`, I have to validate on the backend as well

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
dynco-nym: while I do get all gateways, then filter, this **caches** all gateways retrieved so next request to any of these
- `/gateways`
- `/gateways/entry`
- `/gateways/exit`

fetches from the cache. Also, a MISS on any of those refreshes underlying cache for all of them

nym-node-status-api/nym-node-status-api/src/http/state.rs
dynco-nym: true, added a log

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
dynco-nym: gateways are [filtered](https://github.com/nymtech/nym/blob/f17239075b90a56fed6a04c63d59cff51f14d02e/nym-node-status-api/nym-node-status-api/src/http/models.rs#L68) based on last probe score, then only those countries are returned. That happens outside of DB

nym-node-status-api/nym-node-status-api/src/http/error.rs
dynco-nym: we don't, good catch

nym-node-status-api/nym-node-status-api/src/http/api/dvpn/country.rs
dynco-nym: implemented `celes` validation

nym-node-status-api/nym-node-status-api/src/http/models.rs
dynco-nym: DTO stores it as int, but this is an API model that, for backwards compatibility, needs to be a string

nym-node-status-api/nym-node-status-api/src/http/state.rs
dynco-nym: added `celes`

nym-node-status-api/nym-node-status-api/src/mixnet_scraper/helpers.rs
dynco-nym: this was the key reason for the bug and why a migration was needed. Now that I tested everything thoroughly, I don't dare changing it again, at least this time:)

nym-node-status-api/nym-node-status-api/src/http/models.rs
dynco-nym: changed the other fields, but this one is set from a stringified [Gateway,](https://github.com/nymtech/nym/blob/f17239075b90a56fed6a04c63d59cff51f14d02e/nym-node-status-api/nym-node-status-api/src/http/models.rs#L21) which is in too many db deserializations to change at this moment

nym-node-status-api/nym-node-status-api/src/mixnet_scraper/helpers.rs
jstuczyn: hehe, fair enough

PR #5840, opened 2025-06-11
Removing test-net faucet

tommyv1987:
Part of NET-413

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5840)


jstuczyn (approved):
🔥

PR #5857, opened 2025-06-13
Add build info endpoints

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5857)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/http/state.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/state.rs
@@ -5,15 +5,16 @@ use nym_bin_common::bin_info_owned;
 use nym_contracts_common::NaiveFloat;
 use nym_crypto::asymmetric::ed25519::PublicKey;
 use nym_mixnet_contract_common::NodeId;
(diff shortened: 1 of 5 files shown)


nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -1,14 +1,18 @@
 use cosmwasm_std::Decimal;
 use itertools::Itertools;
 use moka::{future::Cache, Entry};
+use nym_bin_common::bin_info_owned;
 use nym_contracts_common::NaiveFloat;
 use nym_crypto::asymmetric::ed25519::PublicKey;
 use nym_mixnet_contract_common::NodeId;
-use nym_validator_client::nym_api::SkimmedNode;
+use nym_validator_client::{models::BinaryBuildInformationOwned, nym_api::SkimmedNode};
```
jstuczyn: imo nym-node-status-api shouldnt depent on that particular import path for its own models; instead it should re-expose them itself

jstuczyn (approved):

PR #5861, opened 2025-06-16
Make Mix hops optional for Mixnet Client SURBs

jmwample:
Allow SURBs to be sent configured to only use gateways in mixnet return path.

Follow on to #5696

Tested as working with mainnet vpnd/vpnc client for a handful of manual connections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5861)

### DIFF
--- a/common/client-core/config-types/src/lib.rs
+++ b/common/client-core/config-types/src/lib.rs
@@ -445,10 +445,7 @@ impl Default for Traffic {
             primary_packet_size: PacketSize::RegularPacket,
             secondary_packet_size: None,
             packet_type: PacketType::Mix,
-
-            // we should use the legacy format until sufficient number of nodes understand the
-            // improved encoding
-            use_legacy_sphinx_format: true,
+            use_legacy_sphinx_format: false,
             disable_mix_hops: false,
         }
     }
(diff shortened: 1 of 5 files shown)


jmwample:
@jstuczyn I don't know how broadly the `use_legacy_sphinx_format` settings apply. Given that we have looked at the coverage that the update has in node currently Is it okay to update the default to where clients will use the updated sphinx format? Or should we leave the legacy format as default  and make this this be more targeted somehow?

jstuczyn:
> @jstuczyn I don't know how broadly the `use_legacy_sphinx_format` settings apply. Given that we have looked at the coverage that the update has in node currently Is it okay to update the default to where clients will use the updated sphinx format? Or should we leave the legacy format as default and make this this be more targeted somehow?

so the thing is, it will apply to **every packet** constructed by the client. meaning we'd expect every mixnod, gateway and service provider on the route to be compatible with that change (otherwise they'd be dropping the packets or failing to parse them). ideally, maybe for now, we should do it in a more targetted way to just affect the vpn registration?

jmwample:
> we should do it in a more targetted way to just affect the vpn registration

How about this solution: If the client configures `disabled_mix_hops` it overrides the `use_legacy_sphinx_format` setting by necessity, creating its own exception. For now this only happens in the registration mixnet client. This way you have to intentionally enable a feature (that is disabled by default) to use this format.

When we are ready we can look into changing the default setting independently.

jstuczyn (approved):

PR #5864, opened 2025-06-18
Remove old free credential handle

neacsu:
and create some traits for unit testing purposes. Also added a few unit tests, but more will follow in future PRs

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)


service-providers/authenticator/src/main.rs
jstuczyn, line 1: I was under the impression it was useful for testing purposes to be able to run service providers outside the gateway (but if removing this options simplifies things, I'm all for it!)

common/gateway-storage/src/models.rs
```diff
@@ -112,35 +110,26 @@ impl TryFrom<UnverifiedTicketData> for ClientTicket {
...
             public_key: value.public_key.to_string(),
-            preshared_key: value.preshared_key.as_ref().map(|k| k.to_string()),
-            protocol_version: value.protocol_version.map(|v| v as i64),
-            endpoint: value.endpoint.map(|e| e.to_string()),
-            last_handshake: value.last_handshake.map(OffsetDateTime::from),
-            tx_bytes: value.tx_bytes as i64,
-            rx_bytes: value.rx_bytes as i64,
-            persistent_keepalive_interval: value.persistent_keepalive_interval.map(|v| v as i64),
             allowed_ips: bincode::Options::serialize(
                 bincode::DefaultOptions::new(),
```
jstuczyn: very nitty nit (especially since it's a pre-existing code): rather than using defaults at the call site, could we define a seperate method to create the serialiser (like we have with `make_bincode_serializer`), the reason for it is that bincode 2 has some incompatible changes and that way it will be easier to upgrade

common/wireguard/src/lib.rs
```diff
@@ -100,29 +164,13 @@ pub async fn start_wireguard(
...
-        let bandwidth_manager =
-            PeerController::generate_bandwidth_manager(storage.clone(), &peer.public_key)
-                .await?
-                .map(|bw_m| Arc::new(RwLock::new(bw_m)));
-        // Update storage with *x_bytes set to 0, as in kernel peers we can't set those values
-        // so we need to restart counting. Hopefully the bandwidth was counted in available_bandwidth
-        storage
-            .insert_wireguard_peer(peer, bandwidth_manager.is_some())
-            .await?;
+        let bandwidth_manager = Arc::new(RwLock::new(
```
jstuczyn, line 180: that's just some background thinking. given we might, theoretically, have thousands of clients and a single underlying storage that all of them use, I wonder if this will cause bottlenecking. I wonder, should we perhaps add `log_slow_statements` when creating the sqlx db so we could see symptoms of it if they appear? what do you reckon?

common/wireguard/src/peer_controller.rs
```diff
@@ -41,10 +36,6 @@ pub enum PeerControlRequest {
         key: Key,
         response_tx: oneshot::Sender<QueryPeerControlResponse>,
     },
-    QueryBandwidth {
```
jstuczyn, line 44: wouldn't it be useful to retain this functionality so that clients could explicitly ask for their remaining bandwidth? unless this is implemented differently now

common/wireguard/src/peer_handle.rs
```diff
@@ -88,98 +81,78 @@ impl PeerHandle {
...
+                    "Overflow on cached adding bytes: {} + {}",
+                    cached_peer.rx_bytes,
+                    cached_peer.tx_bytes
+                );
                 None
             })?;
 
-        kernel_total.checked_sub(storage_total).or_else(|| {
-            tracing::error!("Overflow on spent bandwidth subtraction: kernel - storage = {kernel_total} - {storage_total}");
+        kernel_total.checked_sub(cached_total).or_else(|| {
```
jstuczyn, line 96: probably a stupid question, but shouldn't it be the other way around given that (at least that's my understanding) kernel peer starts from 0 and cached/storage one resumes from whatever it had at last flush?

common/wireguard/src/peer_handle.rs
```diff
@@ -88,98 +81,78 @@ impl PeerHandle {
...
+        if spent_bandwidth > 0
+            && self
+                .bandwidth_storage_manager
+                .write()
+                .await
+                .try_use_bandwidth(spent_bandwidth)
+                .await
+                .is_err()
+        {
+            tracing::debug!(
```
jstuczyn, line 128: but shouldn't a client be allowed to send a ticket to increase its bandwidth? immediate removal sounds a bit aggressive and I wonder if this could result in some unintentional disconnects

jstuczyn (approved):
it's always amazing to see so much removed code and additional tests!

just two questions that bother me:
- I've seen you removed the logic to do with deciding if storage peer bandwidth should be synced to the storage - we're not doing this on every update now, are we?
- given the removal of the free cap, how will it affect backwards compatibility? as in, how many clients do we expect might break because of this change?

service-providers/authenticator/src/main.rs
neacsu: It should have been, but because of disk and ecash dependencies, it ended up being more of a hassle than a help. Going more towards the trait testing should fill that gap better than a cli testing tool.

neacsu:
> * I've seen you removed the logic to do with deciding if storage peer bandwidth should be synced to the storage - we're not doing this on every update now, are we?

The storage for peer bandwidth is removed altogether since it wasn't really needed. The kernel peer structure couldn't be populated with the restored values from disk anyway, so on gateway restart we just start from 0 the counting of what's to be consumed. The remaining update to storage is now only the one we already have from the `BandwidthStorageManager`, which tracks top-ups and consumption, so that we can get available bandwidth.

> * given the removal of the free cap, how will it affect backwards compatibility? as in, how many clients do we expect might break because of this change?

The free cap is officially dropped since a long time ago, when we released the paid product, and such stale clients are not supported officially anymore. But let me check if we know about any old clients being used in production.

common/wireguard/src/lib.rs
neacsu: That sounds like a very good idea actually. Thanks!

common/wireguard/src/peer_controller.rs
neacsu: It is still exposed by reading on the `ClientBandwidth` lock, but it's done on the caller side instead of on the controller now. I don't think duplicating it makes sense anymore.

common/wireguard/src/peer_handle.rs
neacsu: The cached one doesn't have any relation to storage anymore, wrt byte counting. I've left that solely to the ecash module that was doing it anyway. So this is just last synced value of kernel, and the kernel values always go up. ~~Actually, this is a good point, since `cached` is probably a bad naming, since it's now more like a last seen reference, to know what is left to subtract. I'll rename it.~~ On second thought, cached doesn't seem like a bad name, it's just that is not caching storage in this case, it's caching kernel values.

common/wireguard/src/peer_handle.rs
neacsu: Currently we rely on clients pre-emptively sending tickets so that they don't get into this situation. I agree that a better system would be needed, to find a good mix between unnecessary over spending and dangerously approaching forced disconnections. The mixnet channel's unreliability makes things even more difficult.
I think the better system implementation shouldn't be part of this PR though.

pronebird (changes requested):
> Reviewed 6 of 47 files at r1, 1 of 3 files at r4, 2 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 9 of 48 files reviewed, 8 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/credential-verification/src/ecash/state.rs` line 25 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU6FCV19aRs57o0UUCU:-OU6FCV19aRs57o0UUCV:bigcyh0) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/credential-verification/src/ecash/state.rs#L25)):*
> ```XML
> 
> // state shared by different subtasks dealing with credentials
> pub(crate) struct SharedState {
> ```

I wonder if using `Arc<RwLock<SharedState>>` would remove the need for `dyn_clone`? I don't know much about this code but  this is something that
___
*[`common/gateway-storage/src/traits.rs` line 207 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU92dKr9FmfhIxxto6D:-OU92dKr9FmfhIxxto6E:b-pucgjm) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/traits.rs#L207)):*
> ```XML
> 
>     #[derive(Clone, Default)]
>     pub struct MockGatewayStorage {
> ```

A  dozen of fields use `Arc<RwLock<_>>`.  Would it not be better to use one `Arc<RwLock<MockGatewayStorage>>` instead?

pronebird (changes requested):
> Reviewed 1 of 47 files at r1, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 11 of 48 files reviewed, 9 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/gateway-storage/src/traits.rs` line 153 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU95ldB3cUyiR0Lj4NZ:-OU95ldB3cUyiR0Lj4N_:b-ya0frk) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/traits.rs#L153)):*
> ```XML
>     ) -> Result<i64, GatewayStorageError>;
> 
>     /// Tries to retrieve available bandwidth for the particular peer.
> ```

This does not seem right. The method returns `WireguardPeer`

pronebird (commented):
> Reviewed 1 of 47 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 12 of 48 files reviewed, 9 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 12 of 48 files reviewed, 10 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/gateway-storage/src/models.rs` line 126 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU96pHX3JMN4oSb77SS:-OU96pHX3JMN4oSb77ST:b-z1n5qy) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/models.rs#L126)):*
> ```XML
>             allowed_ips: bincode::Options::serialize(make_bincode_serializer(), &value.allowed_ips)
>                 .map_err(|e| {
>                     crate::error::GatewayStorageError::TypeConversion(format!("allowed ips {e}"))
> ```

Not: 

I'd add two variants instead and map them and let the error printer do the rest:

```rs
#[error("Serialization failure for {field_key})]
Serialize { field_key: &'static str, source: bincode::Error },

#[error("Deserialization failure for {field_key})]
Deserialize { field_key: &'static str, source: bincode::Error },
```

pronebird (changes requested):
> Reviewed 4 of 47 files at r1, 1 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 17 of 48 files reviewed, 11 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/wireguard/src/peer_controller.rs` line 471 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU99smY8EE1JJ0nd-lG:-OU99smY8EE1JJ0nd-lH:b-w9711j) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_controller.rs#L471)):*
> ```XML
>     }
> 
>     // #[tokio::test]
> ```

Should we not delete the dead and commented out code from our repository? In the end we have git history to resurrect it if needed?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 17 of 48 files reviewed, 12 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/wireguard/src/peer_handle.rs` line 148 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9AjT7-j11K3vbSxOV:-OU9AjT7-j11K3vbSxOW:b-cdsd6) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_handle.rs#L148)):*
> ```XML
>             .get(&self.public_key)
>             .ok_or(Error::MissingClientKernelEntry(self.public_key.to_string()))?
>             .into();
> ```

Nit: a bit difficult to say what `.into()` does here and as such `.map(ExpectedType::from)` would give a better insight. I understand that inlay type hints probably infer it properly in the IDE that's why this is a nit comment.

pronebird (changes requested):
> Reviewed 1 of 47 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 18 of 48 files reviewed, 13 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/wireguard/src/peer_storage_manager.rs` line 28 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9BeQP2rEsrsPW6utj:-OU9BeQP2rEsrsPW6utk:bnzrhk3) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_storage_manager.rs#L28)):*
> ```XML
> }
> 
> pub struct CachedPeerManager {
> ```

Would it not be better to remove this type and work with `Option<PeerInformation>` directly at the call site? It seems that this type is superficial

pronebird (changes requested):
> Reviewed 1 of 47 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 19 of 48 files reviewed, 14 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`gateway/src/node/mod.rs` line 40 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9CAPbEGSYG2nlZzrC:-OU9CAPbEGSYG2nlZzrD:b-tdasaj) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/gateway/src/node/mod.rs#L40)):*
> ```XML
> pub use client_handling::active_clients::ActiveClientsStore;
> pub use nym_gateway_stats_storage::PersistentStatsStorage;
> pub use nym_gateway_storage::{error::GatewayStorageError, traits::*, GatewayStorage};
> ```

Perhaps this wildcard is not needed? Also would it not be better to explicitly import what needs to be in the scope?

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 19 of 48 files reviewed, 14 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @pronebird, and @simonwicky)

___
*[`common/credential-verification/src/ecash/state.rs` line 25 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU6FCV19aRs57o0UUCU:-OU9rkkICKYo4__YiFqR:b-t9o7xi) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/credential-verification/src/ecash/state.rs#L25)):*

I wonder if using `Arc<RwLock<SharedState>>` would remove the need for `dyn_clone`? I don't know much about this code but  this is something that
</blockquote></details>

I think this might be a question for @simonwicky . Is there a reason for putting `RwLock`s inside the shared state instead of on the entire state? Performance or functional reasons? I see this complicates things a bit with guards all over the place and `RwLockReadGuard` & `RwLockWriteGuard` returned by some functions, would it be possible/better to just rw lock the entire structure?
I'm a bit reluctant to change things here functionally so that we don't end up with some unexpected dead locks.
___
*[`common/gateway-storage/src/models.rs` line 126 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU96pHX3JMN4oSb77SS:-OU9ugQB0lvOY0QXls3d:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/models.rs#L126)):*

Not: 

I'd add two variants instead and map them and let the error printer do the rest:

```rs
#[error("Serialization failure for {field_key})]
Serialize { field_key: &'static str, source: bincode::Error },

#[error("Deserialization failure for {field_key})]
Deserialize { field_key: &'static str, source: bincode::Error },
```
</blockquote></details>

Done.
___
*[`common/gateway-storage/src/traits.rs` line 153 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU95ldB3cUyiR0Lj4NZ:-OU9uvL40zZXAqxD06OM:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/traits.rs#L153)):*

This does not seem right. The method returns `WireguardPeer`
</blockquote></details>

Done.
___
*[`common/gateway-storage/src/traits.rs` line 207 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU92dKr9FmfhIxxto6D:-OU9yiHaAGfBYRfuWVNf:bnyhzwh) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/gateway-storage/src/traits.rs#L207)):*

A  dozen of fields use `Arc<RwLock<_>>`.  Would it not be better to use one `Arc<RwLock<MockGatewayStorage>>` instead?
</blockquote></details>

That's a less verbose way of implementing it as well. I'll apply the changes, thanks for the suggestion.
___
*[`common/wireguard/src/peer_controller.rs` line 471 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU99smY8EE1JJ0nd-lG:-OU9z-Wz5sVEHRhQPJbD:b-3h95u5) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_controller.rs#L471)):*

Should we not delete the dead and commented out code from our repository? In the end we have git history to resurrect it if needed?
</blockquote></details>

This is a test I was thinking adding here initially, but it seems it worked better in the service provider crate. I'll remove the commented code
___
*[`common/wireguard/src/peer_storage_manager.rs` line 28 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9BeQP2rEsrsPW6utj:-OU9zX-R40Q8GAbbgMgV:b7nhg11) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_storage_manager.rs#L28)):*

Would it not be better to remove this type and work with `Option<PeerInformation>` directly at the call site? It seems that this type is superficial
</blockquote></details>

I find it easier to follow when being wrapped, although the exact reason for this wrapper is the previous implementation.
___
*[`gateway/src/node/mod.rs` line 40 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9CAPbEGSYG2nlZzrC:-OUA0P9A5UZ8rOk_NDDE:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/gateway/src/node/mod.rs#L40)):*

Perhaps this wildcard is not needed? Also would it not be better to explicitly import what needs to be in the scope?
</blockquote></details>

Done.

simonwicky (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 17 of 49 files reviewed, 14 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @neacsu, and @pronebird)

___
*[`common/credential-verification/src/ecash/state.rs` line 25 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU6FCV19aRs57o0UUCU:-OUA4k51CDzq3u3K292U:b-fuexvo) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/credential-verification/src/ecash/state.rs#L25)):*

I think this might be a question for @simonwicky . Is there a reason for putting `RwLock`s inside the shared state instead of on the entire state? Performance or functional reasons? I see this complicates things a bit with guards all over the place and `RwLockReadGuard` & `RwLockWriteGuard` returned by some functions, would it be possible/better to just rw lock the entire structure?
I'm a bit reluctant to change things here functionally so that we don't end up with some unexpected dead locks.
</blockquote></details>

While `git blame` has my name on it, it's a part of one of the "Grand ecash squasheroos" that @jstuczyn did, so I have to call him on the stand for that, as I can't answer that question

jstuczyn:
*[`common/credential-verification/src/ecash/state.rs` line 25 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU6FCV19aRs57o0UUCU:-OUE5GqlC9PAi7hpDDbh:b-x8x8b4) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/credential-verification/src/ecash/state.rs#L25)):*
<details><summary><i>Previously, simonwicky (Simon Wicky) wrote…</i></summary><blockquote>

While `git blame` has my name on it, it's a part of one of the "Grand ecash squasheroos" that @jstuczyn did, so I have to call him on the stand for that, as I can't answer that question
</blockquote></details>

it's been a while so I can't answer with 100% certainty, but I looked at the code history (including coconut iteration) and I think this is just result of evolving code without refactoring it.

i.e. initially it was just 2 strings and a client behind a lock (so no need to put the whole thing behind a lock). then additional 2 fields were added, both put behind their own locks. but in this instance it made sense as they were being accessed and updated independently and often in paralellel. and now we're at the current iteration. 

there is still some benefit for putting data and client behind separate locks (as both of them could be accessed simultanously), but if it were to simplify the code, I think the benefit of sharing it would outweight it. However, if you're going to be doing refactoring, please make sure any chain transaction you make with the client uses **WRITE** lock (even though type signature would allow you to get away with read). if you don't you will mess up your chain sequence numbers and some tx will fail.

neacsu:
> _[`common/credential-verification/src/ecash/state.rs` line 25 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU6FCV19aRs57o0UUCU:-OUE5GqlC9PAi7hpDDbh:b-x8x8b4) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/credential-verification/src/ecash/state.rs#L25)):_
> 
> Previously, simonwicky (Simon Wicky) wrote…
> it's been a while so I can't answer with 100% certainty, but I looked at the code history (including coconut iteration) and I think this is just result of evolving code without refactoring it.
> 
> i.e. initially it was just 2 strings and a client behind a lock (so no need to put the whole thing behind a lock). then additional 2 fields were added, both put behind their own locks. but in this instance it made sense as they were being accessed and updated independently and often in paralellel. and now we're at the current iteration.
> 
> there is still some benefit for putting data and client behind separate locks (as both of them could be accessed simultanously), but if it were to simplify the code, I think the benefit of sharing it would outweight it. However, if you're going to be doing refactoring, please make sure any chain transaction you make with the client uses **WRITE** lock (even though type signature would allow you to get away with read). if you don't you will mess up your chain sequence numbers and some tx will fail.

This is the reason I would refrain from doing those changes in this (already big) PR @pronebird, as unexpected things might break

pronebird (dismissed):
> Reviewed 1 of 47 files at r1, 3 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5864)* status: 20 of 49 files reviewed, 9 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, and @neacsu)

___
*[`common/wireguard/src/peer_storage_manager.rs` line 28 at r5](https://reviewable.io/reviews/nymtech/nym/5864#-OU9BeQP2rEsrsPW6utj:-OUKVzjPAhekc2IKGRYn:b6j8zmz) ([raw file](https://github.com/nymtech/nym/blob/39756a154a053b8e6bb91410cae77e70156cac30/common/wireguard/src/peer_storage_manager.rs#L28)):*

I find it easier to follow when being wrapped, although the exact reason for this wrapper is the previous implementation.
</blockquote></details>

It looks strange but I will leave it up to you

PR #5866, opened 2025-06-20
Return true remaining

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5866)


jstuczyn (approved):
nice catch!

PR #5871, opened 2025-06-25
feat: basic performance contract integration [within Nym API]

jstuczyn:
Further builds up on https://github.com/nymtech/nym/pull/5833 by integrating the performance contract into nym-api.

By setting `.performance_provider.debug.use_performance_contract_data` to true in the api's config, it will attempt to use the contract as the source of node scores for all annotations (and thus endpoints and rewarding). Note that if the contract address is not set or the contract itself is empty, the binary will not start. If the flag is not enabled, the binary will work like before, i.e. it will just use whatever it has in the storage.

For now this flag shouldn't be set, for there's one more missing piece of the puzzle to complete the whole thing. A separate process to actually submit aggregated data to the aforementioned contract.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5871)


nym-api/src/node_performance/provider/contract_provider.rs
```diff
@@ -0,0 +1,94 @@
...
+        let err = match contract_cache.node_routing_score(node_id, epoch_id) {
+            Ok(res) => return Ok(res),
+            Err(err) => err,
+        };
+
+        warn!("failed to retrieve performance score of node {node_id} for epoch {epoch_id}. falling back to at most {} past epochs", self.max_epochs_fallback);
+
+        let threshold = epoch_id.saturating_sub(self.max_epochs_fallback);
+        let start = epoch_id.saturating_sub(1);
+        for epoch_id in start..threshold {
```
durch, line 44: start is > then threshold, this loop will never loop, probably better calling it end, and iterating threshold to end.

For the iteration itself, assuming we want to find the latest available epoch we could iterate in reverse, `(threshold..=start).rev()`

nym-api/src/node_performance/contract_cache/refresher.rs
```diff
@@ -0,0 +1,135 @@
...
+                PerformanceContractEpochCacheData::from_node_performance(performance, epoch);
+            epoch_performance.insert(epoch, per_epoch_performance);
+        }
+
+        self.last_submission = Some(last_submitted);
+
+        Ok(PerformanceContractCacheData { epoch_performance })
+    }
+
+    async fn refresh(&mut self) -> Result<Option<PerformanceContractEpochCacheData>, NyxdError> {
```
durch, line 102: There might be a race condition in here, thinking of a few scenarios:

  Scenario 1: New submission between check and fetch
  1. Refresher: Gets last_submission with block_height: 1000, epoch_id: 100
  2. Refresher: Determines it's different from cached (needs update)
  3. Meanwhile: New data submitted to contract at block_height: 1001, epoch_id: 101
  4. Refresher: Fetches performance data (might get epoch 101 data)
  5. Refresher: Stores last_submission as the old value from step 1
  6. Refresher: Will think there's an update (because stored != actual) but fetch same data

  Scenario 2: Epoch transition during refresh
  1. Refresher: Gets last_submission for epoch 100
  2. Refresher: Gets current_epoch = 100
  3. Meanwhile: Epoch transitions to 101
  4. Refresher: Tries to fetch epoch 100 performance (might fail or get incomplete data)
  5. System could miss the epoch 101 data entirely

I think we should use the actual fetched data to determine if we have an update, not the metadata about when data was last submitted
Maybe we could fetch both pieces of data at the same time:

```rust
// Get current epoch                                                                                                                                                                                                       
let current_epoch = self                                                                                                                                                                                                   
       .mixnet_contract_cache                                                                                                                                                                                                 
       .current_interval()                                                                                                                                                                                                    
       .await                                                                                                                                                                                                                 
       .unwrap()                                                                                                                                                                                                              
       .current_epoch_absolute_id();                                                                                                                                                                                          
                                                                                                                                                                                                                         
// Fetch both submission info and performance data in parallel                                                                                                                                                             
let (last_submitted, performance_result) = tokio::join!(                                                                                                                                                                   
       self.nyxd_client.get_last_performance_contract_submission(),                                                                                                                                                           
       self.nyxd_client.get_full_epoch_performance(current_epoch)                                                                                                                                                             
);                                                                                                                                                                                                                   
```

Then figure out if we've already processed this and update if needed

nym-api/src/node_performance/provider/contract_provider.rs
```diff
@@ -0,0 +1,94 @@
...
+        let contract_cache = self.cached.get().await.map_err(|_| {
+            PerformanceRetrievalFailure::new(
+                *first,
+                epoch_id,
+                "performance contract cache has not been initialised yet",
+            )
+        })?;
+
+        let mut scores = HashMap::new();
+        for node_id in node_ids {
```
durch, line 87: This could be a lot (thousands) of nodes, and it grows exponentially by number of nodes and epochs, something like `O(nodes × fallback_epochs × log(total_epochs_in_cache))`. I think we should turn this on its head, at present we potentially check all nodes for all epochs, what if instead we fetched data for the latest epoch, looked up performance for all the nodes in there, the fetched the next latest, and only looked up data for the nodes that were still missing perf data, something like pseudo code below:

```
  function get_scores_for_nodes(nodes, target_epoch):
      results = {}
      remaining_nodes = set(nodes)

      // Check each epoch ONCE
      for epoch from target_epoch down to (target_epoch - 12):
          if remaining_nodes is empty:
              break  // Found everyone!

          if epoch_data[epoch] exists:
              // Check ALL remaining nodes against this epoch
              for node in remaining_nodes:
                  if epoch_data[epoch] has node:
                      results[node] = epoch_data[epoch][node]
                      remove node from remaining_nodes

      // Mark any unfound nodes as errors
      for node in remaining_nodes:
          results[node] = ERROR("Not found")

      return results
```

durch (changes requested):
I've got one bug, and two suggestion. Apart from fixing the bug, I think implementing the performance optimisation might be worth while, not so sure on the race condition, maybe I'm just a bit pedantic/paranoid here

nym-api/src/node_performance/provider/contract_provider.rs
jstuczyn: such a stupid mistake on my end. nice catch!

nym-api/src/node_performance/contract_cache/refresher.rs
jstuczyn: > I think we should use the actual fetched data to determine if we have an update, not the metadata about when data was last submitted

the problem with this approach is that we'd have to fetch a lot of data, so that's why I introduced the metadata to have a quick way of seeing if there has been any updates since the last time it was polled. remember that we really don't care about data from the past epochs and are only interested if the most recent one is updated. on startup we pull the full information for the however many fallback epochs we care about and then in the loop we check "has there been anything new submitted? if so, update **current** epoch" - if the update was for some previous epochs. tough luck. they should have submitted it earlier (and that data would have only ever been used as fallback anyway). in 99% cases network monitors wouldn't be running behind and would actually be submitting only for the current epoch.

with that in mind, let's consider those race condition scenarios:
1) I'm assuming in this scenario the current epoch is `101`. so we'll just refresh `101` instead of getting anything for `100`. it will be slightly inefficient since we'll just refresh what we already had, but it shouldn't hurt at all.
2) so epoch will transition to `101` and we'll potentially get empty data for the current epoch. but in this instance, when anybody queries for "current" performance, we will fallback to information we already had on `100`. Also, this is an incredibly unlikely scenario as network monitors are meant to (well, once implemented) submit performance measurements for epochs BEFORE the actual transition takes place

also, fetching those at the same time wouldn't have solved anything as in the world of the internet one of those would have been resolved first anyway (plus `get_full_epoch_performance` is actually multiple paged queries)

nym-api/src/node_performance/provider/contract_provider.rs
jstuczyn: as discussed it's actually not that big of a problem : )

durch (approved):

PR #5872, opened 2025-06-25
Set busy_timeout in sqlx

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5872)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/cli/mod.rs
+++ b/nym-node-status-api/nym-node-status-api/src/cli/mod.rs
@@ -45,6 +45,10 @@ pub(crate) struct Cli {
     #[clap(long, env = "DATABASE_URL")]
     pub(crate) database_url: String,
 
+    #[clap(long, default_value = "5", env = "SQLX_BUSY_TIMEOUT_S")]
+    #[arg(value_parser = parse_duration)]
+    pub(crate) sqlx_busy_timeout_s: Duration,
+
     #[clap(
         long,
         default_value = "300",
--- a/nym-node-status-api/nym-node-status-api/src/db/mod.rs
+++ b/nym-node-status-api/nym-node-status-api/src/db/mod.rs
@@ -1,10 +1,11 @@
 use anyhow::{anyhow, Result};
 use sqlx::{
     migrate::Migrator,
+    query,
     sqlite::{SqliteAutoVacuum, SqliteConnectOptions, SqliteSynchronous},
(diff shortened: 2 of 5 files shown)


nym-node-status-api/nym-node-status-api/src/cli/mod.rs
```diff
@@ -45,6 +45,10 @@ pub(crate) struct Cli {
     #[clap(long, env = "DATABASE_URL")]
     pub(crate) database_url: String,
 
+    #[clap(long, default_value = "5", env = "SQLX_BUSY_TIMEOUT_S")]
+    #[arg(value_parser = parse_duration)]
+    pub(crate) sqlx_busy_timeout_s: Duration,
```
jstuczyn, line 50: nit: you could provide parsing via `humantime` crate for more versatile options (so say either `1s` or `500ms` would be valid options)

also, doesn't 5s seem a bit excessive? I mean if it works, then it works. it just seems like it's quite a big value

nym-node-status-api/nym-node-status-api/src/db/mod.rs
```diff
@@ -33,11 +35,37 @@ impl Storage {
 
         MIGRATOR.run(&pool).await?;
 
+        // aftering setting pragma, check whether it was set successfully
+        Self::assert_busy_timeout(pool.clone(), busy_timeout.as_secs() as i64).await?;
```
jstuczyn, line 39: why wouldn't it be?

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/db/mod.rs
dynco-nym: some settings aren't applied globally, but on a per-connection basis. E.g. 
- if you `sqlite3` into a running DB, then run 
- `PRAGMA busy_timeout=5`
- then exit the DB and log in again (different connection), `busy_timeout` don't be the value you set. 

`sqlx` should apply this to the whole connection pool, but this check ensures it

nym-node-status-api/nym-node-status-api/src/cli/mod.rs
dynco-nym: 5s is supposedly the [default set by sqlx](https://docs.rs/sqlx/0.7.4/sqlx/sqlite/struct.SqliteConnectOptions.html#method.busy_timeout) so that was my reference

PR #5876, opened 2025-06-30
bugfix: key-rotation + reply SURBs

jstuczyn:
this PR should hopefully, probably, maybe, I think resolve the issue of suddenly running out of usable reply SURBs when a new key rotation occurs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5876)


pronebird:
It would be better to replace log with tracing or run clippy in a separate PR to avoid such a massive bloat.

jstuczyn:
> It would be better to replace log with tracing or run clippy in a separate PR to avoid such a massive bloat.

in hindsight that's what I should I have done. I got annoyed by red CI and just fixed the issues here.

benedettadavico (approved):

PR #5880, opened 2025-07-03
Allow PG database backend

durch:
Summary

  - Added PostgreSQL database support alongside existing SQLite through Cargo feature flags
  - Implemented runtime query conversion from SQLite ? placeholders to PostgreSQL $1, $2, ... format
  - Single codebase now supports both databases without query duplication

  Implementation

  - Created query wrapper functions (query(), query_as(), query_scalar()) that automatically convert placeholders
  - Converted all sqlx macro calls to use the wrapper functions
  - Added type conversions where needed (SQLite i64 → PostgreSQL i32)
  - Feature-gated database-specific SQL syntax (RETURNING clauses, INSERT OR IGNORE vs ON CONFLICT)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5880)


nym-node-status-api/nym-node-status-api/src/db/queries/gateways.rs
```diff
@@ -153,15 +152,54 @@ pub(crate) async fn insert_gateway_description(
...
+            performance, 
+            self_described, 
+            last_updated_utc
+        ) VALUES (?, ?, ?, ?, ?)
+        RETURNING id"#,
+    )
+    .bind(gateway_identity_key.to_string())
+    .bind(true) // Assume bonded since being tested
+    .bind(0) // Initial performance
+    .bind("null")
```
dynco-nym, line 199: isn't this field NOT NULL on `gateways` table?

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
```diff
@@ -170,6 +183,72 @@ async fn submit_testrun(
...
+            process_testrun_submission(testrun, submission.payload, &mut conn).await
+        }
+        Err(_) => {
+            // External testrun - create records
+            tracing::info!(
+                "Creating external testrun {} for gateway {}",
+                submitted_testrun_id,
+                submission.payload.gateway_identity_key
+            );
+
```
dynco-nym, line 229: why do we create a gateway here? is it because this NS instance might accept testruns that were issued by another NS so its gateway might not be present?

dynco-nym (commented):
CI is failing due to clippy, but I couldn't fix it locally because `cargo build` is failing

dynco-nym:
great work BTW, looking forward to having this merged!

nym-node-status-api/nym-node-status-api/src/http/api/testruns.rs
durch: Yes, the NS API can now accept testruns created by other NS APIs

nym-node-status-api/nym-node-status-api/src/db/queries/gateways.rs
durch: We can probably remove that, will take a peek

PR #5896, opened 2025-07-16
sqlx-pool-guard: allocate more memory on windows

pronebird:
Sentry picked up the following error on Windows:

> imp::check_files_closed() failure: The specified information record length does not match the length required for the specified information class. (os error -805306364)

I have concluded that It's possible for the open file handles table to grow between calls to `NtQuerySystemInformation` resulting in the buffer being too small two times in a row.

In order to alleviate this, we should take a system returned length as a hint and allocate a bit larger buffer to avoid stumbling upon the same error on the next call.

Changes in this PR:

- Allocate 1.5x more memory than reported by the system to provide a safety margin
- Increase number of retry attempts to 5

JIRA-VPN-3688

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5896)

### DIFF
--- a/sqlx-pool-guard/src/windows.rs
+++ b/sqlx-pool-guard/src/windows.rs
@@ -31,40 +31,16 @@ const SYSTEM_HANDLE_INFORMATION_CLASS: SYSTEM_INFORMATION_CLASS = SYSTEM_INFORMA
 /// The number is based on what I observe on a pretty standard Windows 11
 const SYSTEM_HANDLE_INFORMATION_INITIAL_SIZE: usize = 2_500_000;
 
+/// Max retry attempts for querying system handle information before giving up
+const MAX_RETRY_ATTEMPTS: u32 = 5;
+
 /// Check if there are no open handles to the given files.
 ///
 /// Uses undocumented NT API to obtain open handles on the system.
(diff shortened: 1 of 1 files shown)


sqlx-pool-guard/src/windows.rs
```diff
@@ -106,6 +82,43 @@ pub async fn check_files_closed(file_paths: &[&Path]) -> io::Result<bool> {
...
+                handle_table_info.as_mut_ptr() as _,
+                return_len,
+                &mut return_len,
+            )
+        };
+
+        // Buffer is too small, resize memory and retry again.
+        if status == STATUS_INFO_LENGTH_MISMATCH {
+            // Allocate a bit more memory since the size of table can change between calls
+            let resize_to = (return_len as usize) * 3 / 2;
```
durch, line 106: Minor suggestion, make these magic numbers constants.

durch (approved):

sqlx-pool-guard/src/windows.rs
pronebird: Given that expression is evaluated left to right, one constant wouldn't cut it because `3 / 2 = 1` in integer math. I could use floating point to express the 1.5x multiplier but that's a different ballgame :)

PR #5899, opened 2025-07-16
Feature/dkg epoch dealers query

jstuczyn:
NET-499

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5899)


contracts/coconut-dkg/src/dealers/queries.rs
```diff
@@ -82,8 +82,37 @@ pub fn query_dealers_indices_paged(
     Ok(PagedDealerIndexResponse::new(dealers, start_next_after))
 }
 
-pub fn query_current_dealers_paged(
+pub fn query_epoch_dealers_addresses_paged(
```
neacsu, line 85: Can we unit test this?

jstuczyn: we didn't have unit tests for any existing queries so I was being lazy and haven't written any. but you're absolutely right. I will add some.

neacsu (approved):

PR #5900, opened 2025-07-17
Feature/dkg snapshot epoch

jstuczyn:
NET-501

> In order to determine if singer quorum has been down at particular height, we need to know with certainty the dkg epoch id corresponding to given block height.

this PR makes it possible. every time epoch state is changed (due to DKG progress), snapshot is saved and can be queried. this doesn't work for past data, but given mainnet has only had a single DKG instance, that's not an issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5900)

### DIFF
--- a/common/client-libs/validator-client/src/nyxd/contract_traits/dkg_query_client.rs
+++ b/common/client-libs/validator-client/src/nyxd/contract_traits/dkg_query_client.rs
@@ -41,6 +41,11 @@ pub trait DkgQueryClient {
         self.query_dkg_contract(request).await
     }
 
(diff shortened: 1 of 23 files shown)


neacsu (approved):
> Reviewed 19 of 20 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5900)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jstuczyn)

PR #5910, opened 2025-07-23
chore: allow compatibility with 'CDLA-Permissive-2.0'

jstuczyn:
this license is present in the included `webpki-roots`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5910)


pronebird (approved):
Don't know much about that license but it looks like a liability disclaimer more than a license.

PR #5915, opened 2025-07-28
Wireguard private metadata

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5915)


common/credential-verification/src/lib.rs
```diff
@@ -139,3 +140,15 @@ impl CredentialVerifier {
             .await)
     }
 }
+
+#[async_trait]
+pub trait Verifier {
+    async fn verify(&mut self) -> Result<i64>;
```
jstuczyn, line 149: nit: maybe document what this i64 represents, because it's very much not obvious from the method name. after you  verify _something_ you get back an integer

jstuczyn: another nit: perhaps rename that `Verifier` to I'd say `CredentialVerifier` or `TicketVerifier` to be more explicit about what we're checking here

nym-node/nym-node-requests/src/api/v1/gateway/models.rs
```diff
@@ -17,7 +17,10 @@ pub struct Gateway {
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct Wireguard {
     #[cfg_attr(feature = "openapi", schema(example = 51822, default = 51822))]
-    pub port: u16,
+    pub tunnel_port: u16,
```
jstuczyn, line 24: changing those types will cause incompatibility with older versions of software that query for this information.

You will have to unfortunately retain and duplicate the `port` field (but deprecate it)

gateway/src/node/mod.rs
```diff
@@ -492,14 +501,42 @@ impl GatewayTasksBuilder {
...
+            ),
+        ));
+
+        let bind_address = std::net::SocketAddr::new(
+            wireguard_data.inner.config().private_ipv4.into(),
+            wireguard_data.inner.config().announced_metadata_port,
+        );
+        let server = router.build_server(&bind_address).await?;
+
+        let cancel_token: tokio_util::sync::CancellationToken = (*self.shutdown_token).clone();
```
jstuczyn, line 517: why are you dereferencing the shutdown token to tokio's cancellation token? is something missing in the base API? (would love to add it if that's the case : D)

common/wireguard/src/peer_controller.rs
```diff
@@ -192,7 +204,21 @@ impl PeerController {
...
     }
 
-    async fn handle_query_peer(&self, key: &Key) -> Result<Option<Peer>> {
+    async fn ip_to_key(&self, ip: IpAddr) -> Result<Option<Key>> {
+        Ok(self
+            .wg_api
+            .read_interface_data()?
+            .peers
+            .values()
+            .find_map(|peer| {
```
jstuczyn: is there no better way than to iterate through all the peers? because if we have to perform this lookup frequently it might get relatively costly depending on how many peers (and their allowed ips) we're expecting to have

common/wireguard-private-metadata/src/error.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
+pub enum Error {
```
jstuczyn: nit: can you add some prefix to the Error name? it's very confusing to have so many `Error` enums all over the place.

common/wireguard-private-metadata/src/http/router.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// Middleware added here intercepts the request before it gets to other routes.
+    fn finalize_routes(self) -> Router<AppState> {
+        self.unfinished_router
+            .layer(setup_cors())
+            .layer(axum::middleware::from_fn(log_request_info))
+    }
+}
+
+fn setup_cors() -> CorsLayer {
```
jstuczyn, line 60: do we actually need cors here?

common/wireguard-private-metadata/src/models/version_1.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl TryFrom<InnerAvailableBandwidthResponse> for AvailableBandwidthResponse {
+    type Error = Error;
+
+    fn try_from(value: InnerAvailableBandwidthResponse) -> Result<Self, Self::Error> {
+        Ok(AvailableBandwidthResponse {
+            version: VERSION,
+            inner: bincode::serialize(&value)?,
```
jstuczyn: rather  than doing implicit `bincode::serialize` with default config, would you mind explicitly creating the serializer instantiated with specific config? The reason for it is that eventually we'll want to migrate to bincode 2.0 and some defaults had changed there making it easy to miss some of those cases

common/wireguard-private-metadata/src/models/version_1.rs
```diff
@@ -0,0 +1,157 @@
...
+            });
+        }
+        Ok(bincode::deserialize(&value.inner)?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
jstuczyn, line 87: relating to my previous comment, if it's not too much of a hassle, mind adding extra test cases to make sure the serialisation matches expected bytes?

common/wireguard-private-metadata/src/network.rs
```diff
@@ -0,0 +1,77 @@
...
+    extract::{ConnectInfo, Query, State},
+    Json, Router,
+};
+use nym_http_api_common::{FormattedResponse, OutputParams};
+use tower_http::compression::CompressionLayer;
+
+use crate::{
+    http::state::AppState,
+    models::{
+        latest::InnerTopUpRequest, AvailableBandwidthResponse, AxumErrorResponse, AxumResult,
```
jstuczyn, line 16: rather than using `latest::....` use `v1::..` please as if you change anything in serialisation (add new field, rename it, etc.) the old endpoints will no longer be usable since bincode is not self-describing (that's why we had to add some `v2` endpoints to nym-api due because we had to add some fields to skimmed endpoints...)

common/wireguard-private-metadata/src/network.rs
```diff
@@ -0,0 +1,77 @@
...
+)]
+async fn available_bandwidth(
+    ConnectInfo(addr): ConnectInfo<SocketAddr>,
+    Query(output): Query<OutputParams>,
+    State(state): State<AppState>,
+) -> AxumResult<FormattedResponse<AvailableBandwidthResponse>> {
+    let output = output.output.unwrap_or_default();
+    Ok(output.to_response(
+        state
+            .available_bandwidth(addr.ip())
```
jstuczyn, line 47: just thinking from security perspective here. what if somebody spoofed somebody else's ip address. would it cause problems to be able to expose data of other users?
if so, could we somehow authenticate this endpoint? maybe use shared key or something?

nym-node/nym-node-requests/src/api/v1/gateway/models.rs
neacsu: I was just trying to figure out all those models and how will I do to expose the new port to the vpn clients. I guess I also need to modify `WireguardDetails` and also keep and deprecate the `port` field there. Really difficult to track and do changes to those types so information can reach all the way to the vpn clients

gateway/src/node/mod.rs
neacsu: Because I need the `CancellationToken` from the `ShutdownToken`, and the only extractor I could see there is `deref`. The other easier way would be to add an `inner` method there that just gives direct access. Do you want me to add that and use it instead?

common/wireguard/src/peer_controller.rs
neacsu: Right, I think I can fit the information into `bw_storage_managers` hashmap instead

common/wireguard-private-metadata/src/http/router.rs
neacsu: I am not sure, possibly not? I am not very knowledgeable about cors, I've seen it on our other APIs and put it here as well, if it's not harming in any way I don't see why remove it.

common/wireguard-private-metadata/src/models/version_1.rs
neacsu: Sure, I forgot about this, will make the changes 😄

common/wireguard-private-metadata/src/models/version_1.rs
neacsu: No hassle, will add checks

common/wireguard-private-metadata/src/network.rs
neacsu: In https://github.com/nymtech/nym/pull/5943 I extend on versioning, so there `latest` makes more sense in order to make it more clear what code needs to be added/updated when doing a new version. It won't be let to the whims of `bincode`, no worries :D

common/wireguard-private-metadata/src/network.rs
neacsu: The endpoint is implicitly authenticated because it's over the private wireguard network, which maps source IPs to public keys, so a spoofed source private IP would not be able to encrypt those messages with the corresponding keypair. And that wireguard IP is itself allocated after authentication happens in the ol' authenticator.

common/wireguard/src/peer_controller.rs
pronebird: allowed ips represent a mask AFAIK so you might want to build a binary tree to work with IP ranges.

pronebird (commented):
> Reviewed 3 of 37 files at r1, 3 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5915)* status: 6 of 39 files reviewed, 10 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @neacsu)

common/wireguard-private-metadata/src/network.rs
pronebird: I am planning on organizing settings versioning in the daemon the same way. All actual code refers to models via latest. Note sure if that's the same here

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5915)* status: 6 of 39 files reviewed, 10 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @neacsu)

pronebird (commented):
> Reviewed 9 of 37 files at r1, 4 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5915)* status: 19 of 39 files reviewed, 10 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @neacsu)

pronebird (commented):
> Reviewed 1 of 37 files at r1, 1 of 6 files at r2, 2 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5915)* status: 23 of 39 files reviewed, 10 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @neacsu)

jstuczyn (approved):

PR #5943, opened 2025-08-07
Wireguard metadata client library

neacsu:
Depends on https://github.com/nymtech/nym/pull/5915 .  I will rebase on top of develop once that PR is merged there

Tested end to end using https://github.com/nymtech/nym-vpn-client/pull/3316 in canary, with a pair of gateways that has the changes deployed:

```
nym-vpnc connect --enable-two-hop  --entry-gateway-id 7CWjY3QFoA9dgE535u9bQiXCfzgMZvSpJu842GA1Wn42 --exit-gateway-id 7fp3cmzCvgeRgbB1ycTnK6RokjHNqPmCCSBG23gyxshj
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5943)


common/wireguard-private-metadata/Cargo.toml
```diff
@@ -25,19 +26,36 @@ tower-http = { workspace = true, features = [
...
     "output",
-] }
-nym-wireguard = { path = "../wireguard" }
+], optional = true }
+nym-wireguard = { path = "../wireguard", optional = true }
 
-[dev-dependencies]
-async-trait = { workspace = true }
+[features]
+default = ["server", "client"]
```
jstuczyn: conceptually it feels weird that by default the library enables both server and client features. consumers of this code will only use one or the other

jstuczyn: plus the existence of those features makes everything look a bit chaotic. would it perhaps make more sense to split this crate into 3 subcrates: server, client and shared?

common/wireguard-private-metadata/src/http/state.rs
```diff
@@ -21,23 +17,17 @@ impl AppState {
...
+    pub(crate) async fn available_bandwidth(&self, ip: IpAddr) -> Result<i64, Error> {
+        self.transceiver.query_bandwidth(ip).await
     }
 
     pub(crate) async fn topup_bandwidth(
         &self,
         ip: IpAddr,
         credential: CredentialSpendingData,
-    ) -> Result<(), Error> {
+    ) -> Result<i64, Error> {
```
jstuczyn: nit: add comment to show what that i64 represents

common/wireguard-private-metadata/src/lib.rs
```diff
@@ -1,22 +1,91 @@
...
+
+    use crate::{
+        client::WireguardMetadataApiClient, models::ErrorResponse,
+        transceiver::tests::MockVerifier, AppState, PeerControllerTransceiver, RouterBuilder,
+    };
+
+    pub(crate) const VERIFIER_AVAILABLE_BANDWIDTH: i64 = 42;
+
+    pub(crate) async fn spawn_server_and_create_client() -> Client {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
```
jstuczyn: rather than actually spawning server on the socket, wouldn't it make more sense to use `TestServer` from [axum_test](https://docs.rs/axum-test/latest/axum_test/) ? this way you wouldn't need to depend on IO/networking

jstuczyn: we have an example of that in [nym-api](https://github.com/nymtech/nym/blob/dc59149a5dd3bc8a61a70b1ad80de83b98fe1b7c/nym-api/src/ecash/tests/mod.rs#L1367) as a matter of fact I think you were the one who originally introduced it : D

common/wireguard-private-metadata/Cargo.toml
```diff
@@ -25,19 +26,36 @@ tower-http = { workspace = true, features = [
...
+    "tower-http",
+    "utoipa",
+    "utoipauto",
+    "utoipa-swagger-ui",
+    "nym-credentials-interface",
+    "nym-credential-verification",
+    "nym-http-api-common",
+    "nym-wireguard",
+]
+client = ["nym-http-api-client"]
```
jstuczyn: mind also adding 
```toml
[lints]
workspace = true
```

at the bottom? ideally that'd be included in any new crate we make : )

common/wireguard-private-metadata/src/lib.rs
neacsu: Indeed X)

neacsu: Looking more into this, it seems it won't be possible to use the mocked transport of `TestServer`, because the routes need the source IP internally (to map to the wireguard key), and that is only available with real TCP stack. Also, I like this approach a bit more since I can link directly to the client implementation and basically unit test server and client at the same time (although it would have also been possible to do that, if there were a mocked client structure that would have implemented `ApiClientCore`, but again, TCP/IP stack is needed in this case)

see https://docs.rs/axum-test/latest/axum_test/struct.TestServerConfig.html#structfield.transport
```
The default is to use mocking, apart from services built using [axum::extract::connect_info::IntoMakeServiceWithConnectInfo](https://docs.rs/axum/0.8.4/x86_64-unknown-linux-gnu/axum/extract/connect_info/struct.IntoMakeServiceWithConnectInfo.html) (this is because it needs a real TCP stream).
```

neacsu: Fyi following my dive into this, although for this case it is not useful, it might be for other unit tests that don't need IP stack: https://github.com/nymtech/nym/pull/5956

common/wireguard-private-metadata/Cargo.toml
neacsu: I've separated into 3 subcrates + a testing subcrate

pronebird (dismissed):
> Reviewed 31 of 44 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5943)* status: 31 of 45 files reviewed, 5 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @neacsu)

___
*[`common/wireguard-private-metadata/shared/src/models/mod.rs` line 41 at r2](https://reviewable.io/reviews/nymtech/nym/5943#-OY0iyRo-aeOGLwCAsTj:-OY0iyRo-aeOGLwCAsTk:b-gdwyv8) ([raw file](https://github.com/nymtech/nym/blob/b3c777279e0a8b4ff5eb60eda96092b261b645a1/common/wireguard-private-metadata/shared/src/models/mod.rs#L41)):*
> ```XML
> }
> 
> pub trait Extract<T> {
> ```

This looks familiar to what axum already offers:

https://docs.rs/axum/latest/axum/extract/index.html
Can`FromRequest` and `IntoResponse` be leveraged for automatic ser/deserialization?

common/wireguard-private-metadata/Cargo.toml
neacsu: Done.

neacsu (commented):
> @neacsu dismissed @jstuczyn and @pronebird from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/5943)* status: 24 of 48 files reviewed, 2 unresolved discussions (waiting on @durch, @dynco-nym, @jstuczyn, @mfahampshire, and @pronebird)

___
*[`common/wireguard-private-metadata/shared/src/models/mod.rs` line 41 at r2](https://reviewable.io/reviews/nymtech/nym/5943#-OY0iyRo-aeOGLwCAsTj:-OY5qk7429_Ly6a8bo30:bj1pong) ([raw file](https://github.com/nymtech/nym/blob/b3c777279e0a8b4ff5eb60eda96092b261b645a1/common/wireguard-private-metadata/shared/src/models/mod.rs#L41)):*

This looks familiar to what axum already offers:

https://docs.rs/axum/latest/axum/extract/index.html
Can`FromRequest` and `IntoResponse` be leveraged for automatic ser/deserialization?
</blockquote></details>

I wouldn't over automate this, I prefer having more control over extraction/construction functions, as this might not be solely based on typing, but also on the client/server logic (i.e. at some point, the lack of credential in the request wasn't considered an error, but rather it was used for free usage mode).

jstuczyn (approved):

PR #5948, opened 2025-08-11
fix: use WASM compatible time API in client

elsirion:
While working on a [PWA in rust that spawns a Nym client](https://github.com/elsirion/broadnym/blob/ecedf7769d9d6734cf8ece399a7b63a155ef6383/client/src/main.rs#L253-L278) I noticed that it crashed right after client initialization could finish due to time APIs being used that aren't available in WASM. These changes fixed it for me, although there might be further problematic places (I did not do a full sweep of the code base).

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5948)


jstuczyn (approved):

elsirion:
Sorry for missing formatting :/

PR #5955, opened 2025-08-14
chore: updated refs to cheddar rev of nym repo

jstuczyn:
CC @dynco-nym and @simonwicky in case there were some specific, non-obvious, reasons node-status and statistics-api were still pointing to cheddar

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5955)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6967,7 +6967,7 @@ dependencies = [
 
 [[package]]
 name = "nym-statistics-api"
-version = "0.1.4"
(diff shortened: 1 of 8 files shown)


simonwicky (approved):
It was pointing to cheddar because at the time, the `validator-client` from `develop` had types that were different from the deployed API. I guess it's no longer the case

dynco-nym (approved):

PR #5957, opened 2025-08-14
Feature/credential proxy jwt

jstuczyn:
This PR is part of the 'Upgrade Mode' ([NET-341](https://nymtech.atlassian.net/browse/NET-341)) that should allow usage of the network in a situation where ecash credentials are unissuable, because, for example, we have lost signing quorum (i.e. we have fewer than the required number of threshold signers responding to requests).

It partially implements [NET-448](https://nymtech.atlassian.net/browse/NET-488), however, this version is more naive. Instead requesting actual 'emergency credentials' that would have been issued by a subset of ecash signers, the credentials proxy creates a JWT, signed with its key, attesting the upgrade mode has been enabled.

In the current iteration it is expected to work as follows
- there exists an external endpoint hosting an `attestation.json` file that dictates whether the upgrade mode has been enabled. this is expected to be under Nym's control and signed with the corresponding key
- credential proxy periodically queries this endpoint to check whether the upgrade mode has been enabled
- if so, a JWT is signed and stored in credential proxy's state
- now, any subsequent requests for obtaining credential shares will instead return the JWT instead of the actual shares (since they couldn't have been obtained anyway)

this is as far as the flow goes in this PR. once fully implemented (via subsequent changes), it will continue as follows:
- once JWT is obtained by the client, it will stop its internal routine for toping up bandwidth
- it will then send the JWT alongside the actual published attestation to its target gateway
- gateways will have their own watchers for the attestation, but in case there's been a desync, they will do an expedited check once they receive attestation from one of its clients.
- if gateway confirms upgrade mode, it will stop bandwidth metering and keep returning constant base amount of bandwidth for legacy clients (that don't understand upgrade mode), so that they would not attempt to spend any credentials)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5957)

[NET-341]: https://nymtech.atlassian.net/browse/NET-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NET-448]: https://nymtech.atlassian.net/browse/NET-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


common/upgrade-mode-check/src/attestation.rs
```diff
@@ -1,30 +1,32 @@
...
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Copy)]
 #[serde(tag = "type")]
 #[serde(rename = "upgrade_mode")]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct UpgradeModeAttestationContent {
     #[serde(with = "time::serde::timestamp")]
+    #[cfg_attr(feature = "openapi", schema(value_type = i64))]
     pub starting_time: OffsetDateTime,
```
mmsinclair: any chance we could make it an RFC 3339 string instead?

nym-credential-proxy/nym-credential-proxy/src/attestation_watcher.rs
```diff
@@ -0,0 +1,94 @@
...
+use tokio::time::Instant;
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info};
+use url::Url;
+
+pub struct AttestationWatcher {
+    // default polling interval
+    regular_polling_interval: Duration,
+
+    // expedited polling interval once upgrade mode is detected
```
mmsinclair, line 19: Nice, like this 😁

nym-credential-proxy/nym-credential-proxy/src/attestation_watcher.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    async fn try_update_state(&self) {
+        match attempt_retrieve_attestation(
+            self.attestation_url.as_str(),
+            Some(generate_user_agent!()),
+        )
+        .await
+        {
+            Err(err) => error!("failed to retrieve attestation information: {err}"),
```
mmsinclair: when the upgrade mode is not on, there will be nothing at this URL, so there will be lots of "errors" in the log saying there is no attestation mode

think there are two options here:
- change this to `info` and make message say: "there is no upgrade mode attestation"
- have an attestation at that location that says "everything is fine, upgrade mode is not on"

mmsinclair (changes requested):
- in the attestation add:
  - trusted JWT issuer pubkeys

mmsinclair (approved):

PR #5974, opened 2025-08-22
Domain fronting integration

durch:
Summary

This PR completes the migration from the NymApiClient to the nym_http_api_client::Client across the entire Nym monorepo, enabling domain fronting support and providing a consistent HTTP client interface throughout the codebase.

Changes

Core HTTP Client Unification

- Removed the Client re-export from nym_api module that was causing namespace conflicts
- Migrated all code from NymApiClient to the unified nym_http_api_client::Client
- Extended NymApiClientExt trait with missing methods to maintain backward compatibility:
    - get_current_rewarded_set() - for fetching epoch rewarded sets
    - get_all_basic_nodes_with_metadata() - for retrieving all network nodes
    - get_all_described_nodes() - for getting node descriptions
    - get_all_basic_active_mixing_assigned_nodes_with_metadata() - for active mixnodes
    - get_all_basic_entry_assigned_nodes_with_metadata() - for entry gateways
    - get_all_expanded_nodes() - for detailed node information
    - change_base_urls() - for API endpoint rotation
    - nodes_by_addresses() - for querying nodes by IP addresses

Domain Fronting Support

- Unified HTTP client now supports domain fronting configuration
- Client can rotate between multiple API endpoints for resilience (this needs extra review, I'm not sure its right)
- Proper URL type conversions implemented throughout

Crate Updates

- Added nym-http-api_client dependency to:
  - nym-validator-rewarder
  - nym-api
  - nym-network-monitor
  - nym-ecash-signer-check
  - testnet-manager
  - validator-status-check
  - nym-sdk
  - nym-client-core
- Added nym-mixnet-contract-common to nym-client-core for type conversions

Type System Improvements

- Fixed RewardedSetResponse → EpochRewardedSet → CachedEpochRewardedSet conversion chain
- Implemented proper error type conversions using From trait for NymAPIError
- Updated all error handling to use RequestError::new() constructor instead of direct field access

Component-Specific Changes

nym-node

  - Updated NodesQuerier to use unified client
  - Fixed API endpoint rotation logic in NymApisClient
  - Removed deprecated NymApiClient::from() usage
  - Updated broadcast functionality to work with new client

nym-api-provider

  - Refactored NymApiTopologyProvider to use unified client
  - Added support for multiple API URLs with automatic rotation on failure
  - Properly handles bincode encoding preference (though runtime toggle not yet supported)

validator-client

  - Consolidated client implementation to use unified HTTP client
  - Removed duplicate client instantiation code
  - Improved error handling and type safety

SDK and Examples

  - Updated custom_topology_provider example to use new client API
  - Fixed import paths and type conversions
  - Documentation updated to reflect new patterns

Error Handling

  - Consistent error construction using RequestError::new()
  - Proper error propagation with ? operator
  - Better error messages for debugging

Migration Notes

For code using the old NymApiClient:
```
// Before
let client = NymApiClient::new(url);
client.change_nym_api(new_url);

// After
let client = nym_http_api_client::Client::builder(url)?.build()?;
client.change_base_urls(vec![new_url]);

The NymApiClientExt trait must be imported to access Nym-specific methods:
use nym_validator_client::nym_api::NymApiClientExt;
```

Breaking Changes

- NymApiClient is now deprecated (though still available for backward compatibility)
- Client is no longer re-exported from nym_api module
- Some method signatures changed to use &mut self instead of &self (e.g., change_base_urls)

Future Work

- Complete removal of deprecated NymApiClient in next major version
- Runtime bincode toggle support in unified client
- Further optimization of paginated API calls

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5974)


common/http-api-client/src/lib.rs
```diff
@@ -742,6 +824,13 @@ impl ApiClientCore for Client {
...
         let mut rb = RequestBuilder::from_parts(self.reqwest_client.clone(), req);
 
+        // Set Accept header based on serialization preference
+        let accept_header = match self.serialization {
+            SerializationFormat::Json => "application/json",
+            SerializationFormat::Bincode => "application/bincode",
+        };
+        rb = rb.header(reqwest::header::ACCEPT, accept_header);
+
         if let Some(body) = json_body {
```
jstuczyn: this would also have to be conditional in this instance. if we decided on `SerializationFormat::Bincode` then setting json body makes no sense

common/http-api-client/src/lib.rs
```diff
@@ -742,6 +824,13 @@ impl ApiClientCore for Client {
 
         let mut rb = RequestBuilder::from_parts(self.reqwest_client.clone(), req);
 
+        // Set Accept header based on serialization preference
+        let accept_header = match self.serialization {
```
jstuczyn: would it perhaps be cleaner to add method on `SerializationFormat` to return `Option<&str>` (or whatever type is appropriate) for the header? especially if we decided to add more variants

common/http-api-client/src/lib.rs
```diff
@@ -191,6 +193,15 @@ pub type Params<'a, K, V> = &'a [(K, V)];
 /// Empty collection of HTTP Request Parameters.
 pub const NO_PARAMS: Params<'_, &'_ str, &'_ str> = &[];
 
+/// Serialization format for API requests and responses
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SerializationFormat {
```
jstuczyn, line 219: if it's not too much of a hassle, for completion sake I'd also include yaml here (as it exists on `FormattedResponse`.

Perhaps also a good old `text` as sometimes you might want to send a string body without any further serialisation

common/http-api-client/src/lib.rs
```diff
@@ -742,6 +824,13 @@ impl ApiClientCore for Client {
 
         let mut rb = RequestBuilder::from_parts(self.reqwest_client.clone(), req);
 
+        // Set Accept header based on serialization preference
+        let accept_header = match self.serialization {
+            SerializationFormat::Json => "application/json",
+            SerializationFormat::Bincode => "application/bincode",
+        };
+        rb = rb.header(reqwest::header::ACCEPT, accept_header);
```
jstuczyn: would we also have to set `CONTENT_TYPE` (for anything other than `json`, as reqwest is already doing that)?

common/http-api-client/src/lib.rs
```diff
@@ -370,6 +381,7 @@ pub struct ClientBuilder {
     front: Option<fronted::Front>,
 
     retry_limit: usize,
+    serialization: SerializationFormat,
```
jstuczyn, line 558: the more I think about it, the more I wonder: should it be property of a client or a request? because i could conceivably see a situation where you might want to connect to nym-api and requests some data as json and some as bincode

common/http-api-client/src/lib.rs
```diff
@@ -693,26 +764,37 @@ impl Client {
...
+
+                tracing::debug!(
+                    "Domain fronting enabled: routing via CDN {} to actual host {}",
+                    front_host,
+                    actual_host
+                );
+
                 // this should never fail as we are transplanting the host from one url to another
-                r.url_mut().set_host(url.front_str()).unwrap();
+                r.url_mut().set_host(Some(front_host)).unwrap();
```
jstuczyn: wouldn't this blow up if we have invalid hosts? especially since we do `.unwrap_or("")`

maybe add 
```toml
[lints]
workspace = true
```

to cargo.toml to catch all unwraps, expects, etc

common/verloc/src/measurements/measurer.rs
```diff
@@ -135,10 +135,17 @@ impl VerlocMeasurer {
         let mut api_endpoints = self.config.nym_api_urls.clone();
         api_endpoints.shuffle(&mut thread_rng());
         for api_endpoint in api_endpoints {
-            let client = NymApiClient::new_with_user_agent(
-                api_endpoint.clone(),
-                self.config.user_agent.clone(),
-            );
+            let client =
+                match nym_http_api_client::Client::builder(api_endpoint.clone()).and_then(|b| {
```
jstuczyn: given this instance is used by nym-node specifically, would it make sense to add a separate builder to the `Client` such that:
- it wouldn't be using hickory -> no point
- it wouldn't be using domain fronting
?

common/client-core/src/client/base_client/mod.rs
```diff
@@ -749,21 +749,42 @@ where
...
+
+        builder = builder.with_bincode();
+
+        builder
+            .build::<nym_validator_client::models::RequestError>()
+            .map_err(|e| ClientCoreError::NymApiQueryFailure {
+                source: nym_validator_client::nym_api::error::NymAPIError::GenericRequestFailure(
+                    e.to_string(),
+                ),
+            })
```
jstuczyn:
```suggestion
        builder
            .with_bincode()
            .build::<nym_validator_client::models::RequestError>()
            .map_err(|e| ClientCoreError::NymApiQueryFailure {
                source: nym_validator_client::nym_api::error::NymAPIError::GenericRequestFailure(
                    e.to_string(),
                ),
            })
```

common/http-api-client/src/lib.rs
```diff
@@ -500,6 +557,17 @@ impl ClientBuilder {
...
 
+    /// Set the serialization format for API requests and responses
+    pub fn with_serialization(mut self, format: SerializationFormat) -> Self {
+        self.serialization = format;
+        self
+    }
+
+    /// Configure the client to use bincode serialization
+    pub fn with_bincode(self) -> Self {
+        self.with_serialization(SerializationFormat::Bincode)
```
jstuczyn, line 728: what if the target host doesn't support bincode? though I guess it's the caller's fault for being stupid for trying to use it in the first place

common/client-core/src/init/helpers.rs
```diff
@@ -83,6 +84,48 @@ struct GatewayWithLatency<'a, G: ConnectableGateway> {
     latency: Duration,
 }
 
+// Helper to collect all pages of entry nodes - replicates NymApiClient's convenience method
+async fn get_all_basic_entry_nodes_with_metadata(
```
jstuczyn, line 88: given you had to extract it, maybe we should also do a tiny cleanup of `NymApiClient` and move all of those methods to some `Ext` trait on the `nym_http_api_client::Client`?

common/client-libs/validator-client/src/nym_api/mod.rs
```diff
@@ -241,6 +243,162 @@ pub trait NymApiClientExt: ApiClient {
         .await
     }
 
+    async fn get_current_rewarded_set(&self) -> Result<RewardedSetResponse, NymAPIError> {
+        self.get_rewarded_set().await
+    }
+
+    async fn get_all_basic_nodes_with_metadata(
```
jstuczyn, line 252: would the `collect_paged_skimmed_v2` macro help here?

jstuczyn (commented):
looks good! only some minor comments

common/client-core/src/init/helpers.rs
durch: I'd leave that for cleanup number two, there are a bunch of similar tweaks we should do

common/http-api-client/src/lib.rs
durch: Yeah, thats a risk, but some responsibility has to be with the caller

common/client-libs/validator-client/src/nym_api/mod.rs
durch: Oh nice, I've got a bunch of these, I'll ticket them up and clean it!

common/verloc/src/measurements/measurer.rs
durch: Maybe, but then its another single purpose builder, I think we should dogfood this client as much as possible, will find nastiness sooner ;)

common/http-api-client/Cargo.toml
```diff
@@ -23,6 +24,8 @@ url = { workspace = true }
 once_cell = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+serde_yaml = "0.9.34-deprecated"
+serde_plain = "1.0.2"
```
jstuczyn: let's not pollute our crates with standalone dependency versioning. could you set those verisons in the workspace Cargo.toml? 🙏🏻

jstuczyn (approved):

PR #5977, opened 2025-08-25
Remove freshness check on testrun submit

dynco-nym:
- freshness is enforced by a background task that marks testruns as stale after a configured amount of time
- make existing freshness period configurable to avoid code changes in the future
- added `humantime` for parsing

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/5977)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/http/state.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/state.rs
@@ -5,6 +5,7 @@ use nym_bin_common::bin_info_owned;
 use nym_contracts_common::NaiveFloat;
 use nym_crypto::asymmetric::ed25519::PublicKey;
 use nym_mixnet_contract_common::NodeId;
+use nym_node_status_client::auth::VerifiableRequest;
 use nym_validator_client::nym_api::SkimmedNode;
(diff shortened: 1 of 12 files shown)


nym-node-status-api/nym-node-status-api/src/http/state.rs
```diff
@@ -94,6 +101,44 @@ impl AppState {
     pub(crate) fn build_information(&self) -> &BinaryBuildInformationOwned {
         &self.bin_info.build_info
     }
+
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub(crate) fn authenticate_agent_submission(
+        &self,
+        request: &impl VerifiableRequest,
+    ) -> HttpResult<()> {
+        if self.is_registered(request.public_key()) {
```
durch: This is missing the `!`

durch (approved):
Looks good, need to fix the authentication chek

nym-node-status-api/nym-node-status-api/src/http/state.rs
dynco-nym: good spot!

PR #6035, opened 2025-09-15
Bridge proto client params in Self-Described

jmwample:
This PR gives the nym-node a way to expose information about the bridge protocols that the node supports, and the parameters that are necessary to connect using those protocols.

This is meant to be usable by the node status API to be be included into node descriptors that are compiled for the vpn client. 

- Adds a new field to the nym-node config `gateway_tasks.storage_paths.bridge_client_params`
  - IF the new config field is present a new self-described endpoint is available at `/v1/bridges/client-params`
  - IF the new config field is NOT present the endpoint is not exposed.

I arbitrarily chose config v8 as the oldest nym-node configuration version that supports the option. This can probably be propogated further backwards if necessary.

NOTE: The new `/bridges/client-params` endpoint does not have swagger / utopia docs associated. This interface will likely change in several upcoming iterations and serving from file (for now) means that the types are not defined internally.

tested as working on node `3wqfp9eb` both when file is provided in config (sucessful response) and when file is not specified in config (path gives 404). 
https://nym-circ.anonym.tech/api/v1/bridges/client-params

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6035)

### DIFF
--- a/nym-node/nym-node-requests/src/api/v1/gateway/models.rs
+++ b/nym-node/nym-node-requests/src/api/v1/gateway/models.rs
@@ -47,7 +47,6 @@ pub struct WebSockets {
     pub wss_port: Option<u16>,
 }
 
-
 #[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct Bridges {
--- a/nym-node/nym-node-requests/src/api/v1/gateway/models.rs
+++ b/nym-node/nym-node-requests/src/api/v1/gateway/models.rs
@@ -46,3 +46,10 @@ pub struct WebSockets {
 
     pub wss_port: Option<u16>,
 }
+
+
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+pub struct Bridges {
+    pub client_params_path: String,
+}
--- a/nym-node/nym-node-requests/src/lib.rs
+++ b/nym-node/nym-node-requests/src/lib.rs
@@ -41,6 +41,7 @@ pub mod routes {
             pub const GATEWAY: &str = "/gateway";
             pub const MIXNODE: &str = "/mixnode";
             pub const METRICS: &str = "/metrics";
+            pub const BRIDGES: &str = "/bridges";
             pub const NETWORK_REQUESTER: &str = "/network-requester";
             pub const IP_PACKET_ROUTER: &str = "/ip-packet-router";
             pub const AUTHENTICATOR: &str = "/authenticator";
--- a/nym-node/src/config/old_configs/old_config_v10.rs
+++ b/nym-node/src/config/old_configs/old_config_v10.rs
@@ -979,6 +979,9 @@ pub struct GatewayTasksPathsV10 {
 
     /// Path to file containing cosmos account mnemonic used for zk-nym redemption.
     pub cosmos_mnemonic: PathBuf,
+
+    /// Path to file containing bridge client params to be served in the node self-described.
+    pub bridge_client_params: Option<PathBuf>,
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
@@ -1469,7 +1472,7 @@ pub async fn try_upgrade_config_v10<P: AsRef<Path>>(
                 clients_storage: old_cfg.gateway_tasks.storage_paths.clients_storage,
                 stats_storage: old_cfg.gateway_tasks.storage_paths.stats_storage,
                 cosmos_mnemonic: old_cfg.gateway_tasks.storage_paths.cosmos_mnemonic,
-                bridge_client_params: None,
+                bridge_client_params: old_cfg.gateway_tasks.storage_paths.bridge_client_params,
             },
             enforce_zk_nyms: old_cfg.gateway_tasks.enforce_zk_nyms,
             ws_bind_address: old_cfg.gateway_tasks.ws_bind_address,
--- a/nym-node/src/config/old_configs/old_config_v10.rs
+++ b/nym-node/src/config/old_configs/old_config_v10.rs
@@ -1469,6 +1469,7 @@ pub async fn try_upgrade_config_v10<P: AsRef<Path>>(
                 clients_storage: old_cfg.gateway_tasks.storage_paths.clients_storage,
                 stats_storage: old_cfg.gateway_tasks.storage_paths.stats_storage,
                 cosmos_mnemonic: old_cfg.gateway_tasks.storage_paths.cosmos_mnemonic,
+                bridge_client_params: None,
             },
             enforce_zk_nyms: old_cfg.gateway_tasks.enforce_zk_nyms,
             ws_bind_address: old_cfg.gateway_tasks.ws_bind_address,
--- a/nym-node/src/config/old_configs/old_config_v7.rs
+++ b/nym-node/src/config/old_configs/old_config_v7.rs
@@ -1340,7 +1340,6 @@ pub async fn try_upgrade_config_v7<P: AsRef<Path>>(
                 clients_storage: old_cfg.gateway_tasks.storage_paths.clients_storage,
                 stats_storage: old_cfg.gateway_tasks.storage_paths.stats_storage,
                 cosmos_mnemonic: old_cfg.gateway_tasks.storage_paths.cosmos_mnemonic,
-                bridge_client_params: None,
             },
             enforce_zk_nyms: old_cfg.gateway_tasks.enforce_zk_nyms,
             ws_bind_address: {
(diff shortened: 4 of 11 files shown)


jmwample:
I do not think the tests failing in `nym-api/tests/public-api/nym_nodes.rs`  are related to the changes in this pr

durch (approved):

neacsu (commented):
> @neacsu reviewed 8 of 11 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6035)* status: all files reviewed, 2 unresolved discussions (waiting on @aniampio, @jstuczyn, and @mmsinclair)

___
*[`nym-node/src/config/persistence.rs` line 151 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pe3be6PHCXHJzJU7i:-O_pe3be6PHCXHJzJU7j:bwpy5dc) ([raw file](https://github.com/nymtech/nym/blob/816cd62be7ee63c4aff618006ccb6e87d1193650/nym-node/src/config/persistence.rs#L151)):*
> ```XML
> 
>     /// Path to file containing bridge client params to be served in the node self-described.
>     pub bridge_client_params: Option<PathBuf>,
> ```

Aren't we still using sections in `config.toml` for storing params for certain nym-node features? I had the impression that we use paths mostly for referencing keys and databases.
___
*[`nym-node/src/config/old_configs/old_config_v7.rs` line 1343 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pbaH3CkUfpLPiRXbZ:-O_pbaH3CkUfpLPiRXb_:b-bz207d) ([raw file](https://github.com/nymtech/nym/blob/816cd62be7ee63c4aff618006ccb6e87d1193650/nym-node/src/config/old_configs/old_config_v7.rs#L1343)):*
> ```XML
>                 stats_storage: old_cfg.gateway_tasks.storage_paths.stats_storage,
>                 cosmos_mnemonic: old_cfg.gateway_tasks.storage_paths.cosmos_mnemonic,
>                 bridge_client_params: None,
> ```

Is there a reason to modify all those previous `old_config`s instead of doing the same we have been doing so far i.e. create a new config version and writing the migration? Other than being a ton of boilerplate code :sweat_smile: Also, if latest version (v10) wasn't released yet, you could probably just piggy back onto that one

jmwample (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6035)* status: all files reviewed, 2 unresolved discussions (waiting on @aniampio, @jstuczyn, @mmsinclair, and @neacsu)

___
*[`nym-node/src/config/persistence.rs` line 151 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pe3be6PHCXHJzJU7i:-O_w7tGtAXH5ODuPw0n9:b3uhrwv) ([raw file](https://github.com/nymtech/nym/blob/816cd62be7ee63c4aff618006ccb6e87d1193650/nym-node/src/config/persistence.rs#L151)):*

Aren't we still using sections in `config.toml` for storing params for certain nym-node features? I had the impression that we use paths mostly for referencing keys and databases.
</blockquote></details>

I am not sure what you mean by `sections`. I didn't think this fit as a service provider as it is a path to config for an optional feature on the entry gateway. If there is a better place for this in the config I am fully open to it.
___
*[`nym-node/src/config/old_configs/old_config_v7.rs` line 1343 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pbaH3CkUfpLPiRXbZ:-O_w8TCD4XzsnKH_ahQo:bg2w56n) ([raw file](https://github.com/nymtech/nym/blob/816cd62be7ee63c4aff618006ccb6e87d1193650/nym-node/src/config/old_configs/old_config_v7.rs#L1343)):*

Is there a reason to modify all those previous `old_config`s instead of doing the same we have been doing so far i.e. create a new config version and writing the migration? Other than being a ton of boilerplate code :sweat_smile: Also, if latest version (v10) wasn't released yet, you could probably just piggy back onto that one
</blockquote></details>

The field is optional, not a breaking change to something existing or a required field that will stop binaries from launching. As such it doesn't have to go into the new config standard. Also this PR is potentially going to be patched back onto an existing release (feta) and could even be patched further back if needed. The previous config support just means that if an operator configures this in a previous version it gets propogated through when it upgrades.

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6035)* status: all files reviewed, 2 unresolved discussions (waiting on @aniampio, @jmwample, @jstuczyn, and @mmsinclair)

___
*[`nym-node/src/config/persistence.rs` line 151 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pe3be6PHCXHJzJU7i:-Oa-M_Od-qGQ_RsYq_IK:b6a7rty) ([raw file](https://github.com/nymtech/nym/blob/2a67745819ada05f68255034be59451d91adf982/nym-node/src/config/persistence.rs#L151)):*
<details><summary><i>Previously, jmwample (Jack Wampler) wrote…</i></summary><blockquote>

I am not sure what you mean by `sections`. I didn't think this fit as a service provider as it is a path to config for an optional feature on the entry gateway. If there is a better place for this in the config I am fully open to it.
</blockquote></details>

I'm referring to the sections inside https://github.com/nymtech/nym/blob/develop/nym-node/src/config/template.rs . Service providers are just another sections in there, it doesn't mean this should go under it, it can be a new section by itself.
___
*[`nym-node/src/config/old_configs/old_config_v7.rs` line 1343 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pbaH3CkUfpLPiRXbZ:-Oa-N8uj0qJqSqHd4-ts:b-oofhp) ([raw file](https://github.com/nymtech/nym/blob/2a67745819ada05f68255034be59451d91adf982/nym-node/src/config/old_configs/old_config_v7.rs#L1343)):*
<details><summary><i>Previously, jmwample (Jack Wampler) wrote…</i></summary><blockquote>

The field is optional, not a breaking change to something existing or a required field that will stop binaries from launching. As such it doesn't have to go into the new config standard. Also this PR is potentially going to be patched back onto an existing release (feta) and could even be patched further back if needed. The previous config support just means that if an operator configures this in a previous version it gets propogated through when it upgrades.
</blockquote></details>

Yes, creating a new config version doesn't necessarily mean it has to be breaking, just that it's something we didn't have in previous configs. Also, since we're aiming for a new nym-node release soon, I don't see why we'd need to patch existing releases like feta, we can just put it into the latest release. Operators are not patch updating old versions, they usually just get the latest and greatest. The only case I see would be to patch the current RC (isabirra) for a faster than usual release.

jmwample (commented):
I would like to merge this in its current state so that it can be included in develop and in test environment builds. If we need to adjust config paths or config version support we can do so in either another pull request before the next release, or in an updated config version if we need to.

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6035#-:-Oa0mjdi4MfX10O-pIq9:bzhan54)* status: 7 of 11 files reviewed, all discussions resolved (waiting on @aniampio, @jstuczyn, @mmsinclair, and @neacsu)

___
*[`nym-node/src/config/persistence.rs` line 151 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pe3be6PHCXHJzJU7i:-Oa0lH1Q8vo33qS50OKK:b-6wejmk) ([raw file](https://github.com/nymtech/nym/blob/2a67745819ada05f68255034be59451d91adf982/nym-node/src/config/persistence.rs#L151)):*

I'm referring to the sections inside https://github.com/nymtech/nym/blob/develop/nym-node/src/config/template.rs . Service providers are just another sections in there, it doesn't mean this should go under it, it can be a new section by itself.
</blockquote></details>

Adding as a root section in https://github.com/nymtech/nym/blob/eeb0278d13f3fe9a1b7dc32d22f4956004a99c20/nym-node/src/config/mod.rs#L162 didn't seem as fitting (at least for this initial version) as this is a property of only the entry gateways and not a fully configured service / feature / etc. I think at least for now it fits well enough in the gateway_tasks storage_paths. 

If that doesn't fit going forward we can always update it in a next version of the config.
___
*[`nym-node/src/config/old_configs/old_config_v7.rs` line 1343 at r3](https://reviewable.io/reviews/nymtech/nym/6035#-O_pbaH3CkUfpLPiRXbZ:-Oa0m6GIA82KLMaOnepN:b-c7e0gj) ([raw file](https://github.com/nymtech/nym/blob/2a67745819ada05f68255034be59451d91adf982/nym-node/src/config/old_configs/old_config_v7.rs#L1343)):*

Yes, creating a new config version doesn't necessarily mean it has to be breaking, just that it's something we didn't have in previous configs. Also, since we're aiming for a new nym-node release soon, I don't see why we'd need to patch existing releases like feta, we can just put it into the latest release. Operators are not patch updating old versions, they usually just get the latest and greatest. The only case I see would be to patch the current RC (isabirra) for a faster than usual release.
</blockquote></details>

We are planning on releasing this quickly. Based on several features coming out soon there are blocking changes preventing deployment and we are planning patches on feta / isabirra to make this work with them so we aren't waiting for those features to stabilize.

PR #6045, opened 2025-09-17
Hotfix: Explorer-v2 table caching issues

serinko:
This PR should bugfix the issue with nodes selected by correct algo, showing up on the recommended page even hours after they drop their uptime. Removing cache and adding uptime floor as additional safety.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6045)

### DIFF
--- a/explorer-v2/src/app/lib/recommended.ts
+++ b/explorer-v2/src/app/lib/recommended.ts
@@ -1,4 +1,5 @@
-import { cache } from "react";
+// NOTE: removed the cache() import to avoid sticky results across requests
+// import { cache } from "react";
 
 type DeclaredRole = {
   entry?: boolean;
@@ -10,7 +11,7 @@ type DeclaredRole = {
 type ApiNode = {
   node_id?: number;
   total_stake?: number | string;
-  uptime?: number | string;
+  uptime?: number | string; // fraction 0..1
   description?: {
     wireguard?: unknown | null;
     declared_role?: DeclaredRole;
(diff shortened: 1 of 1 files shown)


mfahampshire (approved):

explorer-v2/src/app/lib/recommended.ts
```diff
@@ -69,27 +77,56 @@ function sortByUptimeDescStakeAsc(a: ApiNode, b: ApiNode): number {
...
-      { cache: "no-store" },
-    );
-    if (!res.ok) throw new Error(`Failed to fetch page ${page}`);
-    const json = await res.json();
-    const data: ApiNode[] = Array.isArray(json) ? json : json.data;
-    if (!data || data.length === 0) break;
-    all = [...all, ...data];
-    if (data.length < pageSize) break; // last page
-    page++;
+  const base = "https://api.nym.spectredao.net/api/v1/nodes";
```
yanok87, line 82: Should be moved into the url's file and imported as constant. Also, how come we fetch all nodes from Node Status Api, but recommended from SpectreDao?

explorer-v2/src/app/lib/recommended.ts
```diff
@@ -69,27 +77,56 @@ function sortByUptimeDescStakeAsc(a: ApiNode, b: ApiNode): number {
...
+    }
+    if (all.length) return all;
+  }
+
+  // 3. fallback single-shot
+  const res = await fetch(base, { cache: "no-store" });
+  if (!res.ok) throw new Error(`Failed to fetch nodes: ${res.status}`);
+  const js = await res.json();
+  return Array.isArray(js) ? js : Array.isArray(js?.data) ? js.data : [];
 }
```
yanok87, line 130: All fetch functions could and should be moved into src/app/api/index.ts and given proper names as all the api calls reside in one file. Also, there's a ready fetchNSApiNodes function (for NS nodes though...) in the index.ts that covers the limit/offset/page-size/fallback options with one call. Why do we need 3 fetch functions for the same endpoint?

explorer-v2/src/app/lib/recommended.ts
```diff
@@ -128,5 +167,8 @@ async function fetchRecommendedNodes(): Promise<number[]> {
     .filter((id) => Number.isFinite(id) && id > 0);
 }
 
-export const getRecommendedNodes = cache(fetchRecommendedNodes);
+// keep the same exports the code expects:
+export async function getRecommendedNodes(): Promise<number[]> {
+  return fetchRecommendedNodes();
+}
 export const RECOMMENDED_NODES: Promise<number[]> = getRecommendedNodes();
```
yanok87, line 174: Also better to move the whole recomended.ts into "api" directory that handles all nodes api calls instead of "lib" which handles Strapi only

yanok87 (changes requested):

explorer-v2/src/app/lib/recommended.ts
serinko: It can be refactored to fit the convention of the whole design.
The nodes are fetched from an API I knew will work to get all info I need  at the moment of building this quick fix.

explorer-v2/src/app/lib/recommended.ts
serinko: It can be refactored to fit the convention of the whole design.
The nodes are fetched from an API I knew will work to get all info I need  at the moment of building this quick fix.

explorer-v2/src/app/lib/recommended.ts
serinko: It can be refactored to fit the convention of the whole design.

PR #6048, opened 2025-09-17
chore: remove Option wrapper from ShutdownTracker on the client

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6048)


jstuczyn:
superseded by https://github.com/nymtech/nym/pull/6059 which does the same thing

PR #6060, opened 2025-09-24
[chore] Clippy fix

simonwicky:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6060)

### DIFF
--- a/nym-node/src/config/old_configs/old_config_v10.rs
+++ b/nym-node/src/config/old_configs/old_config_v10.rs
@@ -438,10 +438,6 @@ impl Default for HttpV10 {
     }
 }
 
-#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
-#[serde(deny_unknown_fields)]
-pub struct MixnodePathsV10 {}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct DebugV10 {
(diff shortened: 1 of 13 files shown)


nym-node/src/config/old_configs/old_config_v10.rs
```diff
@@ -438,10 +438,6 @@ impl Default for HttpV10 {
     }
 }
 
-#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
-#[serde(deny_unknown_fields)]
-pub struct MixnodePathsV10 {}
```
neacsu, line 443: Is this all code removal part of a bad rebase?

simonwicky: Tbh, I have no idea. 
Clippy signals it as dead code and removing it doesn't seem to crash the builds. 
Maybe the config changed, these fields are gone, but we stuck to copy paste them when making a new version

simonwicky: I'm intrigued as why it shows up only now

neacsu: Probably new clippy. Also, I suspect it's dead code because it was in previous versions and there are no unit tests for it to be at least used there, since latest code doesn't need it anymore

wasm/zknym-lib/src/vpn_api_client/types.rs
```diff
@@ -11,6 +11,7 @@ use uuid::Uuid;
 #[derive(Tsify, Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq)]
 #[tsify(into_wasm_abi, from_wasm_abi)]
 #[serde(rename_all = "camelCase")]
+#[allow(dead_code)]
```
simonwicky, line 14: I guess these are used for warm/ts binding but I'm not 100% sure

neacsu (approved):

PR #6061, opened 2025-09-24
ns-api: filter out gateways with less than 75% performance from dvpn endpoint

mmsinclair:
Filter out gateways with less than 75% performance for the dVPN endpoint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6061)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6508,7 +6508,7 @@ dependencies = [
 
 [[package]]
 name = "nym-node-status-api"
-version = "4.0.3"
+version = "4.0.4"
 dependencies = [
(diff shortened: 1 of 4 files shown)


neacsu (approved):

pronebird (commented):
Let's not do that. The clients are entirely capable doing this

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6061#-:-Oa0QPZI3dBfm1YHRLXR:b9rhizr)* status: 0 of 4 files reviewed, all discussions resolved

PR #6074, opened 2025-09-30
Feature: Ping probe all nodes /described nodes from a server

serinko:
This script should be ran from a node hosting server. It pings all IPs listed in /described endpoint and returns a file with unreachable IPs. Such list gives operator an idea on IPs potentially blocking their IP.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6074)


mfahampshire (approved):
:boom:

PR #6099, opened 2025-10-09
Bugfix/incompatibility fixes

jstuczyn:
this PR fixes a few incompatibility fixes:
- initialisation of fresh nym-api
- native/socks5 clients that don't persist `ClientRequestSender` channel
- network manager builds with contracts built with the docker image

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6099)

### DIFF
--- a/common/client-core/src/client/mix_traffic/mod.rs
+++ b/common/client-core/src/client/mix_traffic/mod.rs
@@ -62,19 +62,7 @@ impl MixTrafficController {
         gateway_transceiver: Box<dyn GatewayTransceiver + Send>,
         shutdown_token: ShutdownToken,
     ) -> MixTrafficController {
-        let (message_sender, message_receiver) =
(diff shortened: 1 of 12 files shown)


common/client-core/src/client/mix_traffic/mod.rs
```diff
@@ -52,41 +47,42 @@ impl MixTrafficController {
...
-            message_sender,
-            client_sender,
-        )
+
+        MixTrafficController {
+            gateway_transceiver,
+            mix_tx: message_sender,
+            mix_rx: message_receiver,
+            client_rx: client_receiver,
+            client_tx: client_sender.clone(),
```
durch: Wondering about this `clone()`, not sure why we need a clone here, and not in `new()`

durch (approved):
All good, even with the clone comment

common/client-core/src/client/mix_traffic/mod.rs
jstuczyn: just removed the whole duplication instead : D

PR #6108, opened 2025-10-14
Signal shutdown when gateway is considered unreachable

neacsu:
Depends on https://github.com/nymtech/nym/pull/6105 and https://github.com/nymtech/nym/pull/6052/commits/5025c49a0e015d615b1af3b25d36ac5dc59a6f0c being merged in order to fix https://nymtech.atlassian.net/browse/VPN-4214

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6108)

### DIFF
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -892,7 +892,7 @@ where
         // Create a shutdown tracker for this client - either as a child of provided tracker
         // or get one from the registry
         let shutdown_tracker = match self.shutdown {
-            Some(parent_tracker) => parent_tracker.child_tracker(),
+            Some(parent_tracker) => parent_tracker.clone(),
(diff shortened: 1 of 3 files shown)


pronebird (approved):
I'd really prefer a centralized approach with back channel, this way we could notify the caller that something went wrong. Currently this propagates via cancellation token which is ambiguous if we consider that mixnet cancellation token is derived from other token that may as well be cancelled.

pronebird:
Note that we need to patch this call otherwise disconnect won't happen:

https://github.com/nymtech/nym/blob/develop/nym-authenticator-client/src/mixnet_listener.rs#L103

pronebird (changes requested):
Sadly I don't think this is the silver bullet because mixnet cancellation token is being used to determine when to disconnect() mixnet client.

neacsu:
Closed in favor of https://github.com/nymtech/nym/pull/6119

PR #6110, opened 2025-10-14
Add connect timeout

pronebird:
This is a draft PR that adds connect timeout for websocket. Since `tokio::net::TcpSocket` does not support `connect_timeout()` from std, the next close alternative using `tokio::time::timeout()` is used instead.

Connect timeout only limits how much time the call to `TcpSocket::connect()` may take before giving up.

A few notes:
- The structure of project seems to be very vertical which results into dragging variables all the way down 
- I don't understand how to pass connect_timeout in wasm client. It's not that vpn-client needs wasm, but for consistency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6110)

### DIFF
--- a/common/client-core/src/cli_helpers/client_add_gateway.rs
+++ b/common/client-core/src/cli_helpers/client_add_gateway.rs
@@ -140,6 +140,7 @@ where
         available_gateways,
         #[cfg(unix)]
         connection_fd_callback: None,
+        connect_timeout: None,
     };
 
     let init_details =
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -359,6 +359,11 @@ where
         self
     }
 
+    pub fn with_connect_timeout(mut self, timeout: Duration) -> Self {
+        self.connect_timeout = Some(timeout);
+        self
+    }
+
     // note: do **NOT** make this method public as its only valid usage is from within `start_base`
     // because it relies on the crypto keys being already loaded
     fn mix_address(details: &InitialisationResult) -> Recipient {
--- a/common/client-core/src/client/base_client/mod.rs
+++ b/common/client-core/src/client/base_client/mod.rs
@@ -65,6 +65,7 @@ use std::fmt::Debug;
 use std::os::raw::c_int as RawFd;
 use std::path::Path;
 use std::sync::Arc;
+use std::time::Duration;
 use time::OffsetDateTime;
 use tokio::sync::mpsc::Sender;
 use url::Url;
@@ -230,6 +231,7 @@ pub struct BaseClientBuilder<C, S: MixnetClientStorage> {
 
     #[cfg(unix)]
     connection_fd_callback: Option<Arc<dyn Fn(RawFd) + Send + Sync>>,
+    connect_timeout: Option<Duration>,
 
     derivation_material: Option<DerivationMaterial>,
 }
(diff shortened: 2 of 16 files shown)


common/client-core/src/cli_helpers/client_add_gateway.rs
```diff
@@ -140,6 +140,7 @@ where
         available_gateways,
         #[cfg(unix)]
         connection_fd_callback: None,
+        connect_timeout: None,
```
jstuczyn, line 143: I'm not sure connection timeout is a propery of a `GatewaySetup` (I don't like `connection_fd_callback` being there either). but if it has to be there, maybe it should be wrapped into some sort of config (alongside that callback)?

common/client-core/src/client/base_client/mod.rs
```diff
@@ -356,6 +359,11 @@ where
         self
     }
 
+    pub fn with_connect_timeout(mut self, timeout: Duration) -> Self {
```
jstuczyn, line 362: I think it would it be easier for you if rather than setting connection timeout as a separate thing of the client, you put it as part of the base client config. `gateway_connection` section inside the `DebugConfig` sounds like the perfect place for it, wouldn't you say?

common/client-core/src/client/base_client/mod.rs
```diff
@@ -533,6 +541,7 @@ where
         packet_router: PacketRouter,
         stats_reporter: ClientStatsSender,
         #[cfg(unix)] connection_fd_callback: Option<Arc<dyn Fn(RawFd) + Send + Sync>>,
+        connect_timeout: Option<Duration>,
```
jstuczyn, line 544: and then I guess (with my previous comment in mind), all of those extra changes could be avoided

common/client-libs/gateway-client/src/client/websockets.rs
```diff
@@ -17,6 +18,7 @@ use std::net::SocketAddr;
 pub(crate) async fn connect_async(
```
jstuczyn, line 18: while going through the code, I've noticed we have identical `connect_async` function in `common/client-core/src/init/websockets.rs` (don't ask me why - I wish I knew). and actually when you go to where it is being called (the duplicate), it's already wrapped in a timeout - perhaps we could remove some duplication there?

common/client-libs/gateway-client/src/client/websockets.rs
```diff
@@ -64,7 +66,22 @@ pub(crate) async fn connect_async(
             callback.as_ref()(socket.as_raw_fd());
         }
 
-        match socket.connect(sock_addr).await {
+        let connect_res = if let Some(connect_timeout) = connect_timeout {
+            match tokio::time::timeout(connect_timeout, socket.connect(sock_addr)).await {
+                Ok(res) => res,
+                Err(_elapsed) => {
+                    stream = Err(GatewayClientError::NetworkConnectionTimeout {
```
jstuczyn, line 73: wouldn't existing `GatewayConnectionTimeout` variant be more appropriate?

jstuczyn (changes requested):
And as for wasm, that's a tricky one because there's no underlying `connect_async` function as internally it just calls browser's `WebSocket` and relies on callbacks for knowing when the connection has been opened. And gloo's implementation (the library we're using) is using a blocking call. Perhaps you could introduce js' timeout and cancel the socket in that case?

jstuczyn:
is this PR still relevant or has it been made reduntant with @simonwicky changes?

pronebird:
I think we can close it for now.

PR #6119, opened 2025-10-15
Introduce event backchannel

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)


sdk/rust/nym-sdk/src/mixnet.rs
```diff
@@ -45,9 +45,14 @@ pub use native_client::MixnetClient;
...
-        gateways_storage::{ActiveGateway, BadGateway, GatewayRegistration, GatewaysDetailsStore},
-        Ephemeral, MixnetClientStorage, OnDiskPersistent,
+    base_client::{
+        storage::{
+            gateways_storage::{
+                ActiveGateway, BadGateway, GatewayRegistration, GatewaysDetailsStore,
+            },
+            Ephemeral, MixnetClientStorage, OnDiskPersistent,
+        },
+        EventReceiver, EventSender,
```
jstuczyn: it'd be probably useful to also expose the `Event` itself so that sdk consumers could match on it

common/client-core/src/client/base_client/mod.rs
```diff
@@ -83,6 +83,13 @@ pub mod non_wasm_helpers;
 pub mod helpers;
 pub mod storage;
 
+pub enum Event {
```
jstuczyn: just a nit, but could we give it a more meaningful name? `Event` sounds so generic and it will be easy to confuse

common/client-core/src/client/mix_traffic/mod.rs
```diff
@@ -155,6 +164,9 @@ impl MixTrafficController {
                                 error!("Failed to send sphinx packet to the gateway {MAX_FAILURE_COUNT} times in a row - assuming the gateway is dead");
                                 // Do we need to handle the embedded mixnet client case
                                 // separately?
+                                if let Some(event_tx) = self.event_tx.as_mut() {
```
jstuczyn: if the channel is full and the consumer has gone offline, this await will wait forever

nym-registration-client/src/builder/mod.rs
```diff
@@ -37,14 +38,16 @@ impl RegistrationClientBuilder {
...
+        let (event_tx, event_rx) = mpsc::unbounded();
 
         let nyxd_client = get_nyxd_client(&self.config.network_env)?;
 
         let (mixnet_client, bandwidth_controller): (
             MixnetClient,
             Box<dyn BandwidthTicketProvider>,
         ) = if let Some((mixnet_client_storage, credential_storage)) = storage {
-            let builder = MixnetClientBuilder::new_with_storage(mixnet_client_storage);
+            let builder =
```
jstuczyn: I wonder, regarding design choices. should you have to provide `event_tx` when constructing the client, or could the corresponding receiver be always available on the `ClientOutput` (after building the client? then we also wouldn't need to propagate it through so many different builder layers.

common/client-core/src/client/mix_traffic/mod.rs
jstuczyn: also, for usability purposes, it might be a good idea to wrap that channel in additional struct `Foo(Option<EventSender>)` so all the option matching could be done there and all the consumers could just call `send` without having to go through all the boilerplate

common/client-core/src/client/base_client/mod.rs
```diff
@@ -736,10 +751,14 @@ where
...
         shutdown_tracker: &ShutdownTracker,
+        event_tx: Option<EventSender>,
     ) -> (BatchMixMessageSender, ClientRequestSender) {
         info!("Starting mix traffic controller...");
-        let (mut mix_traffic_controller, mix_tx, client_tx) =
-            MixTrafficController::new(gateway_transceiver, shutdown_tracker.clone_shutdown_token());
+        let (mut mix_traffic_controller, mix_tx, client_tx) = MixTrafficController::new(
+            gateway_transceiver,
+            shutdown_tracker.clone_shutdown_token(),
+            event_tx,
```
jstuczyn, line 783: question regarding design: is this `event_tx` meant to be the property of `MixTrafficController` or the client itself?

sdk/rust/nym-sdk/src/mixnet.rs
neacsu: Right, I was consuming it from `nym_client_core` in vpn, but it's better to have it from the sdk.

common/client-core/src/client/base_client/mod.rs
neacsu: Changed it to `MixnetClientEvent`

common/client-core/src/client/mix_traffic/mod.rs
neacsu: This is a good catch, I was actually wanting to send without awaiting. I think it's better to just change it to `unbounded_send`.

neacsu: Although the struct wrapping seems like a good idea.

common/client-core/src/client/base_client/mod.rs
neacsu: It's meant to be passed inside every spawned task of the mixnet client from where we think it's relevant to back propagate some event. In the context of this bug, I'm only looking at one particular code branch, but it can be anything.

nym-registration-client/src/builder/mod.rs
neacsu: I think I understand what you're saying, this was my first attempt actually, instead of passing around the `event_tx`, that the channel would be constructed internally and kept inside the client somewhere. The problem is that in the vpn app, we loop `select!` on the client for incoming packets, so reading the event would also need to be inside that `select!` branch. So in that case, the events would have to be multiplexed with the incoming mix packets. That's an option too, but a more complex one, so I'm not sure how I feel about it

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @jstuczyn and @simonwicky)

___
*[`common/client-core/src/client/shutdown_control.rs` line 31 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-Obgaw8wCRJaZU_SyNQr:-Obgaw8wCRJaZU_SyNQs:b-pudsco) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/shutdown_control.rs#L31)):*
> ```XML
>                 biased;
>                 _ = external_shutdown_token.cancelled() => {
>                     self
> ```

orphan `self`?

pronebird (commented):
> @pronebird reviewed 1 of 11 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: 3 of 11 files reviewed, 6 unresolved discussions (waiting on @jstuczyn, @neacsu, and @simonwicky)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-ObgcJGg5TDKXipF2LH9:-ObgcJGg5TDKXipF2LHA:b-o5she4) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/mix_traffic/mod.rs#L166)):*
> ```XML
>                                 // Do we need to handle the embedded mixnet client case
>                                 // separately?
>                                 if let Some(event_tx) = self.event_tx.as_mut() {
> ```

FYI I hope that there is a copy of `event_tx` tucked away somewhere. Remember that channels close when either all senders or receivers are gone.

pronebird (commented):
> @pronebird reviewed 4 of 11 files at r1, 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @jstuczyn, @neacsu, and @simonwicky)

___
*[`nym-registration-client/src/builder/mod.rs` line 80 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-Obgcu83-UFEdYaJLGtl:-Obgcu83-UFEdYaJLGtm:b-j0t0yk) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/nym-registration-client/src/builder/mod.rs#L80)):*
> ```XML
>             mixnet_client_address,
>             bandwidth_controller,
>             event_rx,
> ```

FYI I guess we expect that `event_rx` is always consumed by the vpn app

pronebird (changes requested):
> @pronebird reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: 9 of 11 files reviewed, 7 unresolved discussions (waiting on @jstuczyn, @neacsu, and @simonwicky)

___
*[`common/client-core/src/client/shutdown_control.rs` line 39 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-ObgdfvM92H6cX9MQFmu:-ObgdfvM92H6cX9MQFmv:b-izy277) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/shutdown_control.rs#L39)):*
> ```XML
>                 }
>                 next_message = self.next() => if let Some(next_message) = next_message {
>                     self.on_message(next_message).await;
> ```

Where does this call into? I don't see `on_message()` in this file. Is it implemented somewhere else?

pronebird (commented):
> @pronebird reviewed 2 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed (commit messages unreviewed), 7 unresolved discussions (waiting on @jstuczyn, @neacsu, and @simonwicky)

pronebird (commented):
> @pronebird reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed, 7 unresolved discussions (waiting on @jstuczyn, @neacsu, and @simonwicky)

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed, 7 unresolved discussions (waiting on @jstuczyn, @pronebird, and @simonwicky)

___
*[`common/client-core/src/client/shutdown_control.rs` line 31 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-Obgaw8wCRJaZU_SyNQr:-Obge2eg4eZ4S3SrshF5:b-8ymi71) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/shutdown_control.rs#L31)):*

orphan `self`?
</blockquote></details>

Ah sorry, this is actually orphan file from a different attempt to implement two shutdown tokens manually, previous to the revert. I'll remove it altogether (the errors weren't showing as it's not declared as module)
___
*[`common/client-core/src/client/shutdown_control.rs` line 39 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-ObgdfvM92H6cX9MQFmu:-ObgeH2B5q_3c2xgmNQF:b-ctrph2) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/shutdown_control.rs#L39)):*

Where does this call into? I don't see `on_message()` in this file. Is it implemented somewhere else?
</blockquote></details>

See above comment

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: 10 of 11 files reviewed, 7 unresolved discussions (waiting on @jstuczyn, @pronebird, and @simonwicky)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-ObgcJGg5TDKXipF2LH9:-ObgjmlV19Akvedj9p4G:bbwae3p) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/common/client-core/src/client/mix_traffic/mod.rs#L166)):*

FYI I hope that there is a copy of `event_tx` tucked away somewhere. Remember that channels close when either all senders or receivers are gone.
</blockquote></details>

Well there isn't in this case, since this is the only place we currently send a relevant event. Whereas to the channel being closed, I would say that the functioning way is that even if all senders get closed, the receivers can still read all the buffered messages. If the reader gets closed, then the senders would just error out on sending.
___
*[`nym-registration-client/src/builder/mod.rs` line 80 at r2](https://reviewable.io/reviews/nymtech/nym/6119#-Obgcu83-UFEdYaJLGtl:-ObgkRxx3ImYp8r2qJeW:b637wma) ([raw file](https://github.com/nymtech/nym/blob/4ace339626fbb307e48f8826a2ad26b9e2dc7edb/nym-registration-client/src/builder/mod.rs#L80)):*

FYI I guess we expect that `event_rx` is always consumed by the vpn app
</blockquote></details>

`event_rx` can be consumed by the SDK user as well. In the case of vpn app, since we use this part of the sdk via the registration client, we have to also it for returning the reader end

pronebird (commented):
> @pronebird reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @jstuczyn and @simonwicky)

pronebird (approved):
Looks good to me. You guys sort out the remainder of issues before merging

> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119#-:-ObgunB15ghsHq1keGx2:b-w4dunf)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @jstuczyn and @simonwicky)

common/client-core/src/client/mix_traffic/mod.rs
pronebird: ps: does sender need to be mutable?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @jstuczyn and @simonwicky)

common/client-core/src/client/mix_traffic/mod.rs
neacsu: Now with `unbounded_send` it doesn't anymore, good point!

neacsu (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @jstuczyn and @simonwicky)

pronebird (commented):
> @pronebird reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6119)* status: all files reviewed (commit messages unreviewed), 5 unresolved discussions (waiting on @jstuczyn and @simonwicky)

jstuczyn (approved):

PR #6129, opened 2025-10-17
Max/sdk streamer

mfahampshire:
This PR grew a bit, it includes: 
- a TcpSocket-like abstraction for the Rust SDK (utilising AsyncRead/AsyncWrite traits) 
- an abstraction built on top of this, which allows for the SDK to use Exit Gateway's IPRs 
- the proof of concept for `mixtcp`, a `smoltcp` device that uses the IPR abstraction for transport. this proof of concept has several example files showing how it can be used for HTTPS requests and doing TLS handshakes via the Mixnet. 

Documentation will be done in a separate PR. 

ToDo: 
- [x] rebase with `develop` 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6129)


jstuczyn:
you might want to rebase it first : )

common/client-core/Cargo.toml
```diff
@@ -10,7 +10,8 @@ license.workspace = true
 
 [dependencies]
 async-trait = { workspace = true }
-base64 = { workspace = true }
+bincode = { workspace = true }
+base64 = "0.21.2"
```
jstuczyn: use workspace versioning

Cargo.toml
```diff
@@ -154,6 +155,7 @@ members = [
     "tools/internal/ssl-inject",
     "tools/internal/testnet-manager",
     "tools/internal/testnet-manager/dkg-bypass-contract",
+    "tools/internal/testnet-manager/dkg-bypass-contract",
```
jstuczyn: it seems that member is duplicated?

Cargo.toml
```diff
@@ -142,6 +141,8 @@ members = [
     "sdk/ffi/go",
     "sdk/ffi/shared",
     "sdk/rust/nym-sdk",
+    "mixtcp",
+    # "service-providers/authenticator",
```
jstuczyn: why are we removing the authenticator?

common/client-core/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
```
jstuczyn: just for the reference, what's deprecated there?

common/client-core/surb-storage/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2022 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+#![allow(deprecated)]
```
jstuczyn: just for the reference, what's deprecated there?

jstuczyn: I'm going to stop putting this under every single instance. this comment applies to all of  those : )

common/nymsphinx/anonymous-replies/src/requests/mod.rs
```diff
@@ -30,7 +31,7 @@ pub enum InvalidAnonymousSenderTagRepresentation {
     InvalidLength { received: usize, expected: usize },
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
```
jstuczyn, line 34: I'm not entirely sure you want to just derive serde on it. it really depends on the context where it's used. for example, I can imagine, in js world stringified version would be way more desirable.

not saying what you've done is wrong. just think about it

common/task/src/connections.rs
```diff
@@ -2,13 +2,14 @@
...
 use futures::channel::mpsc;
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 // const LANE_CONSIDERED_CLEAR: usize = 10;
 
 pub type ConnectionId = u64;
 
-#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, Serialize, Deserialize)]
```
jstuczyn, line 12: that looks like some serious red flag. why would we ever need to serialise TransmissionLane? where are we sending it across?

sdk/rust/nym-sdk/src/lib.rs
```diff
@@ -1,14 +1,17 @@
 //! Rust SDK for the Nym platform
 //!
 //! The main component currently is [`mixnet`].
-//! [`tcp_proxy`] is probably a good place to start for anyone wanting to integrate with existing app code and read/write from a socket.
 //! [`client_pool`] is a configurable client pool.
+//!
+//! TODO OTHER MODULES
```
jstuczyn: well? : )

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -280,6 +304,43 @@ impl MixnetClient {
             }
         }
     }
+    // fn read_buffer_to_slice(
```
jstuczyn: old dead code?

common/nymsphinx/forwarding/src/packet.rs
```diff
@@ -174,13 +179,51 @@ impl MixPacket {
...
+    pub fn into_v2_bytes(&self) -> Result<Vec<u8>, MixPacketFormattingError> {
         Ok(std::iter::once(self.packet_type as u8)
             .chain(std::iter::once(self.key_rotation as u8))
             .chain(self.next_hop.as_bytes())
             .chain(self.packet.to_bytes()?)
             .collect())
     }
+
+    /*
+    * =======
```
jstuczyn: seems to be some old conflicts?

common/nymsphinx/forwarding/src/packet.rs
```diff
@@ -174,13 +179,51 @@ impl MixPacket {
         })
     }
 
-    pub fn into_v2_bytes(self) -> Result<Vec<u8>, MixPacketFormattingError> {
+    pub fn into_v2_bytes(&self) -> Result<Vec<u8>, MixPacketFormattingError> {
```
jstuczyn: why do we need to preserve the old packet?
if there's a reason for it, the new method should be exposed as `to_v2_bytes`. by convention an `into` consumes `self`

common/nymsphinx/forwarding/src/packet.rs
```diff
@@ -174,13 +179,51 @@ impl MixPacket {
...
+            Ok(std::iter::once(self.packet_type as u8)
+    >>>>>>> 1926e2950 (InputMessageCodec, Serde for MixPacket)
+                .chain(self.next_hop.as_bytes())
+                .chain(self.packet.to_bytes()?)
+                .collect())
+    */
+}
+
+// MAX TODO implement for v1 as well for back compat - this was added in the original asyncread/write work when we only had one v
+impl Serialize for MixPacket {
```
jstuczyn, line 192: another possible red flag. why do we need to have `serde` for `MixPacket`? this should never really be serialised directly

common/client-core/src/client/inbound_messages.rs
```diff
@@ -185,4 +199,71 @@ impl InputMessage {
...
+            return Ok(None);
+        }
+        #[allow(clippy::expect_used)]
+        let len = u32::from_le_bytes(
+            buf[0..LENGHT_ENCODING_PREFIX_SIZE]
+                .try_into()
+                .expect("Could not coarce to array"),
+        ) as usize;
+        if buf.len() < len + LENGHT_ENCODING_PREFIX_SIZE {
+            return Ok(None);
```
jstuczyn, line 253: nit: in all those cases where you return none, you could instead pre-reserve the minimum number of bytes you know you'd need, as recommended by the trait itself

common/client-core/src/client/inbound_messages.rs
```diff
@@ -185,4 +199,71 @@ impl InputMessage {
...
+        }
+
+        let decoded = match bincode::deserialize(
```
jstuczyn: don't use the default implementation, as that behaviour is changing in bincode 2.0 (which we eventually want to migrate to). instead create an explicit serialiser (look in our codebase for `make_bincode_serializer` which we define in multiple places for this very reason)

common/client-core/src/client/inbound_messages.rs
```diff
@@ -185,4 +199,71 @@ impl InputMessage {
...
         self
     }
+    #[allow(clippy::expect_used)]
+    pub fn serialized_size(&self) -> u64 {
+        bincode::serialized_size(self).expect("failed to get serialized InputMessage size")
+            + LENGHT_ENCODING_PREFIX_SIZE as u64
+    }
+}
+
+// TODO: Tests
```
jstuczyn: yeah, those would be quite vital especially with the third party serialisation that we will want to eventually change

common/nymsphinx/src/receiver.rs
```diff
@@ -15,10 +18,13 @@ use nym_sphinx_chunking::reconstruction::MessageReconstructor;
...
 };
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tokio_util::bytes::{Buf, BytesMut};
+use tokio_util::codec::{Decoder, Encoder};
 
 // TODO: should this live in this file?
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct ReconstructedMessage {
```
jstuczyn, line 29: Similarly to the `MixPacket`, why do we need serde on this guy?

common/nymsphinx/src/receiver.rs
```diff
@@ -56,6 +62,62 @@ impl From<PlainMessage> for ReconstructedMessage {
...
+
+impl Encoder<ReconstructedMessage> for ReconstructedMessageCodec {
+    type Error = MessageRecoveryError;
+
+    fn encode(
+        &mut self,
+        item: ReconstructedMessage,
+        buf: &mut BytesMut,
+    ) -> Result<(), Self::Error> {
+        let encoded = bincode::serialize(&item).expect("failed to serialize ReconstructedMessage");
```
jstuczyn: if serialisation fails, we shouldn't blow up, we should rather return an error

common/nymsphinx/src/receiver.rs
```diff
@@ -56,6 +62,62 @@ impl From<PlainMessage> for ReconstructedMessage {
...
+        let encoded_len = encoded.len() as u32;
+        let mut encoded_with_len = encoded_len.to_le_bytes().to_vec();
+        encoded_with_len.extend(encoded);
+        buf.reserve(encoded_with_len.len());
+        buf.extend_from_slice(&encoded_with_len);
```
jstuczyn: rather than allocating and copying the whole message in `encoded_with_len`, you can extend the buffer twice instead. once with the length and once with the actual content

common/nymsphinx/src/receiver.rs
```diff
@@ -56,6 +62,62 @@ impl From<PlainMessage> for ReconstructedMessage {
...
+    }
+}
+
+impl Decoder for ReconstructedMessageCodec {
+    type Item = ReconstructedMessage;
+    type Error = MessageRecoveryError;
+
+    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if buf.len() < LENGHT_ENCODING_PREFIX_SIZE {
+            return Ok(None);
```
jstuczyn, line 92: same comments about allocating buffer before returning

common/nymsphinx/src/receiver.rs
```diff
@@ -56,6 +62,62 @@ impl From<PlainMessage> for ReconstructedMessage {
...
+        }
+
+        let decoded = match bincode::deserialize(
+            &buf[LENGHT_ENCODING_PREFIX_SIZE..len + LENGHT_ENCODING_PREFIX_SIZE],
+        ) {
+            Ok(decoded) => decoded,
+            Err(e) => {
+                debug!("Failed to decode the message - {:?}", e);
+                return Ok(None);
```
jstuczyn: if decoding fails, we should terminate the stream, i.e. return an error, rather than returning `None`. None implies we don't have enough bytes to read a message and that the stream should retry later. so we'll just get stuck in a constant retry loop

common/socks5/proxy-helpers/src/proxy_runner/mod.rs
```diff
@@ -35,7 +36,7 @@ impl From<(Vec<u8>, bool)> for ProxyMessage {
     }
 }
 
-pub type MixProxySender<S> = tokio::sync::mpsc::Sender<S>;
+pub type MixProxySender<S> = PollSender<S>;
```
jstuczyn, line 39: what's the reason for changing this into a `PollSender`?

mixtcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+nym-bin-common = { path = "../common/bin-common", features = ["basic_tracing"] }
+nym-sdk = { path = "../sdk/rust/nym-sdk" }
+nym-ip-packet-requests = { path = "../common/ip-packet-requests" }
+thiserror.workspace = true
+reqwest.workspace = true
+dirs.workspace = true
+rustls  = { version = "0.23", default-features = false, features = ["std"] }
+webpki-roots = "1.0.2"
```
jstuczyn: use workspace deps please

mixtcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "mixtcp"
+version = "0.0.1"
+edition = "2021"
+license.workspace = true
+
+[dependencies]
+smoltcp = { version = "0.12", default-features = false, features = [
```
jstuczyn: use workspace deps please

sdk/rust/nym-sdk/src/mixnet/traits.rs
```diff
@@ -16,9 +16,11 @@ pub trait MixnetMessageSender {
         None
     }
 
+    fn sender(&mut self) -> &mut tokio_util::sync::PollSender<InputMessage>;
```
jstuczyn, line 19: this is leaking the internal implementation. why are we exposing the internal channel here? couldn't we instead require/implement methods to avoid it?
for example we have all different variants of `send`, `send_plain`, etc. that are used by everything that implements this trait

service-providers/network-requester/src/core.rs
```diff
@@ -346,7 +350,7 @@ impl NRServiceProvider {
                 socks5_msg = mix_input_reader.recv() => {
                     if let Some(msg) = socks5_msg {
                         let response_message = msg.into_input_message(packet_type);
-                        mixnet_client_sender.send(response_message).await.unwrap();
+                        nym_sdk::mixnet::MixnetMessageSender::send(&mut mixnet_client_sender, response_message).await.unwrap();
```
jstuczyn, line 353: doesn't `MixnetClientSender` implement the required trait? why do we need to be so explicit about it?

common/client-core/src/client/inbound_messages.rs
```diff
@@ -1,16 +1,26 @@
...
+use tokio_util::{
+    bytes::Buf,
+    bytes::BytesMut,
+    codec::{Decoder, Encoder},
+};
+
+use crate::error::ClientCoreError;
 
-pub type InputMessageSender = tokio::sync::mpsc::Sender<InputMessage>;
+pub type InputMessageSender = tokio_util::sync::PollSender<InputMessage>;
```
jstuczyn, line 19: do we have to be using `PollSender` instead of a normal `Sender`? this makes things much more annoying to use because suddenly sending messages, as seen, requires mutable reference what is quite restricting. could we make it somehow optional, i.e. make it "tranform" normal sender into this poll sender on demand, i.e. only for the stream case?

mixtcp/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Mixnet to remote hosts via an IPR.
+///
+/// This function handles the complete setup process:
+/// - Ensures the IPR stream is connected
+/// - Retrieves allocated IP addresses
+/// - Creates communication channels
+/// - Constructs the device and bridge components
+pub async fn create_device(
+    mut ipr_stream: IpMixStream,
+) -> Result<(NymIprDevice, NymIprBridge, IpPair), MixtcpError> {
```
jstuczyn, line 26: given you're returning a tuple with 3 different elements, wouldn't it make more sense to wrap it in some struct instead?

mixtcp/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("Nym SDK error: {0}")]
+    NymSdk(#[from] nym_sdk::Error),
+
+    #[error("TLS handshake failed")]
+    TlsHandshakeFailed,
+
+    #[error("TLS encrypt/decrypt error")]
+    TlsCrypto,
+
+    #[error("DNS err placeholder")]
```
jstuczyn, line 23: is it still a placeholder?

mixtcp/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("TCP connection failed")]
+    TcpConnectionFailed,
+
+    #[error("Response read failed")]
+    ResponseReadFailed,
+
+    #[error("No response received")]
+    NoResponseReceived,
+
+    #[error("Invalid HTTP response")]
```
jstuczyn, line 50: I think some additional information here would be useful. like status code, response, etc

mixtcp/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Create a connected smoltcp device and async bridge for the tunneling packets through the
+/// Mixnet to remote hosts via an IPR.
+///
+/// This function handles the complete setup process:
+/// - Ensures the IPR stream is connected
+/// - Retrieves allocated IP addresses
+/// - Creates communication channels
+/// - Constructs the device and bridge components
+pub async fn create_device(
+    mut ipr_stream: IpMixStream,
```
jstuczyn, line 25: it almost feels as if mixtcp could have been its own independent crate without direct dependency on the sdk. could we perhaps abstract that `IpMixStream` away with some trait instead?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+    }
+
+    Ok(described)
+}
+
+async fn get_random_ipr(client: nym_http_api_client::Client) -> Result<Recipient, Error> {
+    let nodes = retrieve_exit_nodes_with_performance(client).await?;
+    info!("Found {} Exit Gateways", nodes.len());
+
+    // JS: I'm leaving the old behaviour here of choosing node with the highest performance,
```
jstuczyn, line 140: I think this should be addressed

mixtcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+bytes = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+nym-bin-common = { path = "../common/bin-common", features = ["basic_tracing"] }
+nym-sdk = { path = "../sdk/rust/nym-sdk" }
+nym-ip-packet-requests = { path = "../common/ip-packet-requests" }
+thiserror.workspace = true
+reqwest.workspace = true
+dirs.workspace = true
+rustls  = { version = "0.23", default-features = false, features = ["std"] }
```
jstuczyn: it seems some of those dependencies, rust `rustls` are only needed by the examples - perhaps we could prune that section then?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+    ipr_address: Recipient,
+    listener: IprListener,
+    allocated_ips: Option<IpPair>,
+    connection_state: ConnectionState,
+}
+
+impl IpMixStream {
+    // TODO be able to pass in DisconnectedMixnetClient to use as MixStream inner client.
+    pub async fn new() -> Result<Self, Error> {
+        // JS: I think you need some bootstrapping here, something would need to be passed to `new()`
```
jstuczyn: this

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+    peer_surbs: Option<AnonymousSenderTag>, // ..since we might just be using SURBs instead
+}
+
+impl MixStream {
+    /// Create a MixStream instance and immediately connect (convenience method) or pass in a MixSocket (pre-configured DisconnectedMixnetClient) and connect it.
+    // TODO in future take config from MixSocket if exists in Option<> param, else spin up ephemeral client. Just doing ephemeral for initial sketch.
+    pub async fn new(socket: Option<MixSocket>, peer: Recipient) -> Self {
+        let client = match socket {
+            Some(socket) => socket.into_inner(),
+            None => MixnetClient::connect_new().await.unwrap(),
```
jstuczyn: just so we're clear, this is sending cover traffic, poisson, etc.

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+
+pub struct MixStream {
+    pub client: MixnetClient,
+    peer: Option<Recipient>, // We might be accepting incoming messages and replying, so might not have a Nym addr to talk to..
+    peer_surbs: Option<AnonymousSenderTag>, // ..since we might just be using SURBs instead
+}
+
+impl MixStream {
+    /// Create a MixStream instance and immediately connect (convenience method) or pass in a MixSocket (pre-configured DisconnectedMixnetClient) and connect it.
+    // TODO in future take config from MixSocket if exists in Option<> param, else spin up ephemeral client. Just doing ephemeral for initial sketch.
```
jstuczyn: some additional config would definitely be useful. such as being able to run on different networks

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+    /// Following idea of having single client with multiple concurrent connections represented by per-Recipient MixStream instance.
+    pub async fn new() -> Result<Self, Error> {
+        // TODO make this take an option, if Some create with builder pattern, if None make ephemeral like we already are.
+        let inner = MixnetClient::connect_new().await?;
+        Ok(MixSocket { inner })
+    }
+
+    /// Connect to a specific peer (Nym Client) and return a Stream (cf TcpSocket::connect() / TcpStream::new()).
+    pub async fn connect_to(_recipient: Recipient) -> Result<MixStream, Error> {
+        todo!()
```
jstuczyn: deadcode?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+        Self {
+            client,
+            peer: Some(peer),
+            peer_surbs: None,
+        }
+    }
+
+    /// Nym address of Stream's peer (Nym Client it will communicate with).
+    pub fn peer_addr(&self) -> Recipient {
+        let peer = &self.peer.expect("No Peer set");
```
jstuczyn: should it blow up if there's no peer? or would it be better if it returned an option instead

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+        *peer
+    }
+
+    /// Our Nym address.
+    pub fn local_addr(&self) -> &Recipient {
```
jstuczyn, line 379: nit: `Recipient` implements `Copy`

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+        self.client.nym_address()
+    }
+
+    /// Store SURBs in own field.
```
jstuczyn: does it though? you just provide it a tag. no actual surbs are attached

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub fn surbs(&self) -> Option<AnonymousSenderTag> {
+        self.peer_surbs
+    }
+
+    /// Split for concurrent read/write (like TcpStream::Split) into MixnetStreamReader and MixnetStreamWriter.
+    pub fn split(self) -> (MixStreamReader, MixStreamWriter) {
+        debug!("Splitting MixStream");
+        let sender = self.client.split_sender();
+        debug!("Split MixStream into Reader and Writer");
+        let (surb_tx, surb_rx) = oneshot::channel();
```
jstuczyn, line 454: why is it a oneshot? it means we can only send the surb data once. is that what we want?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+            },
+        )
+    }
+
+    /// Convenience method for just piping bytes into the Mixnet.
+    pub async fn write_bytes(&mut self, data: &[u8]) -> Result<(), Error> {
+        let input_message = if self.peer_surbs.is_some() {
+            info!("Writing {} bytes, sending with SURBs", data.len());
+            InputMessage::Reply {
+                recipient_tag: (self.peer_surbs.expect("No Peer SURBs set")),
```
jstuczyn: again, do we want to be blowing up if this is not set?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+                reply_surbs: (10),
+                lane: (nym_task::connections::TransmissionLane::General),
+                max_retransmissions: (Some(5)), // TODO check with Drazen - guessing here
+            }
+        };
+
+        let mut codec = InputMessageCodec {};
+        let mut serialized_bytes = BytesMut::new();
+        codec.encode(input_message, &mut serialized_bytes)?;
+        debug!("Serialized bytes: {:?}", serialized_bytes);
```
jstuczyn: providing actual bytes in the debug output could be quite annoying

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
```diff
@@ -0,0 +1,523 @@
...
+            InputMessage::Anonymous {
+                recipient: (self.peer),
+                data: (data.to_owned()),
+                reply_surbs: (10),
+                lane: (nym_task::connections::TransmissionLane::General),
+                max_retransmissions: (Some(5)), // TODO check with Drazen - guessing here
+            }
+        };
+
+        let mut codec = InputMessageCodec {};
```
jstuczyn, line 823: could we perhaps extract all the code to do with the codec encoding/decoding to a separate place? 1. to avoid duplication 2. to make sure the implementation is consistent

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum ConnectionState {
+    Disconnected,
+    Connecting,
+    Connected,
+}
+
+fn create_nym_api_client(nym_api_urls: Vec<ApiUrl>) -> Result<nym_http_api_client::Client, Error> {
+    // TODO do something proper with this
+    let user_agent = UserAgent {
```
jstuczyn: that will rot quite quickly

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+                        None => {
+                            return Err(Error::IPRClientStreamClosed);
+                        }
+                        Some(Err(e)) => {
+                            return Err(Error::MessageRecovery(e));
+                        }
+                        Some(Ok(reconstructed)) => {
+                            if let Err(e) = check_ipr_message_version(&reconstructed) {
+                                error!("Version check failed: {}", e);
+                                continue;
```
jstuczyn: if version check failed, is there any point in continuing given we won't be able to talk with the IPR?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+                            }
+                            if let Ok(response) = IpPacketResponse::from_reconstructed_message(&reconstructed) {
+                                if response.id() == Some(request_id) {
```
jstuczyn, line 308: what if the id doesn't match?

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
```diff
@@ -0,0 +1,777 @@
...
+    pub async fn send_ip_packet(&mut self, packet: &[u8]) -> Result<(), Error> {
+        if self.connection_state != ConnectionState::Connected {
+            return Err(Error::IprStreamClientNotConnected);
+        }
+        let request = IpPacketRequest::new_data_request(packet.to_vec().into());
+        self.send_ipr_request(request).await
+    }
+
+    pub async fn handle_incoming(&mut self) -> Result<Vec<Bytes>, Error> {
+        let mut framed = FramedRead::new(&mut self.stream, ReconstructedMessageCodec {});
```
jstuczyn: it seems that whenever we read from the stream we always wrap it in the codec. couldn't we instead store the wrapped stream on `self` directly instead?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +349,193 @@ pub struct MixnetClientSender {
...
+            return self.read_buffer_to_slice(buf, cx);
+        }
+
+        let msg = match self.as_mut().poll_next(cx) {
+            Poll::Ready(Some(msg)) => msg,
+            Poll::Ready(None) => return Poll::Ready(Ok(())),
+            Poll::Pending => return Poll::Pending,
+        };
+
+        match codec.encode(msg, &mut self._read.buffer) {
```
jstuczyn, line 352: nit: no need for match, you can just do `if let Err(err) = ....`

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -60,6 +65,24 @@ pub struct MixnetClient {
...
+    fn clear(&mut self) {
+        self.buffer.clear();
+    }
+
+    fn pending(&self) -> bool {
+        !self.buffer.is_empty()
+    }
 }
 
 impl MixnetClient {
```
jstuczyn, line 88: it feels a bit weird that `AsyncRead` and `AsyncWrite` operate on different message abstractions - `InputMessage` and `ReconstructedMessage` - is there no way to somehow unify it those to be just bytes?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +349,193 @@ pub struct MixnetClientSender {
...
+
+        let msg_size = msg.serialized_size();
+
+        let mut fut = pin!(self.client_input.send(msg));
+        match fut.poll_unpin(cx) {
+            Poll::Ready(Ok(())) => Poll::Ready(Ok(msg_size as usize)),
+            Poll::Ready(Err(_)) => Poll::Ready(Err(std::io::Error::other(
+                "failed to send message to mixnet",
+            ))),
+            Poll::Pending => Poll::Pending,
```
jstuczyn, line 394: this branch will be returned in almost 99% of cases. However, we're immediately dropping the sending future so `poll_write` might never be called again. you might have to store the future itself on `self` (or some intermediate struct) so that it could be polled again

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +349,193 @@ pub struct MixnetClientSender {
...
+                    "failed to read message from input",
+                )))
+            }
+            Poll::Pending => return Poll::Pending,
+            Poll::Ready(None) => return Poll::Ready(Ok(0)),
+        };
+
+        let msg_size = msg.serialized_size();
+
+        let mut fut = pin!(self.client_input.send(msg));
```
jstuczyn, line 466: same issue about this future

sdk/rust/nym-sdk/src/ip_packet_client/connect.rs
```diff
@@ -0,0 +1,196 @@
...
+pub type SharedMixnetClient = Arc<tokio::sync::Mutex<Option<MixnetClient>>>;
+
+const IPR_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum ConnectionState {
+    Disconnected,
+    Connecting,
+    Connected,
+    #[allow(unused)]
```
jstuczyn, line 36: if it's unused, what's the point of retaining it?

sdk/rust/nym-sdk/src/ip_packet_client/connect.rs
jstuczyn, line 1: all those unwraps here are quite scary

sdk/rust/nym-sdk/src/ip_packet_client/connect.rs
```diff
@@ -0,0 +1,196 @@
...
+                msgs = mixnet_client.wait_for_messages() => match msgs {
+                    None => {
+                        return Err(Error::NoMixnetMessagesReceived);
+                    }
+                    Some(msgs) => {
+                        for msg in msgs {
+                            // Confirm that the version is correct
+                            if let Err(err) = check_ipr_message_version(&msg) {
+                                tracing::info!("Mixnet message version mismatch: {err}");
+                                continue;
```
jstuczyn: if we have a version mismatch, should we continue?

sdk/rust/nym-sdk/src/ip_packet_client/connect.rs
```diff
@@ -0,0 +1,196 @@
...
+                            }
+
+                            // Then we deserialize the message
+                            tracing::debug!("IprClient: got message while waiting for connect response");
+                            let Ok(response) = IpPacketResponse::from_reconstructed_message(&msg) else {
+                                // This is ok, it's likely just one of our self-pings
```
jstuczyn, line 167: so could we explicitly check for that? otherwise we might have a corrupt stream that is never closed

sdk/rust/nym-sdk/src/ip_packet_client/helpers.rs
```diff
@@ -0,0 +1,368 @@
...
+    ipv6::{Ipv6Packet, MutableIpv6Packet},
+    Packet,
+};
+use std::cmp::Ordering;
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+/**
+ * This function is from the original nym-ip-packet-client crate.
+ */
+pub(crate) fn check_ipr_message_version(message: &ReconstructedMessage) -> Result<()> {
```
jstuczyn, line 31: was there a reason for not importing it directly then?

sdk/rust/nym-sdk/src/ip_packet_client/helpers.rs
```diff
@@ -0,0 +1,368 @@
...
+                Ok(())
+            }
+        }
+    } else {
+        Err(Error::NoVersionInMessage)
+    }
+}
+
+/**
+ * Functions below are from the nym-connection-monitor crate.
```
jstuczyn, line 54: again, couldn't we just use the crate directly (since it's in the monorepo) rather than duplicating the non-trivial code?

sdk/rust/nym-sdk/src/ip_packet_client/listener.rs
```diff
@@ -0,0 +1,133 @@
...
+            }
+            _ => {
+                debug!("Received unexpected request: {request:?}");
+                false
+            }
+        }
+    }
+
+    pub async fn handle_reconstructed_message(
+        &mut self,
```
jstuczyn, line 63: why does it need to be mutable given `IprListener` literally doesn't have anything on it?

mixtcp/src/device.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    pub fn tx_sender(&self) -> mpsc::UnboundedSender<Vec<u8>> {
+        self.tx_sender.clone()
+    }
+
+    /// Get the receiver for external use
+    pub fn rx_receiver(&self) -> mpsc::UnboundedReceiver<Vec<u8>> {
+        // Create a new channel and return the receiver
+        // This is a bit of a hack but necessary for the current architecture
+        let (_tx, rx) = mpsc::unbounded_channel();
```
jstuczyn, line 81: wait. what's the point of a channel that's immediately getting closed?

mixtcp/src/bridge.rs
```diff
@@ -0,0 +1,120 @@
...
+                Some(packet) = self.tx_receiver.recv() => {
+                    info!("Bridge sending {} byte packet to mixnet", packet.len());
+
+                    // Log packet details for debugging
+                    if packet.len() >= 20 {
+                        let version = (packet[0] >> 4) & 0xF;
+                        let proto = packet[9];
+                        let src_ip = &packet[12..16];
+                        let dst_ip = &packet[16..20];
+                        info!(
```
jstuczyn, line 75: should it be `info` log?

mixtcp/src/bridge.rs
```diff
@@ -0,0 +1,120 @@
...
+                            "Outgoing IPv{} packet: proto={}, src={}.{}.{}.{}, dst={}.{}.{}.{}",
+                            version, proto,
+                            src_ip[0], src_ip[1], src_ip[2], src_ip[3],
```
jstuczyn, line 78: well, those logs will only work if it's an ipv4 packet

jstuczyn (commented):
I haven't got very deep into the actual mixtcp implementation, since you've had Drazen help you with that and because there was quite a lot of code to go through. 

I will definitely have a second pass on it all later

common/task/src/connections.rs
mfahampshire: This is coming from @durch 's original asyncread/write PR which never got merged (https://github.com/nymtech/nym/pull/4634/files#diff-c80e5e164df5ab2498f1f3f9198d709f823319eee8492874b346aa514a2cc4c6)

common/socks5/proxy-helpers/src/proxy_runner/mod.rs
mfahampshire: This was also from Drazen's original PR, will check in with him

mixtcp/src/lib.rs
mfahampshire: Good point, will look into it

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
mfahampshire: Yes this is a vague comment, its more storing the tag to use for anonymous replies, will change

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
mfahampshire: the assumption was that since the stream is already with one other client, that we're only having to store the tag once. although maybe i borked something there, will revisit

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
mfahampshire: I could either switch to trace or just remove it, that's true

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
mfahampshire: the deliberate setting of the user agent you mean? will look into whether there's a sensible default we can just pull in

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper_ipr.rs
mfahampshire: good point

sdk/rust/nym-sdk/src/ip_packet_client/helpers.rs
mfahampshire: this is my oversight - I don't think we'd pulled it into the monorepo when i originally wrote this and forgot in my rebasing when getting this ready for review

Cargo.toml
mfahampshire: It was removed in a previous PR / no longer exists in `develop` - I commented it out instead of removing it for some reason. Removing

common/client-core/src/lib.rs
mfahampshire: it removes the clippy err for: `warning: use of deprecated associated function nym_crypto::generic_array::GenericArray::<T, N>::clone_from_slice: please upgrade to generic-array 1.x` - I will add in a comment on the `allow(deprecated)`s what is being ignored. Followed Drazen's advice to ignore instead of start messing with the crypto crate on this PR.

common/nymsphinx/forwarding/src/packet.rs
durch: I think the idea was to make it easy and foolproof to get the pure bytes representation, ie make it easy to do the right thing.

common/nymsphinx/forwarding/src/packet.rs
mfahampshire: Its in there from what I can tell because `serializer.serialize_bytes` expects `&[u8]`

mfahampshire: I can change the method to follow the convention

sdk/rust/nym-sdk/src/mixnet/native_client.rs
mfahampshire: removed

common/nymsphinx/forwarding/src/packet.rs
mfahampshire: removed

common/client-core/surb-storage/src/lib.rs
mfahampshire: added the compiler warning (that becomes a clippy err) that each of these fixes to the ones i've added in this PR

sdk/rust/nym-sdk/src/stream_wrapper/mixnet_stream_wrapper.rs
mfahampshire: Not sure I understand this comment - I would hope it was since its basically just a wrapper around a MixnetClient. Or am I missing something?

mixtcp/src/lib.rs
mfahampshire: If we're assuming that it will always use the Mixnet though, what does implementing a trait give us here? Or do you mean that we could try and abstract `IpMixStream` out from the `nym-sdk` crate?

mixtcp/Cargo.toml
mfahampshire: moved to `dev-dependencies`

sdk/rust/nym-sdk/src/ip_packet_client/helpers.rs
mfahampshire: Ah no, it relies on the sdk, so I'd hit circular dependency. ~~another reason to try pull the IPR streamer out of the sdk?~~

sdk/rust/nym-sdk/src/ip_packet_client/connect.rs
mfahampshire: It was from the original code I copied in, removed

sdk/rust/nym-sdk/src/ip_packet_client/helpers.rs
mfahampshire: See above comment

common/client-core/src/client/inbound_messages.rs
mfahampshire: I think we need it because of the `Sink` impl in `sdk/rust/nym-sdk/src/mixnet/native_client.rs`. 

I can have a look @ whether there's some kind of transformation we could do for the stream case.

common/socks5/proxy-helpers/src/proxy_runner/mod.rs
mfahampshire: See https://github.com/nymtech/nym/pull/6129/files#r2486722873

mfahampshire:
Breaking this up into a few more easily reviewed PRs: 
#6318 - asyncread/write implementation in client core 
#6320 - stream wrappers for SDK 
#6321 - mixtcp poc 

Going to close this PR in place of above.

PR #6130, opened 2025-10-17
VPN-4262: Update `Url` to return `url` and `front` fields.

trojanfoe:
The VPN client is using the `Url` type alot now and in order to avoid double URL-parsing we would like the content of the `Url` type exposed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6130)

### DIFF
--- a/common/http-api-client/src/url.rs
+++ b/common/http-api-client/src/url.rs
@@ -183,6 +183,11 @@ impl Url {
         })
     }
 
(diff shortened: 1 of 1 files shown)


jstuczyn (approved):

trojanfoe:
Thanks @jstuczyn !

PR #6126, opened 2025-10-16
Enable URL rotation and retries for mixnet gateway init

tommyv1987:
Changes:
- Multiple URL fallback with configurable retries (defaults to 3)
- Infallible URL conversion per Andrews feedback (Url::from() instead of parse().ok())
- Non-breaking builder pattern for BuilderConfig per Andrej's "too many arguments" feedback
- Reverted redundant node filtering per Andrew's clarification that API already filters by supported_roles.entry

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6126)

PR #6134, opened 2025-10-20
Domain fronting

tommyv1987:
This cannot be merged until this is done:

Enable URL rotation and retries for mixnet gateway [init](https://github.com/nymtech/nym/pull/6126)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6134)


common/client-core/src/client/base_client/mod.rs
```diff
@@ -263,6 +272,16 @@ where
         self
     }
 
+    /// Set network details for domain fronting support.
+    ///
+    /// When provided, the client will use network details (which include front_hosts)
+    /// to construct HTTP clients with domain fronting enabled.
+    #[must_use]
+    pub fn with_network_details(mut self, network_details: NymNetworkDetails) -> Self {
```
jstuczyn: if it's just for domain fronting, wouldn't it make more sense to only pass `Vec<ApiUrl>`?
this is for few reasons:
1. passing whole `NymNetworkDetails` might give wrong impression as if this allowed you to control, for example, contract endpoints, which is not the intention here
2. `NymNetworkDetails` allows for `None` value for the api urls, which again, is not what we want here
3. it is a more explicit way of stating we meant to only change domain fronting

common/client-core/src/client/base_client/mod.rs
```diff
@@ -863,9 +882,35 @@ where
...
+            network_details.is_some()
+        );
+
+        // If network details are provided, use from_network() which handles domain fronting
+        if let Some(network_details) = network_details {
+            tracing::info!(
+                "Building nym-api client from network details (with domain fronting support)"
+            );
+
+            let mut builder = nym_http_api_client::ClientBuilder::from_network(network_details)
```
jstuczyn: if you apply my suggestion from above, you will need to use a different constructor, which, purely coincidently, I exposed in https://github.com/nymtech/nym/pull/6136

jstuczyn (changes requested):

common/client-core/src/client/base_client/mod.rs
```diff
@@ -263,6 +272,26 @@ where
...
+        self.nym_api_urls = Some(nym_api_urls);
+        self
+    }
+
+    /// Set network details for domain fronting support (deprecated).
+    ///
+    /// Use `with_nym_api_urls()` instead to explicitly pass the API URLs.
+    #[must_use]
+    #[deprecated(note = "use explicit Self::with_nym_api_urls instead")]
+    pub fn with_network_details(self, network_details: NymNetworkDetails) -> Self {
```
jstuczyn: what's the point of adding and immediately deprecating this method?

common/client-core/src/client/base_client/mod.rs
```diff
@@ -863,9 +892,36 @@ where
...
+                builder = builder.with_user_agent(user_agent);
+            }
+
+            return builder.build().map_err(ClientCoreError::from);
+        }
+
+        // Fallback to basic client for backwards compatibility
+        tracing::debug!("Building basic nym-api HTTP client from config endpoints");
+
         let mut nym_api_urls = config.get_nym_api_endpoints();
```
jstuczyn, line 919: given context of #6126 do we also want to remove the randomness and instead use all apis?

tools/echo-server/src/lib.rs
```diff
@@ -226,7 +226,8 @@ mod tests {
                 error!("{err}");
                 // this is not an ideal way of checking it, but if test fails due to networking failures
                 // it should be fine to progress
-                if err.to_string().contains("nym api request failed") {
+                let err_str = err.to_string();
+                if err_str.contains("nym api") || err_str.contains("failed to connect") {
```
jstuczyn, line 230: what's the reason for these changes?

tommyv1987: The tests were failing due to network / connectivity issues with the changes made in the fallback-url branch. Then with BaseClientBuilder work this makes the http client more reliable. This fixes the echo-server test failures that were happening due to our single url selection

common/client-core/src/client/base_client/mod.rs
tommyv1987: Done

jstuczyn (approved):

PR #6142, opened 2025-10-23
LP registration

durch:
Summary

  This PR implements the Lewes Protocol (LP) registration path for direct TCP-based gateway connections, providing a more efficient alternative to mixnet-based registration while maintaining security through Noise protocol
  handshake and credential verification.

  What's Implemented

  Server-Side (Gateway)

  - LP Listener on port 41264 accepting TCP connections
  - Noise protocol handshake (XKpsk3_25519_ChaChaPoly_SHA256) for secure channel establishment
  - Registration handler supporting both dVPN and Mixnet modes
  - Credential verification using existing EcashManager infrastructure
  - WireGuard peer management integration for dVPN registrations
  - Session management with replay protection using bitmap-based counter validation

  Client-Side (RegistrationClient)

  - LpRegistrationClient module with TCP connection handling
  - State machine for LP handshake as initiator
  - Registration flow with request/response handling
  - Transport layer for post-handshake data transfer
  - Integration with existing RegistrationClient via register_lp() method
  - Result types (LpRegistrationResult) matching existing patterns

  Protocol Features

  - Length-prefixed framing for TCP packet boundaries
  - Random PSK generation per registration (32-byte)
  - Replay protection with 1024-packet reorder window
  - Registration modes: dVPN (WireGuard) and Mixnet

  Configuration & Defaults

  - Connection timeout: 10 seconds
  - Handshake timeout: 15 seconds
  - Registration timeout: 30 seconds
  - TCP optimizations: TCP_NODELAY enabled, no keepalive needed
  - No persistent connections: Registration-only protocol

  Architecture Changes

  New Modules

  - common/nym-lp/ - Core LP protocol implementation
  - common/nym-lp-common/ - Shared LP types
  - gateway/src/node/lp_listener/ - Gateway LP listener
  - nym-registration-client/src/lp_client/ - Client LP implementation

  Modified Components

  - NymNode struct now includes optional lp_address field
  - RegistrationClient supports RegistrationMode::Lp
  - Gateway startup includes LP listener initialization

  What's Outstanding

  Testing

  - Unit tests for LP protocol components
  - Integration tests for end-to-end registration flow
  - Real gateway testing with LP-enabled nodes

  Documentation

  - API documentation for LP registration usage
  - Configuration guide for LP-specific settings
  - Protocol specification documentation

  Future Enhancements (Not blocking)

  - Metrics collection for LP operations
  - Gateway probe support for LP endpoints
  - Advanced PSK management (currently using random generation)
  - Error handling improvements and categorization

  Breaking Changes

  None - LP is an additional registration path that doesn't affect existing WireGuard or Mixnet registration modes.

  Performance Impact

  LP registration provides:
  - Lower latency: Direct TCP connection vs multi-hop mixnet
  - Reduced overhead: No Sphinx packet wrapping
  - Faster registration: ~1-2 seconds vs 10-30 seconds for mixnet

  Security Considerations

  - Noise protocol provides mutual authentication and forward secrecy
  - Random PSK per registration prevents replay attacks
  - Credential verification ensures proper authorization
  - No persistent state reduces attack surface

  Migration Path

  No migration required. Clients can opt-in to LP registration by:
  1. Checking if gateway advertises LP address
  2. Using RegistrationMode::Lp when building RegistrationClient
  3. Falling back to existing modes if LP unavailable

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6142)


common/cosmwasm-smart-contracts/mixnet-contract/src/nym_node.rs
```diff
@@ -373,6 +373,11 @@ pub struct NymNode {
     /// Base58-encoded ed25519 EdDSA public key.
     #[cfg_attr(feature = "utoipa", schema(value_type = String))]
     pub identity_key: IdentityKey,
+
+    /// Optional LP (Lewes Protocol) listener address for direct gateway connections.
+    /// Format: "host:port", for example "1.1.1.1:41264" or "gateway.example.com:41264"
+    #[serde(default)]
+    pub lp_address: Option<String>,
```
jstuczyn, line 380: under what situation will lp host be any different from the existing `host` field? shouldn't we be just announcing the lp port (assuming it's non-default)?
as a matter of fact - do we even need to do that at all? why couldn't everything be retrieved via the self-described information that's available on the http endpoint? (like we're doing with say the sphinx keys and whatnot)

common/crypto/src/asymmetric/ed25519/mod.rs
```diff
@@ -334,6 +365,28 @@ impl PrivateKey {
...
+    /// Uses the standard ed25519→x25519 conversion via SHA-512 hash and clamping.
+    /// This is the same approach as libsodium's `crypto_sign_ed25519_sk_to_curve25519`.
+    ///
+    /// # Returns
+    /// The converted X25519 private key
+    pub fn to_x25519(&self) -> crate::asymmetric::x25519::PrivateKey {
+        use sha2::{Digest, Sha512};
+
+        // Hash the Ed25519 secret key with SHA-512
+        let hash = Sha512::digest(self.0);
```
jstuczyn, line 380: we need to zeroize those bytes otherwise we leave private key dangling in memory

common/wireguard/Cargo.toml
```diff
@@ -15,6 +15,9 @@ base64 = { workspace = true }
 defguard_wireguard_rs = { workspace = true }
 futures = { workspace = true }
 ip_network = { workspace = true }
+ipnetwork = { workspace = true }
+log.workspace = true
```
jstuczyn, line 19: that's just a super tiny nit. if we're adding new logs, can we do it via `tracing` instead of `log`?

contracts/mixnet/src/support/tests/mod.rs
```diff
@@ -967,6 +967,7 @@ pub mod test_helpers {
                 host: "1.2.3.4".to_string(),
                 custom_http_port: None,
                 identity_key,
+                lp_address: None,
```
jstuczyn, line 970: as I said before, I'm not sure this is the right place for the lp address

common/nym-kcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "wire_format"
+path = "bin/wire_format/main.rs"
+
+[[bin]]
+name = "session"
+path = "bin/session/main.rs"
+
+[dependencies]
+tokio-util = { workspace = true, features = ["codec"] }
+byte_string = "1.0"
```
jstuczyn, line 21: use workspace dependencies instead

common/nym-kcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+bytes = { workspace = true }
+thiserror = { workspace = true }
+log = { workspace = true }
```
jstuczyn, line 24: nit: if possible use `tracing` rather than `log`

common/nym-kcp/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "nym-kcp"
+version = "0.1.0"
+edition = { workspace = true }
+license = { workspace = true }
+
+[lib]
+name = "nym_kcp"
+path = "src/lib.rs"
+
+[[bin]]
```
jstuczyn, line 11: rather than dedicated bins, those more look like they'd fit better under `[example]`

common/nym-kcp/bin/session/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    // Set an MSS (max segment size) smaller than our data to force fragmentation
+    local_sess.set_mtu(40);
+    remote_sess.set_mtu(40);
+
+    // Some data larger than 30 bytes to demonstrate multi-fragment
+    let big_data = b"The quick brown fox jumps over the lazy dog. This is a test.";
+
+    // --- LOCAL sends data ---
+    info!(
```
jstuczyn, line 18: well, those logs (here and in the `wire_format`) actualy don't do anything because even though you included `env_logger` in dev-dependencies, no logger is actually being set up.

common/nym-kcp/Cargo.toml
jstuczyn: as a matter of fact, this dep doesn't seem to be used at all

common/nym-kcp/src/packet.rs
```diff
@@ -0,0 +1,219 @@
...
+            82 => Ok(KcpCommand::Ack),
+            83 => Ok(KcpCommand::Wask),
+            84 => Ok(KcpCommand::Wins),
+            _ => Err(KcpError::InvalidCommand(value)),
+        }
+    }
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<u8> for KcpCommand {
```
jstuczyn, line 44: well, to appease clippy we could always do `From<KcpCommand> for u8`

common/nym-kcp/src/packet.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn una(&self) -> u32 {
+        self.una
+    }
+}
+
+impl Default for KcpPacket {
+    fn default() -> Self {
+        // We must pick some default command, e.g. `Push`.
+        // Or omit `Default` if you don't need it.
+        KcpPacket {
```
jstuczyn, line 132: I'd perhaps omit the `Default` implementation completely as 
1) it's not being needed anywhere
2) doesn't make sense that `Push` would suddenly be chosen over anything else
3) could lead to unexpected surprise behaviour due to 2) and the fact all packets would suddenly have the same timestamp, the same conversation id, the same, well, everything

common/nym-kcp/src/packet.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+impl KcpPacket {
+    /// Attempt to decode a `KcpPacket` from `src`.
+    /// Returns Ok(Some(pkt)) if fully available, Ok(None) if not enough data,
+    /// or Err(...) if there's an invalid command or other error.
+    pub fn decode(src: &mut BytesMut) -> Result<Option<Self>, KcpError> {
+        trace!("Decoding buffer with len: {}", src.len());
+        if src.len() < KCP_HEADER {
+            // Not enough for even the header, this is usually fine, more data will arrive
```
jstuczyn, line 152: in that case we should allocate the bytes via `src.reserve(KCP_HEADER)`

common/nym-kcp/src/packet.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        let total_needed = KCP_HEADER + len;
+        if src.len() < total_needed {
+            // We don't have the full packet yet
+            debug!(
+                "Not enough data for packet, want {}, have {}",
+                total_needed,
+                src.len()
+            );
+            return Ok(None);
```
jstuczyn, line 177: same thing about reserving sufficient number of bytes for the whole packet. otherwise decoding might be needlessly called with insufficient data

common/nym-kcp/src/packet.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        // Convert the raw u8 into our KcpCommand enum
+        let cmd = KcpCommand::try_from(cmd_byte)?;
```
jstuczyn, line 181: we should probably check if we received a valid command before waiting for the packet data, so we could discard invalid packets more quickly

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+use ansi_term::Color::Yellow;
+use bytes::{Buf, BytesMut};
+use log::{debug, error, info, warn};
+use std::thread;
+
+use super::packet::{KcpCommand, KcpPacket};
+
+/// Minimal KCP session that produces/consumes `KcpPacket`s
+pub struct KcpSession {
+    pub conv: u32,
```
jstuczyn, line 16: it's a bit weird how this field is the only one exposed as public. can we instead do it via a getter, like we do for the other values? it doesn't seem to be mutated (thankfully) anywhere

common/nym-kcp/src/driver.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use crate::{error::KcpError, packet::KcpPacket, session::KcpSession};
+
+pub struct KcpDriver {
+    session: KcpSession,
+    buffer: BytesMut,
+}
+
+impl KcpDriver {
+    pub fn conv_id(&self) -> Result<u32, KcpError> {
```
jstuczyn, line 12: why does this return a `Result` if it's infallible?

common/nym-kcp/src/driver.rs
```diff
@@ -0,0 +1,60 @@
...
+                pkt.frg()
+            );
+            self._input(&pkt)?;
+            pkts.push(pkt);
+        }
+        Ok(pkts)
+    }
+
+    fn _input(&mut self, pkt: &KcpPacket) -> Result<(), KcpError> {
+        self.session.input(pkt);
```
jstuczyn, line 37: same thing, why does this return a Result if it's infallible?

common/nym-kcp/src/driver.rs
```diff
@@ -0,0 +1,60 @@
...
+        trace!(
+            "ts_flush: {}, ts_current: {}",
+            self.session.ts_flush(),
+            self.session.ts_current()
+        );
+        self.session.fetch_outgoing()
+    }
+
+    pub fn update(&mut self, tick: u64) {
+        self.session.update(tick as u32);
```
jstuczyn, line 51: if we have to downcast it (which could theoretically fail), why not update the interface to require `u32` instead?

common/nym-kcp/src/driver.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    pub fn send(&mut self, data: &[u8]) {
+        self.session.send(data);
+    }
+
+    pub fn input(&mut self, data: &[u8]) -> Result<Vec<KcpPacket>, KcpError> {
+        self.buffer.extend_from_slice(data);
+        let mut pkts = Vec::new();
+        while let Ok(Some(pkt)) = KcpPacket::decode(&mut self.buffer) {
```
jstuczyn, line 23: I think this has to be refactored because it will simply ignore any errors which i reckon should be returned back to the caller, no?

common/nym-kcp/src/driver.rs
```diff
@@ -0,0 +1,60 @@
+use bytes::BytesMut;
+use log::{debug, trace};
+
+use crate::{error::KcpError, packet::KcpPacket, session::KcpSession};
+
+pub struct KcpDriver {
```
jstuczyn, line 6: Actually, what's the point of the `KcpDriver`? it doesn't appear to be used anywhere, even on the `drazen/gateway-probe-localnet-mode` branch?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        // Build each fragment
+        for i in 0..count {
+            let size = std::cmp::min(self.mtu, data.len());
+            let chunk = &data[..size];
+
+            // KCP fragment numbering is REVERSED - last fragment has frg=0,
+            // first has frg=count-1. This allows receiver to know total count from first packet.
+            // In KCP, `frg` is set to the remaining fragments in reverse order.
+            // i.e., the last fragment has frg=0, the first has frg=count-1.
+            let frg = (count - i - 1) as u8;
```
jstuczyn, line 176: what if we have a huge data chunk to send, such that `count >= 256`, i.e. the `u8` cast fails?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        } else {
+            total_len.div_ceil(self.mtu)
+        };
+
+        debug!("Will send {} fragments", count);
+
+        // Build each fragment
+        for i in 0..count {
+            let size = std::cmp::min(self.mtu, data.len());
+            let chunk = &data[..size];
```
jstuczyn, line 170: instead of manually pointing to different parts of the original data slice, you could use `data.chunks(self.mtu)` (and then I guess also call `enumerate()` on this iterator -> I think that'd be less error prone and more readable

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+            // KCP fragment numbering is REVERSED - last fragment has frg=0,
+            // first has frg=count-1. This allows receiver to know total count from first packet.
+            // In KCP, `frg` is set to the remaining fragments in reverse order.
+            // i.e., the last fragment has frg=0, the first has frg=count-1.
+            let frg = (count - i - 1) as u8;
+
+            let seg = Segment {
+                sn: self.snd_nxt,
+                frg,
+                ts: 0,
```
jstuczyn, line 181: should new segments use timestamp of 0 or should they instead use `self.current`, i.e. the current time of the session? because, I assume, we'd want subsequent calls to `send()` to produce different ts, no?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+            pkt.sn(),
+            pkt.frg(),
+            pkt.wnd(),
+            pkt.ts(),
+            pkt.una()
+        );
+
+        // Check conv
+        if pkt.conv() != self.conv {
+            error!(
```
jstuczyn, line 218: given this error should really terminate a session, shouldn't the `input` function return a `Result` to let the caller know about it? because it seems with the current setup they might just happily continue consuming wrong packets

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        debug!(
+            "[ConvID: {}, Thread: {:?}] parse_una(una={})",
+            self.conv,
+            thread::current().id(),
+            una
+        );
+        // Remove *all* segments in snd_buf where seg.sn < una
+        // KCP's UNA confirms receipt of all segments *before* it.
+        let original_len = self.snd_buf.len();
+        {
```
jstuczyn, line 470: should we perhaps put this block behind `#[cfg(debug_assertions)]`? it seems a bit wasteful to perform this whole allocation just for one **debug** log

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        }
+
+        // Update remote window
+        self.rmt_wnd = pkt.wnd();
+
+        // Parse UNA first - crucial for clearing snd_buf before processing ACKs/data
+        self.parse_una(pkt.una());
+
+        // Log snd_buf state before specific command processing
+        let pre_cmd_sns: Vec<u32> = self.snd_buf.iter().map(|seg| seg.sn).collect();
```
jstuczyn, line 233: it seems that pattern of getting list of all sn in the buffer followed by a debug log repeats quite frequently - should we perhaps wrap it in another call instead?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        debug!(
+            "[ConvID: {}, Thread: {:?}] Parsing ack, sn: {}, ts: {}",
+            self.conv,
+            thread::current().id(),
+            sn,
+            ts
+        );
+        // find the segment in snd_buf
+        if let Some(pos) = self.snd_buf.iter().position(|seg| seg.sn == sn) {
+            let seg = self.snd_buf.remove(pos).unwrap();
```
jstuczyn, line 531: given we have to remove items from the buffer in random positions, is `VecDeque` really the best structure for it? what about, say `BTreeMap`? we still get ordering, double-ended iterator and random (albeit a bit more expensive) access

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+
+        match pkt.cmd() {
+            KcpCommand::Ack => {
+                self.parse_ack(pkt.sn(), pkt.ts());
+            }
+            KcpCommand::Push => {
+                debug!("Received push, sn: {}, frg: {}", pkt.sn(), pkt.frg());
+                // Data
+                // self.ack_push(pkt.sn(), self.current); // Send ack eventually
+                self.ack_push(pkt.sn(), pkt.ts());
```
jstuczyn, line 250: shouldn't the ack (and the actual data as a matter of fact) be pushed here conditionally, i.e. `if (_itimediff(sn, kcp->rcv_nxt + kcp->rcv_wnd) < 0)` // `sn < rcv_next + pkt.rcv_wnd`?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+        // Check if we have it
+        let mut insert_idx = self.rcv_buf.len();
+        for (i, seg) in self.rcv_buf.iter().enumerate() {
+            #[allow(clippy::comparison_chain)]
+            if pkt.sn() < seg.sn {
+                insert_idx = i;
+                break;
+            } else if pkt.sn() == seg.sn {
+                // duplicate
+                return;
```
jstuczyn, line 576: while this seems to be a sensible choice, if you receive a duplicate packet you won't be checking your buffer, i.e. `self.move_buf_to_queue();` meanwhile in the C implementation (which I assume you based this on) , they still check the queue regardless of the packet being repeated

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+                }
+                expected_sn = expected_sn.wrapping_add(1);
+            }
+
+            // If we didn't find a complete message sequence at the front
+            if end_segment_index.is_none() {
+                break;
+            }
+
+            let end_idx = end_segment_index.unwrap();
```
jstuczyn, line 628:
```suggestion
            let Some(end_idx) = end_segment_index else {
              break
            }

File: common/nym-kcp/src/session.rs [hunk 30]
```diff
@@ -0,0 +1,1778 @@
...
+
+    fn move_buf_to_queue(&mut self) {
+        // Loop as long as we can potentially extract a complete message from the front
+        loop {
+            // Check if the buffer starts with the next expected sequence number
+            if self.rcv_buf.is_empty() || self.rcv_buf[0].sn != self.rcv_nxt {
+                break; // Cannot start assembling a message now
+            }
+
+            // Scan ahead in rcv_buf to find if a complete message exists contiguously
```
jstuczyn, line 603: why are we concerned in delivering full messages as opposed to next fragment in order? should it be KCP's concern to de-fragment data?

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+                seg.xmit += 1;
+                // Exponential backoff: double RTO for this segment
+                seg.rto *= 2;
+                // Clamp to the session's maximum RTO (hardcoded as 60s for now)
+                const MAX_RTO: u32 = 60000; // Same as used in update_rtt
+                if seg.rto > MAX_RTO {
+                    seg.rto = MAX_RTO;
+                }
+                seg.resendts = self.current + seg.rto;
+                info!(
```
jstuczyn, line 429: if left at info, that will be quite noisy, I reckon

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+                seg.xmit = 1;
+                seg.resendts = self.current + seg.rto;
+            } else if self.current >= seg.resendts {
+                // time to retransmit
+                need_send = true;
+                seg.xmit += 1;
+                // Exponential backoff: double RTO for this segment
+                seg.rto *= 2;
+                // Clamp to the session's maximum RTO (hardcoded as 60s for now)
+                const MAX_RTO: u32 = 60000; // Same as used in update_rtt
```
jstuczyn, line 424: nit: in that case could we just move it to the top of the file and not duplicate a magic constant (and so that we could guarantee it's actually the same as update_rtt?)

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+            if seg.xmit == 0 {
+                // never sent
+                need_send = true;
+                seg.xmit = 1;
+                seg.resendts = self.current + seg.rto;
+            } else if self.current >= seg.resendts {
+                // time to retransmit
+                need_send = true;
+                seg.xmit += 1;
+                // Exponential backoff: double RTO for this segment
```
jstuczyn, line 421: wait. wasn't the whole idea of KCP that it does **NOT** have expontential backoff? 
> In TCP, retransmission timeout (RTO) increases exponentially—each failure doubles the timeout interval (RTO × 2). This means three consecutive losses can escalate to RTO × 8, leading to serious transmission delays. KCP, in contrast, uses a more responsive approach in fast mode: the RTO increases by a factor of 1.5x

common/nym-kcp/src/session.rs
```diff
@@ -0,0 +1,1778 @@
...
+                    seg.frg,
+                    self.rcv_wnd,
+                    seg.ts, // original send timestamp
+                    seg.sn,
+                    self.rcv_nxt, // self.rcv_nxt for ack
+                    seg.data.clone(),
+                );
+                self.out_pkts.push(pkt);
+
+                // if too many xmit => dead_link check, etc.
```
jstuczyn, line 452: so that's still to be done, right?

jstuczyn (commented):
just a small set of comments based on a smaller subset of changes ( ~~didn't have a chance to look into LP proper~~ UPDATE: LP-related comments are in the following messages ):

- I'm unsure about the changes to nym-node bonding -> why couldn't LP information be put into the self-described endpoint (which has to be queried anyway to retrieve, for example, the sphinx key)? making changes to the contract is always quite tricky especially when we want to ensure backwards compatibility
- I think there might be an issue with the KCP driver and the way it handles new input data. it is defined as follows:

```rust
pub fn input(&mut self, data: &[u8]) -> Result<Vec<KcpPacket>, KcpError> {
    self.buffer.extend_from_slice(data);
    let mut pkts = Vec::new();
	while let Ok(Some(pkt)) = KcpPacket::decode(&mut self.buffer) {
	    debug!(
	        "Decoded packet, cmd: {}, sn: {}, frg: {}",
	        pkt.command(),
	        pkt.sn(),
	        pkt.frg()
	    );
	    self._input(&pkt)?;
	    pkts.push(pkt);
	}
	Ok(pkts)
}
```

However, if we receive malformed data, instead of returning error back the caller and terminating the processing, the driver will happily continue with `Ok(Vec::new())` on all subsequent calls and will never advance its buffer (and continue filling it up with new, unprocessable data!). For each instance of `KcpPacket::decode` we should check for error and continue processing until we hit a `None`
- might be my misunderstanding of KCP, but should new `Segment` have `ts` of 0 or should they use the current value within the `KCPSession`?
- I think we're putting acks onto the send queue too eagerly without making all the checks
- `parse_data` - on duplicates immediately return, meanwhile in C they still continue with moving items from the buffer onto the queue
- some other minor nits regarding dependencies used: ideally all versions of the deps should be specified in the workspace and then be referred to via `workspace = true`. also, if possible, we shouldn't be including `log` and instead use `tracing` for consistency with other non-legacy crates

common/nym-lp/src/packet.rs
```diff
@@ -0,0 +1,197 @@
...
+        write!(f, "{}", format_debug_bytes(&self.debug_bytes())?)
+    }
+}
+
+impl LpPacket {
+    pub fn new(header: LpHeader, message: LpMessage) -> Self {
+        Self {
+            header,
+            message,
+            trailer: [0; TRAILER_LEN],
```
jstuczyn, line 46: I can't seem to find any code path that sets it to anything else but 0 bytes. Aren't they meant to be used as an authentication tag?

common/nym-lp/src/replay/validator.rs
```diff
@@ -0,0 +1,879 @@
...
+            Some(Ok(()))
+        } else if too_far_back {
+            Some(Err(ReplayError::OutOfWindow))
+        } else if duplicate {
+            Some(Err(ReplayError::DuplicateCounter))
+        } else {
+            Some(Ok(()))
+        };
+
+        // Unwrap the option (always Some)
```
jstuczyn, line 128: then why is it even wrapped in an option? makes no sense

common/nym-lp/src/replay/validator.rs
```diff
@@ -0,0 +1,879 @@
...
+        let too_far_back = if counter > u64::MAX - (N_BITS as u64) {
+            // If adding N_BITS would overflow, it can't be too far back
+            false
+        } else {
+            counter + (N_BITS as u64) < self.next
+        };
+
+        let duplicate = self.check_bit_branchless(counter);
+
+        // Using Option to avoid early returns
```
jstuczyn, line 117: why?

common/nym-lp/src/message.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+impl LpMessage {
+    pub fn payload(&self) -> &[u8] {
+        match self {
+            LpMessage::Busy => &[],
+            LpMessage::Handshake(payload) => payload.0.as_slice(),
+            LpMessage::EncryptedData(payload) => payload.0.as_slice(),
+            LpMessage::ClientHello(_) => unimplemented!(), // Structured data, serialized in encode_content
```
jstuczyn, line 130: given ClientHelloData is defined, this shouldn't be difficult to include, no?

common/nym-lp/src/message.rs
```diff
@@ -0,0 +1,277 @@
...
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+use bytes::{BufMut, BytesMut};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use serde::{Deserialize, Serialize};
+
+/// Data structure for the ClientHello message
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ClientHelloData {
+    /// Client's LP x25519 public key (32 bytes) - derived from Ed25519 key
```
jstuczyn, line 12: if you want to be returning a nice slice in `LpMessage::payload` you might have to unfortunately get rid if the explicit structure in favour of a single [u8; 96] and then just return slices to corresponding bits

common/nym-lp/src/message.rs
```diff
@@ -0,0 +1,277 @@
...
+        client_lp_public_key: [u8; 32],
+        client_ed25519_public_key: [u8; 32],
+    ) -> Self {
+        use std::time::{SystemTime, UNIX_EPOCH};
+
+        // Generate salt: timestamp + nonce
+        let mut salt = [0u8; 32];
+
+        // First 8 bytes: current timestamp as u64 little-endian
+        let timestamp = SystemTime::now()
```
jstuczyn, line 38: didn't we want to make it full sans-io, i.e. shouldn't the time be injected externally?

common/nym-lp/src/message.rs
```diff
@@ -0,0 +1,277 @@
...
+            LpMessage::KKTResponse(payload) => payload.0.is_empty(),
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        match self {
+            LpMessage::Busy => 0,
+            LpMessage::Handshake(payload) => payload.0.len(),
+            LpMessage::EncryptedData(payload) => payload.0.len(),
+            LpMessage::ClientHello(_) => 97, // 32 bytes x25519 key + 32 bytes ed25519 key + 32 bytes salt + 1 byte bincode overhead
```
jstuczyn, line 152: do we even have to bother with bincode given all the data lenghts are static and known ahead of time, i.e. can't we just concatenate it all? we might even **have to** if we want to provide the implementation for `payload(&self)`

common/nym-lp/src/packet.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /// Compute a hash of the message payload
+    ///
+    /// This can be used for message integrity verification or deduplication
+    pub fn hash_payload(&self) -> [u8; 32] {
+        use sha2::{Digest, Sha256};
+
+        let mut hasher = Sha256::new();
+        let mut buffer = BytesMut::new();
```
jstuczyn, line 57: that buffer is completely redundant. you can just update `hasher` twice with the relevant content, i.e.

```rust
        hasher.update(&(self.message.typ() as u16).to_le_bytes());
        hasher.update(self.payload());
        hasher.finalize().into()
```

common/nym-lp/src/replay/validator.rs
```diff
@@ -0,0 +1,879 @@
...
+    /// - `Ok(())` if the counter is acceptable
+    /// - `Err(ReplayError::InvalidCounter)` if the counter is invalid (too far back)
+    /// - `Err(ReplayError::DuplicateCounter)` if the counter has already been received
+    #[inline(always)]
+    pub fn will_accept_branchless(&self, counter: u64) -> ReplayResult<()> {
+        // Calculate conditions
+        let is_growing = counter >= self.next;
+
+        // Handle potential overflow when adding N_BITS to counter
+        let too_far_back = if counter > u64::MAX - (N_BITS as u64) {
```
jstuczyn, line 108: well, that's not very branchless, is it? : D

common/nym-lp/src/packet.rs
```diff
@@ -0,0 +1,197 @@
...
+        guard.mark_did_receive_branchless(self.header.counter)?;
+        Ok(())
+    }
+}
+
+// VERSION [1B] || RESERVED [3B] || SENDER_INDEX [4B] || COUNTER [8B]
+#[derive(Debug, Clone)]
+pub struct LpHeader {
+    pub protocol_version: u8,
+    pub reserved: u16,
```
jstuczyn, line 129: given reserved is 3 bytes, putting u16 there can be quite deceiving and could potentially backfire

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+use std::str::FromStr;
+
+use nym_sphinx::{PrivateKey as SphinxPrivateKey, PublicKey as SphinxPublicKey};
+use serde::Serialize;
+use utoipa::ToSchema;
+
+use crate::LpError;
+
+#[derive(Clone)]
+pub struct PrivateKey(SphinxPrivateKey);
```
jstuczyn, line 12: do we want to be piggy backing on the Sphinx key here? why should LP be concerned with sphinx' key rotation for example?

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl Deref for PrivateKey {
+    type Target = SphinxPrivateKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Default for PrivateKey {
```
jstuczyn, line 28: I'd argue this `Default` doesn't make sense

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+        let private_key = SphinxPrivateKey::random();
+        Self(private_key)
+    }
+
+    pub fn to_base58_string(&self) -> String {
+        bs58::encode(self.0.to_bytes()).into_string()
+    }
+
+    pub fn from_base58_string(s: &str) -> Result<Self, LpError> {
+        let bytes: [u8; 32] = bs58::decode(s).into_vec()?.try_into().unwrap();
```
jstuczyn, line 45: what if that conversion fails (i.e. somebody provided too short or too long input)? it will blow up

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl PublicKey {
+    pub fn to_base58_string(&self) -> String {
+        bs58::encode(self.0.as_bytes()).into_string()
+    }
+
+    pub fn from_base58_string(s: &str) -> Result<Self, LpError> {
+        let bytes: [u8; 32] = bs58::decode(s).into_vec()?.try_into().unwrap();
```
jstuczyn, line 84: same issue as with the PrivateKey - this will blow up on malicious inputs

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        self.0.as_bytes()
+    }
+}
+
+impl Default for PublicKey {
+    fn default() -> Self {
+        let private_key = PrivateKey::default();
```
jstuczyn, line 99: ok, that's kinda pointless. you're making a random public key that can realistically be never used since you immediately get rid of the associated private key

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    pub fn public_key(&self) -> &PublicKey {
+        &self.public_key
+    }
+}
+
+impl From<KeypairReadable> for Keypair {
+    fn from(keypair: KeypairReadable) -> Self {
+        Self {
+            private_key: PrivateKey::from_base58_string(&keypair.private).unwrap(),
```
jstuczyn, line 152: can we guarantee this won't fail?

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+    type Err = LpError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        PrivateKey::from_base58_string(s)
+    }
+}
+
+impl Display for PrivateKey {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.to_base58_string())
```
jstuczyn, line 176: so we're redacting the content in `Debug` impl but leaving it in `Display`? shouldn't we redact it in both cases?

common/nym-lp/src/keypair.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl Display for PublicKey {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.to_base58_string())
+    }
+}
+
+#[derive(Serialize, serde::Deserialize, Clone, ToSchema, Debug)]
+pub struct KeypairReadable {
+    private: String,
```
jstuczyn, line 188: private key should get zeroized on drop. either wrap it in `Zeroizing` or derive `ZeroizeOnDrop` trait

common/nym-lp/src/codec.rs
```diff
@@ -0,0 +1,575 @@
...
+        buf_partial_trailer.extend_from_slice(&42u32.to_le_bytes()); // Sender index
+        buf_partial_trailer.extend_from_slice(&123u64.to_le_bytes()); // Counter
+        buf_partial_trailer.extend_from_slice(&MessageType::Handshake.to_u16().to_le_bytes()); // Handshake type
+        let payload = vec![42u8; 20]; // Assume 20 byte payload
+        buf_partial_trailer.extend_from_slice(&payload);
+        buf_partial_trailer.extend_from_slice(&[0; TRAILER_LEN - 1]); // Missing last byte of trailer
+
+        // Total length = 16 + 2 + 20 + 15 = 53. Min size = 34. This passes.
+        // Payload size = 53 - 34 = 19. <--- THIS IS WRONG. The parser assumes the length dictates payload.
+        // Let's fix the parser logic slightly.
```
jstuczyn, line 314: well, let's then

common/nym-lp/src/psk.rs
```diff
@@ -0,0 +1,702 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+//! PSK (Pre-Shared Key) derivation for LP sessions using Blake3 KDF.
+//!
+//! This module implements identity-bound PSK derivation where both client and gateway
+//! derive the same PSK from their LP keypairs.
+//!
+//! Two approaches are supported:
+//! - **Legacy ECDH-only** (`derive_psk`) - Simple but no post-quantum security
```
jstuczyn, line 10: it seems no legacy ECDH is actually present in the code?

common/nym-lp/src/psk.rs
```diff
@@ -0,0 +1,702 @@
...
+        PsqX25519::encapsulate_psq(kem_pk, PSQ_SESSION_CONTEXT, &mut rng)
+            .map_err(|e| LpError::Internal(format!("PSQ encapsulation failed: {:?}", e)))?;
+
+    // Step 3: Combine ECDH + PSQ via Blake3 KDF
+    let mut combined = Vec::with_capacity(64 + psq_psk.len());
+    combined.extend_from_slice(ecdh_secret.as_bytes());
+    combined.extend_from_slice(&psq_psk); // psq_psk is [u8; 32], need &
+    combined.extend_from_slice(salt);
+
+    let final_psk = nym_crypto::kdf::derive_key_blake3(PSK_PSQ_CONTEXT, &combined, &[]);
```
jstuczyn, line 127: it feels a bit weird how you're putting salt in the combined content and then explicitly providing an empty salt to the kdf

common/nym-lp/src/psk.rs
```diff
@@ -0,0 +1,702 @@
...
+
+    // Serialize ciphertext using TLS encoding for transport
+    let ct_bytes = ciphertext
+        .tls_serialize_detached()
```
jstuczyn, line 131: why are we using TLS encoding here? seems like an extra unnecessary dependency. couldn't we just use ciphertext bytes directly instead?

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+/// Generates a fresh salt for PSK derivation.
+///
+/// Salt format: 8 bytes timestamp (u64 LE) + 24 bytes random nonce
+///
+/// This ensures each session derives a unique PSK, even with the same key pairs.
+/// The timestamp provides temporal uniqueness while the random nonce prevents collisions.
+///
+/// # Returns
+/// A 32-byte array containing fresh salt material
+pub fn generate_fresh_salt() -> [u8; 32] {
```
jstuczyn, line 179: this seems to use exactly  the same logic as `ClientHelloData::new_with_fresh_salt` => should those be combined? also the same comment about time injection holds

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+    psq_state: Mutex<PSQState>,
+
+    /// PSK handle from responder (ctxt_B) for future re-registration
+    psk_handle: Mutex<Option<Vec<u8>>>,
+
+    /// Counter for outgoing packets
+    sending_counter: AtomicU64,
+
+    /// Validator for incoming packet counters to prevent replay attacks
+    receiving_counter: Mutex<ReceivingKeyCounterValidator>,
```
jstuczyn, line 144: I haven't looked exhaustively through all possible codepath, but all mutating data access to the fields behind mutexes are happening via `LpStateMachine::process_input`. and that one does use `&mut self`, thus do we even need the overhead of mutexes at all? they don't seem to be providing any value

common/nym-lp/src/noise_protocol.rs
```diff
@@ -0,0 +1,327 @@
...
+                    .set_psk(index as usize, psk)
+                    .map_err(NoiseError::ProtocolError)?;
+                Ok(())
+            }
+            _ => Err(NoiseError::IncorrectStateError),
+        }
+    }
+}
+
+pub fn create_noise_state(
```
jstuczyn, line 289: this, alongside `create_noise_state_responder` doesn't appear to be used anywhere, but looking at the content, I assume, it was used in some test code somewhere?

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+    ) -> Result<Self, LpError> {
+        // XKpsk3 pattern requires remote static key known upfront (XK)
+        // and PSK mixed at position 3. This provides forward secrecy with PSK authentication.
+        let pattern_name = "Noise_XKpsk3_25519_ChaChaPoly_SHA256";
+        let psk_index = 3;
+
+        let params = pattern_name.parse()?;
+        let builder = Builder::new(params);
+
+        let local_key_bytes = local_x25519_key.to_bytes();
```
jstuczyn, line 251: private key bytes should get zeroized (just wrap it with `Zeroizing::new()`)

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+        let builder = builder.local_private_key(&local_key_bytes);
+
+        let remote_key_bytes = remote_x25519_key.to_bytes();
+        let builder = builder.remote_public_key(&remote_key_bytes);
+
+        // Initialize with dummy PSK - real PSK will be injected via set_psk() during handshake
+        // when PSQ runs using X25519 as DHKEM
+        let dummy_psk = [0u8; 32];
+        let builder = builder.psk(psk_index, &dummy_psk);
```
jstuczyn, line 260: do we have to inject dummy non-sense data? looking at snow's builder, it woudn't error out of no psk being provided

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+            is_initiator,
+            noise_state: Mutex::new(noise_protocol),
+            kkt_state: Mutex::new(kkt_state),
+            psq_state: Mutex::new(psq_state),
+            psk_handle: Mutex::new(None),
+            sending_counter: AtomicU64::new(0),
+            receiving_counter: Mutex::new(ReceivingKeyCounterValidator::default()),
+            psk_injected: AtomicBool::new(false),
+            // Ed25519 keys don't impl Clone, so convert to bytes and reconstruct
+            local_ed25519_private: ed25519::PrivateKey::from_bytes(
```
jstuczyn, line 291: ugh. yeah. is there no way to instead wrap the keys with an `Arc` and avoid those hacky copies?

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+    }
+
+    /// Returns current packet statistics for monitoring.
+    ///
+    /// # Returns
+    ///
+    /// A tuple containing:
+    /// * The next expected counter value for incoming packets
+    /// * The total number of received packets
+    pub fn current_packet_cnt(&self) -> (u64, u64) {
```
jstuczyn, line 368: nit: I think it'd be more readable (and less error prone) if that tuple was wrapped in a struct. it's non-obvious and easy to confuse `.0` and `.1` values

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+    /// * `Ok(LpMessage::KKTResponse)` - Signed KKT response ready to send
+    /// * `Err(LpError)` - Signature verification failed or invalid request
+    pub fn process_kkt_request(
+        &self,
+        request_bytes: &[u8],
+        responder_kem_pk: &EncapsulationKey,
+    ) -> Result<LpMessage, LpError> {
+        use nym_kkt::{frame::KKTFrame, kkt::handle_kem_request};
+
+        let mut kkt_state = self.kkt_state.lock();
```
jstuczyn, line 553: shouldn't we also ensure here that we're in the correct initial state so, for example, we wouldn't overwrite a completed state?

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+    /// For initiators, PSQ always runs on the first message:
+    /// 1. Converts X25519 keys to DHKEM format
+    /// 2. Generates PSQ payload and derives PSK
+    /// 3. Injects PSK into Noise HandshakeState
+    /// 4. Embeds PSQ payload in first handshake message as: [u16 len][psq_payload][noise_msg]
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(None)` if no message needs to be sent currently (e.g., waiting for peer, or handshake complete).
+    /// * `Err(LpError)` if there's an error within the Noise protocol or PSQ.
```
jstuczyn, line 595: isn't this the other way around given the actual response type is `Option<Result<..>>` as opposed to `Result<Option<..>>`?

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+
+                    let psq_len = u16::from_le_bytes([payload[0], payload[1]]) as usize;
+
+                    if payload.len() < 2 + psq_len {
+                        return Err(LpError::NoiseError(NoiseError::Other(
+                            "Payload length mismatch for PSQ extraction".to_string(),
+                        )));
+                    }
+
+                    let psq_payload = &payload[2..2 + psq_len];
```
jstuczyn, line 743: I think there should be a separate type, or at least set of function for serialising and deserialising initial PSQ handshake message so it'd be
1) easier to test
2) easier to read
3) less error prone

common/nym-lp/src/session.rs
```diff
@@ -0,0 +1,1898 @@
...
+                        libcrux_kem::Algorithm::X25519,
+                        local_public_bytes,
+                    )
+                    .map_err(|e| {
+                        LpError::KKTError(format!(
+                            "Failed to convert X25519 public key to libcrux PublicKey: {:?}",
+                            e
+                        ))
+                    })?;
+                    let enc_key = EncapsulationKey::X25519(libcrux_public_key);
```
jstuczyn, line 772: given this is happening in each lp session, can't we just provide encapsulated keys straightaway in the constructor (and ideally Arc'd) to prevent this constant conversion?

common/nym-lp/src/state_machine.rs
```diff
@@ -0,0 +1,1045 @@
...
+    }
+    /// Processes an input event and returns a list of actions to perform.
+    pub fn process_input(&mut self, input: LpInput) -> Option<Result<LpAction, LpError>> {
+        // 1. Replace current state with a placeholder, taking ownership of the real current state.
+        let current_state = mem::take(&mut self.state);
+
+        let mut result_action: Option<Result<LpAction, LpError>> = None;
+
+        // 2. Match on the owned current_state. Each arm calculates and returns the NEXT state.
+        let next_state = match (current_state, input) {
```
jstuczyn, line 211: I think it'd help readibility if we split that match given it's over 300 lines long : D

common/nym-lp/src/state_machine.rs
```diff
@@ -0,0 +1,1045 @@
...
+                // Check if packet lp_id matches our session
+                if packet.header.session_id() != session.id() {
+                    result_action = Some(Err(LpError::UnknownSessionId(packet.header.session_id())));
+                    // Don't change state, return the original state variant
+                    LpState::Handshaking { session }
+                } else {
+                    // --- Inline handle_handshake_packet logic ---
+                    // 1. Check replay protection *before* processing
+                    if let Err(e) = session.receiving_counter_quick_check(packet.header.counter) {
+                         let _reason = e.to_string();
```
jstuczyn, line 367: what's with this? does string conversion of the error have any side effects?

jstuczyn (commented):
another set of comments, this time regarding `nym-lp`:
- Lp Packet trailer: it's meant to be authentication tag, but it seems it's always set to 0?
- Lp Packet Client Hello: contains unimplemented!() macro placeholders; uses internal system clock which goes against the sans-io idea
- should LP keypair be using sphinx' keys? like why should it care about sphinx key rotation?
- LP key recovery might blow up
- there's no support for no legacy PSK, just PSQ, even though comments would suggest otherwise
- private keys are a bit mistreated in `LpSession` and converted back and forth between bytes to clone them - do they need to be cloned?
- no initial state validation when processing KKT request (`process_kkt_request`)
- other bits and pieces as pointed out by specific code comments

common/registration/src/lp_messages.rs
```diff
@@ -0,0 +1,270 @@
...
+pub struct LpRegistrationResponse {
+    /// Whether registration succeeded
+    pub success: bool,
+
+    /// Error message if registration failed
+    pub error: Option<String>,
+
+    /// Gateway configuration data (same as returned by authenticator)
+    /// This matches what WireguardRegistrationResult expects
+    pub gateway_data: Option<GatewayData>,
```
jstuczyn, line 59: would it make sense to maybe combine those fields into a single enum to remove possibility (and required matches) of denegerate cases where both are set (or not set)?

docs/LP_DEPLOYMENT.md
```diff
@@ -0,0 +1,845 @@
...
+chmod 600 ~/.nym/gateways/<id>/keys/lp_x25519.pem
+
+# Backup keys securely (encrypted)
+gpg -c ~/.nym/gateways/<id>/keys/lp_x25519.pem
+# Store lp_x25519.pem.gpg in secure location
+```
+
+### 4. Database Configuration
+
+LP requires PostgreSQL for credential tracking:
```
jstuczyn, line 124: huh, since when?

docs/LP_DEPLOYMENT.md
```diff
@@ -0,0 +1,845 @@
...
+```
+
+**Database Maintenance:**
+
+```sql
+-- Index for nullifier lookups (critical for performance)
+CREATE INDEX idx_nullifiers ON spent_credentials(nullifier);
+
+-- Periodic cleanup of old nullifiers (run daily via cron)
+DELETE FROM spent_credentials WHERE expiry < NOW() - INTERVAL '30 days';
```
jstuczyn, line 146: you really don't want to be manually removing spent credentials

common/wireguard/src/ip_pool.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// # Errors
+    /// Returns `IpPoolError::NoFreeIp` if no IPs are available
+    pub async fn allocate(&self) -> Result<IpPair, IpPoolError> {
+        let mut pool = self.allocations.write().await;
+
+        // Find a free IP and allocate it
+        let free_ip = pool
+            .iter_mut()
+            .filter(|(_, state)| matches!(state, AllocationState::Free))
+            .choose(&mut rand::thread_rng())
```
jstuczyn, line 112: wouldn't this form of allocation be quite costly as we'd have to iterate the whole map? why not
1) have two structures, one for allocated and one for free IpPair
2) allocate addresses sequentially from the free pool

common/wireguard/src/peer_controller.rs
```diff
@@ -393,6 +487,14 @@ impl PeerController {
 
                     *self.host_information.write().await = host;
                 }
+                _ = self.ip_cleanup_interval.next() => {
+                    // Periodically cleanup stale IP allocations
+                    let freed = self.ip_pool.cleanup_stale(DEFAULT_IP_STALE_AGE).await;
```
jstuczyn, line 492: that seems quite drastic given the staleness is set to 1h. what if there's still a connected client using the addresses and then the pair is re-allocated?

common/wireguard/src/lib.rs
```diff
@@ -247,9 +269,38 @@ pub async fn start_wireguard(
...
+    // Mark existing peer IPs as used in the pool
+    for peer in &peers {
+        for allowed_ip in &peer.allowed_ips {
+            // Extract IPv4 and IPv6 from peer's allowed_ips
+            if let IpAddr::V4(ipv4) = allowed_ip.ip {
+                // Find corresponding IPv6
+                if let Some(ipv6_mask) = peer
+                    .allowed_ips
+                    .iter()
+                    .find(|ip| matches!(ip.ip, IpAddr::V6(_)))
```
jstuczyn, line 290: don't we also have to mark them even if they don't have an ipv6?

nym-gateway-probe/src/bandwidth_helpers.rs
```diff
@@ -155,3 +157,93 @@ pub(crate) async fn acquire_bandwidth(
...
+///
+/// Gateway with --lp-use-mock-ecash accepts any credential without verification,
+/// so we only need to provide properly structured data with correct types.
+///
+/// This is useful for local testing without requiring blockchain access or funded accounts.
+///
+/// This uses a pre-serialized test credential from the wireguard tests - since MockEcashManager
+/// doesn't verify anything, any valid CredentialSpendingData structure will work.
+#[allow(clippy::expect_used)] // Test helper with hardcoded valid data
+pub(crate) fn create_dummy_credential(
```
jstuczyn, line 171: can we mark it as only available in test code?

nym-node/src/cli/helpers.rs
```diff
@@ -479,6 +508,12 @@ impl EntryGatewayArgs {
         if let Some(upgrade_mode_attester_public_key) = self.upgrade_mode_attester_public_key {
             section.upgrade_mode.attester_public_key = upgrade_mode_attester_public_key
         }
+        if let Some(enable_lp) = self.enable_lp {
```
jstuczyn, line 511: we'd probably want to put it in config template.rs

nym-node/src/config/gateway_tasks.rs
```diff
@@ -46,6 +46,9 @@ pub struct GatewayTasksConfig {
 
     pub upgrade_mode: UpgradeModeWatcher,
 
+    #[serde(default)]
+    pub lp: nym_gateway::node::LpConfig,
```
jstuczyn, line 50: imo the LP config should be defined in this package, because once we have to change it, backwards compatibility and migrations will be quite annoying and cumbersome

gateway/src/node/lp_listener/handler.rs
```diff
@@ -0,0 +1,996 @@
...
+        // Send the packet
+        self.send_lp_packet(&packet).await
+    }
+
+    /// Receive an LP packet from the stream with proper length-prefixed framing
+    async fn receive_lp_packet(&mut self) -> Result<LpPacket, GatewayError> {
+        use nym_lp::codec::parse_lp_packet;
+
+        // Read 4-byte length prefix (u32 big-endian)
+        let mut len_buf = [0u8; 4];
```
jstuczyn, line 371: tihs length prefixing looks like something that should have been handled by a codec and not dealt with manually

gateway/src/node/lp_listener/handshake.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        debug!(
+            "Sent LP packet ({} bytes + 4 byte header)",
+            packet_buf.len()
+        );
+        Ok(())
+    }
+
+    /// Receive an LP packet from the stream with proper length-prefixed framing
+    async fn receive_packet(&self, stream: &mut TcpStream) -> Result<LpPacket, GatewayError> {
```
jstuczyn, line 146: why are we duplicating packet serialisation/deserialisation (from `handler.rs`)? Can't we wrap `TcpStream` with a codec?

gateway/src/node/lp_listener/handshake.rs
```diff
@@ -0,0 +1,178 @@
...
+                        e
+                    )));
+                }
+            }
+        }
+
+        // Continue handshake until complete
+        loop {
+            // Read incoming packet
+            let packet = self.receive_packet(stream).await?;
```
jstuczyn, line 74: this is one instance of many, but we're not checking for shutdown - what if we end up waiting until the timeout and block the whole process?

gateway/src/node/lp_listener/registration.rs
```diff
@@ -0,0 +1,420 @@
...
+            true,
+        ),
+    );
+
+    // Track credential verification attempts
+    inc!("lp_credential_verification_attempts");
+
+    // For mock ecash mode (local testing), skip cryptographic verification
+    // and just return a dummy bandwidth value since we don't have blockchain access
+    let allocated = if ecash_verifier.is_mock() {
```
jstuczyn, line 106: I think the explicit `is_mock()` is a bit redundant. couldn't just the `MockVerifier` immediately return the mock amount in its `verify()` method?

gateway/src/node/lp_listener/registration.rs
```diff
@@ -0,0 +1,420 @@
...
+    state: &LpHandlerState,
+) -> LpRegistrationResponse {
+    let session_id = rand::random::<u32>();
+    let registration_start = std::time::Instant::now();
+
+    // Track total registration attempts
+    inc!("lp_registration_attempts_total");
+
+    // 1. Validate timestamp for replay protection
+    if !request.validate_timestamp(30) {
```
jstuczyn, line 142: wasn't the timestamp skew set in the config? shouldn't we use that value instead?

jstuczyn (commented):
penultimate comments:
- `IpPool` allocation seems quite inefficient - essentially we have to iterate through the entire pool (both free and allocated entries) for every iteration
- `IpPair` addresses are automatically released after 1h, even if in use!
- Peer addresses might not get marked as used if they don't have a valid ipv6 address
- LpPacket wire serialisation/deserialisation is duplicated in multiple places (say `handshake.rs` and `handler.rs`)
- we're not listening for shutdown anyway within the LP handling -> we could very easily just block there

nym-gateway-probe/src/nodes.rs
```diff
@@ -118,16 +129,189 @@ impl DirectoryNode {
...
+                    wireguard,
+                    mixnet_websockets: WebSockets {
+                        ws_port: websockets.ws_port,
+                        wss_port: websockets.wss_port,
+                    },
+                };
+
+                // Create NymNodeDescription
+                let described = NymNodeDescription {
+                    node_id: 0, // We don't have a node_id from direct query
```
jstuczyn, line 294: which depending on what you're going to use it for might be a problem (like if you converted it to a 'RoutingNode')

nym-registration-client/src/builder/mod.rs
```diff
@@ -46,7 +50,7 @@ impl RegistrationClientBuilder {
             let builder = MixnetClientBuilder::new_with_storage(mixnet_client_storage)
                 .event_tx(EventSender(event_tx));
             let mixnet_client = tokio::time::timeout(
-                self.config.mixnet_client_startup_timeout,
+                MIXNET_CLIENT_STARTUP_TIMEOUT,
```
jstuczyn, line 53: nit-ish: shouldn't this still be configurable like before? I can imagine different scenarios might require different timeouts

nym-registration-client/src/lib.rs
```diff
@@ -150,34 +121,23 @@ impl RegistrationClient {
...
-        let (entry, exit) = Box::pin(
-            self.cancel_token
-                .run_until_cancelled(async { tokio::join!(entry_fut, exit_fut) }),
-        )
-        .await
-        .ok_or(RegistrationError {
-            mixnet_client: None,
-            source: RegistrationClientError::Cancelled,
-        })?;
+        let (entry, exit) = Box::pin(async { tokio::join!(entry_fut, exit_fut) }).await;
```
jstuczyn, line 124: you seem to have removed checking for shutdown here

nym-registration-client/src/lib.rs
```diff
@@ -192,24 +152,160 @@ impl RegistrationClient {
...
-            Ok(result) => Ok(result),
-            Err(error) => {
-                debug!("Registration failed");
-                if let Some(mixnet_client) = error.mixnet_client {
-                    debug!("Shutting down mixnet client");
-                    mixnet_client.disconnect().await;
-                }
-                Err(error.source)
+        // Register exit gateway via LP
+        let exit_fut = {
```
jstuczyn, line 222: aren't we technically performing direct registration with the exit here as opposed to telescoping via entry? isn't this a bit against the idea of the whole thing?

nym-registration-client/src/lp_client/client.rs
```diff
@@ -0,0 +1,801 @@
...
+        Ok(())
+    }
+
+    /// Sends an LP packet over the TCP stream with length-prefixed framing.
+    ///
+    /// Format: 4-byte big-endian u32 length + packet bytes
+    ///
+    /// # Errors
+    /// Returns an error if serialization or network transmission fails.
+    async fn send_packet(stream: &mut TcpStream, packet: &LpPacket) -> Result<()> {
```
jstuczyn, line 353: this has been duplicated in so many places before, if it was just once or twice, fine, but I think I've seen the same logic like 5 times already - this should really be exported elsewhere and preferably wrapped in a Codec or something

nym-registration-client/src/lp_client/client.rs
```diff
@@ -0,0 +1,801 @@
...
+    /// Establishes TCP connection to the gateway's LP listener.
+    ///
+    /// This must be called before attempting handshake or registration.
+    ///
+    /// # Errors
+    /// Returns `LpClientError::TcpConnection` if the connection fails or times out.
+    ///
+    /// # Implementation Note
+    /// This is implemented in nym-78. The handshake (nym-79) and registration
+    /// (nym-80, nym-81) will be added in subsequent tasks.
```
jstuczyn, line 130: is there any reference to those tasks anywhere? they don't appear to correspond to jira ticket or anything like that

jstuczyn (commented):
- when registering with the exit gateway (LP), we seem to connect to that node directly - doesn't that go against the whole idea of 2-hop? This comment, however, might be completely redundant because I can assume the telescoping PR changes this behaviour
- `LpTranport` could easily be wrapped into `Framed<TcpStream, LpCodec>`, because that's essentially what `send_packet` and `receive_packet` are doing. (the same is true for the `LpRegistrationClient`) plus I already mentioned duplication of this very functionality in different places.

common/nym-lp/src/packet.rs
durch: Yeah, we did that in one of the follow ups

common/nym-lp/src/replay/validator.rs
durch: This is ancient, at this point I honestly have no idea, maybe made some sense at the time :/

common/nym-lp/src/session.rs
durch: Documentation rot

common/nym-lp/src/state_machine.rs
durch: This was probably some sort of debugging exercise

docs/LP_DEPLOYMENT.md
durch: no, these are mostly notes on things, I think this is all sorted

gateway/src/node/lp_listener/registration.rs
durch: config probably came later, nice catch

nym-gateway-probe/src/nodes.rs
durch: Thats addressed in the telescoping branch, this was an early version

nym-registration-client/src/lib.rs
durch: Might be a messed up rebase

nym-registration-client/src/lp_client/client.rs
durch: Yeah, this got sorted eventually

nym-registration-client/src/lp_client/client.rs
durch: No, I have an internal thing where I toss things, these docs reference that, it all can be cleaned up

common/nym-kcp/src/session.rs
durch: Yeah, thats what that does, in case it got stuff out of order it will break, and wait for the next one. The fragment will remain, and wait its turn

common/nym-kcp/src/session.rs
durch: Thats interesting, if i remember correctly in the mixnet its too noisy to use the defaults, but maybe this is not the right fix

common/nym-kcp/src/session.rs
durch: That was supposed to come when we implement backpressure, we just never got to it

common/nym-lp/src/message.rs
durch: Sorted it in a less bad way later on, ie just returning &[]. Tension here is that early message types were just Vecs, but then I added these structured ones, and now that whole thing is neither here nor there

common/nym-lp/src/message.rs
durch: This is/was a special case, since the ClientHello has very specific purpose, and  is used only in the registration flow but good point nevertheless, I think I mixed concearns there.

gateway/src/node/lp_listener/handler.rs
durch: The issue is that codec is designed for datagram parsing, and not stream parsing, so this is a sort of adapter where we convert the stream to what the codec wants, and the use it

gateway/src/node/lp_listener/handshake.rs
durch: Looking at it again, this is debt from moving from TcpStreams to one Tcp connection per packet (to remain sans-io-ish), it can be consolidated nicely

durch: And after more careful looking, I'd just leave it, if we want a generalized helper, and if we want it to live in nym-lp where it should be, then we need to drag tokio in there, all in all, it seems simpler to just special case it, I'm also a bit uncomfortable with 4 places where we duplicate this, but it seems simpler to just leave it

jstuczyn:

PR #6143, opened 2025-10-23
Bugfix: Add circuit breaker

tommyv1987:
When the mixnet client's mix_tx channel closes during network drops, `OutQueueControl`
would retry sending packets through the closed channel, flooding logs and 
hanging the daemon. This affects all clients (VPN, SOCKS5, native clients), not just VPN.

Solution

Cancelling the root token from the MixTrafficController

Additionally, reduced `MAX_FAILURE_COUNT` from 100 → 20 to detect dead gateways faster 
(~1-2s instead of ~6s), improving reconnection speed during mobile network drops and 
sleep/wake cycles.

Example Logs (Before Fix)
```
2025-10-23T14:18:26.772232Z ERROR nym_client_core::client::mix_traffic: Failed to send sphinx packet to the gateway 100 times in a row - assuming the gateway is dead
2025-10-23T14:18:26.772239Z DEBUG nym_client_core::client::mix_traffic: MixTrafficController: Exiting
2025-10-23T14:18:26.773309Z ERROR nym_client_core::client::real_messages_control::real_traffic_stream: failed to send mixnet packet due to closed channel (outside of shutdown!)
2025-10-23T14:18:26.773382Z ERROR nym_client_core::client::real_messages_control::real_traffic_stream: failed to send mixnet packet due to closed channel (outside of shutdown!)
2025-10-23T14:18:26.774273Z ERROR nym_client_core::client::real_messages_control::real_traffic_stream: failed to send mixnet packet due to closed channel (outside of shutdown!)
... (continues infinitely)
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)

### DIFF
--- a/common/task/src/lib.rs
+++ b/common/task/src/lib.rs
@@ -24,6 +24,6 @@ pub use crate::runtime_registry::RegistryAccessError;
 
 /// Get or create a ShutdownTracker for SDK use.
 /// This provides automatic task management without requiring manual setup.
-pub fn get_sdk_shutdown_tracker() -> Result<ShutdownTracker, RegistryAccessError> {
-    Ok(runtime_registry::RuntimeRegistry::get_or_create_sdk()?.shutdown_tracker_owned())
+pub fn create_sdk_shutdown_tracker() -> Result<ShutdownTracker, RegistryAccessError> {
+    Ok(runtime_registry::RuntimeRegistry::create_sdk()?.shutdown_tracker_owned())
 }
--- a/common/task/src/runtime_registry.rs
+++ b/common/task/src/runtime_registry.rs
@@ -19,30 +19,45 @@ pub(crate) struct RuntimeRegistry {
 pub enum RegistryAccessError {
     #[error("the runtime registry is poisoned")]
     Poisoned,
+
+    #[error("The SDK ShutdownManager already exists")]
+    ExistingShutdownManager,
+
+    #[error("No existing SDK ShutdownManager")]
+    MissingShutdownManager,
 }
 
 impl RuntimeRegistry {
-    /// Get or create a ShutdownManager for SDK use.
+    /// Create a ShutdownManager for SDK use.
     /// This manager doesn't listen to OS signals, making it suitable for library use.
-    pub(crate) fn get_or_create_sdk() -> Result<Arc<ShutdownManager>, RegistryAccessError> {
+    /// This function overwrite any existing manager!
+    pub(crate) fn create_sdk() -> Result<Arc<ShutdownManager>, RegistryAccessError> {
+        let mut guard = REGISTRY
+            .sdk_manager
+            .write()
+            .map_err(|_| RegistryAccessError::Poisoned)?;
+
+        Ok(guard
+            .insert(Arc::new(
+                ShutdownManager::new_without_signals().with_cancel_on_panic(),
+            ))
+            .clone())
+    }
+
+    /// Get the  ShutdownManager for SDK use.
+    /// This manager doesn't listen to OS signals, making it suitable for library use.
+    /// Not yet used, but maybe in the future
+    #[allow(dead_code)]
+    pub(crate) fn get_sdk() -> Result<Arc<ShutdownManager>, RegistryAccessError> {
         let guard = REGISTRY
             .sdk_manager
             .read()
             .map_err(|_| RegistryAccessError::Poisoned)?;
         if let Some(manager) = guard.as_ref() {
-            return Ok(manager.clone());
+            Ok(manager.clone())
+        } else {
+            Err(RegistryAccessError::MissingShutdownManager)
         }
-        drop(guard);
-
-        let mut guard = REGISTRY
-            .sdk_manager
-            .write()
-            .map_err(|_| RegistryAccessError::Poisoned)?;
-        Ok(guard
-            .get_or_insert_with(|| {
-                Arc::new(ShutdownManager::new_without_signals().with_cancel_on_panic())
-            })
-            .clone())
     }
 
     /// Check if an SDK manager has been created.
@@ -85,10 +100,13 @@ mod tests {
 
         assert!(!RuntimeRegistry::has_sdk_manager().unwrap());
 
-        let manager1 = RuntimeRegistry::get_or_create_sdk().unwrap();
+        // Error if nothing was created
+        assert!(RuntimeRegistry::get_sdk().is_err());
+
+        let manager1 = RuntimeRegistry::create_sdk().unwrap();
         assert!(RuntimeRegistry::has_sdk_manager().unwrap());
 
-        let manager2 = RuntimeRegistry::get_or_create_sdk().unwrap();
+        let manager2 = RuntimeRegistry::get_sdk().unwrap();
         // Should return the same instance
         assert!(Arc::ptr_eq(&manager1, &manager2));
 
--- a/sdk/rust/nym-sdk/src/mixnet/client.rs
+++ b/sdk/rust/nym-sdk/src/mixnet/client.rs
@@ -736,15 +736,11 @@ where
             base_builder = base_builder.with_topology_provider(topology_provider);
         }
 
-        // Use custom shutdown if provided, otherwise get from registry
-        let shutdown_tracker = match self.custom_shutdown {
-            Some(custom) => custom,
-            None => {
-                // Auto-create from registry for SDK use
-                nym_task::get_sdk_shutdown_tracker()?
-            }
-        };
-        base_builder = base_builder.with_shutdown(shutdown_tracker);
+        // Use custom shutdown if provided, otherwise the sdk one will be used later down the line
+        if let Some(shutdown_tracker) = self.custom_shutdown {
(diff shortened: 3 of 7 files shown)


common/task/src/lib.rs
```diff
@@ -24,6 +24,6 @@ pub use crate::runtime_registry::RegistryAccessError;
 
 /// Get or create a ShutdownTracker for SDK use.
 /// This provides automatic task management without requiring manual setup.
-pub fn get_sdk_shutdown_tracker() -> Result<ShutdownTracker, RegistryAccessError> {
-    Ok(runtime_registry::RuntimeRegistry::get_or_create_sdk()?.shutdown_tracker_owned())
+pub fn create_sdk_shutdown_tracker() -> Result<ShutdownTracker, RegistryAccessError> {
+    Ok(runtime_registry::RuntimeRegistry::create_sdk()?.shutdown_tracker_owned())
```
jstuczyn, line 28: why did you change the existing behaviour to overwrite a pre-existing shutdown manager? this could be potentially dangereous, especially if it had already registered some signals, tasks, etc

common/task/src/runtime_registry.rs
```diff
@@ -19,32 +19,47 @@ pub(crate) struct RuntimeRegistry {
...
-            .get_or_insert_with(|| {
-                Arc::new(ShutdownManager::new_without_signals().with_cancel_on_panic())
-            })
+            .insert(Arc::new(
+                ShutdownManager::new_without_signals().with_cancel_on_panic(),
+            ))
             .clone())
     }
 
+    /// Get the  ShutdownManager for SDK use.
```
jstuczyn, line 47: 1. if it's not yet used, just remove it
2. don't leak SDK needs into the common task library (it's like exposing VPN-specific methods in the monorepo)

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -736,15 +736,11 @@ where
...
-            None => {
-                // Auto-create from registry for SDK use
-                nym_task::get_sdk_shutdown_tracker()?
-            }
-        };
-        base_builder = base_builder.with_shutdown(shutdown_tracker);
+        // Use custom shutdown if provided, otherwise the sdk one will be used later down the line
+        if let Some(shutdown_tracker) = self.custom_shutdown {
+            base_builder = base_builder.with_shutdown(shutdown_tracker);
+        }
```
jstuczyn, line 742: so why do we no longer get the default static one?

common/task/src/lib.rs
simonwicky: Having a global shutdown means that if you cancel the underlying token, you can never spin up a new mixnet client again, because the underlying token will be cancelled to start with. 

The SDK manager is only used if no other shutdown manager is provided

sdk/rust/nym-sdk/src/mixnet/client.rs
simonwicky: We still need it, but it was set a first time here, and then later in the base client startup too. There is no point in setting it in two places

common/task/src/runtime_registry.rs
simonwicky: 2. I merely modified the existing SDK shutdown manager. It's more of an internal shutdown manager rather than a specific one imo. Either you give your client a custom one, or it creates one for internal use. We could have an entire discussion on the shutdown process which I still think could use a proper look into

sdk/rust/nym-sdk/src/mixnet/client.rs
simonwicky: Reference : https://github.com/nymtech/nym/blob/4f0974fcf19929b1adc07c0dc900a4c9def87676/common/client-core/src/client/base_client/mod.rs#L1006

I'd be fine with setting it up there and remove it here mentioned. Then `BaseClientBuilder.shutdown` should no longer be optional

pronebird (commented):
> @pronebird reviewed 6 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)* status: 6 of 7 files reviewed, 4 unresolved discussions (waiting on @durch and @jstuczyn)

___
*[`common/client-core/src/client/real_messages_control/real_traffic_stream.rs` line 283 at r2](https://reviewable.io/reviews/nymtech/nym/6143#-OcZvmqJCphMkgaqAZXe:-OcZvmqJCphMkgaqAZXf:b-1rdrbz) ([raw file](https://github.com/nymtech/nym/blob/fd5a95fa4d5ae991b18d86aa0e95b4040730582e/common/client-core/src/client/real_messages_control/real_traffic_stream.rs#L283)):*
> ```XML
>         };
> 
>         let sending_res = tokio::select! {
> ```

NIT: This could be written as:

```rs
let sending_res = self.shutdown_token.run_until_cancelled(self.mix_tx.send(vec![next_message])).await;
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @durch, @jstuczyn, and @tommyv1987)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6143#-OcZyPE3Dw0OQS4TLzgg:-OcZyPE3Dw0OQS4TLzgh:b-o53xkl) ([raw file](https://github.com/nymtech/nym/blob/fd5a95fa4d5ae991b18d86aa0e95b4040730582e/common/client-core/src/client/mix_traffic/mod.rs#L166)):*
> ```XML
>                             // Gateway is dead, we have to shut down currently
>                             error!("Signalling shutdown from the MixTrafficController");
>                             self.shutdown_token.cancel();
> ```

Would it not be easier to pass a channel from parent task and bubble error back up? Then let parent cancel everything and exit cleanly.

simonwicky (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @durch, @jstuczyn, @pronebird, and @tommyv1987)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6143#-OcZyPE3Dw0OQS4TLzgg:-Oc_1Efu8KqO7ao99u0h:b-4pcmjk) ([raw file](https://github.com/nymtech/nym/blob/fd5a95fa4d5ae991b18d86aa0e95b4040730582e/common/client-core/src/client/mix_traffic/mod.rs#L166)):*

Would it not be easier to pass a channel from parent task and bubble error back up? Then let parent cancel everything and exit cleanly.
</blockquote></details>

Easier no, much better yes, hence my comment about it. We need to come back on that, that PR is just a bandaid over a bigger issue

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @durch, @jstuczyn, and @tommyv1987)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6143#-OcZyPE3Dw0OQS4TLzgg:-Oc_MlGT0t_Jl_yqK_eu:b-u8wocr) ([raw file](https://github.com/nymtech/nym/blob/fd5a95fa4d5ae991b18d86aa0e95b4040730582e/common/client-core/src/client/mix_traffic/mod.rs#L166)):*
<details><summary><i>Previously, simonwicky (Simon Wicky) wrote…</i></summary><blockquote>

Easier no, much better yes, hence my comment about it. We need to come back on that, that PR is just a bandaid over a bigger issue
</blockquote></details>

If we have a clear parent task, to me it seems logical to eliminate bandaid and just add a simple channel, bubble up the message and then cancel all child tasks from above but not from below.

simonwicky (commented):
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6143)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @durch, @jstuczyn, @pronebird, and @tommyv1987)

___
*[`common/client-core/src/client/mix_traffic/mod.rs` line 166 at r2](https://reviewable.io/reviews/nymtech/nym/6143#-OcZyPE3Dw0OQS4TLzgg:-Oc_ONx45aF_ZGqDCtKx:bpuqqh8) ([raw file](https://github.com/nymtech/nym/blob/fd5a95fa4d5ae991b18d86aa0e95b4040730582e/common/client-core/src/client/mix_traffic/mod.rs#L166)):*

If we have a clear parent task, to me it seems logical to eliminate bandaid and just add a simple channel, bubble up the message and then cancel all child tasks from above but not from below.
</blockquote></details>

Bubbling up messages from everywhere is gonna be real ugly, we need a signalling channel within the custom shutdown mechanism we have, and I don't have the time to do it properly

tommyv1987:
Closing with more work being undertaken here: https://nymtech.atlassian.net/browse/NET-688

PR #6144, opened 2025-10-24
Add weighted scoring to NS API

benedettadavico:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6144)

### DIFF
--- a/nym-node-status-api/nym-node-status-api/src/http/models.rs
+++ b/nym-node-status-api/nym-node-status-api/src/http/models.rs
@@ -387,13 +387,14 @@ fn calculate_mixnet_score(gateway: &Gateway) -> ScoreValue {
     }
 }
 
-/// calculates a visual score for the gateway
+/// calculates a visual score for the gateway using weighted metrics
 fn calculate_score(gateway: &Gateway, probe_outcome: &LastProbeResult) -> ScoreValue {
     let mixnet_performance = gateway.performance as f64 / 100.0;
-    let ping_ips_performance = probe_outcome
+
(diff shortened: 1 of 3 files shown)


nym-node-status-api/nym-node-status-api/src/http/models.rs
```diff
@@ -419,18 +420,19 @@ fn calculate_score(gateway: &Gateway, probe_outcome: &LastProbeResult) -> ScoreV
                 0.1
             };
 
-            // combine the scores
-            file_download_score * ping_ips_performance
+            (file_download_score, ping_ips_performance)
         })
-        .unwrap_or(0f64);
+        .unwrap_or((0.0, 0.0));
```
jstuczyn: I'd say define a helper struct, like 
```rust
struct NodeScore {
  download_speed_score: f64,
  ping_ips_score: f64
}
```

rather than passing an (f64, f64) tuple, because somebody, at some point, will accidentally mix them up

jstuczyn: or maybe you could even include there `mixnet_performance` and have `fn calculate_weighted_score(&self) -> f64` so that it would be slightly more maintainable and easier to adjust the inidividual weights

jstuczyn (approved):

Jehado1991 (commented):
reviews/2025-11-edera-container-runtime-securityreview.pdf

Jehado1991 (commented):
reviews/2025-11-edera-container-runtime-securityreview.pdf

PR #6158, opened 2025-10-30
[bugfix] Disconnect mixnet client if registration fails

simonwicky:
# Description
Currently the registration client does its job and then handles everything to the vpn-client, who then takes care on the mixnet client.

However, if the registration fails, nothing is disconnecting the mixnet client, which can lead to errors of the kind of `There is already an open connection to this client` 

This PR addresses that

## Changes
- The owner of the mixnetClient is responsible for it.
- In 2-hop mode, once the `mixnetClient` is handed over to the `AuthClientMixnetListener`, it becomes its responsibility. In that spirit, it now disconnects it when it stops (before the call was explicit)
- In the registration fails, the `mixnetClient` is handed back to the top `fn` who shuts it down before returning the underlying error
- `IprConnect::new` is no longer `async` because it never needed to be `async` in the first place
- For a better `mixnetClient` management, registration timeouts have been moved further down the call stack, where the `async` stuff actually happens

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6158)


tommyv1987:
LGTM - What Was Tested

1. Mixnet Client Cleanup
Verified clients are disconnected when listener stops
Tested cancellation triggers proper shutdown

2. API Change Validation
Confirmed IprClientConnect::new() works as synchronous (removed async)
Client retrieval via into_mixnet_client() works correctly

3. Error Path Resource Management
All error scenarios properly clean up mixnet clients
Ownership tracking works: knows who owns client at each stage
Cancellation tokens propagate correctly (parent -> child)

Validation
-  Mixnet clients always disconnected when appropriate
- No double-disconnect issues after ownership transfer

tommyv1987 (approved):

PR #6172, opened 2025-10-31
Data Observatory

mmsinclair:
This PR adds the Data Observatory that is:

- chain scraper
- parses Cosmos SDK messages
- parses Cosmwasm messages
- stores data in pgsql

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6172)


nym-data-observatory/.gitignore
```diff
@@ -0,0 +1,2 @@
+0001_metadata.sql
```
jstuczyn: why are we ignoring the sql migrations and why would they even exist outside the dedicated migrations directory?

nym-validator-rewarder/src/config/persistence/paths.rs
```diff
@@ -46,7 +46,10 @@ impl ValidatorRewarderPaths {
 impl Default for ValidatorRewarderPaths {
     fn default() -> Self {
         ValidatorRewarderPaths {
-            nyxd_scraper: default_data_directory().join(DEFAULT_SCRAPER_DB_FILENAME),
+            // validator rewarder uses sqlite
+            nyxd_scraper: (default_data_directory().join(DEFAULT_SCRAPER_DB_FILENAME))
+                .to_string_lossy()
```
jstuczyn: do we want to convert to a lossy string or just blow up if it's not a valid string/path instead?

nym-data-observatory/src/config/data_observatory.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct DataObservatoryConfig {
+    pub webhooks: Vec<WebhookConfig>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WebhookConfig {
+    pub id: String,
+    pub description: Option<String>,
+    pub webhook_url: String,
```
jstuczyn: would it make sense to use more strongly typed `Url` from the get go to avoid parsing (and possibly failing) later?

nym-data-observatory/src/config/template.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+
+// While using normal toml marshalling would have been way simpler with less overhead,
+// I think it's useful to have comments attached to the saved config file to explain behaviour of
+// particular fields.
+// Note: any changes to the template must be reflected in the appropriate structs.
+pub(crate) const CONFIG_TEMPLATE: &str = r#"
+# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
```
jstuczyn, line 11: template is missing `chain_scraper_connection_string` which doesn't have any default value, so deserialisation of saved config will fail

nym-data-observatory/src/config/template.rs
```diff
@@ -0,0 +1,29 @@
...
+[data_observatory_config]
+{{#each data_observatory_config.webhooks }}
+[[webhooks]]
+id={{this.id}}
+description='{{this.description}}'
+webhook_url='{{this.webhook_url}}'
```
jstuczyn, line 17: don't we also need to attach other fields here, i.e. `watch_for_chain_message_types` and `authentication`?

nym-data-observatory/src/db/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+    pool: DbPool,
+}
+
+impl Storage {
+    pub async fn init(connection_url: String) -> Result<Self> {
+        let connect_options = PgConnectOptions::from_str(&connection_url)?;
+
+        let pool = DbPool::connect_with(connect_options)
+            .await
+            .map_err(|err| anyhow!("Failed to connect to {}: {}", &connection_url, err))?;
```
jstuczyn, line 23: not saying to use it here, but a nice trick for converting into `anyhow` errors is to use `.context("...")` - it works for both `Result` and `Option`

nym-data-observatory/src/error.rs
```diff
@@ -0,0 +1,45 @@
+use std::io;
+use std::path::PathBuf;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum NymDataObservatoryError {
+    // #[error("failed to save config file using path '{}'. detailed message: {source}", path.display())]
```
jstuczyn, line 7: if those are unused, can we just 🔥  them?

nym-data-observatory/src/http/api/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    pub(crate) fn with_state(self, state: AppState) -> RouterWithState {
+        RouterWithState {
+            router: self.finalize_routes().with_state(state),
+        }
+    }
+
+    fn finalize_routes(self) -> Router<AppState> {
```
jstuczyn, line 46: just making sure if that's intended: there's no auth required for any routes and it's fine it's all public, right?

nym-data-observatory/src/logging.rs
```diff
@@ -0,0 +1,43 @@
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{EnvFilter, filter::Directive};
+
+pub(crate) fn setup_tracing_logger() {
```
jstuczyn, line 4: any reason we can't use the default global `setup_tracing_logger` / `build_tracing_logger` instead?
I think it would actually make sense to move those filtered crates there by default too. if I'm running, say, nym-node in `debug` I actually want to see some useful debug logs rather than being slammed with `h2` or other `tendermint` stuff

nym-data-observatory/src/modules/wasm.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl WasmModule {
+    pub fn new(connection_pool: DbPool) -> Self {
+        WasmModule { connection_pool }
+    }
+}
+
+#[async_trait]
+impl MsgModule for WasmModule {
```
jstuczyn, line 30: if I could be nitpicky on this fine Friday evening, that's not really a `Wasm` module,  but just `WasmExecute` module : )

nym-data-observatory/src/modules/wasm.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    async fn handle_msg(
+        &mut self,
+        index: usize,
+        msg: &Any,
+        tx: &ParsedTransactionResponse,
+        _storage_tx: &mut dyn NyxdScraperTransaction,
+    ) -> Result<(), ScraperError> {
+        let message = serde_json::to_value(tx.parsed_messages.get(&index)).unwrap_or_default();
```
jstuczyn, line 42: I think rather tuan doing so many `unwrap_or_default()` you should return from this method instead otherwise we might run into some weird issues down the line (like the silent serialisation failures in node status api). while not ideal, maybe return `Ok(())` instead?

nym-data-observatory/src/modules/wasm.rs
```diff
@@ -0,0 +1,117 @@
...
+            .clone()
+            .into_iter()
+            .map(|x| DbCoin {
+                amount: x.amount.to_string(),
+                denom: x.denom.to_string(),
+            })
+            .collect();
+
+        let offset_datetime: OffsetDateTime = tx.block.header.time.into();
+        let executed_at = PrimitiveDateTime::new(offset_datetime.date(), offset_datetime.time());
```
jstuczyn, line 59: why are we mixing `OffsetDateTime` and `PrimitiveDateTime`? can we just stick with one otherwise, I fear, we'll run into weird time inconsistencies

nym-data-observatory/src/modules/wasm.rs
```diff
@@ -0,0 +1,117 @@
...
+        Some(value) => match value.as_object() {
+            Some(map) => map.keys().next().cloned(),
+            None => None,
+        },
+        None => None,
+    }
+}
+
+fn get_wasm_message_type(value: &Value) -> String {
+    get_first_field_name(value.get("msg")).unwrap_or_default()
```
jstuczyn, line 116: oof. uhm. are we guaranteed that the json output of the message has its fields in constant order? can we try to get that field by proper name instead?

nym-data-observatory/src/chain_scraper/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+use std::fs;
+use tracing::{info, warn};
+
+pub(crate) mod webhook;
+
+pub(crate) async fn run_chain_scraper(
+    config: &crate::config::Config,
+    connection_pool: DbPool,
+) -> anyhow::Result<PostgresNyxdScraper> {
+    let websocket_url = std::env::var("NYXD_WS").expect("NYXD_WS not defined");
```
jstuczyn: rather than explicitly trying to read from env here, could we perhaps pass those values ass part of config or cli args (and then let clap handle env for us)?

nym-data-observatory/src/chain_scraper/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let rpc_url = std::env::var("NYXD").expect("NYXD not defined");
+    let websocket_url = reqwest::Url::parse(&websocket_url)?;
+    let rpc_url = reqwest::Url::parse(&rpc_url)?;
+
+    // why are those not part of CLI? : (
```
jstuczyn: hehe, indeed, my past self must have been onto something!

nym-data-observatory/src/chain_scraper/webhook.rs
```diff
@@ -0,0 +1,136 @@
...
+        };
+
+        println!(
+            "->>>>>>>>>>>>>>>>>>>>>>>>> {}",
+            serde_json::to_string(&payload).unwrap()
+        );
+
+        for webhook in self.webhooks.clone() {
+            let payload = payload.clone();
+            tokio::spawn(async move {
```
jstuczyn: rather than spawning every request as we get them, would it maybe make sense to have a `TaskTracker` or `JoinSet` on `WebhookModule` itself to keep track of all of those?
perhaps with some kind of queue/limit so we wouldn't suddenly spawn 1000 webhooks and starve everything else

nym-data-observatory/src/chain_scraper/webhook.rs
```diff
@@ -0,0 +1,136 @@
...
+            transaction_hash: tx.hash.to_string(),
+            message,
+        };
+
+        println!(
+            "->>>>>>>>>>>>>>>>>>>>>>>>> {}",
+            serde_json::to_string(&payload).unwrap()
+        );
+
+        for webhook in self.webhooks.clone() {
```
jstuczyn: note, we're completing ignoring the `watch_for_chain_message_types` in the webhook itself. is that fine? if so, perhaps we should remove that field then?

nym-data-observatory/src/chain_scraper/webhook.rs
```diff
@@ -0,0 +1,136 @@
...
+            .map(|watcher_cfg| Webhook::new(watcher_cfg.clone()))
+            .collect::<anyhow::Result<Vec<_>>>()?;
+        Ok(Self { webhooks })
+    }
+}
+
+#[async_trait]
+impl MsgModule for WebhookModule {
+    fn type_url(&self) -> String {
+        <MsgSend as Msg>::Proto::type_url()
```
jstuczyn: just so we're on the same page, `WebhookModule` will **only** see `MsgSend`, nothing else

nym-data-observatory/src/chain_scraper/webhook.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    async fn handle_msg(
+        &mut self,
+        index: usize,
+        _msg: &Any,
+        tx: &ParsedTransactionResponse,
+        _storage_tx: &mut dyn NyxdScraperTransaction,
+    ) -> Result<(), ScraperError> {
+        let message = serde_json::to_value(tx.parsed_messages.get(&index)).ok();
```
jstuczyn: why are we not using the `msg` instead?

nym-data-observatory/.gitignore
mmsinclair: it copies the shared ones in from the common folder (unless we can use multiple directories for migrations?) have added some comments and notes for this

nym-validator-rewarder/src/config/persistence/paths.rs
mmsinclair: yup, will replace with a path

nym-data-observatory/src/config/template.rs
mmsinclair: iirc they are optional, so will check and add them in as args / defaults if necessary

nym-validator-rewarder/src/config/persistence/paths.rs
mmsinclair: changed to clap arg

nym-data-observatory/src/config/data_observatory.rs
mmsinclair: changed to Url

nym-data-observatory/src/config/template.rs
mmsinclair: is optional and errors now

nym-data-observatory/src/chain_scraper/webhook.rs
mmsinclair: changed to TxModule and webhook processing is per message

nym-data-observatory/src/chain_scraper/webhook.rs
mmsinclair: changed to use this

nym-data-observatory/src/chain_scraper/webhook.rs
mmsinclair: added a TODO for future us to look back at the this wisdom

nym-data-observatory/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+tower-http = { workspace = true, features = ["cors", "trace"] }
+url = { workspace = true }
+utoipa = { workspace = true, features = ["axum_extras", "time"] }
+utoipa-swagger-ui = { workspace = true, features = ["axum"] }
+utoipauto = { workspace = true }
+
+
+[build-dependencies]
+anyhow = { workspace = true }
+glob = "0.3.3"
```
jstuczyn: can we avoid using explicit versioning and instead use whatever is defined in the workspace cargo.toml?

nym-data-observatory/migrations/0103_wasm.sql
jstuczyn, line 1: given we're copying common migrations to this directory during build, I think we should rename `0003_wasm` to say `0100_XXX` (as you have with the other nym-data-observatory specific migrations) as the moment we add a `0003` migration to the common, sqlx is going to complain about name clashes and inconsistencies

nym-data-observatory/build.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+use glob::glob;
+use std::env;
+use std::path::Path;
+
+fn main() {
+    // copy common manifest files from "../common/nyxd-scraper-psql/sql_migrations/* migrations"
+    println!("Copying common migrations...");
```
jstuczyn: I don't remember from the top of my head, but are we allowed to have debug `println!` inside `build.rs`? I think the compiler is actually using the STDOUT, like with this: `println!("cargo:rustc-env=DATABASE_URL={database_url}");`

nym-data-observatory/migrations/0103_wasm.sql
mmsinclair: I'll make this work better

nym-data-observatory/build.rs
mmsinclair: it works

nym-data-observatory/Cargo.toml
mmsinclair: done

nym-data-observatory/migrations/0103_wasm.sql
mmsinclair:
```
Applied 1/migrate metadata (3.288958ms)
Applied 2/migrate cosmos (45.50475ms)
Applied 103/migrate wasm (4.540417ms)
Applied 1000/migrate startup info (1.711042ms)
Applied 1101/migrate price data (1.493833ms)
Running sqlx prepare for PostgreSQL...
```

jstuczyn (approved):
let's ship this beast!

nym-data-observatory/src/modules/wasm.rs
mmsinclair: this is a good question

nym-data-observatory/src/modules/wasm.rs
mmsinclair: yup, can't remember my reason for doing this

nym-data-observatory/src/modules/wasm.rs
mmsinclair: it is ok to have `null` as the default, because we can always reprocess blocks later

you are right about the silent failures though, however, this is what bdjuno does, so going to keep this behaviour going, and you can tell me "I told you so" later 😬

nym-data-observatory/src/chain_scraper/mod.rs
mmsinclair: Part of config now

nym-data-observatory/src/modules/wasm.rs
mmsinclair: will address this later in another PR

nym-data-observatory/src/modules/wasm.rs
mmsinclair: will investigate and resolve if necessary in another PR

PR #6176, opened 2025-11-05
[bugfix] Distinguish authenticator errors by credential spent

simonwicky:
This PR introduces a distinction between authenticator client errors that are happening before a credential was taken out of storage, and after

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6176)


nym-authenticator-client/src/error.rs
```diff
@@ -42,10 +42,19 @@ pub enum Error {
 
     #[error("unknown authenticator version number")]
     UnsupportedAuthenticatorVersion,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    NoCredentialSpent(InnerError), // This intentionnally doesn't use `from` to avoid random ? operator to land here when they shouldn't
```
jstuczyn: perhaps put the inner value into named `source` field like in the `CredentialSpent` variant for consistency sake?

nym-authenticator-client/src/lib.rs
```diff
@@ -25,7 +26,7 @@ mod error;
 mod helpers;
 mod mixnet_listener;
 
-pub use crate::error::{Error, Result};
+pub use crate::error::{Error as RegistrationError, InnerError as AuthenticationClientError};
```
jstuczyn: if we're using those type aliases anyway, could we just rename the underlying errors themselves?

nym-authenticator-client/src/error.rs
```diff
@@ -42,10 +42,19 @@ pub enum Error {
...
 
-    #[error("failed to wait on AuthenticatorClientListener")]
-    FailedToJoinOnTask(#[from] tokio::task::JoinError),
+    #[error("an error occured after a credential was spent : {source}")]
+    CredentialSpent {
```
jstuczyn: nit: perhaps  change it to `CredentialSent`? because technically we don't know for sure if it was spent : )

nym-authenticator-client/src/lib.rs
```diff
@@ -158,10 +159,14 @@ impl AuthenticatorClient {
         &mut self,
         controller: &dyn BandwidthTicketProvider,
         ticketbook_type: TicketType,
-    ) -> Result<GatewayData> {
+    ) -> std::result::Result<GatewayData, Error> {
         debug!("Registering with the wg gateway...");
         let init_message = match self.auth_version {
```
jstuczyn, line 163: would it perhaps make sense to split this function into 2: one for all things happening before registration (i.e. `Error::NoCredentialSpent` and after (i.e. `Error::CredentialSpent`)?

then we could do something like
```
let foo = stuff_before(...).await.map_err(Error::NoCredentialSpent)?;
let bar = stuff_after(foo).await.map_err(Error::CredentialSpent)?;
...
```

nym-authenticator-client/src/lib.rs
simonwicky: 198fdbe6ef1fa8b7bc73d10ce7a68040-1061753530

nym-authenticator-client/src/error.rs
simonwicky: `source` doesn't work with `transparent` and I don't see a point of adding a layer on that variant

nym-authenticator-client/src/lib.rs
simonwicky: fixed in https://github.com/nymtech/nym/pull/6176/commits/b3864a18dc342fb3d8bae4e6a4aefe1ef59098b1

nym-authenticator-client/src/error.rs
simonwicky: Fixed in https://github.com/nymtech/nym/pull/6176/commits/b3864a18dc342fb3d8bae4e6a4aefe1ef59098b1

nym-authenticator-client/src/lib.rs
simonwicky: Current fn structure makes it very impractical

tommyv1987:
Testing steps:

Kill gateway before InitMessage
- Invalid gateway authenticator address
- Network disconnect before registration starts
Expected: Error shown but credential count unchanged

Test post-credential errors (SHOULD waste):
- Kill gateway after receiving PendingRegistration
- Corrupt FinalMessage
Expected: CredentialWastedOnEntryGateway error, credential count decreases

GetTicket errors (SHOULD waste):
- Empty credential database
- Insufficient bandwidth

`sudo sqlite3 /var/lib/nym-vpnd/mainnet/credentials_database.db \
  "SELECT ticketbook_type, COUNT(*) FROM ecash_ticketbook GROUP BY ticketbook_type"`
  
  Ticket books were being correctly spent / wasted on changes and selection of nodes

tommyv1987 (approved):

tommyv1987:
Tesing comments provided [here](https://github.com/nymtech/nym/pull/6176#pullrequestreview-3429210236)

PR #6174, opened 2025-11-03
feat: merge intermediate upgrade mode changes

jstuczyn:
this is a stepping stone towards the the naive versions [NET-451](https://nymtech.atlassian.net/browse/NET-451) and [NET-450](https://nymtech.atlassian.net/browse/NET-450) as part of [NET-341](https://nymtech.atlassian.net/browse/NET-341
and supersedes [#5986](https://github.com/nymtech/nym/pull/5986).

this PR contains all of the ongoing work that can be merged into develop to reduce code rot whilst additional work continues on a separate branch.

to fully understand the context behind the changes, it is crucial to familiarise yourself with the following document: https://nymtech.atlassian.net/wiki/spaces/KB/pages/263880725/Implementation#

as long as nothing gets put on the url controlled by `UPGRADE_MODE_ATTESTATION_URL` (by default `"https://nym.com/upgrade-mode/attestation.json"` on mainnet), there should be no changes in behaviour in our clients or nodes.

for reference for @simonwicky: this PR does **NOT** contain changes to `BandwidthTicketProvider` trait. this is still ongoing work on another branch

## Changelog

the following changes are by no means exhaustive, because, due to the fault of my own, I have severely underestimated the scope of the work and haven't been keeping track of all the relevant updates and had to attempt to produce the list on a later date.

#### Added

- `UPGRADE_MODE_ATTESTATION_URL` to control url used for polling attestation endpoint
- `UPGRADE_MODE_ATTESTER_ED25519_BS58_PUBKEY` to control expected key used for signing the attestation - @benedettadavico @tommyv1987: please visit our vault in order to retrieve private keys for sandbox and canary

#### Added

- `lints.workspace = true` section to `Cargo.toml` files of additional crates and fixed associated warnings
- `nym-credentials-interface`: `BandwidthCredential` to act as a wrapper for different types of credentials. currently it includes the old `CredentialSpendingData` (i.e. zk-nym ticketbooks) and JWT tokens issued during upgrade mode. in the future it is expected to get expanded to include emergency mode credentials issued by signer quorum

#### Changed

- multiple struct naming typos, e.g. `RegistredData` => `RegisteredData` (including all conversion impls) or `RegistredAndFree` => `RegisteredAndFree`

### Nym Node Changes

- there is now additional upgrade-mode related shared state between:
   - Wireguard Metadata Endpoint
   - WebSocket Listener
   - Authenticator Service Provider
   - Attestation Watcher
to control upgrade-mode related features
[...]

PR #6189, opened 2025-11-12
feat: upgrade mode: VPN adjustments

jstuczyn:
this PR further builds up on https://github.com/nymtech/nym/pull/6174 to include changes required by the VPN-client to fully support the upgrade mode. please refer to the aforementioned pull requests for more details regarding the scope of the changes.

what is relevant here is that this PR modifies the credential storage to allow it to storage an opaque `emergency credential` that lets it be shared between sessions (if it is still valid)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6189)


common/client-libs/gateway-client/src/socket_state.rs
```diff
@@ -27,6 +27,7 @@ use tokio::net::TcpStream;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio_tungstenite::{MaybeTlsStream, WebSocketStream};
 
+use crate::client::config::BandwidthTickets;
```
simonwicky: nit : move that import with the other `crate` ones

common/credential-storage/migrations/20251030120000_upgrade_mode_token.sql
```diff
@@ -0,0 +1,20 @@
...
+CREATE TABLE emergency_credential
+(
+    id         INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+
+    type       TEXT    NOT NULL,
+
+    -- don't define any strict schema on the content as it might be implementation-dependant
+    content    BLOB    NOT NULL,
+
+    expiration TIMESTAMP WITHOUT TIME ZONE
```
simonwicky, line 12: Ngl, all those empty lines are making it look weird

common/credential-storage/src/backends/memory.rs
```diff
@@ -29,13 +32,21 @@ struct EcashCredentialManagerInner {
     master_vk: HashMap<u64, VerificationKeyAuth>,
     coin_indices_sigs: HashMap<u64, Vec<AnnotatedCoinIndexSignature>>,
     expiration_date_sigs: HashMap<(u64, Date), Vec<AnnotatedExpirationDateSignature>>,
-    _next_id: i64,
+    emergency_credentials: HashMap<String, Vec<EmergencyCredential>>,
+    _next_ticketbook_id: i64,
+    _next_emergency_credential_id: i64,
```
simonwicky: Why the leading _ ? They are used no?

common/credential-storage/src/models.rs
```diff
@@ -3,6 +3,7 @@
 
 use nym_credentials::{IssuanceTicketBook, IssuedTicketBook};
 use nym_ecash_time::Date;
+use sqlx::types::time::OffsetDateTime;
```
simonwicky: We're using `time::OffsetDateTime` (almost) everywhere else. I know it's the same underlying one, but I reckon using the `time` one would be better

gateway/src/node/upgrade_mode/mod.rs
```diff
@@ -2,3 +2,5 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 pub(crate) mod watcher;
+
+pub use nym_credential_verification::upgrade_mode::UpgradeModeState;
```
simonwicky: That re-export looks out of place 🤔

simonwicky (approved):

common/credential-storage/src/backends/memory.rs
jstuczyn: a bit to imply this is a field that is not really meant to be used, but I guess since there's two of them now, I will put them alongside in an explicit wrapper

gateway/src/node/upgrade_mode/mod.rs
jstuczyn: oops. my lazy-self left it there during quick experimentation, fixed it properly!

PR #6200, opened 2025-11-13
HTTP API resilience enable & domain rotation conditions

jmwample:
Changes to make sure fallback domains are updated / enabled only for relevant errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6200)

### DIFF
--- a/common/http-api-client/src/lib.rs
+++ b/common/http-api-client/src/lib.rs
@@ -395,6 +395,13 @@ pub enum HttpClientError {
     #[error("failed to resolve request to {url} due to data inconsistency: {details}")]
     InternalResponseInconsistency { url: ::url::Url, details: String },
 
+    #[cfg(not(target_arch = "wasm32"))]
+    #[error("encountered dns failure: {inner}")]
+    DnsLookupFailure {
+        #[from]
+        inner: ResolveError,
+    },
+
     #[error("Failed to encode bincode: {0}")]
     Bincode(#[from] bincode::Error),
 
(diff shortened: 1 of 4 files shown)


common/http-api-client/src/lib.rs
```diff
@@ -1048,8 +1086,27 @@ impl ApiClientCore for Client {
                 .build()
                 .map_err(HttpClientError::reqwest_client_build_error)?;
             self.apply_hosts_to_req(&mut req);
+            let url = Url::parse(req.url().as_str()).unwrap();
```
durch: While this can't fail, we have an `From<reqwest::Url for Url>`, so something like `let url: Url = req.url().clone().into();` should work -> `common/http-api-client/src/url.rs`

common/http-api-client/src/lib.rs
```diff
@@ -1048,8 +1086,27 @@ impl ApiClientCore for Client {
...
+
+            // try an explicit DNS resolution - if successful then it will be in cache when reqwest
+            // goes to execute the request. If failure then we get to handle the DNS lookup error.
             #[cfg(not(target_arch = "wasm32"))]
-            let url = req.url().clone();
+            if self.using_secure_dns
+                && let Some(hostname) = req.url().domain()
+                && let Err(err) = HickoryDnsResolver::default().resolve_str(hostname).await
```
durch, line 1097: This `default` is non-obviously delegating to the shared resolver, maybe just add a little blurb for future us

durch (approved):
Two nits, overall good to go

PR #6203, opened 2025-11-14
KKT + PSQ

durch:
## Overview

This PR enhances the Lewes Protocol (LP) with post-quantum cryptographic security through two complementary mechanisms:

1. KKT (KEM Key Transport): A pre-handshake protocol for securely exchanging authenticated KEM public keys
2. PSQ (Post-Quantum PSK): Integration with libcrux-psq to derive post-quantum secure pre-shared keys for the Noise XKpsk3 handshake

Together, these additions provide defense-in-depth against quantum adversaries while maintaining backward compatibility with classical cryptography through hybrid approaches.

## Architecture

Protocol Flow

Client <--------------------------> Gateway
  |------------------------------------------|
  |--- TCP Connection -------------------->|
  |------------------------------------------|
  |--- KKT Request (KEM key request) ----> |
  |<-- KKT Response (signed KEM key) ----  |
  |------------------------------------------ |
  |--- PSQ Handshake (derive PQ-PSK) --->  |
  |<-- PSQ Response --------------------->  |
  |------------------------------------------ |
  |--- Noise XKpsk3 (with PQ-PSK) ------->  |
  |<-- Noise Handshake -------------------> |
  |-------------------------------------------|
  |--- Encrypted Data Transport -----------> |

### State Machine Updates

The LP state machine now includes a new KKTExchange state between ReadyToHandshake and Handshaking:

ReadyToHandshake → KKTExchange → Handshaking → Transport

Both KKT and PSQ states are tracked independently within the session, allowing fine-grained control over the handshake phases.

### KKT Protocol

- Authenticated Key Exchange: KEM public keys are signed with Ed25519 to prevent MITM attacks
- Hash Verification: Client verifies KEM key hash against trusted directory data
- Flexible KEM Support: ML-KEM-768, XWing, X25519, Classic McEliece

### PSQ Integration

- Post-Quantum PSK Derivation: Uses libcrux-psq for quantum-resistant pre-shared key generation
- Hybrid Security: Combines classical ECDH with PQ-KEM encapsulation

### New Components

- common/nymkkt/: New crate implementing the KKT protocol
  - Ciphersuite negotiation (KEM, hash, signature algorithms)
  - KKT frame serialization/deserialization
  - Key utilities for KEM keypair generation and hashing
  - Session management for initiator/responder roles
- common/nym-lp/src/kkt_orchestrator.rs: High-level KKT orchestration functions
  - create_request(): Client-side KKT request generation
  - process_response(): Client-side response validation
  - handle_request(): Gateway-side request handling

### Modified Components

- common/nym-lp/src/session.rs: Extended session management
  - Added KKTState enum to track KKT exchange progress
  - Added PSQState enum to track PSQ handshake progress
  - Integrated KKT request/response methods into LpSession
  - PSQ-aware handshake message processing
- common/nym-lp/src/state_machine.rs: State machine enhancements
  - New KKTExchange state for pre-handshake KEM exchange
  - Updated transitions to include KKT phase
  - Ed25519-only constructor with internal X25519 derivation
- common/nym-lp/src/codec.rs: Protocol message updates
  - MessageType::KKTRequest and MessageType::KKTResponse
  - Updated ClientHelloData to include Ed25519 public key
  - Frame serialization for KKT messages

## Dependencies

- libcrux: Cryptographic primitives (ML-KEM, ChaCha20Poly1305, HKDF, etc.)
- classic-mceliece-rust: Classic McEliece KEM implementation
- tls_codec: Serialization for cryptographic structures

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6203)


common/nym-lp/Cargo.toml
```diff
@@ -14,13 +14,23 @@ bytes = { workspace = true }
 dashmap = { workspace = true }
 sha2 = { workspace = true }
 ansi_term = { workspace = true }
+log = { workspace = true }
```
jstuczyn: can we avoid using `log` library and instead use `tracing` like we do in all new crates we introduce?

common/nym-lp/Cargo.toml
```diff
@@ -14,13 +14,23 @@ bytes = { workspace = true }
...
 nym-crypto = { path = "../crypto", features = ["hashing", "asymmetric"] }
+nym-kkt = { path = "../nymkkt" }
 nym-lp-common = { path = "../nym-lp-common" }
 nym-sphinx = { path = "../nymsphinx" }
 
+# libcrux dependencies for PSQ (Post-Quantum PSK derivation)
+libcrux-psq = { git = "https://github.com/cryspen/libcrux", features = ["test-utils"] }
+libcrux-kem = { git = "https://github.com/cryspen/libcrux" }
+libcrux-traits = { git = "https://github.com/cryspen/libcrux" }
+tls_codec = "0.4"
```
jstuczyn: can we put this in the workspace to keep versioning consistent??

common/nym-lp/Cargo.toml
```diff
@@ -14,13 +14,23 @@ bytes = { workspace = true }
 dashmap = { workspace = true }
 sha2 = { workspace = true }
 ansi_term = { workspace = true }
+log = { workspace = true }
 utoipa = { workspace = true, features = ["macros", "non_strict_integers"] }
 rand = { workspace = true }
+# rand 0.9 for KKT integration (nym-kkt uses rand 0.9)
+rand09 = { package = "rand", version = "0.9.2" }
```
jstuczyn, line 22: while using old versions is not an ideal choice for new crates, but is there anything in particular blocking us from using rand 0.8 in `nym-kkt` given we control its content?

common/nym-lp/src/message.rs
```diff
@@ -72,6 +80,8 @@ impl MessageType {
             0x0001 => Some(MessageType::Handshake),
             0x0002 => Some(MessageType::EncryptedData),
             0x0003 => Some(MessageType::ClientHello),
+            0x0004 => Some(MessageType::KKTRequest),
```
jstuczyn: suggestion for this, and other variants to not duplicate the magic values:
```suggestion
            n if n == MessageType::KKTRequest as 16 => Some(MessageType::KKTRequest),
```

common/nym-lp/src/message.rs
```diff
@@ -82,6 +92,8 @@ impl MessageType {
             MessageType::Handshake => 0x0001,
             MessageType::EncryptedData => 0x0002,
             MessageType::ClientHello => 0x0003,
+            MessageType::KKTRequest => 0x0004,
+            MessageType::KKTResponse => 0x0005,
```
jstuczyn: you can remove the whole match and replace it with `(*self) as u16` instead

common/nym-lp/src/message.rs
```diff
@@ -118,6 +142,8 @@ impl LpMessage {
             LpMessage::Handshake(payload) => payload.0.as_slice(),
             LpMessage::EncryptedData(payload) => payload.0.as_slice(),
             LpMessage::ClientHello(_) => unimplemented!(), // Structured data, serialized in encode_content
```
jstuczyn, line 130: missing implementation

common/nym-lp/src/message.rs
```diff
@@ -135,7 +163,9 @@ impl LpMessage {
             LpMessage::Busy => 0,
             LpMessage::Handshake(payload) => payload.0.len(),
             LpMessage::EncryptedData(payload) => payload.0.len(),
-            LpMessage::ClientHello(_) => 65, // 32 bytes key + 1 byte version + 32 bytes salt
+            LpMessage::ClientHello(_) => 97, // 32 bytes x25519 key + 32 bytes ed25519 key + 32 bytes salt + 1 byte bincode overhead
```
jstuczyn, line 152: can we perhaps define this as a constant (in terms of its components, i.e. `const CLIENT_HELLO_SIZE = X25519_PUBKEY_SIZE + ED25519_PUBKEY_SIZE + ...` on the top of the file? otherwise I fear it might rot

common/nym-lp/src/message.rs
```diff
@@ -163,6 +195,12 @@ impl LpMessage {
                     bincode::serialize(data).expect("Failed to serialize ClientHelloData");
```
jstuczyn, line 181: two things here:
1. can we return a `Result` in case serialisation fails? we want to avoid having random panics inside crypto code if something changes
2. can we have some explicit function defining the serialiser, like we have `fn make_bincode_serializer() -> impl bincode::Options` in multiple places for, for example, http-api-common, gateway-storage, etc. the reason for it is that the default options in bincode 2.0 are changed so the explicit serialiser will allow us to migrate more easily without breaking compatibility

common/nym-kkt/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+libcrux-traits = { git = "https://github.com/cryspen/libcrux" }
+libcrux-kem = { git = "https://github.com/cryspen/libcrux" }
+libcrux-psq = { git = "https://github.com/cryspen/libcrux", features = ["test-utils"] }
+libcrux-sha3 = { git = "https://github.com/cryspen/libcrux" }
+libcrux-ml-kem = { git = "https://github.com/cryspen/libcrux" }
+libcrux-ecdh = { git = "https://github.com/cryspen/libcrux", features = ["codec"]}
+
+rand = "0.9.2"
+curve25519-dalek = {version = "4.1.3", features = ["rand_core", "serde"] }
```
jstuczyn, line 34: nit: move to workspace

nym-registration-client/src/builder/mod.rs
```diff
@@ -46,7 +50,7 @@ impl RegistrationClientBuilder {
             let builder = MixnetClientBuilder::new_with_storage(mixnet_client_storage)
                 .event_tx(EventSender(event_tx));
             let mixnet_client = tokio::time::timeout(
-                self.config.mixnet_client_startup_timeout,
+                MIXNET_CLIENT_STARTUP_TIMEOUT,
```
jstuczyn, line 53: why are we moving away from the value set in the config?

common/nym-lp/src/error.rs
```diff
@@ -48,6 +49,9 @@ pub enum LpError {
     #[error("Deserialization error: {0}")]
     DeserializationError(String),
 
+    #[error("KKT protocol error: {0}")]
+    KKTError(String),
```
jstuczyn, line 53: nit: could we perhaps have structured variants for possible KKT errors?

common/nym-lp/src/psk.rs
```diff
@@ -45,6 +94,14 @@ const PSK_CONTEXT: &str = "nym-lp-psk-v1";
 /// let client_public = /* from ClientHello */;
 /// let psk = derive_psk(&gateway_private, &client_public, &salt);
 /// ```
+///
+/// # Deprecation Warning
+/// This function provides no post-quantum security. Use `derive_psk_with_psq_initiator`
+/// or `derive_psk_with_psq_responder` for HNDL (Harvest-Now, Decrypt-Later) resistance.
+#[deprecated(
+    since = "0.1.0",
```
jstuczyn: nit: in that case we should probably bump up the version in `Cargo.toml` : )

gateway/src/node/lp_listener/handler.rs
```diff
@@ -278,7 +259,16 @@ impl LpConnectionHandler {
     }
 
     /// Receive client's public key and salt via ClientHello message
-    async fn receive_client_hello(&mut self) -> Result<(PublicKey, [u8; 32]), GatewayError> {
+    async fn receive_client_hello(
+        &mut self,
+    ) -> Result<
+        (
+            PublicKey,
+            nym_crypto::asymmetric::ed25519::PublicKey,
```
jstuczyn, line 255: can we perhaps wrap that return tuple in some struct? it's very much non-obvious looking at the signature what each field represents

gateway/src/node/lp_listener/handler.rs
```diff
@@ -104,36 +104,14 @@ impl LpConnectionHandler {
...
-        let gateway_keypair = Keypair::from_keys(lp_private, lp_public);
+        // AIDEV-NOTE: Gateway LP handshake with Ed25519-only API
+        // The state machine now accepts only Ed25519 keys and internally derives X25519 keys.
+        // This simplifies the API by removing manual key conversion from the caller.
+        // Gateway's Ed25519 identity is used for both PSQ authentication and X25519 derivation.
 
         // Receive client's public key and salt via ClientHello message
         // The client initiates by sending ClientHello as first packet
-        let (client_pubkey, salt) = match self.receive_client_hello().await {
+        let (client_pubkey, client_ed25519_pubkey, salt) = match self.receive_client_hello().await {
```
jstuczyn: client's pubkey (i.e. the x25519) seems to be unused? is this expected?

jstuczyn: ah. right. we derive it from ed25519 - in that case, do we still need to include them in the ClientHello?

docs/LP_PROTOCOL.md
jstuczyn, line 1: the file seems to have slightly rotten compared to the actual implementation

docs/LP_PROTOCOL.md
```diff
@@ -0,0 +1,990 @@
...
+```rust
+pub struct RegistrationRequest {
+    pub mode: RegistrationMode,
+    pub credential: EcashCredential,
+    pub gateway_identity: String,
+}
+
+pub enum RegistrationMode {
+    Dvpn {
+        wg_public_key: [u8; 32],
```
jstuczyn, line 292: for  example (regarding the rotting comments), this field does not exist in the actual implementation

common/nym-kkt/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "nym-kkt"
+version = "0.1.0"
+authors = ["Georgio Nicolas <georgio@nymtech.net>"]
+edition = "2021"
+license.workspace = true
+
+[dependencies]
+arc-swap = { workspace = true }
```
jstuczyn, line 9: it seems that most of these dependencies, apart from `blake3` and `thiserror` are actually unused

common/nym-kkt/src/ciphersuite.rs
```diff
@@ -0,0 +1,311 @@
...
+                // 64 bytes
+                SignatureScheme::Ed25519 => ed25519::SIGNATURE_LENGTH,
+            },
+        })
+    }
+    pub fn encode(&self) -> [u8; 4] {
+        // [kem, hash, hashlen, sig]
+        [
+            match self.kem {
+                KEM::XWing => 0,
```
jstuczyn, line 210: to avoid duplicating magic values, could we keep the encodings on the definitions itself, i.e. either include `repr(u8)` on those enums or expose `fn as_u8(&self) -> u8`

common/nym-kkt/src/ciphersuite.rs
```diff
@@ -0,0 +1,311 @@
...
+                _ => self.hash_length,
+            },
+            match self.signature_scheme {
+                SignatureScheme::Ed25519 => 0,
+            },
+        ]
+    }
+    pub fn decode(encoding: &[u8]) -> Result<Self, KKTError> {
+        if encoding.len() == 4 {
+            let kem = match encoding[0] {
```
jstuczyn, line 232: that's just personal opinion, but I think it'd be more readable if there was `TryFrom<u8> for KEM` (and other parts of the ciphersuite), then the whole decoding could be simplified to:
```rust
if encoding.len() != 4 {
  return Err(KKTError::CiphersuiteDecodingError {
      info: format!(
          "Incorrect Encoding Length: actual: {} != expected: {}",
          encoding.len(),
          CIPHERSUITE_ENCODING_LEN
      ),
  })

let kem = encoding[0].try_into()?;
let hash_function = encoding[1].try_into()?;
let custom_hash_length = ... // todo
let signature_scheme = encoding[3].try_into()?;
Self::resolve_ciphersuite(kem, hash_function, signature_scheme, custom_hash_length)
```

common/nymkkt/src/ciphersuite.rs
```diff
@@ -0,0 +1,311 @@
...
+        )
+    }
+}
+
+pub const fn map_kem_to_libcrux_kem(kem: KEM) -> Algorithm {
+    match kem {
+        KEM::MlKem768 => Algorithm::MlKem768,
+        KEM::XWing => Algorithm::XWingKemDraft06,
+        KEM::X25519 => Algorithm::X25519,
+        _ => unreachable!(),
```
jstuczyn: I disagree with this being unreachable. since the function is declared as public, anyone could theoretically call it with an invalid argument, so either make it return an Option/Result or ensure it can never be called within an invalid context

common/nym-kkt/src/session.rs
```diff
@@ -0,0 +1,223 @@
...
+
+                    match Signature::from_bytes(remote_frame.signature_ref()) {
+                        Ok(sig) => match remote_verif_key.verify(bytes_to_verify, &sig) {
+                            Ok(()) => {
+                                // using own_context here because maybe for whatever reason we want to ignore the remote kem key
+                                match own_context.mode() {
+                                    KKTMode::OneWay => Ok((own_context, None)),
+                                    KKTMode::Mutual => {
+                                        match expected_hash {
+    Some(expected_hash) =>{
```
jstuczyn: uhm. not sure what happened here, but can we somehow reformat it manually? (for some reason `cargo fmt` sees no issue with this O.o). i wonder if this is due too many levels of match statements. perhaps it'd be worth to do some early returns instead?

common/nym-kkt/src/kkt.rs
```diff
@@ -0,0 +1,355 @@
...
+/// // Send response_frame back to client
+/// ```
+pub fn handle_kem_request<'a>(
+    request_frame: &KKTFrame,
+    initiator_vk: Option<&ed25519::PublicKey>,
+    responder_signing_key: &ed25519::PrivateKey,
+    responder_kem_key: &EncapsulationKey<'a>,
+) -> Result<KKTFrame, KKTError> {
+    // Parse context from the request frame
+    let request_bytes = request_frame.to_bytes();
```
jstuczyn, line 126: I'm a bit confused here. we're converting the KKTFrame to bytes only to recover it from bytes immediately? can't we just extract the context directly from the frame instead?

common/nymkkt/src/frame.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+pub struct EncryptedKKTRequest {
+    ephermeral_key: PublicKey,
+    body: Vec<u8>,
+}
+impl EncryptedKKTRequest {
+    pub fn to_bytes(mut self) -> Vec<u8> {
+        self.body.extend(self.ephermeral_key.to_bytes());
```
jstuczyn: just the nittest of nits: I'd move `self.body` to a different binding to make it more readable. in a pre-coffee state I was confused reading this thinking: "wait, why is he mutating the body content"
```suggestion
    pub fn to_bytes(self) -> Vec<u8> {
        let mut bytes = self.body;
        bytes.extend(self.ephermeral_key.to_bytes());
        bytes
    }
```

common/nym-kkt/src/context.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    pub fn header_len(&self) -> usize {
+        KKT_CONTEXT_LEN
+    }
+
+    pub fn session_id_len(&self) -> usize {
+        // match self.role {
+        //     KKTRole::Initiator | KKTRole::Responder => SESSION_ID_LENGTH,
+        // It doesn't make sense to send a session_id if we send messages in the clear
+        //     KKTRole::AnonymousInitiator => 0,
```
jstuczyn, line 138: remove dead comments

common/nym-kkt/src/frame.rs
```diff
@@ -0,0 +1,146 @@
...
+
+pub struct KKTFrame {
+    context: Vec<u8>,
+    session_id: Vec<u8>,
+    body: Vec<u8>,
+    signature: Vec<u8>,
+}
+
+// if oneway and message coming from initiator => body is empty, signature contains signature of context + session id (64 bytes).
+// if message coming from anonymous initiator => body is empty, there is no signature.
```
jstuczyn, line 25: for consistency sake, can you also update the comment to say that `session_id` is indeed included even for an anonymous initiator?

common/nymkkt/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+impl From<libcrux_ecdh::Error> for KKTError {
+    fn from(err: libcrux_ecdh::Error) -> Self {
+        match err {
+            libcrux_ecdh::Error::InvalidPoint => KKTError::KEMError {
+                info: "Invalid Remote Public Key",
+            },
+            _ => todo!(),
```
jstuczyn: remaining todo.

could you perhaps add 
```
[lints]
workspace = true
```
section to your cargo.toml to make sure all those instances are always caught?

common/nymkkt/src/context.rs
```diff
@@ -0,0 +1,258 @@
...
+        if self.message_sequence >= 1 << 4 {
+            return Err(KKTError::MessageCountLimitReached);
+        }
+
+        header_bytes.push((KKT_VERSION << 4) + self.message_sequence);
+
+        header_bytes.push(
+            match self.status {
+                KKTStatus::Ok => 0,
+                KKTStatus::InvalidRequestFormat => 0b001_000_00,
```
jstuczyn: similarly to the ciphersuite, can we remove the duplication of the magic bytes from `encode` and `decode` and instead define those directly on the enum variants?

common/nymkkt/src/context.rs
```diff
@@ -0,0 +1,258 @@
...
+
+            // We only check if stuff is valid here, not necessarily if it's compatible
+
+            if (kkt_version >> 4) > KKT_VERSION {
+                return Err(KKTError::FrameDecodingError {
+                    info: format!("Header - Invalid KKT Version: {}", kkt_version >> 4),
+                });
+            }
+
+            let status = match header_bytes[1] & 0b111_000_00 {
```
jstuczyn: and similarly to ciphersuite (though in this case a `TryFrom` wouldn't be too appropriate), I think we would increase readibility with separate functions for those

common/nym-kkt/src/context.rs
```diff
@@ -0,0 +1,258 @@
...
+                KKTMode::Mutual => 0b000_001_00,
+            } + match self.role {
+                KKTRole::Initiator => 0,
+                KKTRole::Responder => 1,
+                KKTRole::AnonymousInitiator => 2,
+            },
+        );
+
+        header_bytes.extend_from_slice(&self.ciphersuite.encode());
+        header_bytes.push(0);
```
jstuczyn, line 176: what's the point of the trailing 0? it seems to get ignored in the decoding anyway

common/nym-kkt/src/encryption.rs
jstuczyn, line 1: what's the purpose of this file - it's not included at all? is this for a future implementation of psq-proper (since it's also missing from the module declaration)

common/nym-kkt/src/encryption.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+fn decrypt(key: [u8; 32], ciphertext: Vec<u8>) -> Vec<u8> {
+    // The empty nonce is fine since we use the key once.
+    let nonce = Nonce::<Aes256>::from_slice(&[]);
+
+    let ciphertext =
+        nym_crypto::symmetric::aead::encrypt::<Aes256GcmSiv>(&key.into(), nonce, message)?;
+
+    key.zeroize();
```
jstuczyn, line 92: if encryption fails, then zeroize won't get triggered as the function would return. instead I suggest to move the key into a `Zeroizing` wrapper, i.e.
```suggestion
let key = Zeroizing::new(key);
// The empty nonce is fine since we use the key once.
let nonce = Nonce::<Aes256>::from_slice(&[]);
// ..
```

common/nym-kkt/src/encryption.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+fn encrypt(mut key: [u8; 32], message: &[u8]) -> Result<Vec<u8>, KKTError> {
+    // The empty nonce is fine since we use the key once.
+    let nonce = Nonce::<Aes256GcmSiv>::from_slice(&[]);
+
+    let ciphertext =
+        nym_crypto::symmetric::aead::encrypt::<Aes256GcmSiv>(&key.into(), nonce, message)?;
+
+    key.zeroize();
```
jstuczyn, line 80: if decryption fails, then zeroize won't get triggered as the function would return. instead I suggest to move the key into a `Zeroizing` wrapper, i.e.
```suggestion
let key = Zeroizing::new(key);
// The empty nonce is fine since we use the key once.
let nonce = Nonce::<Aes256>::from_slice(&[]);
// ..
```

common/nym-kkt/src/encryption.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let mut hasher = Hasher::new();
+    hasher.update(&gsb);
+    gsb.zeroize();
+    let key: [u8; 32] = hasher.finalize().as_bytes().to_owned();
+
+    hasher.update(b.public_key().as_bytes());
+    hasher.update(gs.as_bytes());
+
+    // This runs in constant time
```
jstuczyn, line 63: no it doesn't : )
if we really care about constant time equality, you'd have to call `ct_eq` from `subtle::ConstantTimeEq` like here: https://github.com/nymtech/nym/blob/7d59a2477a93466279fd1a6371aca5652d5d165f/common/gateway-requests/src/types/text_request/authenticate.rs#L83

common/nym-kkt/src/session.rs
```diff
@@ -0,0 +1,223 @@
...
+    let signature = signing_key.sign(bytes_to_sign).to_bytes();
+
+    Ok(KKTFrame::new(&context_bytes, &body, session_id, &signature))
+}
+
+fn check_compatibility(
+    _own_context: &KKTContext,
+    _remote_context: &KKTContext,
+) -> Result<(), KKTError> {
+    // todo: check ciphersuite/context compatibility
```
jstuczyn, line 232: well, let's : )

common/nym-kkt/src/session.rs
```diff
@@ -0,0 +1,223 @@
...
+    let (frame, remote_context) = KKTFrame::from_bytes(message_bytes)?;
+
+    check_compatibility(own_context, &remote_context)?;
+    match remote_context.status() {
+        KKTStatus::Ok => {
+            let mut bytes_to_verify: Vec<u8> = Vec::with_capacity(
+                remote_context.full_message_len() - remote_context.signature_len(),
+            );
+            bytes_to_verify.extend_from_slice(&remote_context.encode()?);
+            bytes_to_verify.extend_from_slice(frame.body_ref());
```
jstuczyn, line 93: nit: in multiple places you're creating `bytes_to_verify` and `bytes_to_sign` - I think it could be potentially useful to extract that to some separate helper function to make sure it's always consistent (in case something changes in implementation)

common/nym-kkt/src/session.rs
```diff
@@ -0,0 +1,223 @@
...
+            bytes_to_verify.extend_from_slice(frame.session_id_ref());
+
+            match Signature::from_bytes(frame.signature_ref()) {
```
jstuczyn, line 96: shouldn't we technically match on the scheme inside the ciphersuite provided within the context? what if (once we have multiple enum variants) we used non ed25519 signature?

common/nym-kkt/src/session.rs
jstuczyn: e.g.
```rust

pub fn responder_ingest_message<'a>(
    remote_context: &KKTContext,
    remote_verification_key: Option<&ed25519::PublicKey>,
    expected_hash: Option<&[u8]>,
    remote_frame: &KKTFrame,
) -> Result<(KKTContext, Option<EncapsulationKey<'a>>), KKTError> {
    let own_context = remote_context.derive_responder_header()?;

    match remote_context.role() {
        KKTRole::AnonymousInitiator => return Ok((own_context, None)),
        KKTRole::Responder => {
            return Err(KKTError::IncompatibilityError {
                info: "Responder received a request from another responder.",
            })
        }
        KKTRole::Initiator => {}
    }

    let Some(remote_verification_key) = remote_verification_key else {
        return Err(KKTError::FunctionInputError {
            info: "Remote Signature Verification Key Not Provided",
        });
    };

    let mut bytes_to_verify: Vec<u8> =
        Vec::with_capacity(own_context.full_message_len() - own_context.signature_len());
    bytes_to_verify.extend_from_slice(remote_frame.context_ref());
    bytes_to_verify.extend_from_slice(remote_frame.body_ref());
    bytes_to_verify.extend_from_slice(remote_frame.session_id_ref());

    let Ok(signature) = Signature::from_bytes(remote_frame.signature_ref()) else {
        return Err(KKTError::SigConstructorError);
    };

    if remote_verification_key
        .verify(bytes_to_verify, &signature)
        .is_err()
    {
        return Err(KKTError::SigVerifError);
    }

    // using own_context here because maybe for whatever reason we want to ignore the remote kem key
    match own_context.mode() {
        KKTMode::OneWay => Ok((own_context, None)),
        KKTMode::Mutual => {
            let Some(expected_hash) = expected_hash else {
                return Err(KKTError::FunctionInputError {
                    info: "Expected hash of the remote encapsulation key is not provided.",
                });
            };
            let received_encapsulation_key =
                EncapsulationKey::decode(own_context.ciphersuite().kem(), remote_frame.body_ref())?;
            if validate_encapsulation_key(
                &own_context.ciphersuite().hash_function(),
                own_context.ciphersuite().hash_len(),
                remote_frame.body_ref(),
                expected_hash,
            ) {
                Ok((own_context, Some(received_encapsulation_key)))
            }
            // The key does not match the hash obtained from the directory
            else {
                Err(KKTError::KEMError { info: "Hash of received encapsulation key does not match the value stored on the directory." })
            }
        }
    }
}
```

common/nym-lp/src/message.rs
```diff
@@ -20,9 +22,12 @@ impl ClientHelloData {
...
     /// # Arguments
-    /// * `client_lp_public_key` - Client's x25519 public key
-    /// * `protocol_version` - Protocol version number
-    pub fn new_with_fresh_salt(client_lp_public_key: [u8; 32]) -> Self {
+    /// * `client_lp_public_key` - Client's x25519 public key (derived from Ed25519)
+    /// * `client_ed25519_public_key` - Client's Ed25519 public key (for PSQ authentication)
+    pub fn new_with_fresh_salt(
+        client_lp_public_key: [u8; 32],
+        client_ed25519_public_key: [u8; 32],
+    ) -> Self {
```
jstuczyn, line 31: could we maybe also inject current timestamp here to make the function independent of the outside world and also make it easier to unit test? I guess the same could apply to the rng instance

common/nym-lp/src/message.rs
```diff
@@ -8,8 +8,10 @@ use serde::{Deserialize, Serialize};
 /// Data structure for the ClientHello message
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ClientHelloData {
-    /// Client's LP x25519 public key (32 bytes)
+    /// Client's LP x25519 public key (32 bytes) - derived from Ed25519 key
     pub client_lp_public_key: [u8; 32],
```
jstuczyn, line 13: would it be beneficial if instead of using byte arrays, we used properly strongly typed keys and salt (if new type was to be created)? then we could validate the content at the deserialisation level

nym-registration-client/src/lp_client/client.rs
```diff
@@ -243,7 +254,7 @@ impl LpRegistrationClient {
             client_hello_data.extract_timestamp()
         );
 
-        // Step 2: Send ClientHello as first packet (before Noise handshake)
+        // Step 3: Send ClientHello as first packet (before Noise handshake)
         let client_hello_header = nym_lp::packet::LpHeader::new(
```
jstuczyn, line 257: there haven't been any changes in the corresponding `packet.rs`, but after glancing at it, I have few comments about it:
- nit1: it'd be useful if protocol_version was defined in some kind of global constant instead of setting it via a magic number, i.e. `protocol_version: 1,`
- nit2: you defined `reserved` field as `u16` yet, as you said yourself, it's a 3-byte value, so it's a bit misleading. given this field is not actually used at all, should we just drop it completely? from the struct definition that is. it should still be included in serialisation obviously

common/nym-lp/src/state_machine.rs
```diff
@@ -118,44 +129,76 @@ impl LpStateMachine {
...
+        salt: &[u8; 32],
     ) -> Result<Self, LpError> {
-        // Calculate the shared lp_id// Calculate the shared lp_id
-        let lp_id = make_lp_id(local_keypair.public_key(), remote_public_key);
+        // AIDEV-NOTE: Ed25519→X25519 conversion for API simplification
+        // We use standard RFC 7748 conversion to derive X25519 keys from Ed25519 identity keys.
+        // This allows callers to provide only Ed25519 keys (which they already have for signing/identity)
+        // without needing to manage separate X25519 keypairs.
+        //
+        // Security: Ed25519→X25519 conversion is cryptographically sound (RFC 7748).
```
jstuczyn, line 160: whilst cryptographically sound, it's usually quite a bad practice to reuse keys for different purposes, but I'll defer to @aniampio's opinion here

common/nym-lp/src/session.rs
```diff
@@ -50,24 +209,36 @@ impl LpSession {
...
+    /// * `remote_ed25519_key` - Peer's Ed25519 public key for PSQ authentication
+    /// * `remote_x25519_key` - Peer's X25519 public key for Noise protocol and DHKEM
+    /// * `salt` - Salt for PSK derivation
     pub fn new(
         id: u32,
         is_initiator: bool,
-        local_private_key: &[u8],
-        remote_public_key: &[u8],
-        psk: &[u8],
+        local_ed25519_keypair: (&ed25519::PrivateKey, &ed25519::PublicKey),
```
jstuczyn, line 237: what's wrong with using the `ed25519::KeyPair` wrapper that was created for the purpose of avoid such tuples?

common/nym-lp/src/session_manager.rs
```diff
@@ -168,12 +169,17 @@ impl SessionManager {
 
     pub fn create_session_state_machine(
         &self,
-        local_keypair: &Keypair,
-        remote_public_key: &PublicKey,
+        local_ed25519_keypair: (&ed25519::PrivateKey, &ed25519::PublicKey),
```
jstuczyn, line 169: any reason for replacing `KeyPair` with the explicit tuple?

common/nym-lp/src/session_manager.rs
```diff
@@ -168,12 +169,17 @@ impl SessionManager {
...
-        let sm = LpStateMachine::new(is_initiator, local_keypair, remote_public_key, psk)?;
+        let sm = LpStateMachine::new(
+            is_initiator,
+            local_ed25519_keypair,
+            remote_ed25519_key,
+            salt,
+        )?;
         let sm_id = sm.id()?;
 
         self.state_machines.insert(sm_id, sm);
```
jstuczyn, line 182: I couldn't find any instances of this (`create_session_state_machine`)  being called outside tests, so I'm not entirely sure what's the full expected flow, but given the session id is derived purely from public keys (local and remote), what if remote uses a malicious public key, i.e. of another party, to overwrite their session data? what I'm saying here is that we should make some additional checks before inserting new state machine, i.e. check if an entry already exists (perhaps also some staleness checks)?

common/nym-lp/src/state_machine.rs
```diff
@@ -118,44 +129,76 @@ impl LpStateMachine {
...
+        // The derived X25519 keys are used for:
+        // - Noise protocol ephemeral DH
+        // - PSQ ECDH baseline security (pre-quantum)
+        // - lp_id calculation (session identifier)
+
+        // Convert Ed25519 keys to X25519 for Noise protocol
+        let local_x25519_private = local_ed25519_keypair.0.to_x25519();
```
jstuczyn, line 167: nit: private key gets dropped without getting zeroized

common/nym-lp/src/state_machine.rs
```diff
@@ -118,44 +129,76 @@ impl LpStateMachine {
...
+        let lp_public = LpPublicKey::from_bytes(local_x25519_public.as_bytes())?;
+        let lp_remote_public = LpPublicKey::from_bytes(remote_x25519_public.as_bytes())?;
+
+        // Create X25519 keypair for Noise and lp_id calculation
+        let local_x25519_keypair = Keypair::from_keys(lp_private, lp_public);
 
-        let local_private_key = local_keypair.private_key().to_bytes();
-        let remote_public_key = remote_public_key.as_bytes();
+        // Calculate the shared lp_id using derived X25519 keys
+        let lp_id = make_lp_id(local_x25519_keypair.public_key(), &lp_remote_public);
```
jstuczyn, line 186: what's with the internal magic `0xCC`?

common/nym-lp/src/state_machine.rs
```diff
@@ -118,44 +129,76 @@ impl LpStateMachine {
...
 
-        // Create the session immediately
+        // Create the session with both Ed25519 (for PSQ auth) and derived X25519 keys (for Noise)
```
jstuczyn, line 188: this looks like a big red flag to me: using the same underlying keys for two completely different purposes. @aniampio / @georgio mind taking a look at this?

common/nym-lp/src/state_machine.rs
```diff
@@ -170,35 +213,147 @@ impl LpStateMachine {
...
+                                Ok(libcrux_public_key) => {
+                                    let responder_kem_pk = EncapsulationKey::X25519(libcrux_public_key);
+
+                                    match session.process_kkt_request(&kkt_request.0, &responder_kem_pk) {
+                                        Ok(kkt_response_message) => {
+                                            match session.next_packet(kkt_response_message) {
+                                                Ok(response_packet) => {
+                                                    result_action = Some(Ok(LpAction::SendPacket(response_packet)));
+                                                    // After KKT exchange, move to Handshaking
+                                                    LpState::Handshaking { session }
```
jstuczyn, line 285: so many different levels of indentation makes it quite hard to read - could we perhaps split up this function? maybe separate handler per state?

common/nym-lp/src/psk.rs
```diff
@@ -57,6 +114,349 @@ pub fn derive_psk(
...
+    let (psq_psk, ciphertext) = PsqX25519::encapsulate_psq(&kem_pk, PSQ_SESSION_CONTEXT, &mut rng)
+        .map_err(|e| LpError::Internal(format!("PSQ encapsulation failed: {:?}", e)))?;
+
+    // Step 3: Combine ECDH + PSQ via Blake3 KDF
+    let mut combined = Vec::with_capacity(64 + psq_psk.len());
+    combined.extend_from_slice(ecdh_secret.as_bytes());
+    combined.extend_from_slice(&psq_psk); // psq_psk is [u8; 32], need &
+    combined.extend_from_slice(salt);
+
+    let final_psk = nym_crypto::kdf::derive_key_blake3(PSK_PSQ_CONTEXT, &combined, &[]);
```
jstuczyn, line 127: nit: you explicitly defined a `salt` argument in `derive_key_blake3`, yet decided to trait it as the key_material instead. I mean sure, the end result will be identical, but in that case, why separate them?

common/nym-lp/src/psk.rs
```diff
@@ -57,6 +114,349 @@ pub fn derive_psk(
...
+
+    // Serialize ciphertext using TLS encoding for transport
+    let ct_bytes = ciphertext
+        .tls_serialize_detached()
```
jstuczyn, line 131: doesn't this leak a bit of implementation and make us a bit inflexible? i.e. why should derivation of psk give us tls encoded bytes? what if used a different transport layer? it seems like those should perhaps be split into different steps

common/nym-lp/src/psk.rs
```diff
@@ -57,6 +114,349 @@ pub fn derive_psk(
...
+                "Only X25519 KEM is currently supported for PSQ".to_string(),
+            ))
+        }
+    };
+
+    // Convert nym Ed25519 keys to libcrux format
+    type Ed25519VerificationKey = <Ed25519 as Authenticator>::VerificationKey;
+    let ed25519_sk_bytes = client_ed25519_sk.to_bytes();
+    let ed25519_pk_bytes = client_ed25519_pk.to_bytes();
+    let ed25519_verification_key = Ed25519VerificationKey::from_bytes(ed25519_pk_bytes);
```
jstuczyn, line 261: could we perhaps include feature-locked conversions with the `From` trait from nym types into libcrux?

common/nym-lp/src/psk.rs
```diff
@@ -57,6 +114,349 @@ pub fn derive_psk(
...
+
+    // Use PSQ v1 API with Ed25519 authentication
+    let mut rng = rand09::rng();
+    let (state, initiator_msg) = Initiator::send_initial_message::<Ed25519, PsqX25519>(
+        session_context,
+        Duration::from_secs(3600), // 1 hour expiry
```
jstuczyn, line 267: why?

common/nym-lp/src/session.rs
```diff
@@ -6,19 +6,118 @@
...
+    InitiatorWaiting { ciphertext: Vec<u8> },
+
+    /// Responder is ready to receive and decapsulate PSQ ciphertext.
+    ResponderWaiting,
+
+    /// PSQ handshake completed successfully.
+    /// The PSK has been derived and registered with the Noise protocol.
+    Completed {
+        /// The derived post-quantum PSK
+        psk: [u8; 32],
```
jstuczyn, line 105: nit: I guess that shouldn't be exposed in the `Debug` impl

common/nym-lp/src/session.rs
```diff
@@ -77,11 +248,16 @@ impl LpSession {
...
 
-        let builder = builder.remote_public_key(remote_public_key);
+        let remote_key_bytes = remote_x25519_key.to_bytes();
+        let builder = builder.remote_public_key(&remote_key_bytes);
 
-        let builder = builder.psk(psk_index, psk);
+        // Initialize with dummy PSK - real PSK will be injected via set_psk() during handshake
+        // when PSQ runs using X25519 as DHKEM
+        let dummy_psk = [0u8; 32];
+        let builder = builder.psk(psk_index, &dummy_psk);
```
jstuczyn, line 260: rather than using a dummy value, could we perhaps be more explicit about it and wrap it within an `Option` to competely remove any chances of accidentally using it?

common/nym-lp/src/session.rs
```diff
@@ -91,12 +267,40 @@ impl LpSession {
...
             id,
             is_initiator,
             noise_state: Mutex::new(noise_protocol),
+            kkt_state: Mutex::new(kkt_state),
+            psq_state: Mutex::new(psq_state),
+            psk_handle: Mutex::new(None),
             sending_counter: AtomicU64::new(0),
             receiving_counter: Mutex::new(ReceivingKeyCounterValidator::default()),
+            psk_injected: AtomicBool::new(false),
+            // Ed25519 keys don't impl Clone, so convert to bytes and reconstruct
```
jstuczyn, line 290: that's very much by design. can't we instead have a single instance and share it via `Arc`s?

common/nym-lp/src/session.rs
```diff
@@ -91,12 +267,40 @@ impl LpSession {
...
+            local_ed25519_private: ed25519::PrivateKey::from_bytes(
+                &local_ed25519_keypair.0.to_bytes(),
```
jstuczyn, line 292: the ephemeral bytes created by `to_bytes()` call are not zeroized

jstuczyn:
```suggestion
                &Zeroizing::new(local_ed25519_keypair.0.to_bytes()),
```

common/nym-lp/src/session.rs
```diff
@@ -166,20 +370,325 @@ impl LpSession {
...
+            KKTState::InitiatorWaiting { context } => *context,
+            _ => {
+                return Err(LpError::Internal(
+                    "KKT response received in invalid state".to_string(),
+                ))
+            }
+        };
+
+        // Determine hash to validate against
+        let hash_for_validation: Vec<u8>;
```
jstuczyn, line 494: I didn't know rust allows you to declare an initialised value it looks odd to say the least and I think it could cause problems down the line. sure. it is only used in the `None` branch below, but why not instead just declare it there directly?

common/nym-lp/src/session.rs
```diff
@@ -6,19 +6,118 @@
...
+/// **Initiator path:**
+/// ```text
+/// NotStarted → InitiatorWaiting → Completed
+/// ```
+///
+/// **Responder path:**
+/// ```text
+/// NotStarted → ResponderProcessed
+/// ```
+pub enum KKTState {
```
jstuczyn, line 39: I wonder, would it be beneficial to have an explicit Error state?

common/nym-lp/src/session.rs
```diff
@@ -29,11 +128,71 @@ pub struct LpSession {
...
     noise_state: Mutex<NoiseProtocol>,
 
+    /// KKT (KEM Key Transfer) exchange state
+    kkt_state: Mutex<KKTState>,
+
+    /// PSQ (Post-Quantum Secure PSK) handshake state
+    psq_state: Mutex<PSQState>,
+
+    /// PSK handle from responder (ctxt_B) for future re-registration
+    psk_handle: Mutex<Option<Vec<u8>>>,
```
jstuczyn, line 138: it doesn't seem like all of those have to actually be behind mutexes, what about using a cheaper wrapper instead, like a `Cell`?

common/nym-lp/src/session.rs
```diff
@@ -166,20 +370,325 @@ impl LpSession {
...
+            let noise_msg = match noise_state.get_bytes_to_send() {
+                Some(Ok(msg)) => msg,
+                Some(Err(e)) => return Some(Err(LpError::NoiseError(e))),
+                None => return None, // Should not happen if is_my_turn, but handle gracefully
+            };
+
+            // Combine: [u16 psq_len][psq_payload][noise_msg]
+            let psq_len = psq_payload.len() as u16;
+            let mut combined = Vec::with_capacity(2 + psq_payload.len() + noise_msg.len());
+            combined.extend_from_slice(&psq_len.to_be_bytes());
```
jstuczyn: it seems we're being quite inconsistent, for example `session_context` was converted with `le_bytes`, yet here we're using `be_bytes` -> let's stick with a single convention

common/nym-lp/src/session.rs
```diff
@@ -192,23 +701,157 @@ impl LpSession {
...
+                            return noise_state
+                                .read_message(noise_payload)
+                                .map_err(LpError::NoiseError);
+                        }
+                    }
+                    // If no valid handle found, fall through to normal processing
+                }
+
+                // Normal flow (no PSQ, or PSQ already completed)
+                drop(psq_state); // Release lock
```
jstuczyn: isn't this quite redundant given the lock will be released once this function exits (which is literally after the next call) and `noise_state.read_message()` does not read psq state?

jstuczyn (changes requested):
few additional comments that didn't apply to any single line of code:

1) the following packages:
- `nym-kcp`
- `nym-lp`
- `nym-lp-common`
have an empty license field (i.e. `lience = ""`) thus `cargo deny check licenses` fails

2) there are some clippy issues that need to be resolved before merging
3) there seem to be an issue with the chosen version of `sha2` crate making contract builds fail - perhaps we need to upgrade/downgrade it somewhere

common/nym-lp/Cargo.toml
durch: Yeah, libcrux uses 0.9.2, and we import stuff from there, so its a whole chain

common/nym-lp/src/session.rs
durch: We could, but it would change the API, and I don't think we'd gain much

common/nym-lp/src/session.rs
durch: Does not live long enough if constructed in the branch, so its either this or a mut.

common/nym-lp/src/message.rs
durch: Found a num_enum crate which is exactly what we need here

common/nym-lp/src/session.rs
durch: We immediatly do something else on errors, so we'd not be in the state long...

PR #6213, opened 2025-11-17
chore: don't rederive wallet keys on every tx

jstuczyn:
A small change I've been meaning to do for a while now. the cosmrs' trait bounds on `EcdsaSigner` got updated to include `Send` and `Sync`, meaning we no longer need to derive private keys on every transaction and instead we can just do it once, on construction.

note that I had to deprecate some builder methods as they're no longer safe (i.e. they can possibly panic). They were not removed so it wouldn't break compatibility with the library consumers (_looking at VPN-core_)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6213)

### DIFF
--- a/common/client-libs/validator-client/src/nyxd/mod.rs
+++ b/common/client-libs/validator-client/src/nyxd/mod.rs
@@ -19,7 +19,7 @@ use crate::signing::signer::NoSigner;
 use crate::signing::signer::OfflineSigner;
 use crate::signing::tx_signer::TxSigner;
 use crate::signing::AccountData;
-use crate::{DirectSigningReqwestRpcNyxdClient, QueryReqwestRpcNyxdClient, ReqwestRpcClient};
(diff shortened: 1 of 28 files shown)


common/client-libs/validator-client/src/nyxd/mod.rs
```diff
@@ -158,12 +158,13 @@ impl NyxdClient<HttpClient> {
     }
 }
 
-impl NyxdClient<ReqwestRpcClient> {
+#[allow(deprecated)]
```
simonwicky, line 161: I guess you know what you're doing but why the deprecated `fn` use? Are there no replacements?

simonwicky (approved):

common/client-libs/validator-client/src/nyxd/mod.rs
jstuczyn: resolved on zulip : )

PR #6223, opened 2025-11-20
Granular controls and bug fixes for internal DNS

jmwample:
See NET-713 & NET-742

## Changes

* Add controls to allow a configuration change that prevents hickory from using IPv6 nameservers.

  * When IPv6 is disabled, the internal resolver should not use IPv6 resolvers.
  * Using these controls clears the cache of the internal resolver on change

  ```rust
  use nym_http_api_client::dns::{HickoryDnsResolver, HostnameIpLookupStrategy, NameServerIpVersionPolicy};
  
  let mut resolver = HickoryDnsResolver::default();
  
  // Configure the resolver to use only Ipv4 nameservers and to resolve
  // addresses to Ipv4 (A)  records only.
  resolver.set_ipv4_only();

  // More explicit controls for setting the nameserver policy for IP version and lookup target
  resolver.set_hostname_ip_version_lookup_strategy(HostnameIpLookupStrategy::A_then_AAAA);
  resolver.set_nameserver_ip_version_strategy(NameServerIpVersionPolicy::Ipv4AndIpv6);
  ```

* Add a toggle that allows the order in which the static table is checked to be configured by the downstream user of the library. The behavior of the shared resolver can be modified as well using this control.
  ```rust
  let mut resolver = HickoryDnsResolver::default();
  // tell the resolver to check the static table before using the internal resolver. 
  resolver.set_check_static_fallback_first(true);
  ```

* Add option to trial the DNS nameservers and rebuild the resolver to use only working nameservers.  [#6228](https://github.com/nymtech/nym/pull/6228)

  ```rust
  impl HickoryDnsResolver {
    /// Do a trial resolution using each nameserver individually to test which are working and which
    /// fail to complete a lookup.
    pub async fn trial_nameservers(&self)

    /// Do a trial resolution using each nameserver individually to test which are working and which
    /// fail to complete a lookup. If one or more of the resolutions succeeds, rebuild the resolver
    /// using only the nameservers that successfully completed the lookup.
    pub async fn trial_nameservers_and_reconfigure(&mut self) -> Result<(), ResolveError>
    // ...
  }
  ```

  ```rust
  let resolver = HickoryDnsResolver::default();
  resolver.trial_nameservers_and_reconfigure().await.unwrap();
  ```

* QoL features for using the `HickoryDnsResolver` directly.  [#6263](https://github.com/nymtech/nym/pull/6263) for details.
  ```rust
  impl HickoryDnsResolver {
    /// Get the list of currently available nameserver configs.
    pub fn all_configured_name_servers(&self) -> Vec<NameServerConfig> 
    /// Get the list of currently used nameserver configs. 
    pub fn active_name_servers(&self) -> Vec<NameServerConfig>
    /// Sets the available nameservers for use by this resolver to the provided list
    pub fn set_name_servers(&mut self, name_servers: impl AsRef<[NameServerConfig]>)
    // ...
  }
  ```

* Initialize that static table uesd by the (default) shared resolver properly

### IPv6

In the VPN client there are settings specifically intended to control how the library interacts with the IPv6 network stack. This is because some platforms do weird things and actually throw errors if IPv6 is used, so preventing the client from attempting IPv6 can be important in some cases.

This is a challenge with the custom DNS resolver as the [hickory-resolver](https://docs.rs/hickory-resolver/latest/hickory_resolver/) crate that we use provides no controls to indicate that an outgoing lookup should use only IPv4 nameservers or only IPv6 (or both). 

Also there is an issue of randomness even when IPv6 doesn’t cause explicit problems where the resolver selecting nameservers to do a resolution could select IPv6 nameservers (because they are configured) even when the IPv6 network will be unreachable - i.e. the networking stack supports IPv6 but the device isn’t connection with an IPv6 address.

### Static table check

The static lookup fallback added in https://github.com/nymtech/nym/pull/6178 checks the configured table after the inner resolver fails, however this has the issue that you will likely already have waited for the timeout making the subsequent use for the addresses also likely to timeout.

We can change this so that the check in the static table is done first, but the static addresses are currently VERY static, and making all clients relient on them not changing in order to access things like the API makes me nervous.

### Trial Based Reconfigure

Adds a mechanism for a user of the internal DNS resolver to trigger an asynchronous test of the internal nameservers. This can be done as for debugging purposes if (for example) the vpn gets into a bad state and we want to gather baseline information to asses wether DNS accessibility is the root cause.
Extending this, the caller can use the results of the test to reconfigure the resolver to use only the nameservers that successfully completed the lookup. This is done asynchronously and (likely) requires the full lookup timeout to allow non-functional resolvers to fail the lookup before reconfiguring.

This is meant to short-cut the https://github.com/hickory-dns/hickory-dns/issues/1702 process.

This could also be used as an early test to check that at least one of the configured nameservers can succeed before we move forward with VPN connections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6223)


jmwample:
Hmmm. Some of the DNS related tests seem either flakey or like they are interleaving in a way that is doing bad things to the shared resolver.

jmwample:
After changing the default to use IPv4 Only nameservers the errors in the tests run as part of the github actions have gone away. However, I do not fully understand the errors so I can't say for sure that this fixed the issue and that we didn't just get lucky this time. 

It seems like when the resolver chooses IPv6 in the tests in the gh action runner it times out instead of returning a `network unreachable` error. Which then causes other things to time out. which breaks the things I am actually trying to test.

If this is the case though, how did we not run into this issue before?

jmwample:
closing this PR as the stable elements of the PR are included in # 6267 (#6279 for develop). The internal mutability features create unnecessary challenges for now.

PR #6224, opened 2025-11-21
Add peer-controller timeouts and lock tweaks

tommyv1987:
Ad a 5 s timeout around every PeerManager call into the WireGuard peer controller (recv_with_timeout). Instead of blocking forever when the controller is stuck after suspend, the authenticator now surfaces PeerInteractionTimeout and keeps processing other clients.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6224)


gateway/src/node/internal_service_providers/authenticator/mixnet_listener.rs
```diff
@@ -383,27 +394,33 @@ impl MixnetListener {
...
-            self.keypair().private_key(),
-            remote_public.inner(),
-            *private_ip_ref.0,
-            nonce,
-        );
-        let registration_data = latest::registration::RegistrationData {
-            nonce,
-            gateway_data: gateway_data.clone(),
-            wg_port: self.config.authenticator.tunnel_announced_port,
+        let (registration_data, private_ips) = {
```
neacsu: I don't understand why is it needed to create them via a tuple instead of each one individually, as previously. Not against it, but I don't see any difference.

gateway/src/node/internal_service_providers/authenticator/config/mod.rs
```diff
@@ -57,6 +57,10 @@ pub struct Authenticator {
     /// The prefix denoting the maximum number of the clients that can be connected via Wireguard using IPv6.
     /// The maximum value for IPv6 is 128
     pub private_network_prefix_v6: u8,
+
+    /// Timeout (in milliseconds) to wait for responses from the peer controller before failing.
+    /// Helps the authenticator recover from suspend/resume scenarios where peer RPCs hang.
+    pub peer_interaction_timeout_ms: u64,
```
neacsu: You could probably put a `Duration` here directly, to not have to specify that it's a number representing milliseconds and later to transform it correctly in that `Duration`

gateway/src/node/internal_service_providers/authenticator/mod.rs
```diff
@@ -151,13 +151,15 @@ impl Authenticator {
                 }
             })
             .collect();
+        let peer_timeout = std::cmp::max(1, self.config.authenticator.peer_interaction_timeout_ms);
```
neacsu: Hmm. Why? Why not just set the value directly?

nym-node/src/config/mod.rs
```diff
@@ -944,6 +945,7 @@ pub struct Wireguard {
 
     /// Tunnel port announced to external clients wishing to connect to the wireguard interface.
     /// Useful in the instances where the node is behind a proxy.
+    #[serde(alias = "announced_port")]
```
neacsu, line 947: This seems like it should be part of a separate PR, if it's a fix for something.

nym-node/src/config/old_configs/old_config_v10.rs
```diff
@@ -63,6 +66,7 @@ pub struct WireguardV10 {
 
     /// Port announced to external clients wishing to connect to the wireguard interface.
     /// Useful in the instances where the node is behind a proxy.
+    #[serde(alias = "announced_tunnel_port")]
```
neacsu, line 66: This seems like it should be part of a separate PR, if it's a fix for something.

nym-node/src/config/old_configs/old_config_v10.rs
```diff
@@ -75,6 +79,11 @@ pub struct WireguardV10 {
 
     /// Paths for wireguard keys, client registries, etc.
     pub storage_paths: WireguardPathsV10,
+
+    /// Optional override for the peer interaction timeout expressed in milliseconds.
+    /// Accepts either a plain integer value or a humantime string such as "5s".
+    #[serde(default, deserialize_with = "deserialize_optional_duration_ms")]
+    pub peer_interaction_timeout_ms: Option<u64>,
```
neacsu: Those `old_config_v*` files are not really to be modified this way. They show what was in a previous released version, so they are immutable. If you want to add something, you'll need to create a new `old_config_v` file with the current structures and modify the `mod` structures to add what you want. Pining @jstuczyn just in case that has changed

nym-node/src/config/template.rs
```diff
@@ -151,6 +151,10 @@ announced_tunnel_port = {{ wireguard.announced_tunnel_port }}
 # Useful in the instances where the node is behind a proxy.
 announced_metadata_port = {{ wireguard.announced_metadata_port }}
 
+# Timeout to wait for responses from the peer controller before aborting the operation.
+# Accepts standard humantime duration strings such as `200ms`, `5s`.
+peer_interaction_timeout_ms = '{{ wireguard.peer_interaction_timeout_ms }}'
```
neacsu: I'm not so sure this should be made into a configurable value. If someone sets here a value that's too small, it could wrongfully error out with timeout and corrupt the peer tables, especially for the `add_peer` operations.

neacsu (changes requested):
> Instead of blocking forever when the controller is stuck after suspend, the authenticator now surfaces PeerInteractionTimeout and keeps processing other clients.

What suspend are you referring to exactly? I'm curious about the logs that caused this bug, could you reference them in this PR as well please?

Also, if the controller is stuck for some reason, there's not really much reason to process other clients on the mixnet listener, because there won't be any working controller that can process the requests. In that case, we should probably just stop the whole authenticator, because otherwise we would just get continuous error logs.

gateway/src/node/internal_service_providers/authenticator/mod.rs
tommyv1987: Done 👍

nym-node/src/config/mod.rs
tommyv1987: Removed from here as you rightly mentioned

nym-node/src/config/template.rs
tommyv1987: Removed as its not needed

nym-node/src/config/old_configs/old_config_v10.rs
tommyv1987: Removed from this PR so resolving these

nym-node/src/config/old_configs/old_config_v10.rs
tommyv1987: Removed :)

gateway/src/node/internal_service_providers/authenticator/config/mod.rs
tommyv1987: Done 👍

tommyv1987:
> > Instead of blocking forever when the controller is stuck after suspend, the authenticator now surfaces PeerInteractionTimeout and keeps processing other clients.
> 
> What suspend are you referring to exactly? I'm curious about the logs that caused this bug, could you reference them in this PR as well please?
> 
> Also, if the controller is stuck for some reason, there's not really much reason to process other clients on the mixnet listener, because there won't be any working controller that can process the requests. In that case, we should probably just stop the whole authenticator, because otherwise we would just get continuous error logs.

Added comments in the code explaining this. We don't have specific log snippets in this PR, but I'll try to source them again from testing.

Stopping the authenticator - Agreed. This PR adds the timeout to prevent blocking. Failure tracking and shutdown on repeated failures.

A follow up PR  is probably needed to address the shutdown of the authenticator.  A good thing though is that the timeout logs warnings so operators can see the issue.

PR #6237, opened 2025-11-26
Initial changes to support extra configurable parameters and to print…

nick1231321:
This branch adds support for the additional configurable parameters introduced in  
`nicolas/sdk-param-support-debug` in the nym vpn client branch and also debugging messages to verify that it works. 

## Ticket
[JIRA-VPN-4558](https://nymtech.atlassian.net/browse/VPN-4558)

Further details can be found in [**Configurable Mixnet Params – Design**](https://nymtech.atlassian.net/wiki/x/nACJDw).

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6237)

### DIFF
--- a/nym-registration-client/src/builder/config.rs
+++ b/nym-registration-client/src/builder/config.rs
@@ -217,21 +217,22 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
     // 2. The loop cover traffic average delay is set to 0ms (special value meaning "disable")
     let mut disable_loop_cover_stream = mixnet_client_config.disable_background_cover_traffic;
 
-    if let Some(delay) = mixnet_client_config.loop_cover_traffic_average_delay {
-        if delay.is_zero() {
-            disable_loop_cover_stream = true;
-        }
+    if let Some(delay) = mixnet_client_config.loop_cover_traffic_average_delay
+        && delay.is_zero()
+    {
+        disable_loop_cover_stream = true;
     }
 
     debug_config.cover_traffic.disable_loop_cover_traffic_stream = disable_loop_cover_stream;
 
     // Only apply a custom loop cover delay if the stream is enabled.
     // If the stream is disabled, the delay value is ignored.
-    if !disable_loop_cover_stream {
-        if let Some(delay) = mixnet_client_config.loop_cover_traffic_average_delay {
-            debug_config.cover_traffic.loop_cover_traffic_average_delay = delay;
-        }
+    if !disable_loop_cover_stream
+        && let Some(delay) = mixnet_client_config.loop_cover_traffic_average_delay
+    {
+        debug_config.cover_traffic.loop_cover_traffic_average_delay = delay;
     }
+
     log_mixnet_client_config(&debug_config);
     debug_config
 }
--- a/nym-registration-client/src/builder/config.rs
+++ b/nym-registration-client/src/builder/config.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
-
+ 
 use nym_credential_storage::persistent_storage::PersistentStorage;
 use nym_registration_common::NymNode;
 use nym_sdk::{
@@ -10,24 +10,24 @@ use nym_sdk::{
         MixnetClientStorage, OnDiskPersistent, ReplyStorageBackend, StoragePaths, x25519::KeyPair,
     },
 };
-
+ 
 #[cfg(unix)]
 use std::os::fd::RawFd;
 use std::{path::PathBuf, sync::Arc, time::Duration};
 use tokio_util::sync::CancellationToken;
 use typed_builder::TypedBuilder;
-
+ 
 use crate::error::RegistrationClientError;
-
+ 
 const VPN_AVERAGE_PACKET_DELAY: Duration = Duration::from_millis(15);
 const MIXNET_CLIENT_STARTUP_TIMEOUT: Duration = Duration::from_secs(30);
-
+ 
 #[derive(Clone)]
 pub struct NymNodeWithKeys {
     pub node: NymNode,
     pub keys: Arc<KeyPair>,
 }
-
+ 
 #[derive(TypedBuilder)]
 pub struct BuilderConfig {
     pub entry_node: NymNodeWithKeys,
@@ -44,27 +44,29 @@ pub struct BuilderConfig {
     #[cfg(unix)]
     pub connection_fd_callback: Arc<dyn Fn(RawFd) + Send + Sync>,
 }
-
+ 
 #[derive(Clone, Default, Debug, Eq, PartialEq)]
 pub struct MixnetClientConfig {
     /// Disable Poission process rate limiting of outbound traffic.
-    pub disable_poisson_rate: bool,
-
+    pub disable_real_traffic_poisson_process: bool,
+ 
     /// Disable constant rate background loop cover traffic
     pub disable_background_cover_traffic: bool,
-
+ 
     /// The minimum performance of mixnodes to use.
     pub min_mixnode_performance: Option<u8>,
-
+ 
     /// The minimum performance of gateways to use.
     pub min_gateway_performance: Option<u8>,
+ 
     ///Setting optionally the poisson rate for cover traffic stream
-    pub poisson_rate: Option<u32>,
+    pub loop_cover_traffic_average_delay: Option<Duration>,
+ 
     /// Average packet delay in milliseconds.
-    pub average_packet_delay: Option<u32>,
-
+    pub average_packet_delay: Option<Duration>,
+ 
     /// Average message sending delay in milliseconds.
-    pub message_sending_average_delay: Option<u32>,
+    pub message_sending_average_delay: Option<Duration>,
 }
 impl BuilderConfig {
     pub fn mixnet_client_debug_config(&self) -> DebugConfig {
@@ -74,16 +76,16 @@ impl BuilderConfig {
             mixnet_debug_config(&self.mixnet_client_config)
         }
     }
-
+ 
     pub async fn setup_storage(
         &self,
     ) -> Result<Option<(OnDiskPersistent, PersistentStorage)>, RegistrationClientError> {
         if let Some(path) = &self.data_path {
             tracing::debug!("Using custom key storage path: {}", path.display());
-
+ 
             let storage_paths = StoragePaths::new_from_dir(path)
                 .map_err(|err| RegistrationClientError::BuildMixnetClient(Box::new(err)))?;
-
+ 
             let mixnet_client_storage = storage_paths
                 .initialise_persistent_storage(&self.mixnet_client_debug_config())
                 .await
@@ -92,13 +94,13 @@ impl BuilderConfig {
                 .persistent_credential_storage()
                 .await
                 .map_err(|err| RegistrationClientError::BuildMixnetClient(Box::new(err)))?;
-
+ 
             Ok(Some((mixnet_client_storage, credential_storage)))
         } else {
             Ok(None)
         }
     }
-
+ 
     pub async fn build_and_connect_mixnet_client<S>(
         self,
         builder: MixnetClientBuilder<S>,
@@ -118,7 +120,7 @@ impl BuilderConfig {
         } else {
             RememberMe::new_mixnet()
         };
-
+ 
         let builder = builder
             .with_user_agent(self.user_agent)
             .request_gateway(self.entry_node.node.identity.to_string())
@@ -128,10 +130,10 @@ impl BuilderConfig {
             .no_hostname(true)
             .with_remember_me(remember_me)
             .custom_topology_provider(self.custom_topology_provider);
-
+ 
         #[cfg(unix)]
         let builder = builder.with_connection_fd_callback(self.connection_fd_callback);
-
+ 
         builder
             .build()
             .map_err(|err| RegistrationClientError::BuildMixnetClient(Box::new(err)))?
@@ -140,12 +142,12 @@ impl BuilderConfig {
             .map_err(|err| RegistrationClientError::ConnectToMixnet(Box::new(err)))
     }
 }
-
+ 
 fn two_hop_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig {
     let mut debug_config = DebugConfig::default();
-
+ 
     debug_config.traffic.average_packet_delay = VPN_AVERAGE_PACKET_DELAY;
-
+ 
     // We disable mix hops for the mixnet connection.
     debug_config.traffic.disable_mix_hops = true;
     // Always disable poisson process for outbound traffic in wireguard.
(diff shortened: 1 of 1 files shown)


common/client-core/src/client/real_messages_control/mod.rs
```diff
@@ -205,7 +205,6 @@ impl RealMessagesController<OsRng> {
             stats_tx,
             shutdown_token.clone(),
         );
-
         RealMessagesController {
```
simonwicky, line 209: I guess there was a `println` there? 
No harm in that but the whole change can be discarded, please undo that line deletion

common/client-core/src/client/real_messages_control/real_traffic_stream.rs
```diff
@@ -248,7 +248,6 @@ where
                         return;
                     }
                 };
-
                 // SAFETY: our topology must be valid at this point
```
simonwicky, line 252: Same comment on line deletion

common/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -179,15 +178,14 @@ impl LoopCoverTrafficStream<OsRng> {
...
-
         let cover_message = match generate_loop_cover_packet(
             &mut self.rng,
             self.use_legacy_sphinx_format,
             topology_ref,
             &self.ack_key,
             &self.our_full_destination,
             self.average_ack_delay,
-            self.cover_traffic.loop_cover_traffic_average_delay,
+            self.average_ack_delay,
```
simonwicky: I'm pretty sure that's wrong. Why the change?

common/client-core/src/client/cover_traffic_stream.rs
```diff
@@ -244,14 +236,12 @@ impl LoopCoverTrafficStream<OsRng> {
...
         }
-
         // we should set initial delay only when we actually start the stream
         let sampled = sample_poisson_duration(
             &mut self.rng,
             self.cover_traffic.loop_cover_traffic_average_delay,
         );
         self.set_next_delay(sampled);
-
         while self.next().await.is_some() {
```
simonwicky, line 259: Same comment on line deletion ;)

nym-registration-client/src/builder/config.rs
```diff
@@ -45,7 +41,7 @@ pub struct BuilderConfig {
     pub connection_fd_callback: Arc<dyn Fn(RawFd) + Send + Sync>,
 }
 
-#[derive(Clone, Default, Debug, Eq, PartialEq)]
+#[derive(Clone, Default, Debug, PartialEq)]
 pub struct MixnetClientConfig {
```
simonwicky, line 49: Why is `Eq` gone? Is it failing to derive?

nym-registration-client/src/builder/config.rs
```diff
@@ -58,6 +54,13 @@ pub struct MixnetClientConfig {
...
     /// The minimum performance of gateways to use.
     pub min_gateway_performance: Option<u8>,
+    ///Setting optionally the poisson rate for cover traffic stream
+    pub poisson_rate: Option<u32>,
+    /// Average packet delay in milliseconds.
+    pub average_packet_delay: Option<u32>,
+
+    /// Average message sending delay in milliseconds.
+    pub message_sending_average_delay: Option<u32>,
 }
```
simonwicky, line 70: Keep the same formatting as above, with the extra line between a variable and the next comment. 
Also the `poisson_rate` variable isn't very explicit, how about `loop_cover_traffic_average_delay` ?
If you need duration later, why not taking it as duration directly?

nym-registration-client/src/builder/config.rs
```diff
@@ -155,7 +158,15 @@ fn two_hop_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfi
...
     }
+    if let Some(avg_packet_ms) = mixnet_client_config.average_packet_delay {
+        debug_config.traffic.average_packet_delay = Duration::from_millis(avg_packet_ms as u64);
+        debug_config.acknowledgements.average_ack_delay = Duration::from_millis(avg_packet_ms as u64);
 
+    }
+
+    if let Some(msg_delay_ms) = mixnet_client_config.message_sending_average_delay {
+        debug_config.traffic.message_sending_average_delay = Duration::from_millis(msg_delay_ms as u64);
+    }
```
simonwicky: Two hop mode has the main poisson process disabled. This parameter is ignored anyway.

nym-registration-client/src/builder/config.rs
```diff
@@ -155,7 +158,15 @@ fn two_hop_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfi
     if let Some(min_gateway_performance) = mixnet_client_config.min_gateway_performance {
         debug_config.topology.minimum_gateway_performance = min_gateway_performance;
     }
+    if let Some(avg_packet_ms) = mixnet_client_config.average_packet_delay {
+        debug_config.traffic.average_packet_delay = Duration::from_millis(avg_packet_ms as u64);
+        debug_config.acknowledgements.average_ack_delay = Duration::from_millis(avg_packet_ms as u64);
 
+    }
```
simonwicky: @aniampio do we want to change the packet delay for two-hop mode. Does it make sense?
If yes, set those parameters as `Option<Duration>` directly in the `MixnetClientConfig`

nym-registration-client/src/builder/config.rs
```diff
@@ -170,14 +181,26 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
 
     debug_config.cover_traffic.disable_loop_cover_traffic_stream =
         mixnet_client_config.disable_background_cover_traffic;
+    if let Some(poisson_rate) = mixnet_client_config.poisson_rate {
+            let duration = std::time::Duration::from_millis((poisson_rate as f64).round() as u64);
+        debug_config.cover_traffic.loop_cover_traffic_average_delay = duration;
```
simonwicky: What is that type gymnastic? Take the parameters as Duration in the config please

nym-registration-client/src/builder/config.rs
```diff
@@ -170,14 +181,26 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
     }
+    if let Some(avg_packet_ms) = mixnet_client_config.average_packet_delay {
+        debug_config.traffic.average_packet_delay = Duration::from_millis(avg_packet_ms as u64);
+    debug_config.acknowledgements.average_ack_delay=Duration::from_millis(avg_packet_ms as u64);
+    }
+
+    if let Some(msg_delay_ms) = mixnet_client_config.message_sending_average_delay {
+        debug_config.traffic.message_sending_average_delay = Duration::from_millis(msg_delay_ms as u64);
+    }
     log_mixnet_client_config(&debug_config);
```
simonwicky, line 212: Please add those new options to `log_mixnet_client_config` so we can see what's happening?

nym-registration-client/src/builder/config.rs
```diff
@@ -170,14 +181,26 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
     if let Some(min_mixnode_performance) = mixnet_client_config.min_mixnode_performance {
         debug_config.topology.minimum_mixnode_performance = min_mixnode_performance;
     }
 
     if let Some(min_gateway_performance) = mixnet_client_config.min_gateway_performance {
         debug_config.topology.minimum_gateway_performance = min_gateway_performance;
     }
+    if let Some(avg_packet_ms) = mixnet_client_config.average_packet_delay {
+        debug_config.traffic.average_packet_delay = Duration::from_millis(avg_packet_ms as u64);
+    debug_config.acknowledgements.average_ack_delay=Duration::from_millis(avg_packet_ms as u64);
```
simonwicky: Same comments as above for durations

.cargo-ok
simonwicky, line 1: What is that file doing here?

simonwicky (changes requested):
In general, double check the changes you have in your PR. I usually do a soft reset to the latest `develop` to go through all of them. It will allow you to catch those line deletion for example.

For formatting, setup your IDE to format (I have `format on save` with `rust-analyser` on VSCode), we can help you with other setups if needed. If you don't want to/can't do that, run `cargo fmt --all` to format everything.

nym-registration-client/src/builder/config.rs
```diff
@@ -170,14 +181,26 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+    }
+
+    if let Some(msg_delay_ms) = mixnet_client_config.message_sending_average_delay {
+        debug_config.traffic.message_sending_average_delay = Duration::from_millis(msg_delay_ms as u64);
```
simonwicky: We need a bound check on that too. Because if this `0`, we will exhaust the CPU by sending with everything we have. 
Either don't accept `0`, or disable the poisson process if you have `0` here

simonwicky (changes requested):

common/client-core/src/client/cover_traffic_stream.rs
simonwicky: After internal discussions, indeed there is a bug here. Let's tackle this in a different PR though

nym-registration-client/src/builder/config.rs
nick1231321: Fixed

nym-registration-client/src/builder/config.rs
nick1231321: Fixed

nym-registration-client/src/builder/config.rs
nick1231321: Fixed

nym-registration-client/src/builder/config.rs
```diff
@@ -58,8 +58,16 @@ pub struct MixnetClientConfig {
 
     /// The minimum performance of gateways to use.
     pub min_gateway_performance: Option<u8>,
-}
 
+    ///Setting optionally the poisson rate for cover traffic stream
```
simonwicky: nit : space before comment
```suggestion
    /// Setting optionally the poisson rate for cover traffic stream
```

nym-registration-client/src/builder/config.rs
```diff
@@ -179,6 +195,30 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+        debug_config.acknowledgements.average_ack_delay = avg_packet_ms;
+    }
+
+    if let Some(msg_delay_ms) = mixnet_client_config.message_sending_average_delay {
+        if msg_delay_ms.is_zero() {
+            debug_config
+                .traffic
+                .disable_main_poisson_packet_distribution = true;
+        } else {
+            debug_config.traffic.message_sending_average_delay = msg_delay_ms;
```
simonwicky: That makes weird interaction if somehow we set `disable_real_traffic_poisson_process = true` and `message_sending_average_delay = Some(50)`, where the stream ends up not disabled. How about merging the setting for both loops?

If delay is set to zero, disable the loop, if it's set to something else, change the delay, if it's not set, default values?

nym-registration-client/src/builder/config.rs
```diff
@@ -207,6 +247,36 @@ fn log_mixnet_client_config(debug_config: &DebugConfig) {
         "mixnet client minimum gateway performance: {}",
         debug_config.topology.minimum_gateway_performance,
     );
+
+    tracing::info!(
+        "mixnet client loop cover traffic average delay: {} ms",
```
simonwicky: I'd say display those only if the loop isn't disabled to avoid confusion

nym-registration-client/src/builder/config.rs
```diff
@@ -207,6 +247,36 @@ fn log_mixnet_client_config(debug_config: &DebugConfig) {
...
+            .as_millis()
+    );
+
+    tracing::info!(
+        "mixnet client average packet delay: {} ms",
+        debug_config.traffic.average_packet_delay.as_millis()
+    );
+
+    tracing::info!(
+        "mixnet client message sending average delay: {} ms",
```
simonwicky: same here

simonwicky (changes requested):
The open comments still holds

nym-registration-client/src/builder/config.rs
aniampio: @simonwicky So the 2-hop mode does not do any mixing (since it is WG), but I assume that you mean for the registration client, which relies on the mixnet client. As far as I remember (and please @jmwample correct me if I'm wrong) - we changed it, and the mixnet client used in the registration actually doesn't use any mixnodes, hence this parameter doesn't have any impact. Regardless, I would not change anything about it. 

If you mean something totally different, please let me know! :)

nym-registration-client/src/builder/config.rs
nick1231321: Fixed

nym-registration-client/src/builder/config.rs
nick1231321: fixed

nym-registration-client/src/builder/config.rs
nick1231321: fixed

nym-registration-client/src/builder/config.rs
nick1231321: fixed

nym-registration-client/src/builder/config.rs
```diff
@@ -207,6 +262,40 @@ fn log_mixnet_client_config(debug_config: &DebugConfig) {
...
+    {
+        tracing::info!(
+            "mixnet client message sending average delay: {} ms",
+            debug_config
+                .traffic
+                .message_sending_average_delay
+                .as_millis()
+        );
+    }
+    tracing::info!(
```
simonwicky: That's already logged at the fn start, no need to log it again. The rest is good

nym-registration-client/src/builder/config.rs
```diff
@@ -179,6 +181,59 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+    // We disable it if:
+    // 1. The user explicitly requested it via `disable_real_traffic_poisson_process`, OR
+    // 2. The user set `message_sending_average_delay` to 0ms (special value meaning "disable")
+    let mut disable_real_traffic_poisson =
+        mixnet_client_config.disable_real_traffic_poisson_process;
+
+    if let Some(delay) = mixnet_client_config.message_sending_average_delay
+        && delay.is_zero()
+    {
+        disable_real_traffic_poisson = true;
```
simonwicky: nit: How about :
```suggestion
    debug_config
        .traffic
        .disable_main_poisson_packet_distribution =
        mixnet_client_config.disable_real_traffic_poisson_process;
    if let Some(delay) = mixnet_client_config.message_sending_average_delay
        && delay.is_zero()
    {
        debug_config
            .traffic
            .disable_main_poisson_packet_distribution = true;
    }
```

instead of the whole block. It is correct but we don't really need the intermediate mutable state

nym-registration-client/src/builder/config.rs
```diff
@@ -179,6 +181,59 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+    {
+        debug_config.traffic.message_sending_average_delay = delay;
+    }
+
+    // Decide whether the loop cover traffic stream should be disabled.
+    //
+    // We disable it if:
+    // 1. The user explicitly disabled background cover traffic, OR
+    // 2. The loop cover traffic average delay is set to 0ms (special value meaning "disable")
+    let mut disable_loop_cover_stream = mixnet_client_config.disable_background_cover_traffic;
```
simonwicky: Same as comment with the suggestion above

nym-registration-client/src/builder/config.rs
```diff
@@ -179,6 +181,59 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+        disable_real_traffic_poisson = true;
+    }
+
+    debug_config
+        .traffic
+        .disable_main_poisson_packet_distribution = disable_real_traffic_poisson;
+
+    // Only apply a custom message sending delay if the Poisson process is enabled.
+    // If Poisson is disabled, any delay value is ignored.
+    if !disable_real_traffic_poisson
```
simonwicky: nit: It's correct, but we don't really need that extra condition on `disable_real_traffic_poisson` imo. If it's disabled, it will get ignored anyway

nym-registration-client/src/builder/config.rs
```diff
@@ -179,6 +181,59 @@ fn mixnet_debug_config(mixnet_client_config: &MixnetClientConfig) -> DebugConfig
...
+        && delay.is_zero()
+    {
+        disable_loop_cover_stream = true;
+    }
+
+    debug_config.cover_traffic.disable_loop_cover_traffic_stream = disable_loop_cover_stream;
+
+    // Only apply a custom loop cover delay if the stream is enabled.
+    // If the stream is disabled, the delay value is ignored.
+    if !disable_loop_cover_stream
```
simonwicky: Same as comment above

simonwicky (approved):
Some duplicate log to remove and some empty lines that needs to stay there ;)
Once that's done, ship it.

Your call on the couple of nits

nym-registration-client/src/builder/config.rs
nick1231321: Ok so i assign it without checking anything

PR #6238, opened 2025-11-26
fix: fix assertion

YichiZhang0613:
The inconsistency among comment, code and assertion cause confusion here, which should be fixed.
The assumption("ASSUMPTION: our validity period is AT LEAST 2 days") indicates that TICKETBOOK_VALIDITY_DAYS >= 2 days, which indicates TICKETBOOK_VALIDITY_DAYS - 1 >= 1 day. So target_secs should be >= 86400.
The comment will cause the confusion("making sure it's no greater than 1 day"). It indicate that it should be <= 86400, which is wrong semantically and should be changed it into "no lower than 1 day". 
The assertion message requires target_secs >= 86400("no lower than"), while the predicate is target_secs > 86400, so I changed it into `assert!(target_secs >= 86400,"the maximum time between redemption can't be lower than 1 day!");`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6238)

### DIFF
--- a/nym-node/src/config/gateway_tasks.rs
+++ b/nym-node/src/config/gateway_tasks.rs
@@ -133,7 +133,7 @@ impl ZkNymTicketHandlerDebug {
     pub const DEFAULT_MINIMUM_API_QUORUM: f32 = 0.7;
     pub const DEFAULT_MINIMUM_REDEMPTION_TICKETS: usize = 100;
 
-    // use min(4/5 of max validity, validity - 1), but making sure it's no greater than 1 day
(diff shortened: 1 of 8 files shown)


YichiZhang0613:
@pronebird @jstuczyn @jmwample Hi, sorry for the interruption. I see you've been active reviewers recently. Could someone please review this PR when you have time? It's been open for two weeks and I'd appreciate any feedback. Thank you very much.

jstuczyn (approved):

PR #6240, opened 2025-11-26
LP reg telescoping

durch:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6240)


common/nym-lp/README.md
```diff
@@ -1,71 +1,309 @@
...
+
+```
+TCP Connect → Receive Packet → Process → Send Response → TCP Close
+```
+
+**State Storage:**
+- `handshake_states`: Maps `session_id → LpStateMachine` (during handshake)
+- `session_states`: Maps `session_id → LpSession` (after handshake complete)
+
+Both maps use TTL-based cleanup to remove stale entries (default: 5 min handshake, 1 hour session).
```
jstuczyn, line 129: I hope that's not unconditional because sessions can easily last more than 1h

common/nym-lp/README.md
```diff
@@ -1,71 +1,309 @@
...
+```
+┌──────────┐
+│  Client  │
+└────┬─────┘
+     │ Outer LP Session (established, encrypted)
+     │
+     ▼
+┌────────────────┐
+│ Entry Gateway  │  Sees: Client IP
+│                │  Doesn't see: Exit destination
```
jstuczyn, line 164: but it must see exit destination, otherwise how would it forward the packet?

common/nym-lp/src/message.rs
```diff
@@ -98,6 +117,41 @@ pub struct KKTRequestData(pub Vec<u8>);
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct KKTResponseData(pub Vec<u8>);
 
+/// Packet forwarding request with embedded inner LP packet
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ForwardPacketData {
+    /// Target gateway's Ed25519 identity (32 bytes)
+    pub target_gateway_identity: [u8; 32],
```
jstuczyn, line 124: it doesn't seem like we're using the identity at all when forwarding the packet and only rely on the `lp_address` -> is it actually redundant or will it be used further down the PR stack?

common/nym-lp/README.md
```diff
@@ -1,71 +1,309 @@
...
+| `ClientHello` | 0x0003 | Initial session negotiation |
+| `KKTRequest` | 0x0004 | KEM Key Transfer request |
+| `KKTResponse` | 0x0005 | KEM Key Transfer response |
+| `ForwardPacket` | 0x0006 | Nested session forwarding |
+
+## Session Establishment
+
+### Session ID
+
+Sessions are identified by a deterministic 32-bit ID computed from both parties' X25519 public keys:
```
jstuczyn, line 57: this contradicts what's written in `DESIGN.md`:

> The client generates a random 4-byte receiver_index and includes it in ClientHello. The gateway uses this as the session lookup key. This replaces the previous approach of computing a deterministic session_id from both parties' keys.

gateway/src/node/lp_listener/handler.rs
```diff
@@ -101,90 +89,519 @@ impl LpConnectionHandler {
...
+        // Check both handshake_states (in-progress) and session_states (established)
+        if self.state.handshake_states.contains_key(&receiver_index)
+            || self.state.session_states.contains_key(&receiver_index)
+        {
+            warn!("Receiver index collision: {} from {}", receiver_index, self.remote_addr);
+            inc!("lp_receiver_index_collision");
+
+            // Send Collision response to tell client to retry with new receiver_index
+            // No outer key - this is before PSK derivation
+            let collision_packet = LpPacket::new(
```
jstuczyn, line 224: question for @georgio for whether this is an actual problem. couldn't client technically perform an enumeration attack to figure indices of other connections? what would happen if they actually were successful?

common/nym-lp/src/message.rs
```diff
@@ -106,6 +160,21 @@ pub enum LpMessage {
     ClientHello(ClientHelloData),
     KKTRequest(KKTRequestData),
     KKTResponse(KKTResponseData),
+    ForwardPacket(ForwardPacketData),
+    /// Receiver index collision - client should retry with new receiver_index
+    Collision,
+    /// Acknowledgment - gateway confirms receipt of message
+    Ack,
```
jstuczyn, line 167: can we rename it to `ClientHelloAck` ? initially I got a bit confused thinking it was some sort of Ack for data packets

common/nym-lp/src/codec.rs
```diff
@@ -3,105 +3,382 @@
...
+        }
+        MessageType::SubsessionReady => {
+            let data: SubsessionReadyData = bincode::deserialize(content)
+                .map_err(|e| LpError::DeserializationError(e.to_string()))?;
+            Ok(LpMessage::SubsessionReady(data))
+        }
+        MessageType::SubsessionAbort => {
+            // Empty signal message - no content to deserialize
+            if !content.is_empty() {
+                return Err(LpError::DeserializationError(
```
jstuczyn, line 169: a bit of a nit: we seem to be inconsistent in erroring. sometimes when content is non-empty (and we expect no data), we return `LpError::InvalidPayloadSize` and sometimes, like here, `LpError::DeserializationError`

common/nym-lp/DESIGN.md
```diff
@@ -0,0 +1,365 @@
...
+| payload | variable | Message type (2B) + content |
+| trailer | 16 bytes | Zeros (no PSK) or AEAD Poly1305 tag (with PSK) |
+
+### Wire Format
+
+Length-prefixed over TCP:
+
+```
+┌────────────────────┬─────────────────────────────────────────────────────┐
+│ length (4B BE u32) │ LpPacket                                            │
```
jstuczyn, line 57: given we want LpPackets to fit inside UDP datagrams, if we're optimising things, isn't using u32 for length a wasteful overkill? wouldn't u16 been more than enough?

common/nym-lp/src/codec.rs
```diff
@@ -3,105 +3,382 @@
...
+pub fn serialize_lp_packet(
+    item: &LpPacket,
+    dst: &mut BytesMut,
+    outer_key: Option<&OuterAeadKey>,
+) -> Result<(), LpError> {
+    // Total size: outer_header(12) + inner_prefix(4) + msg_type(2) + content + trailer(16)
+    let total_size = OUTER_HEADER_SIZE + INNER_PREFIX_SIZE + 2 + item.message.len() + TRAILER_LEN;
+    dst.reserve(total_size);
+
+    // 1. Write outer header (always cleartext) - 12 bytes
```
jstuczyn, line 327: we could also put the plaintext inner header and content here. then with no key we'd just attach the empty trailer and with the key, we'd continue with encryption with no changes

common/nym-lp/src/codec.rs
```diff
@@ -3,105 +3,382 @@
...
+///
+/// 2. **Key derivation**: `outer_key = Blake3_KDF("lp-outer-aead", PSK)`.
+///    Different PSK → different outer_key → nonce reuse impossible.
+///
+/// 3. **No PSK persistence**: PSK handles are not stored/reused across sessions.
+///    Each connection performs fresh KKT+PSQ handshake.
+///
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
+pub struct OuterAeadKey {
+    key: [u8; 32],
```
jstuczyn, line 47: would it make sense to just explicitly hold here an instance of `chacha20poly1305::Key`?

common/nym-lp/src/message.rs
```diff
@@ -98,6 +117,41 @@ pub struct KKTRequestData(pub Vec<u8>);
...
+pub struct ForwardPacketData {
+    /// Target gateway's Ed25519 identity (32 bytes)
+    pub target_gateway_identity: [u8; 32],
+
+    /// Target gateway's LP address (IP:port string)
+    pub target_lp_address: String,
+
+    /// Complete inner LP packet bytes (serialized LpPacket)
+    /// This is the CLIENT→EXIT gateway packet, encrypted for exit
+    pub inner_packet_bytes: Vec<u8>,
```
jstuczyn, line 131: just leaving a question here: do we ensure that when we construct inner bytes we account for smaller MTU due additional hop overhead?

common/nym-lp/src/message.rs
```diff
@@ -127,9 +204,17 @@ impl LpMessage {
             LpMessage::Busy => &[],
             LpMessage::Handshake(payload) => payload.0.as_slice(),
             LpMessage::EncryptedData(payload) => payload.0.as_slice(),
-            LpMessage::ClientHello(_) => unimplemented!(), // Structured data, serialized in encode_content
+            LpMessage::ClientHello(_) => &[], // Structured data, serialized in encode_content
```
jstuczyn, line 207: that's a bit misleading given that `ClientHello` is actually **NOT** empty (as most of the other instances below)

common/nym-lp/src/message.rs
```diff
@@ -149,9 +242,22 @@ impl LpMessage {
             LpMessage::Busy => 0,
             LpMessage::Handshake(payload) => payload.0.len(),
             LpMessage::EncryptedData(payload) => payload.0.len(),
-            LpMessage::ClientHello(_) => 97, // 32 bytes x25519 key + 32 bytes ed25519 key + 32 bytes salt + 1 byte bincode overhead
+            // 4 bytes receiver_index + 32 bytes x25519 key + 32 bytes ed25519 key + 32 bytes salt + bincode overhead
```
jstuczyn, line 245: I really don't like the fact we have some magic serialiser specific overhead. what if the  implementation changes? the client hello data is simple enough to get rid of it (we could just concat all fields since they have constant lengths)

common/nym-lp/src/message.rs
```diff
@@ -149,9 +242,22 @@ impl LpMessage {
...
+            LpMessage::ClientHello(_) => 101,
             LpMessage::KKTRequest(payload) => payload.0.len(),
             LpMessage::KKTResponse(payload) => payload.0.len(),
+            LpMessage::ForwardPacket(data) => {
+                32 + data.target_lp_address.len() + data.inner_packet_bytes.len() + 10
```
jstuczyn, line 250: what are those magic values here?

common/nym-lp/src/message.rs
```diff
@@ -149,9 +242,22 @@ impl LpMessage {
...
+            }
+            LpMessage::Collision => 0,
+            LpMessage::Ack => 0,
+            LpMessage::SubsessionRequest => 0,
+            // Variable length: bincode overhead (~8 bytes for Vec length) + payload
```
jstuczyn, line 255: do we have tests to ensure that overhead is static?

common/nym-lp/src/message.rs
```diff
@@ -187,6 +301,30 @@ impl LpMessage {
             LpMessage::KKTResponse(payload) => {
                 dst.put_slice(&payload.0);
             }
+            LpMessage::ForwardPacket(data) => {
+                let serialized =
+                    bincode::serialize(data).expect("Failed to serialize ForwardPacketData");
```
jstuczyn, line 306: we really should have an explicit function creating bincode serialiser to ensure fixed int encoding

common/nym-lp/src/session.rs
```diff
@@ -858,6 +951,49 @@ impl LpSession {
         self.noise_state.lock().is_handshake_finished()
     }
 
+    /// Returns the PQ shared secret (K_pq) if available.
+    ///
+    /// This is the raw KEM output from PSQ before Blake3 KDF combination.
+    /// Used for deriving subsession PSKs to maintain PQ protection.
+    pub fn pq_shared_secret(&self) -> Option<[u8; 32]> {
+        self.pq_shared_secret.lock().as_ref().map(|s| *s.as_bytes())
```
jstuczyn, line 959: we should return an `Option<PqSharedSecret>` instead here to ensure data is zeroized. otherwise there are some codepaths that will leave those bytes hanging around

gateway/src/node/lp_listener/mod.rs
```diff
@@ -186,6 +315,29 @@ pub struct LpHandlerState {
 
     /// LP configuration (for timestamp validation, etc.)
     pub lp_config: LpConfig,
+
+    /// In-progress handshakes keyed by session_id
```
jstuczyn, line 319: comment rot. it's keyed by receiver index which is no longer computed deterministically from the keys but rather proposed by the initiator

common/nym-lp/src/state_machine.rs
```diff
@@ -452,43 +513,108 @@ impl LpStateMachine {
...
+                                 Ok(subsession) => {
+                                     // Process KK1
+                                     match subsession.process_message(&kk1_data.payload) {
+                                         Ok(_) => {
+                                             // Prepare KK2 response
+                                             match subsession.prepare_message() {
+                                                 Ok(kk2_payload) => {
+                                                     let kk2_msg = LpMessage::SubsessionKK2(SubsessionKK2Data { payload: kk2_payload });
+                                                     match session.next_packet(kk2_msg) {
+                                                         Ok(response_packet) => {
```
jstuczyn, line 538: I think I suggested it before, but early error returns would really help with readibility here. 13 levels of indentation is a bit much

common/nym-lp/src/state_machine.rs
```diff
@@ -452,43 +513,108 @@ impl LpStateMachine {
...
+                                 // 2. Decrypt data
+                                 match session.decrypt_data(&packet.message) {
+                                     Ok(plaintext) => {
+                                         // 3. Mark counter as received
+                                         if let Err(e) = session.receiving_counter_mark(packet.header.counter) {
+                                             result_action = Some(Err(e));
+                                             LpState::Transport { session }
+                                         } else {
+                                             // 4. Deliver data
+                                             result_action = Some(Ok(LpAction::DeliverData(BytesMut::from(plaintext.as_slice()))));
```
jstuczyn, line 587: does it make sense to store the data in `BytesMut` given they should be delivered as they are, i.e. no further mutability? (ultimately they seem to be always just converted into `Vec<u8>` or dereferences into &[u8] to be further deserialised)

gateway/src/node/lp_listener/handler.rs
```diff
@@ -101,90 +89,519 @@ impl LpConnectionHandler {
...
+            // Established session - should always have PSK
+            // session_states now stores LpStateMachine (not LpSession) for subsession support
+            session_entry
+                .value()
+                .state
+                .session()
+                .ok()
+                .and_then(|s| s.outer_aead_key())
+        } else {
+            // Unknown session - will error during routing, parse cleartext
```
jstuczyn, line 133: so why not just error out immediately and abandon this handler?

common/nym-lp/src/message.rs
```diff
@@ -46,6 +49,7 @@ impl ClientHelloData {
         rand::thread_rng().fill_bytes(&mut salt[8..]);
 
         Self {
+            receiver_index: rand::random(), // Auto-generate random receiver index
```
jstuczyn, line 52: we can't use `rand:random()` directly as (even though incredibly unlikely) it could be set to `BOOTSTRAP_RECEIVER_IDX` and thus it would be incorrectly handled

nym-registration-client/src/lp_client/nested_session.rs
```diff
@@ -0,0 +1,520 @@
...
+///     "2.2.2.2:41264".to_string(),
+///     client_keypair,
+///     exit_public_key,
+/// );
+///
+/// let gateway_data = nested.handshake_and_register(&mut outer_client, ...).await?;
+/// ```
+pub struct NestedLpSession {
+    /// Exit gateway's Ed25519 identity (32 bytes)
+    exit_identity: [u8; 32],
```
jstuczyn, line 57: why not keep it more strongly typed, i.e. as `ed25519::PublicKey`?

nym-registration-client/src/lp_client/nested_session.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    /// Exit gateway's LP address (e.g., "2.2.2.2:41264")
+    exit_address: String,
```
jstuczyn, line 60: same here, but as `SocketAddr`

jstuczyn (commented):
Another batch of comments from me : )

- LpPacket is length-prefixed when put on the wire. but given we want them to fit inside UDP datagrams, isn't using u32 for that purpose a wasteful overkill? wouldn't u16 been more than enough? especially since we do actually make explicit checks to see if data is not bigger than what would have fit in u16 (plus 1):
```
  const MAX_PACKET_SIZE: usize = 65536; // 64KB max
    if packet_len > MAX_PACKET_SIZE {
        return Err(GatewayError::LpProtocolError(format!(
            "Packet size {} exceeds maximum {}",
            packet_len, MAX_PACKET_SIZE
        )));
    }
```
- I'm really not a fan of `ClientHello` serialisation and depending on magic value denoting bincode overhead. the payload structure is simple enough to get rid of it (we could just concat all fields since they have constant lengths)
- if we're going to use bincode for lp messages serialisation, we really should have an explicit function creating the serialiser to ensure fixed int encoding
- when generating receiver_index we have to ensure it's never the same as `BOOTSTRAP_RECEIVER_IDX` otherwise we risk incorrect handling of data
- I'm not convinced by the single packet per connection approach for registration - I can imagine it might cause some friction down the line especially if attempting to register with gateways far away from the user. then the overhead of creating multiple connections to complete the exchange will be very much non-negligible

jstuczyn:

PR #6249, opened 2025-11-27
[Feature] Fallback gateway listener and remove legacy key support

simonwicky:
This PR removes legacy keys support, started by @jstuczyn in #5531 and #5553.
It also adds a secondary ws gateway listener if available and update them every 7 days

Client side has been smoke tested, gateway side not at all

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6249)


common/client-core/gateways-storage/src/backend/fs_backend/manager.rs
```diff
@@ -144,13 +144,10 @@ impl StorageManager {
...
         gateway_id: &str,
     ) -> Result<RawRemoteGatewayDetails, sqlx::Error> {
-        sqlx::query_as!(
-            RawRemoteGatewayDetails,
-            "SELECT * FROM remote_gateway_details WHERE gateway_id_bs58 = ?",
-            gateway_id
-        )
-        .fetch_one(&self.connection_pool)
-        .await
+        sqlx::query_as("SELECT * FROM remote_gateway_details WHERE gateway_id_bs58 = ?")
```
jstuczyn, line 149: nothing wrong with it, just wondering, why did you switch away from the macro?

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+CREATE TABLE remote_gateway_details_temp
```
jstuczyn, line 6: in your older iteration you had a separate table just dedicated to keys - any reason for changing the design?

common/client-core/gateways-storage/src/lib.rs
```diff
@@ -60,10 +59,10 @@ pub trait GatewaysDetailsStore {
         details: &GatewayRegistration,
     ) -> Result<(), Self::StorageError>;
 
-    async fn upgrade_stored_remote_gateway_key(
+    /// Update the gateway details
+    async fn update_gateway_details(
```
jstuczyn: this method is quite misleading because you don't overwrite the key in the implementation, meaning if there was a bug somewhere and your new details had a different key, the whole thing would be unusable. imo create a different wrapper for the passed argument, something that doesn't contain the key and **only** has the fields you're actually end up updating.

what you really care about are the listeners, so how about `update_gateway_listeners` or for more futureproofing, `update_published_gateway_data` (or something along those lines)?

common/client-core/gateways-storage/src/backend/fs_backend/mod.rs
```diff
@@ -134,18 +132,20 @@ impl GatewaysDetailsStore for OnDiskGatewaysDetails {
...
-            )
-            .await?;
+        match &details.details {
+            GatewayDetails::Remote(remote_details) => {
+                let raw_details = &remote_details.into();
+                self.manager
+                    .update_remote_gateway_details(raw_details)
+                    .await?;
+            }
+            GatewayDetails::Custom(_) => {}
```
jstuczyn: it feels a bit weird that even though we're supposedly updating gateway detaiils, there's absolutely nothing being done for `Custom` variant. imo the storage function should be made less generic and should only apply to the data we're actually changing

common/client-core/gateways-storage/src/types.rs
```diff
@@ -2,20 +2,21 @@
...
 use std::str::FromStr;
 use std::sync::Arc;
+use time::Duration;
 use time::OffsetDateTime;
 use url::Url;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 pub const REMOTE_GATEWAY_TYPE: &str = "remote";
 pub const CUSTOM_GATEWAY_TYPE: &str = "custom";
+const GATEWAY_DETAILS_TTL: Duration = Duration::days(7);
```
jstuczyn, line 19: should we make it configurable instead via client config?

common/gateway-requests/src/shared_key.rs
jstuczyn, line 1: given this is the only file left in this module, imo just move it from `shared_key/mod.rs` into `shared_key.rs`

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
```diff
@@ -0,0 +1,22 @@
...
+(
+    gateway_id_bs58                          TEXT NOT NULL UNIQUE PRIMARY KEY REFERENCES registered_gateway (gateway_id_bs58),
+    derived_aes256_gcm_siv_key               BLOB NOT NULL,
+    gateway_listener                         TEXT NOT NULL,
+    fallback_listener                        TEXT,
+    expiration_timestamp                     TIMESTAMP WITHOUT TIME ZONE NOT NULL
+);
+
+
+INSERT INTO remote_gateway_details_temp SELECT gateway_id_bs58, derived_aes256_gcm_siv_key, gateway_listener, NULL, datetime(0, 'unixepoch') FROM remote_gateway_details;
```
jstuczyn: in some cases this might fail if `derived_aes256_gcm_siv_key` is `null` (somebody registered with some gateway ages ago and still used new key); you have to update that query to only included entries where this new key actually does exist

common/client-core/gateways-storage/src/types.rs
```diff
@@ -186,49 +204,11 @@ impl TryFrom<RawRemoteGatewayDetails> for RemoteGatewayDetails {
...
-                AccountId::from_str(raw_owner).map_err(|source| {
-                    BadGateway::MalformedGatewayOwnerAccountAddress {
-                        gateway_id: value.gateway_id_bs58.clone(),
-                        raw_owner: raw_owner.clone(),
-                        source,
-                    }
-                })
-            })
-            .transpose()?;
+        let shared_key = SharedSymmetricKey::try_from_bytes(&value.derived_aes256_gcm_siv_key)
```
jstuczyn, line 259: I love how we were able to simplify all of that : )

common/client-core/gateways-storage/src/types.rs
```diff
@@ -186,49 +204,11 @@ impl TryFrom<RawRemoteGatewayDetails> for RemoteGatewayDetails {
...
+            .map_err(|source| BadGateway::MalformedSharedKeys {
+                gateway_id: value.gateway_id_bs58.clone(),
+                source,
+            })?;
 
         let gateway_listener = Url::parse(&value.gateway_listener).map_err(|source| {
```
jstuczyn: suggestion: I think it might be more readable if you implemented `TryFrom RawGatewayListeners for GatewayListeners` (optional) and `TryFrom RawRemoteGatewayDetails for RemoteGatewayDetails`
and then modified the db types to:

```rust
#[derive(Debug, Zeroize, ZeroizeOnDrop, Serialize, Deserialize)]
#[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
pub struct RawGatewayListeners {
    pub gateway_listener: String,
    pub fallback_listener: Option<String>,
}

#[derive(Debug, Zeroize, ZeroizeOnDrop, Serialize, Deserialize)]
#[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
pub struct RawRemoteGatewayDetails {
    pub gateway_id_bs58: String,
    pub derived_aes256_gcm_siv_key: Vec<u8>,

    #[cfg_attr(feature = "sqlx", sqlx(flatten))]
    pub gateway_listeners: RawGatewayListeners,
    #[zeroize(skip)]
    pub expiration_timestamp: OffsetDateTime,
}
```

common/client-core/gateways-storage/src/types.rs
```diff
@@ -237,30 +217,42 @@ impl TryFrom<RawRemoteGatewayDetails> for RemoteGatewayDetails {
...
+            gateway_listeners: GatewayListeners {
+                primary: gateway_listener,
+                fallback: fallback_listener,
+            },
+            expiration_timestamp: value.expiration_timestamp,
         })
     }
 }
 
 impl<'a> From<&'a RemoteGatewayDetails> for RawRemoteGatewayDetails {
```
jstuczyn, line 273: oh right. we have that already

common/client-core/src/init/mod.rs
```diff
@@ -71,21 +71,28 @@ where
     let mut rng = OsRng;
 
     let selected_gateway = match selection_specification {
-        GatewaySelectionSpecification::UniformRemote { must_use_tls } => {
+        GatewaySelectionSpecification::UniformRemote {
+            must_use_tls,
+            no_hostname,
```
jstuczyn, line 76: why do we have to explicitly provide that argument? couldn't we infeer it based on the retrieved url?

common/client-core/src/cli_helpers/client_list_gateways.rs
```diff
@@ -56,14 +56,16 @@ where
                 identity: remote_details.gateway_id,
                 active: active_gateway == Some(remote_details.gateway_id),
                 typ: GatewayType::Remote.to_string(),
-                endpoint: Some(remote_details.gateway_listener),
+                endpoint: Some(remote_details.gateway_listeners.primary),
```
jstuczyn: would it work if we clumped both primary and secondary listeners under on `endpoint` field?

common/client-core/src/client/base_client/mod.rs
```diff
@@ -593,31 +585,13 @@ where
         // we need to:
         // - perform handshake (reg or auth)
         // - check for key upgrade
-        // - maybe perform another upgrade handshake
         // - check for bandwidth
         // - start background tasks
-        let auth_res = gateway_client
+        let _auth_res = gateway_client
```
jstuczyn: nit: in that case you no longer need to bind to the result

common/client-core/src/client/base_client/mod.rs
```diff
@@ -593,31 +585,13 @@ where
         // we need to:
         // - perform handshake (reg or auth)
         // - check for key upgrade
```
jstuczyn: you removed it : )

common/client-core/src/client/base_client/mod.rs
```diff
@@ -975,8 +947,7 @@ where
         )
         .await?;
 
-        let (reply_storage_backend, credential_store, details_store) =
-            self.client_store.into_runtime_stores();
+        let (reply_storage_backend, credential_store, _) = self.client_store.into_runtime_stores();
```
jstuczyn, line 949: something feels off that you don't need gateways store when starting up the client O.o
I mean the code compiles and what-not, but it feels wrong... I will have a second look through that

common/client-core/src/client/base_client/storage/helpers.rs
```diff
@@ -85,6 +85,22 @@ where
         })
 }
 
+pub async fn update_stored_gateway_details<D>(
```
jstuczyn: I've already left a comment with the actual db code how I think it might be changed

common/client-core/src/client/base_client/storage/migration_helpers.rs
```diff
@@ -2,210 +2,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub mod v1_1_33 {
-    use crate::client::base_client::{
```
jstuczyn, line 5: don't change old migration code. you have to create your own version and then add additional path there

common/client-core/src/client/base_client/storage/migration_helpers.rs
```diff
@@ -2,210 +2,18 @@
...
-            &gateway_registration.details.gateway_id().to_base58_string(),
-        )
-        .await?;
-
-        remove_old_gateway_details(old_storage_paths).map_err(|err| {
-            ClientCoreError::UpgradeFailure {
-                message: format!("failed to remove old data: {err}"),
-            }
-        })
+        Err(ClientCoreError::UnsupportedMigration(
```
jstuczyn, line 15: are you 100% sure 1.1.33 didn't already support new keys? if so, then I guess this update to the migration is actually fine

common/client-core/src/init/mod.rs
```diff
@@ -150,6 +152,43 @@ where
     })
 }
 
+pub async fn update_gateway_details<D>(
```
jstuczyn: rather than update, how about `refresh`?

common/topology/src/node.rs
```diff
@@ -89,6 +89,23 @@ impl RoutingNode {
         self.ws_entry_address_no_tls(prefer_ipv6)
     }
 
+    pub fn ws_entry_address_no_hostname(&self, prefer_ipv6: bool, index: usize) -> Option<String> {
```
jstuczyn: I'm not sure using `index` here is a good idea because I don't think the address ordering is deterministic. maybe if we pre-sorted them

common/gateway-storage/migrations/20251126120000_remove_aes128ctr_key.sql
```diff
@@ -0,0 +1,23 @@
...
+-- make aes256gcm column non-nullable and drop any clients that still use the legacy keys
+CREATE TABLE shared_keys_tmp
+(
+    client_id                  INTEGER NOT NULL PRIMARY KEY REFERENCES clients (id),
+    client_address_bs58        TEXT    NOT NULL UNIQUE,
+    derived_aes256_gcm_siv_key BLOB    NOT NULL,
+    last_used_authentication TIMESTAMP WITHOUT TIME ZONE
+);
+
+INSERT INTO shared_keys_tmp (client_id, client_address_bs58, derived_aes256_gcm_siv_key, last_used_authentication)
```
jstuczyn, line 16: same note as if the client migrations: you must ensure you only select entries where the key is actually not null

common/client-core/src/init/types.rs
```diff
@@ -42,24 +39,47 @@ impl SelectedGateway {
     pub fn from_topology_node(
         node: RoutingNode,
         must_use_tls: bool,
+        no_hostname: bool,
```
jstuczyn, line 42: given we have to already weave it through quite a long call stack, how about we also add that `prefer_ipv6` that is hardcoded there and wrap it in some `Config` struct? (because I've got a gut feeling, we'll have to add extra things here soon...)

common/client-core/src/init/types.rs
```diff
@@ -145,20 +167,36 @@ impl InitialisationResult {
...
+        }
+    }
 }
 
 #[derive(Clone, Debug)]
 pub enum GatewaySelectionSpecification {
     /// Uniformly choose a random remote gateway.
-    UniformRemote { must_use_tls: bool },
+    UniformRemote {
+        must_use_tls: bool,
```
jstuczyn, line 178: as I mentioned with the `Config` above, given all of those fields appear in every single enum variant, how about we wrap them in some wrapper?

common/client-core/src/init/types.rs
```diff
@@ -42,24 +39,47 @@ impl SelectedGateway {
...
+        let (gateway_listener, fallback_listener) = if must_use_tls {
+            // WSS main, no fallback
+            let primary =
+                node.ws_entry_address_tls()
+                    .ok_or(ClientCoreError::UnsupportedWssProtocol {
+                        gateway: node.identity_key.to_base58_string(),
+                    })?;
+            (primary, None)
+        } else if no_hostname {
+            // First IP address for main, second if it exists for fallback
```
jstuczyn: what if node has announced 2 addreses: an ipv6 and ipv4 with the ipv4 being announced as the first one, i.e. `[1.1.1.1, ::1] if `prefer_ipv6 = true`, then you will retrieve that ipv6 as primary: `::1`(because of the `filter` call), but when you query for the fallback, there won't be `nth(1)` ipv6 in the filtered list. so you will go back to the `let any_ip = entry.ip_addresses.get(index)?;` call retrieving addresses from unmodified list. which will give you exactly the same ip again, thus your `primary == fallback` which is not what you want

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -205,10 +201,10 @@ impl<C, St> GatewayClient<C, St> {
     pub async fn establish_connection(&mut self) -> Result<(), GatewayClientError> {
         debug!(
             "Attempting to establish connection to gateway at: {}",
-            self.gateway_address
+            self.gateway_addresses.primary
```
jstuczyn: well, or potentially the fallback, no?

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -221,7 +217,7 @@ impl<C, St> GatewayClient<C, St> {
 
     #[cfg(target_arch = "wasm32")]
     pub async fn establish_connection(&mut self) -> Result<(), GatewayClientError> {
-        let ws_stream = match JSWebsocket::new(&self.gateway_address) {
+        let ws_stream = match JSWebsocket::new(&self.gateway_addresses.primary.as_ref()) {
```
jstuczyn: what about the fallback here?

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -525,75 +516,6 @@ impl<C, St> GatewayClient<C, St> {
...
-        };
-
-        // attempt to decrypt it using NEW key
-        let Ok(response) = SensitiveServerResponse::decrypt(&ciphertext, &nonce, &updated_key)
-        else {
-            return Err(GatewayClientError::FatalKeyUpgradeFailure);
-        };
-
-        match response {
-            SensitiveServerResponse::KeyUpgradeAck { .. } => {
```
jstuczyn, line 584: we might as well remove that enum variant too

common/client-libs/gateway-client/src/client/websockets.rs
```diff
@@ -36,6 +38,8 @@ pub(crate) async fn connect_async(
         Host::Ipv6(addr) => vec![SocketAddr::new(addr.into(), port)],
         Host::Domain(domain) => {
             // Do a DNS lookup for the domain using our custom DNS resolver
+            println!("I'M TIRED BOSS");
```
jstuczyn: I know, I know, but remove that debug code, m'kay?

common/client-libs/gateway-client/src/error.rs
```diff
@@ -117,7 +120,7 @@ pub enum GatewayClientError {
     MixnetMsgSenderFailedToSend,
 
     #[error("Attempted to negotiate connection with gateway using incompatible protocol version. Ours is {current} and the gateway reports {gateway:?}")]
```
jstuczyn: nit:
```suggestion
    #[error("Attempted to negotiate connection with gateway using incompatible protocol version. Ours is {current} and the gateway reports {gateway}")]
```

common/gateway-requests/src/registration/handshake/client.rs
```diff
@@ -18,11 +18,11 @@ impl<S, R> State<'_, S, R> {
         R: CryptoRng + RngCore,
     {
         // 1. if we're using non-legacy, i.e. aes256gcm-siv derivation, generate initiator salt for kdf
-        let maybe_hkdf_salt = self.maybe_generate_initiator_salt();
+        let hkdf_salt = self.generate_initiator_salt();
 
         // 1. send ed25519 pubkey alongside ephemeral x25519 pubkey and a hkdf salt if we're using non-legacy client
         // LOCAL_ID_PUBKEY || EPHEMERAL_KEY || MAYBE_SALT
```
jstuczyn: nit: it's no longer a `MAYBE`

common/gateway-requests/src/registration/handshake/messages.rs
```diff
@@ -21,13 +21,13 @@ pub trait HandshakeMessage {
 pub struct Initialisation {
     pub identity: ed25519::PublicKey,
     pub ephemeral_dh: x25519::PublicKey,
-    pub initiator_salt: Option<Vec<u8>>,
+    pub initiator_salt: Vec<u8>,
```
jstuczyn, line 24: nit: we could now use constant size arrays again

common/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -96,31 +93,27 @@ impl<'a, S, R> State<'a, S, R> {
...
-            Some(salt)
-        } else {
-            None
-        }
+        let mut salt = vec![0u8; KDF_SALT_LENGTH];
+        self.rng.fill_bytes(&mut salt);
+        salt
     }
 
     // LOCAL_ID_PUBKEY || EPHEMERAL_KEY || MAYBE_SALT
```
jstuczyn: nit: no longer `MAYBE`

common/gateway-requests/src/types/helpers.rs
```diff
@@ -83,30 +68,33 @@ impl<'a> BinaryData<'a> {
...
         self,
-        key: &SharedGatewayKey,
+        key: &SharedSymmetricKey,
     ) -> Result<BinaryRequest, GatewayRequestsError> {
         let kind = BinaryRequestKind::from_repr(self.kind)
             .ok_or(GatewayRequestsError::UnknownRequestKind { kind: self.kind })?;
 
         let plaintext = if self.encrypted {
-            &*key.decrypt(self.data, self.maybe_nonce)?
+            let raw_nonce = self.maybe_nonce.unwrap_or(&[]);
```
jstuczyn: wouldn't in this instance the nonce always be set?

common/gateway-requests/src/types/helpers.rs
```diff
@@ -117,13 +105,16 @@ impl<'a> BinaryData<'a> {
...
         self,
-        key: &SharedGatewayKey,
+        key: &SharedSymmetricKey,
     ) -> Result<BinaryResponse, GatewayRequestsError> {
         let kind = BinaryResponseKind::from_repr(self.kind)
             .ok_or(GatewayRequestsError::UnknownResponseKind { kind: self.kind })?;
 
         let plaintext = if self.encrypted {
-            &*key.decrypt(self.data, self.maybe_nonce)?
+            let raw_nonce = self.maybe_nonce.unwrap_or(&[]);
```
jstuczyn: same here

common/wasm/client-core/src/storage/core_client_traits.rs
```diff
@@ -156,17 +154,12 @@ impl GatewaysDetailsStore for ClientStorage {
...
-        updated_key: &SharedSymmetricKey,
+        details: &GatewayRegistration,
     ) -> Result<(), Self::StorageError> {
-        self.update_remote_gateway_key(
-            &gateway_id.to_base58_string(),
-            None,
-            Some(updated_key.as_bytes()),
-        )
-        .await
+        // Will overwrite value, which is what we want
```
jstuczyn: well, not necessarily. we don't want the key to change

gateway/src/node/client_handling/websocket/connection_handler/fresh.rs
```diff
@@ -90,6 +80,9 @@ pub(crate) enum InitialAuthenticationError {
     #[error("Experienced connection error: {0}")]
     ConnectionError(Box<WsError>),
 
+    #[error("Attempted to negotiate connection with client using incompatible protocol version. Ours is {current} and the client reports {client:?}")]
```
jstuczyn: nit:
```suggestion
    #[error("Attempted to negotiate connection with client using incompatible protocol version. Ours is {current} and the client reports {client}")]
```

gateway/src/node/client_handling/websocket/connection_handler/fresh.rs
```diff
@@ -802,9 +694,6 @@ impl<R, S> FreshHandler<R, S> {
             .await?;
         let shared_keys = handshake_result.derived_key;
 
-        // populate the negotiated protocol for future uses
-        self.negotiated_protocol = Some(handshake_result.negotiated_protocol);
```
jstuczyn, line 806: well no. you should take the result from the handshake because that's where the negotiation occurs.  the passed value is just an initial proposal

nym-registration-client/src/builder/config.rs
```diff
@@ -119,6 +119,7 @@ impl BuilderConfig {
             .network_details(self.network_env)
             .debug_config(debug_config)
             .credentials_mode(true)
+            .no_hostname(true)
```
jstuczyn, line 122: wait. should this be the default?

jstuczyn (changes requested):

common/client-libs/gateway-client/src/client/websockets.rs
simonwicky: Oopsie

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
simonwicky: Discussed on zulip. Makes load and store a bit awkward for no real gain

common/client-core/gateways-storage/src/types.rs
simonwicky: I don't know, I haven't thought that far

common/client-core/src/client/base_client/storage/migration_helpers.rs
simonwicky: That comes directly from your commit in #5531

common/client-core/src/client/base_client/storage/migration_helpers.rs
simonwicky: That comes directly from your commit in #5531

common/topology/src/node.rs
simonwicky: We were using `.first()` anyway already. That's just to get the first or second one without duplicating code

nym-registration-client/src/builder/config.rs
simonwicky: For the vpn-client yes, firewall stuff and avoiding DNS we talked about on zulip

common/client-core/gateways-storage/src/backend/fs_backend/manager.rs
simonwicky: SQLx shenanigans, but I was able to fix it.

common/gateway-requests/src/types/helpers.rs
simonwicky: Indeed

common/client-core/src/client/base_client/mod.rs
simonwicky: clippy tells me I still have to

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
simonwicky: Good point. Fixed by dropping those registrations (included the registered_gateways record)

common/client-core/src/client/base_client/mod.rs
simonwicky: It was required by the GatewayClient for the key_upgrade. The keys are loaded right before that in `initialise_keys_and_gateway`

common/gateway-storage/migrations/20251126120000_remove_aes128ctr_key.sql
simonwicky: I'm already doing that

common/client-core/gateways-storage/src/types.rs
simonwicky: In any case, having something to update that doesn't have the key is gonna be very useful

common/client-core/gateways-storage/src/backend/fs_backend/mod.rs
simonwicky: Fixed

common/client-core/gateways-storage/src/types.rs
simonwicky: Fixed using a similar idea that should be future proof

common/client-core/src/cli_helpers/client_list_gateways.rs
simonwicky: I guess so, but it's just used for display atm

common/client-core/src/init/mod.rs
simonwicky: We want to able to bypass the hostname priority if both hostname and IP are available

common/client-libs/gateway-client/src/client/mod.rs
simonwicky: IIRC wasm client requires WSS and that means no IP fallback possible, because we need a hostname for TLS

common/client-libs/gateway-client/src/client/mod.rs
simonwicky: Fine

common/wasm/client-core/src/storage/core_client_traits.rs
simonwicky: Fixed along the other changes of details update

common/client-core/src/init/types.rs
simonwicky: You know the gif I want to send here...

common/client-core/src/init/types.rs
simonwicky: Indeed, fixed in cce2d448dcc069a5a89c411419c13fb932a0f4d9

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
```diff
@@ -0,0 +1,25 @@
...
+    gateway_listener                         TEXT NOT NULL,
+    fallback_listener                        TEXT,
+    expiration_timestamp                     DATETIME NOT NULL
+);
+
+-- keep only registrations with a non null aes256 key
+INSERT INTO remote_gateway_details_temp SELECT gateway_id_bs58, derived_aes256_gcm_siv_key, gateway_listener, NULL, datetime(0, 'unixepoch') FROM remote_gateway_details WHERE derived_aes256_gcm_siv_key IS NOT NULL;
+
+-- delete others
+DELETE FROM registered_gateway WHERE gateway_id_bs58 IN ( SELECT gateway_id_bs58 FROM remote_gateway_details WHERE derived_aes256_gcm_siv_key IS NULL);
```
jstuczyn: this will fail with foreign key constraint because `remote_gateway_details` might still reference some of those entries.

you have to first drop `remote_gateway_details`, then delete entries from `registered_gateway`

common/client-core/gateways-storage/src/backend/fs_backend/mod.rs
```diff
@@ -134,17 +132,13 @@ impl GatewaysDetailsStore for OnDiskGatewaysDetails {
         Ok(())
     }
 
-    async fn upgrade_stored_remote_gateway_key(
+    async fn update_gateway_published_data(
         &self,
-        gateway_id: ed25519::PublicKey,
-        updated_key: &SharedSymmetricKey,
+        gateway_id: &str,
```
jstuczyn: why did we decrease our type safety?

common/client-core/gateways-storage/src/types.rs
```diff
@@ -88,13 +87,30 @@ impl GatewayDetails {
...
     }
 
+    pub fn details_exipration(&self) -> Option<OffsetDateTime> {
+        match self {
+            GatewayDetails::Remote(details) => Some(details.published_data.expiration_timestamp),
+            GatewayDetails::Custom(_) => None,
+        }
+    }
+
+    // pub fn update_remote_listeners(&mut self, new_listeners: GatewayListeners) {
```
jstuczyn: dead code?

simonwicky: Yup :homer-bush:

common/client-core/gateways-storage/src/backend/fs_backend/mod.rs
simonwicky: No. These are two different methods and `diff` doesn't really see it.
Other methods also use `&str` to select the id in the database e.g. `load_gateway_details`

common/client-core/gateways-storage/fs_gateways_migrations/20251126120000_remove_legacy_add_fallback.sql
simonwicky: Nice catch

jstuczyn (approved):

PR #6256, opened 2025-12-01
tunnel-manager: add support for IPv6 uplink detection

nymers:
Add IPv6 uplink interface detection to network tunnel manager, supporting environments where IPv4 and IPv6 traffic routes through different interfaces.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6256)

### DIFF
--- a/scripts/nym-node-setup/network-tunnel-manager.sh
+++ b/scripts/nym-node-setup/network-tunnel-manager.sh
@@ -103,6 +103,7 @@ detect_uplink_interface() {
 
 # uplink device detection, can be overridden
 NETWORK_DEVICE="${NETWORK_DEVICE:-}"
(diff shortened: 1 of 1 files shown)


heri16:
"ip -6 -o route show default" is not reliable. On my machine (Debian 13), it requires "print $3" while this script uses "print $5". See:

```sh
nonroot@localhost:~$ ip -6 -o route show default
default dev eth1.1 proto static metric 1024 pref medium
```

benedettadavico:
Another one, @serinko to review this please

serinko:
Hey
As we discussed in DMs, this had been included in the new version of NTM since then. Thank you for your contribution and raising up important points.

PR #6265, opened 2025-12-03
[Stats API] Active device endpoint and exit country code

simonwicky:
Add an active device endpoints to ping for long sessions and add an optional exit_country_code to the session report

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6265)


nym-statistics-api/migrations/004_sessions_exit_cc.sql
```diff
@@ -0,0 +1,49 @@
...
+
+    -- session info
+    start_day_utc         DATE,
+    connection_time_ms    INTEGER,
+    tunnel_type           TEXT,
+    retry_attempt         INTEGER,
+    session_duration_min  INTEGER,
+    disconnection_time_ms INTEGER,
+    exit_id               TEXT,
+    exit_cc               TEXT,   -- new column
```
jstuczyn, line 28: if it's just a new column that can also be nullable, why not simply do `ALTER TABLE report_v2 ADD COLUMN exit_cc` rather than having to copy all data over, recreate indexes, etc?

nym-statistics-api/src/http/api/stats.rs
```diff
@@ -76,6 +79,53 @@ async fn submit_stats_report(
...
+    State(mut state): State<AppState>,
+    TypedHeader(user_agent): TypedHeader<UserAgent>,
+    insecure_ip_addr: InsecureClientIp,
+    Json(report): Json<ActiveDeviceReport>,
+) -> HttpResult<Json<()>> {
+    let now = time::OffsetDateTime::now_utc();
+
+    let gateway_record = state
+        .network_view()
+        .get_country_by_ip(&insecure_ip_addr.0)
```
jstuczyn, line 103: well, given the nature of this ip address, can we get a reliable record?

nym-statistics-api/migrations/004_sessions_exit_cc.sql
simonwicky: My preference was on keeping a meaningful ordering. I could theoretically not move any data, since nothing has hit that table anywhere yet, but it would ba a bit misleading imo. What do you think of that second option?

nym-statistics-api/src/http/api/stats.rs
simonwicky: It's not used for important data point, and for now it doesn't even work on mainnet because the reverse proxy doesn't set the proper header. If it does, it's just used for a small sanity check at the moment so no big deal

simonwicky: Both deployments have a reverse proxy in front so this is the only ip available

jstuczyn (approved):

PR #6270, opened 2025-12-04
Max/crates io prep v2

mfahampshire:
This PR makes the necessary modifications to be able to publish the majority of the Nym monorepo workspace crates to https://crates.io (essentially everything aside from binaries, internal tools, and a few select other things). 

Overall, this PR:
- defines a workspace version for these crates; we want to keep them all in line with each other, even if that means bumping versions frequently and when there hasn't necessarily been a change in all crates. 
~~adds a script to use `cargo release` to publish the SDK + its dependencies - `cargo release` deals with creating them in the correct order~~
- excludes crates we dont want to publish via each crates' `cargo.toml` file 
- adds a global version definition for crates in the root `cargo.toml` 
- changes most imports from `path` to using the `workspace` version 

Todo:
- [x] fix issues re `git` crates 
   - [x] `bls` crate: upload fork to crates.io
   - [x] merge https://github.com/nymtech/nym/pull/6315 branch ~~to make sure that there is no issue with new clients/gws and legacy ones after the dependency upgrade. If all good, change `git` import to `0.8.0` from crates as per PR~~ Testing has been done on a NymVPN instance that was compiled with the new dependency interacting with mainnet gateways. The reverse situation will be tested by QA.   
- [x] checkin on 'max todo' descriptions 
- [ ] ~~experiment: remove separate `contracts/` workspace - TEMP in max/crates-io-prep-v2-contract-experiment~~ EDIT: publish crates then just use crates.io imports for `contracts/`
- [x] fix `nym-credential-verification` compilation errs in dryrun
- [ ] add `publishing.md` writeup

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6270)


Cargo.toml
```diff
@@ -196,6 +196,7 @@ edition = "2024"
 license = "Apache-2.0"
 rust-version = "1.85"
 readme = "README.md"
+version = "1.20.0" # this version is used for the Rust SDK and its workspace dependencies, it is not applied globally
```
jstuczyn: by "its workspace dependencies", do you been exlusively things that are used by rust SDK or all common crates (I think it'd be good to keep it consistent for everything, especially if we were to add an extra dep to rust sdk)

Cargo.toml
```diff
@@ -373,6 +374,73 @@ zeroize = "1.7.0"
 
 prometheus = { version = "0.14.0" }
 
+# sdk related deps so we can pull in workspace versions of these in other crates' cargo files, and not have to define the version everywhere
+# nym-api = { version = "1.2.0", path = "nym-api" }
+nym-bandwidth-controller = { version = "1.2.0", path = "common/bandwidth-controller" }
```
jstuczyn: I'm not sure you have to specify the version on top here (plz double check), but if not, i'd remove it so  we wouldn't cause some inconsistencies if we forget to update those

nym-sqlx-pool-guard/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "sqlx-pool-guard"
-version = "0.1.0"
+name = "nym-sqlx-pool-guard"
```
jstuczyn, line 2: since you've renamed this, after the PR gets merged, create a draft on the vpn repo to update their dependencies (as they were importing it too)

publish-sdk.sh
```diff
@@ -0,0 +1,87 @@
...
+for pkg in "${PACKAGES[@]}"; do
+  PACKAGE_FLAGS="$PACKAGE_FLAGS -p $pkg"
+done
+
+cargo release \
+  $PACKAGE_FLAGS \
+  --prev-tag-name "" \
+  --no-push \
+  --no-tag \
+  --no-publish \
```
jstuczyn: well, if we actually want to publish it, we probably want to remove this flag : )

Cargo.toml
mfahampshire: Its the crates that are used by the SDK only currently - I've added a note in the publishing.md file in the SDK re: adding workspace deps to the SDK

Cargo.toml
mfahampshire: This was the hacky way I found which allows us to define the workspace.version for the constituent crates. Maybe I can get rid of those.. Will experiment.

publish-sdk.sh
mfahampshire: Yeah will remove once we're happy with the dryrun and setup - was being doubly cautious so I didn't accidentally publish them when testing

Cargo.toml
mfahampshire: my previous comment has become moot with my previous commits making basically everything aside from the binaries/contracts/tools follow the versioning structure, sidesteps this potential issue.

publish-sdk.sh
mfahampshire: moved to a diff tool so this script no longer exists - resolving

nym-sqlx-pool-guard/Cargo.toml
mfahampshire: ~~note to self - check where this occurs elsewhere~~ Looks like that was the only spot. 

PR created, in draft until this PR is merged.

Cargo.toml
mfahampshire: Looks like we have to keep the top level version - but `cargo workspaces` should keep it all in line

.github/workflows/publish-crates-io.yml
```diff
@@ -0,0 +1,47 @@
+name: Publish to crates.io
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
```
pronebird, line 6: Does this require a separate validation or `cargo workspaces` handles it properly?

Cargo.toml
```diff
@@ -389,11 +390,112 @@ zeroize = "1.7.0"
 
 prometheus = { version = "0.14.0" }
 
+# Workspace dep definitions required by crates.io publication - we need a workspace version since `cargo workspaces` doesn't work with path imports from crate manifests
+nym-api-requests = { version = "1.20.1", path = "nym-api/nym-api-requests" }
```
pronebird, line 394: I think you don't have to specify version in here, since workspace already has it and workspace crates take it from `workspace.package.version` anyway.

It's enough to leave `path = ..` in here and remove duplicate `version` definitions

pronebird (changes requested):

Cargo.toml
mfahampshire: We need it for `cargo workspaces`, otherwise when we try and publish we get errors like this: 

```
error: failed to verify manifest at `/home/m/dev/work/nym/common/cosmwasm-smart-contracts/coconut-dkg/Cargo.toml`

Caused by:
  all dependencies must have a version requirement specified when publishing.
  dependency `nym-contracts-common` does not specify a version
  Note: The published dependency will use the version from crates.io,
  the `path` specification will be removed from the dependency declaration.
  ```

.github/workflows/publish-crates-io.yml
mfahampshire: From my understanding, crates.io would reject the upload if e.g. someone fat-fingered the release version to a previous version, or something like that. 

What other validation were you thinking of?

mfahampshire: I can add validation to make the version is properly formatted etc

mfahampshire: Added semver validation action

pronebird (approved):

PR #6271, opened 2025-12-04
Fix migrations in the Data Observatory

mmsinclair:
This PR fixes the migrations used for the Data Observatory by:

- making the migrations in the base chain scraper configurable (default is on)
- disabling the default migrations in the Data Observatory
- using custom migrations for the Data Observatory that are manually synced from the base chain scraper (and checked by the build.rs script)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6271)


common/nyxd-scraper-psql/src/storage/block_storage.rs
```diff
@@ -22,7 +22,10 @@ pub struct PostgresScraperStorage {
 
 impl PostgresScraperStorage {
     #[instrument]
-    pub async fn init(connection_string: &str) -> Result<Self, PostgresScraperError> {
+    pub async fn init(
+        connection_string: &str,
+        run_migrations: &bool,
```
jstuczyn, line 27: why pass bool as a reference? I'm surprised clippy hasn't complained

nym-data-observatory/src/chain_scraper/mod.rs
```diff
@@ -13,11 +12,6 @@ pub(crate) async fn run_chain_scraper(
 ) -> anyhow::Result<PostgresNyxdScraper> {
     let use_best_effort_start_height = args.start_block_height.is_some();
 
-    if args.nuke_db {
```
jstuczyn, line 16: I thought this was quite useful 😅

jstuczyn (approved):

nym-data-observatory/src/chain_scraper/mod.rs
mmsinclair: it only works with sqlite as storage, for pgsql would need another connection string to `DROP DATABASE`

common/nyxd-scraper-psql/src/storage/block_storage.rs
mmsinclair: let me see what it says if I change it...

PR #6274, opened 2025-12-05
Gateway probe localnet mode

durch:
Add gateway-probe localnet mode with LP-based two-hop WireGuard testing

  Summary

  - Add localnet mode to gateway-probe for testing LP registration and two-hop WireGuard tunnels in a containerized environment
  - Implement two-hop WireGuard tunnel support (Client → Entry Gateway → Exit Gateway → Internet)
  - Add mock ecash credentials for local testing without real blockchain
  - Add Docker/Container-based localnet infrastructure for multi-gateway testing

  Key Changes

  Gateway Probe Localnet Mode (nym-gateway-probe/)
  - New --localnet flag to test against container-based local network
  - TestMode enum to separate mainnet/localnet behavior
  - Two-hop WireGuard via UDP forwarder pattern (ported from VPN client)
  - LP-based registration flow testing with nested sessions
  - Mock ecash support via --mock-ecash flag

  Localnet Infrastructure (docker/localnet/)
  - localnet.sh - Orchestrates multi-container test network (nym-api, validators, gateways)
  - Dockerfile.localnet - Container image with nym binaries + test tooling
  - build_topology.py - Generates network topology from running containers
  - NAT/iptables configuration for proper gateway routing

  Gateway LP Registration (gateway/src/node/lp_listener/)
  - Removed unnecessary PSK from WireGuard peer registration (was breaking handshakes)

  Known Issues

  - Container networking to external internet becomes unstable after probe runs. Internal container traffic continues working. Workaround: restart containers between probe runs. Needs validation on real machines

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6274)


nym-gateway-probe/src/mode/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    SingleHop,
+    /// Entry LP + Exit LP (nested session forwarding) + WireGuard tunnel
+    TwoHop,
+    /// LP registration only - test handshake and registration, skip WireGuard
+    LpOnly,
+}
+
+impl TestMode {
+    /// Infer test mode from legacy boolean flags (backward compatibility)
+    pub fn from_flags(
```
jstuczyn, line 32: rather than doing this overcomplicated flag combination matching, why not just expose a `--test-mode` flag in CLI instead? it would then directly parse into a `TestMode`

nym-gateway-probe/src/mode/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+            TestMode::LpOnly
+        } else if test_lp_wg {
+            if has_exit_gateway {
+                TestMode::TwoHop
+            } else {
+                TestMode::SingleHop
+            }
+        } else if only_wireguard {
+            // WireGuard via authenticator (still uses mixnet path)
+            TestMode::Mixnet
```
jstuczyn, line 48: so what's the point of `only-wireguard` if in the end it ends up with the mixnet mode?

scripts/probe-localnet.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "Entry gateway: $ENTRY_ID"
+echo "Exit gateway:  $EXIT_ID"
+echo "Mode: $MODE"
+echo "---"
+
+cargo run -p nym-gateway-probe -- run-local \
+    --entry-gateway-identity "$ENTRY_ID" \
+    --entry-lp-address '127.0.0.1:41264' \
+    --exit-gateway-identity "$EXIT_ID" \
+    --exit-lp-address '192.168.65.6:41264' \
```
jstuczyn, line 33: why is exit api on the local host, but this one assumes address on a container network (that I assume is unstable and `X.X.65.6` will change once restarted?

common/nym-lp/src/session.rs
```diff
@@ -280,10 +286,40 @@ impl LpSession {
     ///
     /// Callers should use `None` for packet encryption/decryption during
     /// the handshake phase, and use the returned key for transport phase.
+    ///
+    /// Note: For sending packets during handshake, use `outer_aead_key_for_sending()`
+    /// which checks PSQ state to avoid encrypting before the responder can decrypt.
     pub fn outer_aead_key(&self) -> Option<OuterAeadKey> {
```
jstuczyn, line 295: whilst this code hasn't been changed directly, I'm not sure if it's a good practice to clone the key -> wouldn't we be able to get away by returning a reference instead?

gateway/src/node/lp_listener/handler.rs
```diff
@@ -408,7 +451,10 @@ impl LpConnectionHandler {
             .map_err(|e| GatewayError::LpProtocolError(format!("State machine error: {}", e)))?;
 
         // Get outer key before releasing borrow
-        let outer_key = state_machine.session().ok().and_then(|s| s.outer_aead_key());
+        let outer_key = state_machine
+            .session()
+            .ok()
```
jstuczyn, line 456: I wonder if this pattern is not going to cause us issues later on. we fail to get a session (i.e. we're in either `Closed` or `Processing` state), we ignore it and then just try continue with what is going to be an empty key. why not just stop here because realistically we can't do anything useful? 
I assume the same thing would apply in multiple places

gateway/src/node/lp_listener/handler.rs
```diff
@@ -811,22 +877,23 @@ impl LpConnectionHandler {
...
-                )));
-            }
-            Err(_) => {
-                inc!("lp_forward_failed");
-                return Err(GatewayError::LpConnectionError(
-                    "Target gateway connection timeout".to_string(),
-                ));
-            }
-        };
+        let mut target_stream =
```
jstuczyn, line 880: I've just traced the callstack of this guy - do I read it right that for every received client packet, we open a **new connection** to the target gateway **every single time**? 
even ignoring hitting the open file limit pretty quickly, that's rather inefficient. couldn't we keep connections alive, at least for some time, from previous attempts?

nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -135,6 +135,277 @@ func wgFreePtr(ptr unsafe.Pointer) {
...
+	exitLogger := device.NewLogger(device.LogLevelError, "exit: ")
+	exitDev := device.NewDevice(exitTun, conn.NewDefaultBind(), exitLogger)
+	defer exitDev.Close()
+
+	// Configure exit device - endpoint is the forwarder, NOT the actual exit gateway
+	var exitIpc strings.Builder
+	exitIpc.WriteString("private_key=")
+	exitIpc.WriteString(req.ExitPrivateKey)
+	// Set listen_port so the forwarder knows which port to accept packets from
+	exitIpc.WriteString(fmt.Sprintf("\nlisten_port=%d", DEFAULT_EXIT_WG_CLIENT_PORT))
```
jstuczyn, line 296: should we be hardcoding the default exit port? what if the gateway has configured it to something else? (assuming we allow such configuration to happen in the first place)

nym-gateway-probe/src/lib.rs
```diff
@@ -159,6 +186,10 @@ pub struct Probe {
     direct_gateway_node: Option<DirectoryNode>,
     /// Pre-queried exit gateway node (used when --exit-gateway-ip is specified for LP forwarding)
     exit_gateway_node: Option<DirectoryNode>,
+    /// Localnet entry gateway info (used when --entry-gateway-identity is specified)
+    localnet_entry: Option<TestedNodeDetails>,
```
jstuczyn, line 190: rather than calling it "localnet", how about something that implies you provide explicit details yourself, i.e. you could very well run it on mainnet but provide all the information manually to avoid the extra queries

nym-registration-client/src/lp_client/nested_session.rs
```diff
@@ -280,6 +270,161 @@ impl NestedLpSession {
...
+
+        // Step 3: Build registration request (credential already provided)
+        let wg_public_key = PeerPublicKey::new(wg_keypair.public_key().to_bytes().into());
+        let request =
+            LpRegistrationRequest::new_dvpn(wg_public_key, credential, ticket_type, client_ip);
+
+        tracing::trace!("Built registration request: {:?}", request);
+
+        // Step 4: Serialize the request
+        let request_bytes = bincode::serialize(&request).map_err(|e| {
```
jstuczyn, line 316: rathen than using the opaque `bincode::serialize`, can you make a helper that explicitly creates `impl bincode::Options` like we do elsewhere in such situations? this is because there are some changes in default behaviour between bincode 1.X and 2.X which would be quite painful to debug once (if) we migrate

jstuczyn:

PR #6282, opened 2025-12-11
Extend performance contract with custom measurements

dynco-nym:
New APIs
- define measurement
- retire measurement

Existing APIs
- extended so that measurement kind needs to be specified
- all performance metrics now need to be defined under some measurement kind

To do
- more unit testing of edge cases
- handle staleness check before deploying (commented in the code about it)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6282)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ members = [
     "common/nym-common",
     "common/config",
     "common/cosmwasm-smart-contracts/coconut-dkg",
+    "contracts/performance",
     "common/cosmwasm-smart-contracts/contracts-common",
     "common/cosmwasm-smart-contracts/contracts-common-testing",
     "common/cosmwasm-smart-contracts/easy_addr",
--- a/contracts/performance/src/queries.rs
+++ b/contracts/performance/src/queries.rs
@@ -1,16 +1,19 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::storage::{retrieval_limits, NYM_PERFORMANCE_CONTRACT_STORAGE};
+use std::collections::HashMap;
+
+use crate::storage::{NYM_PERFORMANCE_CONTRACT_STORAGE, retrieval_limits};
 use cosmwasm_std::{Addr, Deps, Order, StdResult};
 use cw_controllers::AdminResponse;
 use cw_storage_plus::Bound;
 use nym_performance_contract_common::{
-    EpochId, EpochMeasurementsPagedResponse, EpochNodePerformance, EpochPerformancePagedResponse,
-    FullHistoricalPerformancePagedResponse, HistoricalPerformance, LastSubmission,
-    NetworkMonitorInformation, NetworkMonitorResponse, NetworkMonitorsPagedResponse, NodeId,
-    NodeMeasurement, NodeMeasurementsResponse, NodePerformance, NodePerformancePagedResponse,
-    NodePerformanceResponse, NymPerformanceContractError, RetiredNetworkMonitorsPagedResponse,
+    AllNodeMeasurementsResponse, EpochId, EpochMeasurementsPagedResponse, EpochNodePerformance,
+    EpochPerformancePagedResponse, FullHistoricalPerformancePagedResponse, HistoricalPerformance,
+    LastSubmission, NetworkMonitorInformation, NetworkMonitorResponse,
+    NetworkMonitorsPagedResponse, NodeId, NodeMeasurement, NodeMeasurementsResponse,
+    NodePerformancePagedResponse, NodePerformanceResponse, NodePerformanceSpecific,
+    NymPerformanceContractError, RetiredNetworkMonitorsPagedResponse,
 };
 
 pub fn query_admin(deps: Deps) -> Result<AdminResponse, NymPerformanceContractError> {
@@ -20,6 +23,7 @@ pub fn query_admin(deps: Deps) -> Result<AdminResponse, NymPerformanceContractEr
         .map_err(Into::into)
 }
 
+// TODO dz return measurement_kind in response
 pub fn query_node_performance(
     deps: Deps,
     epoch_id: EpochId,
@@ -40,6 +44,51 @@ pub fn query_node_measurements(
         .results
         .may_load(deps.storage, (epoch_id, node_id))?;
     Ok(NodeMeasurementsResponse { measurements })
+}
+
+// TODO dz unit test
+pub fn query_node_measurements_specific(
+    deps: Deps,
+    epoch_id: EpochId,
+    node_id: NodeId,
+    measurement_kind: String,
+) -> Result<NodeMeasurementsResponse, NymPerformanceContractError> {
+    let measurements = NYM_PERFORMANCE_CONTRACT_STORAGE
+        .performance_results
+        .results_specific;
+
+    let key = (epoch_id, node_id, measurement_kind);
+    let measurements = measurements.may_load(deps.storage, key)?;
+
+    Ok(NodeMeasurementsResponse { measurements })
+}
+
+// TODO dz unit test
+pub fn query_all_node_measurements(
+    deps: Deps,
+    epoch_id: EpochId,
+    node_id: NodeId,
+) -> Result<AllNodeMeasurementsResponse, NymPerformanceContractError> {
+    let measurements = NYM_PERFORMANCE_CONTRACT_STORAGE
+        .performance_results
+        .results_specific;
+
+    // retrieve a list of currently defined measurements, only return results for those
+    // (storage may contain measurements that have since been deleted by admin -
+    // this way, they won't be retrieved)
+    let possible_measurements = NYM_PERFORMANCE_CONTRACT_STORAGE
+        .performance_results
+        .defined_measurements(deps.storage)?;
+    let mut node_measurements = HashMap::new();
+    for measure_name in possible_measurements {
+        let key = (epoch_id, node_id, measure_name.clone());
+        let node_measurement = measurements.may_load(deps.storage, key)?;
+        node_measurements.insert(measure_name, node_measurement);
+    }
+
+    Ok(AllNodeMeasurementsResponse {
+        measurements: node_measurements,
+    })
 }
 
 pub fn query_node_performance_paged(
@@ -97,6 +146,7 @@ pub fn query_node_performance_paged(
     })
 }
 
+// TODO dz unit test changes
 pub fn query_epoch_performance_paged(
     deps: Deps,
     epoch_id: EpochId,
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -34,6 +34,7 @@ members = [
     "common/nym-common",
     "common/config",
     "common/cosmwasm-smart-contracts/coconut-dkg",
+    "contracts/performance",
```
jstuczyn, line 37: don't add it to the main workspace - all contracts have their own sub workspace

contracts/performance/src/storage.rs
```diff
@@ -497,6 +559,76 @@ impl PerformanceResultsStorage {
...
+    }
+
+    /// assumes authorization had been performed
+    pub fn define_new_measurement_kind(
+        &self,
+        storage: &mut dyn Storage,
+        measurement_kind: MeasurementKind,
+    ) -> Result<(), NymPerformanceContractError> {
+        if self.is_measurement_defined(storage, measurement_kind.clone())? {
+            return Err(NymPerformanceContractError::InvalidInput(format!(
```
jstuczyn: that's a nit, so feel free to ignore it, but I think it would be cleaner if `MeasurementAlreadyDefined` was its own enum variant rather than being put alongside different errors within the opaque `InvalidInput`

contracts/performance/src/storage.rs
```diff
@@ -497,6 +559,76 @@ impl PerformanceResultsStorage {
...
+    }
+
+    /// assumes authorization had been performed
+    pub fn retire_measurement_kind(
+        &self,
+        storage: &mut dyn Storage,
+        measurement_kind: MeasurementKind,
+    ) -> Result<(), NymPerformanceContractError> {
+        if !self.is_measurement_defined(storage, measurement_kind.clone())? {
+            return Err(NymPerformanceContractError::InvalidInput(format!(
```
jstuczyn: same nit to do with the error variant

contracts/performance/src/transactions.rs
```diff
@@ -21,6 +21,50 @@ pub fn try_update_contract_admin(
...
+        .contract_admin
+        .assert_admin(deps.as_ref(), sender)?;
+
+    // validation
+    if measurement_kind.len() < 2
+        || measurement_kind.len() > 20
+        || !measurement_kind.is_ascii()
+        || measurement_kind.contains(char::is_whitespace)
+    {
+        return Err(NymPerformanceContractError::InvalidInput(format!(
```
jstuczyn: I think it'd be useful to return error message saying which particular check failed

contracts/performance/src/transactions.rs
```diff
@@ -311,4 +365,200 @@ mod tests {
...
+            },
+        };
+
+        #[allow(clippy::panic)]
+        #[test]
+        fn add_requires_admin() {
+            let mut tester = init_contract_tester();
+            let admin = tester.admin_msg();
+            let new_measurement = MeasurementKind::from("new-measurement");
+
```
jstuczyn, line 401: nit-ish. technically do validate it does indeed require admin privileges, you have to check that it does error out from any other account

contracts/performance/src/transactions.rs
```diff
@@ -311,4 +365,200 @@ mod tests {
...
+            let first_attempt = try_define_measurement_kind(
+                tester.deps_mut(),
+                &admin.sender,
+                new_measurement.clone(),
+            );
+            assert!(matches!(first_attempt, Ok(..)));
```
jstuczyn:
```suggestion
            assert!(first_attempt.is_ok())`
```

contracts/performance/src/storage.rs
```diff
@@ -482,10 +542,12 @@ impl PerformanceResultsStorage {
         epoch_id: EpochId,
         data: &NodePerformance,
     ) -> Result<(), NymPerformanceContractError> {
+        self.assert_measurement_defined(storage, data.measurement_kind.clone())?;
```
jstuczyn: this function should only really be concerned about writing data to storage and not checking whether it's actually allowed to insert it (i.e. ensuring kind exists). imo that should be done by the caller

contracts/performance/src/storage.rs
```diff
@@ -192,12 +194,21 @@ impl NymPerformanceContractStorage {
 
         let mut accepted_scores = 0;
         let mut non_existent_nodes = Vec::new();
+        let mut non_existent_measurement_kind = Vec::new();
 
         // 3. submit it
         if self.node_bonded(deps.as_ref(), first.node_id)? {
```
jstuczyn, line 202: this is just me thinking aloud. I wonder, would it perhaps simplify things if batch submission was always guaranteed to be for one kind at once?

contracts/performance/src/storage.rs
```diff
@@ -482,10 +542,12 @@ impl PerformanceResultsStorage {
         epoch_id: EpochId,
         data: &NodePerformance,
     ) -> Result<(), NymPerformanceContractError> {
+        self.assert_measurement_defined(storage, data.measurement_kind.clone())?;
+
         let performance = data.performance;
 
-        let key = (epoch_id, data.node_id);
-        let updated = match self.results.may_load(storage, key)? {
+        let key = (epoch_id, data.node_id, data.measurement_kind.clone());
```
jstuczyn, line 555: you could remove all the kind cloning all over the place if you changed your map key to `pub(crate) results: Map<(EpochId, NodeId, &'static str), NodeResults>,`

contracts/performance/src/queries.rs
```diff
@@ -107,18 +140,19 @@ pub fn query_epoch_performance_paged(
         .unwrap_or(retrieval_limits::NODE_EPOCH_PERFORMANCE_DEFAULT_LIMIT)
         .min(retrieval_limits::NODE_EPOCH_PERFORMANCE_MAX_LIMIT) as usize;
 
-    let start = start_after.map(Bound::exclusive);
+    let start = start_after.map(|node_id| Bound::exclusive((node_id + 1, String::new())));
```
jstuczyn, line 143: 1) wait, why are we skipping one node?
2) if you have to provide a rather opaque value to the bound, i.e. `String::new()`, perhaps the whole query signature should be changed? look at queries in other contracts, for pagination the caller is responsible for providing all elements of the subkey

contracts/performance/src/queries.rs
```diff
@@ -142,27 +176,51 @@ pub fn query_epoch_measurements_paged(
         .unwrap_or(retrieval_limits::NODE_EPOCH_MEASUREMENTS_DEFAULT_LIMIT)
         .min(retrieval_limits::NODE_EPOCH_MEASUREMENTS_MAX_LIMIT) as usize;
 
-    let start = start_after.map(Bound::exclusive);
+    let start = start_after.map(|node_id| Bound::exclusive((node_id + 1, String::new())));
 
+    // because API aggregates per NodeId, and the storage doesn't, we have to
+    // first collect all different measurements for a node and use an
+    // intermediary struct to map from storage to the object returned on the API
+    let mut measurements_per_node: HashMap<NodeId, Vec<(MeasurementKind, NodeResults)>> =
```
jstuczyn, line 184: I think we're just overcomplicating it here. we should just read the data as it's in the storage and let the caller sort it themselves

contracts/performance/src/queries.rs
```diff
@@ -176,26 +234,63 @@ pub fn query_full_historical_performance_paged(
...
+    // so we need an intermediary struct that collects all measurements
+    // per (epoch_id, node_id) pair (and calculates a performance)
+    let mut res_per_epoch_and_node: HashMap<(EpochId, NodeId), HashMap<MeasurementKind, Percent>> =
+        HashMap::new();
+    NYM_PERFORMANCE_CONTRACT_STORAGE
         .performance_results
         .results
         .range(deps.storage, start, None, Order::Ascending)
-        .take(limit)
+        // we can't cut a pagination limit here becasue we don't want to
```
jstuczyn, line 251: why not. just change the query parameters and let the caller resume from where it's meant to resume from

contracts/performance/src/storage.rs
dynco-nym: I don't think you can do that: this string is user-defined at runtime, you can't store it into `'static str`

contracts/performance/src/queries.rs
dynco-nym: `String::new()` is intended to start at `""` onwards for `MeasurementKind`. I.e. it's supposed to capture all measurement kinds under a certain node & epoch. 

In other pagination cases in the contract, you can "cut" anywhere across nodes & epochs. This signature guarantees that you cannot cut across measurement kinds of a node.

dynco-nym: I don't see how this would be more ergonomic to paginate through

page 1
```
(epoch_1, node_1, measurement_1)
(epoch_1, node_1, measurement_2)
(epoch_1, node_1, measurement_3)
(epoch_1, node_2, measurement_1)
(epoch_1, node_2, measurement_2)
```

page 2 
```
(epoch_1, node_2, measurement_3)
(epoch_1, node_3, measurement_1)
(epoch_1, node_3, measurement_2)
(epoch_1, node_3, measurement_3)
(epoch_1, node_4, measurement_1)
```

but I'm open to it if you think it's more valuable

contracts/performance/src/queries.rs
dynco-nym: storage layout is a bit clunky due to cw-storage constraints. Don't we care about API being nicer?

contracts/performance/src/storage.rs
dynco-nym: done

contracts/performance/src/transactions.rs
dynco-nym: done

contracts/performance/src/transactions.rs
dynco-nym: fixed in several places

contracts/performance/src/storage.rs
jstuczyn: you totally can thanks to how cosmwasm contracts work : )
look at for example the `NetworkMonitorsStorage`. internally it uses `pub(crate) authorised: Map<&'static Addr, NetworkMonitorDetails>,`. those keys are also defined at runtime as new entries are being added

contracts/performance/src/queries.rs
jstuczyn: > In other pagination cases in the contract, you can "cut" anywhere across nodes & epochs. This signature guarantees that you cannot cut across measurement kinds of a node.

I think it's perfectly fine to allow cutting across measurements kinds of a node. after all this is a paged query to get all measurements of a node and multiple calls are expected

contracts/performance/src/queries.rs
jstuczyn: what I meant (I think, it's been 3 weeks : D) is that we shouldn't bother pre-sorting values in the response (i.e. assigning it to per node). instead we should just return raw rows as they are in the storage and leave it to the caller to deal with it like we do with all other queries

PR #6286, opened 2025-12-12
LP Registration + Telescoping + Gateway Probe Localnet Mode

durch:
## Summary
- Combines LP registration protocol implementation
- Adds telescoping/nested sessions support
- Adds localnet mode for gateway-probe testing
- Integrates KKT + PSQ cryptographic primitives

This PR consolidates work from:
- drazen/lp-reg
- drazen/lp-reg-telescoping  
- drazen/gateway-probe-localnet-mode

## Test plan
- [ ] Run localnet with LP mode enabled
- [ ] Test gateway-probe with `--gateway-ip` flag against localnet
- [ ] Verify WireGuard tunnel establishment through LP registration
- [ ] Run existing integration tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6286)


nym-registration-client/src/lp_client/nested_session.rs
```diff
@@ -32,6 +33,15 @@ use nym_wireguard_types::PeerPublicKey;
 use std::net::IpAddr;
 use std::sync::Arc;
 
+/// Create explicit bincode options for consistent serialization across versions.
+///
+/// Using explicit options future-proofs against bincode 1.x/2.x default changes.
+fn bincode_options() -> impl Options {
```
jstuczyn: And it'd be amazing to extend it to other bincode instances in LP, for example inside `LpMessage::encode_content`

common/nym-lp/src/session.rs
```diff
@@ -409,6 +432,7 @@ impl LpSession {
             subsession_counter: AtomicU64::new(0),
             read_only: AtomicBool::new(false),
             successor_session_id: Mutex::new(None),
+            negotiated_version: std::sync::atomic::AtomicU8::new(1), // Default to version 1
```
jstuczyn, line 435: shouldn't this be somehow derived from the exchange rather than being hardcoded? just thinking of how backwards compatibility is going to work once we bump it up to v2

common/nym-lp/src/codec.rs
jstuczyn, line 1: do we really expect more than 65k message types? I'm not sure that's  really realistic. if anything, u8 would have probably been enough. and worst case we can use the reserved bytes to bump up the capacity

jstuczyn (approved):
image
how it feels to approve it now (with more work to come)

PR #6287, opened 2025-12-12
Client local storage for gateway details

jmwample:
Follow on to #6275 (changes applied to develop as part of #6249).

When we open a wss connection to the gateway we we do a DNS lookup that  could be prevented be cause we  digest away the information about the nodes IPs that clients keep track of.

We digest all of the relevant information about the node and its hostname, ports, and IPs
https://github.com/nymtech/nym/blob/develop/common/topology/src/node.rs 

```rust
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct EntryDetails {
    // to allow client to choose ipv6 preference, if available
    pub ip_addresses: Vec<IpAddr>,
    pub clients_ws_port: u16,
    pub hostname: Option<String>,
    pub clients_wss_port: Option<u16>,
}
```

into a single URL / String and then ignore the rest.
```
format!("wss://{hostname}:{wss_port}")

format!("ws://{any_ip}:{}", entry.clients_ws_port)
```

Then when we go to use this url if it has a hostname we resolve it again. 

In order to avoid the resolution the previous fix swapped to using ws only and preventing wss usage. 

## Strategy

This PR keeps the entry details, only creating the url at the last minute when we use the information so that we have all port, hostname, IP information to be used whenever it is needed.
This allows wss usage while preventing the DNS lookup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6287)


jmwample:
I am working on a vpn-client change to use this and test vpnc/vpnd with.

common/client-core/gateways-storage/fs_gateways_migrations/20251212120000_store_gateway_details.sql
```diff
@@ -0,0 +1,21 @@
...
+
+CREATE TABLE remote_gateway_details_temp
+(
+    gateway_id_bs58                          TEXT NOT NULL UNIQUE PRIMARY KEY REFERENCES registered_gateway (gateway_id_bs58),
+    derived_aes256_gcm_siv_key               BLOB NOT NULL,
+    gateway_details                          TEXT NOT NULL CHECK (json_valid(gateway_details)),
+    expiration_timestamp                     DATETIME NOT NULL
+);
+
+-- keep none, the gateways listener URL does not contain the gateway details information
```
jstuczyn, line 14: but then we're losing our shared keys

common/client-core/gateways-storage/src/backend/fs_backend/manager.rs
```diff
@@ -156,15 +156,16 @@ impl StorageManager {
         &self,
         remote: &RawRemoteGatewayDetails,
     ) -> Result<(), sqlx::Error> {
+        let details =
+            serde_json::to_string(&remote.published_data.gateway_details).expect("AHHHHHHHHH");
```
jstuczyn, line 160: seems something is missing here : )

common/client-core/gateways-storage/src/types.rs
```diff
@@ -188,16 +187,16 @@ impl GatewayPublishedData {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
 pub struct RawGatewayPublishedData {
-    pub gateway_listener: String,
-    pub fallback_listener: Option<String>,
+    #[cfg_attr(feature = "sqlx", sqlx(json))]
```
jstuczyn, line 190: does json actually work with sqlx + sqlite? i.e. have you run it before? I remember I had problems with that combination in the past and json was only supported with postgres. but that was in an older version so it might have changed

common/client-core/gateways-storage/src/types.rs
```diff
@@ -188,16 +187,16 @@ impl GatewayPublishedData {
...
+    pub gateway_details: EntryDetails,
     pub expiration_timestamp: OffsetDateTime,
 }
 
 impl<'a> From<&'a GatewayPublishedData> for RawGatewayPublishedData {
     fn from(value: &'a GatewayPublishedData) -> Self {
         Self {
-            gateway_listener: value.listeners.primary.to_string(),
-            fallback_listener: value.listeners.fallback.as_ref().map(|uri| uri.to_string()),
+            // fallback_listener: value.listeners.fallback.as_ref().map(|uri| uri.to_string()),
```
jstuczyn, line 198: dead code

common/client-core/gateways-storage/src/types.rs
```diff
@@ -207,28 +206,15 @@ impl TryFrom<RawGatewayPublishedData> for GatewayPublishedData {
...
-            })
-            .transpose()?;
+        // let details = serde_json::from_str(&value.gateway_listener).map_err(|source| {
+        //     BadGateway::MalformedDetailsNoId  {
+        //         raw_details: value.gateway_listener.clone(),
+        //         source,
+        //     }
+        // })?;
 
         Ok(GatewayPublishedData {
```
jstuczyn, line 216: this now looks to be infallible, so change the `TryFrom` into a normal `From` instead

common/client-core/gateways-storage/src/types.rs
```diff
@@ -188,16 +187,16 @@ impl GatewayPublishedData {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
 pub struct RawGatewayPublishedData {
-    pub gateway_listener: String,
-    pub fallback_listener: Option<String>,
+    #[cfg_attr(feature = "sqlx", sqlx(json))]
+    pub gateway_details: EntryDetails,
```
jstuczyn, line 191: I think, for the purposes of the database schema, it'd make more sense to redefine that type, even if it would duplicate all the data. The reason is that `EntryDetails` is an external type that can change at any time and then the database would suddenly break due to changes in serialisation.

common/client-core/src/cli_helpers/client_add_gateway.rs
```diff
@@ -168,7 +167,6 @@ where
         identity: gateway_details.gateway_id,
         active: common_args.set_active,
         typ: gateway_registration.details.typ().to_string(),
-        endpoint: Some(gateway_details.published_data.listeners.primary.clone()),
-        fallback_endpoint: gateway_details.published_data.listeners.fallback.clone(),
+        endpoint: Some(gateway_details.published_data.details.clone()),
```
jstuczyn, line 170: that's a nit, but I think it'd make more sense to rename that field. it's no longer just an `endpoint`

common/client-core/src/init/helpers.rs
```diff
@@ -201,25 +200,28 @@ async fn connect(endpoint: &str) -> Result<WsConn, ClientCoreError> {
 }
 
 #[cfg(target_arch = "wasm32")]
-async fn connect(endpoint: &str) -> Result<WsConn, ClientCoreError> {
-    JSWebsocket::new(endpoint).map_err(|_| ClientCoreError::GatewayJsConnectionFailure)
+async fn connect(endpoint: &EntryDetails) -> Result<WsConn, ClientCoreError> {
+    let uri = endpoint
+        .ws_entry_address(false)
```
jstuczyn, line 205: why are we defaulting to no ipv6 here?

common/client-core/src/init/types.rs
```diff
@@ -39,52 +39,33 @@ impl SelectedGateway {
...
+
+        if must_use_tls
+            && (gateway_details.hostname.is_none() || gateway_details.clients_wss_port.is_none())
+        {
+            return Err(ClientCoreError::UnsupportedWssProtocol {
+                gateway: node.identity_key.to_base58_string(),
+            });
+        }
+
+        if prefer_ipv6 && gateway_details.ip_addresses.iter().all(|i| i.is_ipv4()) {
```
jstuczyn, line 59: should we be erroring out here though? prefer ipv6 != require ipv6. it never did. it would just choose ipv6 over ipv4 **if available**

common/client-core/src/init/websockets.rs
```diff
@@ -1,43 +1,31 @@
...
 
 #[cfg(not(target_arch = "wasm32"))]
 pub(crate) async fn connect_async(
-    endpoint: &str,
+    endpoint: &EntryDetails,
 ) -> Result<(WebSocketStream<MaybeTlsStream<TcpStream>>, Response), ClientCoreError> {
-    let resolver = HickoryDnsResolver::default();
-    let uri = Url::parse(endpoint).map_err(|_| ClientCoreError::InvalidUrl(endpoint.to_owned()))?;
+    let uri = endpoint
+        .ws_entry_address(false)
```
jstuczyn, line 16: same thing

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -199,19 +193,13 @@ impl<C, St> GatewayClient<C, St> {
...
-            );
-        }
-        let (ws_stream, _) = connect_async_with_fallback(
-            &self.gateway_addresses,
+        debug!(
+            "Attempting to establish connection to gateway at: {:?}",
+            self.gateway_details
+        );
+
+        let (ws_stream, _) = connect_async(
```
jstuczyn, line 201: now even though gateway might have multiple ips, we will not attempt a fallback - is this what we want?

common/topology/src/node.rs
```diff
@@ -59,80 +89,17 @@ pub struct RoutingNode {
...
-    }
-
     pub fn identity(&self) -> ed25519::PublicKey {
         self.identity_key
     }
 }
 
+impl std::fmt::Display for EntryDetails {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
```
jstuczyn, line 99: I don't think it makes sense to use `Debug` implementation for `Display`. imo it should either be properly formatted or not implemented at all if it doesn't make sense

nym-api/src/network_monitor/monitor/preparer.rs
```diff
@@ -352,7 +352,7 @@ impl PacketPreparer {
             .collect();
 
         GatewayPackets::new(
-            route.gateway_clients_address(),
+            route.gateway_details().expect("AHHHHHHH"),
```
jstuczyn, line 355: ?

nym-api/src/network_monitor/monitor/sender.rs
```diff
@@ -14,13 +14,14 @@ use nym_bandwidth_controller::BandwidthController;
...
+use nym_gateway_client::client::GatewayConfig;
 use nym_gateway_client::error::GatewayClientError;
 use nym_gateway_client::{
     AcknowledgementReceiver, GatewayClient, MixnetMessageReceiver, PacketRouter, SharedSymmetricKey,
 };
 use nym_sphinx::forwarding::packet::MixPacket;
 use nym_task::ShutdownToken;
+use nym_topology::EntryDetails;
 use pin_project::pin_project;
 use sqlx::__rt::timeout;
```
jstuczyn, line 26: I know it's not your change, but I've just noticed it. wtf is this import

nym-api/src/network_monitor/monitor/sender.rs
```diff
@@ -30,14 +31,13 @@ use std::sync::Arc;
 use std::task::Poll;
 use std::time::Duration;
 use tracing::{debug, error, info, trace, warn};
-use url::Url;
 
 const TIME_CHUNK_SIZE: Duration = Duration::from_millis(50);
 
 pub(crate) struct GatewayPackets {
     /// Network address of the target gateway if wanted to be accessed by the client.
```
jstuczyn, line 38: the comment no longer makes sense

jstuczyn (changes requested):

jstuczyn:
oh. and quick question I forgot to ask: have you checked whether it works in WASM with the new expected data structures?

PR #6298, opened 2025-12-18
Socks5 in GW probe

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6298)


nym-node-status-api/nym-node-status-agent/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "nym-node-status-agent"
-version = "1.0.7"
+version = "1.0.8-gw-socks5"
```
jstuczyn: does it make sense to annotate the version with gw-socks5 suffix? it will either have it or not. i.e. there wouldn't be `1.0.8` that does not have it

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -601,6 +601,10 @@ where
         );
 
         let available_gateways = self.available_gateways().await?;
+        info!("Listing all available gateways in topology:");
+        for node in available_gateways.iter() {
+            info!("{}", node.identity_key.to_base58_string());
```
jstuczyn: I don't think this should be at `info` level. on mainnet we have +300 gateways so on client startup you will see over 300 logs every single time. for debug/trace I think that's useful information though. maybe also include `node_id` there for debugging purposes?

nym-gateway-probe/src/run.rs
```diff
@@ -83,6 +88,11 @@ enum Commands {
         #[arg(long)]
         config_dir: Option<PathBuf>,
     },
+    Socks5 {
```
jstuczyn: shouldn't socks5 test be part of the "normal" testing suite? maybe just introduce a flag in "normal" args to toggle it

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+}
+
+pub struct HttpsConnectivityTest {
+    test_count: u64,
+    mixnet_client_timeout: Duration,
+}
+
+/// endpoint to test against
+/// https://www.quicknode.com/docs/ethereum/web3_clientVersion
+const TARGET_URL: &str = "https://docs-demo.quiknode.pro";
```
jstuczyn: why are we testing against this quite random website? why not something we control or something that's more used for this purpose

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+
+    async fn perform_https_request(&self, client: &reqwest::Client) -> HttpsConnectivityResult {
+        use tokio::time::Instant;
+
+        let mut result = HttpsConnectivityResult::default();
+        let start = Instant::now();
+        match tokio::time::timeout(
+            self.mixnet_client_timeout,
+            client
+                .post(TARGET_URL)
```
jstuczyn: wouldn't a simple `get` be easier? then we could, say, test against our own website

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+
+impl HttpsConnectivityTest {
+    pub fn new(test_count: u64, mixnet_client_timeout: u64) -> Self {
+        Self {
+            test_count: std::cmp::max(test_count, 1),
+            mixnet_client_timeout: Duration::from_secs(mixnet_client_timeout),
+        }
+    }
+
+    pub async fn run_tests(self, socks5_url: String) -> HttpsConnectivityResult {
```
jstuczyn: that's a bit of nit (or maybe not), but should the testing logic be defined in `types.rs`?

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+                result.https_latency_ms = Some(elapsed.as_millis() as u64);
+                debug!(
+                    "HTTPS test completed: status={}, latency={}ms",
+                    status.as_u16(),
+                    elapsed.as_millis()
+                );
+            }
+            Ok(Err(e)) => {
+                warn!("HTTPS request failed: {}", e);
+                if result.error.is_none() {
```
jstuczyn: but if the error is not none, we'd presumably also want to attach this information, no?

jstuczyn: actually, looking deeper at the code, there's no codepath that sets the error beforehand, so those assignments here should be unconditional

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+                && let Some(latency_ms) = interim_res.https_latency_ms
+            {
+                successful_runs += 1;
+                result.https_latency_ms = Some(
+                    result
+                        .https_latency_ms
+                        .map_or(latency_ms, |existing| existing + latency_ms),
+                );
+                result.https_success = true;
+                result.https_status_code = interim_res.https_status_code;
```
jstuczyn: what if (though unlikely) the status code has changed between runs? we'll only retain the last one which might or might not be what we want

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+                info!("{}/{} latency: {}ms", i, self.test_count, latency_ms);
+            } else if let Some(new_error) = interim_res.error {
+                result.error = Some(result.error.map_or(new_error.clone(), |existing| {
```
jstuczyn: I think it'd be useful to attach information which run has failed because there's be no way to distinguish say failure at `1,4`, vs `8,9`

nym-gateway-probe/src/types.rs
```diff
@@ -122,6 +126,169 @@ impl Exit {
...
+                    format!("{},{}", existing, new_error)
+                }))
+            }
+
+            // too many failed runs: return early
+            if successful_runs < 2 && i - successful_runs > 2 {
```
jstuczyn: nit: it would improve readibility to assign `let unsuccessful_runs = i - successful_runs`

nym-gateway-probe/src/nodes.rs
```diff
@@ -199,6 +202,16 @@ impl NymApiDirectory {
             .map(|(id, _)| *id)
     }
 
+    pub fn random_exit_with_nr(&self) -> anyhow::Result<NodeIdentity> {
+        info!("Selecting random gateway with NR enabled");
+        self.nodes
+            .iter()
+            .filter(|(_, n)| n.described.description.ip_packet_router.is_some())
+            .choose(&mut rand::thread_rng())
+            .ok_or(anyhow!("no gateways running NR available"))
```
jstuczyn: anyhow's `context` works for `Option`:

```suggestion
            .context("no gateways running NR available")
```

nym-gateway-probe/src/lib.rs
```diff
@@ -170,20 +185,74 @@ impl Probe {
...
 
+    pub async fn test_socks5_only(
+        self,
+        directory: NymApiDirectory,
+        gateway_key: Option<String>,
+        network_details: NymNetworkDetails,
+    ) -> anyhow::Result<ProbeResult> {
+        let exit_gateway = match gateway_key {
+            Some(gateway_key) => NodeIdentity::from_base58_string(gateway_key)?,
+            None => directory.random_exit_with_nr()?,
```
jstuczyn: what's the point of testing a random gateway for socks5?

nym-gateway-probe/src/lib.rs
```diff
@@ -475,9 +549,34 @@ impl Probe {
             WgProbeResults::default()
         };
 
+        // test failure doesn't stop further tests
+        let socks5_outcome = {
+            if let Some(ref nr_details) = node_info.network_requester_details {
```
jstuczyn: nit: you have exactly the same match logic in multiple places

nym-gateway-probe/src/lib.rs
```diff
@@ -475,9 +549,34 @@ impl Probe {
...
+                match do_socks5_connectivity_test(
+                    &nr_details.address,
+                    NymNetworkDetails::new_from_env(),
```
jstuczyn: what if env is not set?

nym-gateway-probe/src/lib.rs
```diff
@@ -775,6 +875,109 @@ async fn do_ping_exit(
...
+        Err(e) => {
+            error!("Failed to establish SOCKS5 connection: {}", e);
+            results.https_connectivity =
+                HttpsConnectivityResult::with_error(format!("SOCKS5 connection failed: {}", e));
+            return Ok(results);
+        }
+    };
+
+    info!("Waiting for network topology to be ready...");
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
jstuczyn: couldn't we wait for some condition rather than always wait 10s? what if topology takes longer than that? (or way less than that)?

nym-gateway-probe/src/lib.rs
```diff
@@ -775,6 +875,109 @@ async fn do_ping_exit(
     listen_for_icmp_ping_replies(mixnet_client, our_ips).await
 }
 
+/// Creates a SOCKS5 proxy connection through the mixnet to the exit GW
+/// and performs necessary tests.
+#[instrument(level = "info", name = "socks5_test", skip_all)]
+async fn do_socks5_connectivity_test(
```
jstuczyn, line 1684: I think we should still investigate just creating a single client and sending NR packets to exit gateways rather than starting local socks5 listeners.

nym-node-status-api/nym-node-status-agent/Cargo.toml
dynco-nym: this won't stay, I needed a version which wouldn't conflict with what we'll use in prod to build & deploy in a test env

sdk/rust/nym-sdk/src/mixnet/client.rs
dynco-nym: corrected

nym-gateway-probe/src/run.rs
dynco-nym: this is purely cosmetic: `#[command(flatten)]` ensures the listed struct "flattens" so its fields are directly configurable as CLI arguments. 

Follows the same pattern as `CredentialArgs` and `NetstackArgs`

nym-gateway-probe/src/types.rs
mmsinclair: we want it to simulate an HTTP POST to an RPC - @dynco-nym plans to use a public Cloudflare hosted one where the availability will be the best it can get

nym-gateway-probe/src/lib.rs
mmsinclair: yes we can do that later, this will be ok for now, because it is the simplest for deployment

nym-gateway-probe/src/types.rs
dynco-nym: moved to its own module

nym-gateway-probe/src/types.rs
dynco-nym: refactored that part

nym-gateway-probe/src/lib.rs
dynco-nym: refactored

nym-gateway-probe/src/lib.rs
dynco-nym: I exposed a method from SDK to poll topology, check it out

nym-gateway-probe/src/lib.rs
dynco-nym: this pattern was used all across the probe, but I refactored all instances of it

now it's only initialized once at probe start, so it fails early if misconfigured

nym-gateway-probe/src/lib.rs
dynco-nym: it's the same functionality for non-socks5 test: if no gw provided, test a random one

nym-gateway-probe/src/types.rs
dynco-nym: added fallbacks

nym-gateway-probe/src/types.rs
dynco-nym: that's fine: in case of error, info here is just a hint at what's wrong

nym-node-status-api/nym-node-status-agent/Cargo.toml
dynco-nym: TODO revert before merge

nym-gateway-probe/src/types.rs
dynco-nym: same logic as [here](https://github.com/nymtech/nym/pull/6298#discussion_r2705638284)

nym-gateway-probe/src/lib.rs
```diff
@@ -290,6 +320,44 @@ impl Probe {
         self
     }
 
+    pub async fn test_socks5_only(
```
jstuczyn: this function doesn't seem to be used anywhere

nym-gateway-probe/src/lib.rs
jstuczyn: sure, but is there any use case for such behaviour? i.e. **only** socks5 for a completely random node?

nym-gateway-probe/src/lib.rs
```diff
@@ -695,11 +800,15 @@ impl Probe {
...
         only_wireguard: bool,
         use_mock_ecash: bool,
+        network_details: NymNetworkDetails,
     ) -> anyhow::Result<ProbeResult>
     where
         T: MixnetClientStorage + Clone + 'static,
         <T::CredentialStore as CredentialStorage>::StorageError: Send + Sync,
     {
+        let Some(directory) = directory else {
+            bail!("You need to provide NYM API through environment")
```
jstuczyn, line 820: I think we should always just require passing `NymApiDirectory` - even if we mock it for local purposes. because this error can be a bit misleading imo

nym-gateway-probe/src/lib.rs
```diff
@@ -1545,6 +1666,130 @@ async fn do_ping_exit(
...
+    info!(
+        "Network Requester identity: {}",
+        nr_recipient.identity().to_base58_string()
+    );
+
+    // create ephemeral SOCKS5 client
+    let socks5_config = Socks5::new(network_requester_address.to_string());
+
+    // Create debug config similar to main probe
+    let mut debug_config = nym_client_core::config::DebugConfig::default();
```
jstuczyn: can't we then just extra the `DebugConfig` creation to a separate function given we're making it the same way and it's quite verbose

nym-gateway-probe/src/lib.rs
```diff
@@ -1545,6 +1666,130 @@ async fn do_ping_exit(
...
+    debug_config.cover_traffic.disable_loop_cover_traffic_stream = true;
+    debug_config.topology.ignore_egress_epoch_role = true;
+    // since we define both entry & exit gateways to be the same tested GW,
+    // this shouldn't negatively affect mixnet layers but it will force route
+    // construction in case GW would get filtered out on topology refresh
+    debug_config.topology.minimum_gateway_performance = 0;
+
+    // Verify the NR gateway exists in the directory with exit_nr role
+    let nr_gateway_id = nr_recipient.gateway();
+    if let Err(e) = directory.exit_gateway_nr(&nr_gateway_id) {
```
jstuczyn, line 1737: how can we even get here? if we have `network_requester_address`, the gateway MUST have a NR, no? maybe we should restructure the code a bit to avoid this weird case?

nym-gateway-probe/src/lib.rs
```diff
@@ -1545,6 +1666,130 @@ async fn do_ping_exit(
...
+        Err(e) => {
+            error!("Failed to establish SOCKS5 connection: {}", e);
+            results.https_connectivity =
+                HttpsConnectivityResult::with_error(format!("SOCKS5 connection failed: {}", e));
+            return Ok(results);
+        }
+    };
+
+    info!("Waiting for network topology to be ready...");
+    let topology_timeout = Duration::from_secs(60);
```
jstuczyn: rather than waiting for topology, can't you call `with_topology_provider` on the builder and set your hardcoded topology that you had already pulled when starting the probe?

nym-gateway-probe/src/lib.rs
```diff
@@ -1545,6 +1666,130 @@ async fn do_ping_exit(
...
+    }
+
+    let test =
+        HttpsConnectivityTest::new(test_run_count, mixnet_client_timeout, json_rpc_endpoints);
+    results.https_connectivity = test
+        .run_tests(socks5_client.socks5_url(), failure_count_cutoff)
+        .await;
+
+    // cleanup
+    socks5_client.disconnect().await;
```
jstuczyn, line 1808: we need to make sure this is always called. for example if `socks5_client` errors out waiting for topology, it performs an early return without shutdown, meaning it will leave some tokio tasks dangling

nym-gateway-probe/src/socks5_test.rs
```diff
@@ -0,0 +1,351 @@
...
+        final_result
+    }
+
+    async fn perform_https_request(&self, client: &reqwest::Client) -> SingleHttpsTestResult {
+        use tokio::time::Instant;
+
+        let start = Instant::now();
+        let mut error_msg = String::new();
+
+        for endpoint in self.json_rpc_test_endpoints.iter() {
```
jstuczyn: given those endpoints are set by cli, how can we guarantee all of them are going to accept `eth_chainid` request?

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -601,6 +601,14 @@ where
         );
 
         let available_gateways = self.available_gateways().await?;
+        debug!("Listing all available gateways in topology:");
```
simonwicky: Is that a leftover debug print?

nym-gateway-probe/src/lib.rs
dynco-nym: removed, we can addd it if necessary

nym-gateway-probe/src/lib.rs
dynco-nym: this was not my idea. Look at the [code](https://github.com/nymtech/nym/blob/e2be2b0b34ee483947318911f1733b69d7925218/nym-gateway-probe/src/run.rs#L323) on `develop`, sometimes it doesn't pass directory, so I need to handle the optionality

considering it's [constructed](https://github.com/nymtech/nym/blob/e2be2b0b34ee483947318911f1733b69d7925218/nym-gateway-probe/src/run.rs#L379) from `NymNetworkDetails` (which we always load from env), can we just load it from env in every case and remove the `Option` ?

nym-gateway-probe/src/lib.rs
dynco-nym: reused existing function

nym-gateway-probe/src/socks5_test.rs
dynco-nym: `eth_chainid` should be universally supported. There's some variability between what provider offers which endpoint, but this is a pretty basic one. `socks5` error field should note the kind of error. Do you have another suggestion?

nym-gateway-probe/src/lib.rs
dynco-nym: added

sdk/rust/nym-sdk/src/mixnet/client.rs
dynco-nym: yes, removed

nym-gateway-probe/src/lib.rs
dynco-nym: since these come from different fields from node's description, the question becomes: is it possible for a node to have `description.network_requester` set, but not to have its role `description.declared_role` be `exit_nr` ?

nym-gateway-probe/src/lib.rs
dynco-nym: removed wait for topology so now all socks5 testing is done within a single function. 

After it returns (regardless of outcome), we disconnect

nym-gateway-probe/src/socks5_test.rs
dynco-nym: added a check at node startup, it will fail if misconfigured

sdk/rust/nym-sdk/src/mixnet/client.rs
simonwicky: What about the loop after it? ( https://github.com/nymtech/nym/pull/6298/changes#diff-27cedba5bb3066c497edd6dd7bfa9eef6de3800f7c1492b01283d7ccde456f15R605)

dynco-nym: will remove, must've missed this

simonwicky (approved):

dynco-nym:
After rebasing on `develop`, probe wasn't compatible with the network anymore due to the inclusion of `get_all_described_nodes_v2` and related work (which hasn't been released yet). 

So I built `1.0.8` version off this non-rebased branch because that's the version we've been testing are will release as `1.0.8`. 

I don't want to overwrite this branch with a rebased one, I will open a [separate PR](https://github.com/nymtech/nym/pull/6393) for it & merge that. 

This is kept for future reference if we need to rebuild it or see what went into `1.0.8`.

PR #6316, opened 2026-01-08
Update nix to v0.30.1

pronebird:
- Update `nix` to `v0.30.1`
- Use new `Flock` instead of deprecated `flock()`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6316)

### DIFF
--- a/tools/nymvisor/src/error.rs
+++ b/tools/nymvisor/src/error.rs
@@ -178,12 +178,6 @@ While the stored info point to:\n{stored_info:#?}"
         source: io::Error,
     },
 
-    #[error("coundn't unlock the lock at {lock_path} with error code {libc_code}")]
-    LockFileUnlockFailure {
-        lock_path: PathBuf,
-        libc_code: Errno,
-    },
-
     #[error("could not remove the lock file at {} after performing binary upgrade: {source}", path.display())]
     LockFileRemovalFailure {
         path: PathBuf,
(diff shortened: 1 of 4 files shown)


tools/nymvisor/src/error.rs
```diff
@@ -178,6 +178,12 @@ While the stored info point to:\n{stored_info:#?}"
         source: io::Error,
     },
 
+    #[error("coundn't unlock the lock at {lock_path} with error code {libc_code}")]
```
jstuczyn: typo:
```suggestion
    #[error("couldn't unlock the lock at {lock_path} with error code {libc_code}")]
```

tools/nymvisor/src/upgrades/mod.rs
```diff
@@ -115,6 +114,13 @@ pub(crate) async fn perform_upgrade(config: &Config) -> Result<UpgradeResult, Ny
         source,
     })?;
 
+    locked_file
```
jstuczyn: do we need an explicit removal of the flock if we had just removed the file? 
additionally, `Flock` itself will perform the unlock on `drop` anyway

pronebird: We don't. But I thought it would be nice to catch the error. Maybe I am overthinking this?

simonwicky (approved):

tools/nymvisor/src/upgrades/mod.rs
pronebird: Applied changes based on your suggestion.

PR #6318, opened 2026-01-11
Max/asyncread asyncwrite nym client

mfahampshire:
Breaking up #6129 into more reasonably sized chunks. 

This adds `AsyncRead/Write` to the client and adds a streaming mode to the `MIxnetClient` utilising it via substreams. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6318)


common/nym-connection-monitor/src/sync_self_ping.rs
```diff
@@ -22,23 +21,22 @@ pub async fn self_ping_and_wait(
...
-        })
-        .collect::<Vec<_>>()
-        .await;
+async fn send_self_pings(
+    our_address: Recipient,
+    mixnet_client: &mut MixnetClient,
+) -> Result<Vec<u64>> {
+    let mut request_ids = Vec::with_capacity(3);
+
+    for _ in 1..=3 {
```
jstuczyn: note that this changes semantics of the original call. the former was performing all 3 pings concurrently. the updated variant does it sequentially - make sure this is what you want

common/client-core/src/client/inbound_messages.rs
```diff
@@ -185,4 +200,336 @@ impl InputMessage {
         self.set_max_retransmissions(max_retransmissions);
         self
     }
+    #[allow(clippy::expect_used)]
+    pub fn serialized_size(&self) -> u64 {
```
jstuczyn: this doesn't seem to be used anywhere

common/client-core/src/client/inbound_messages.rs
```diff
@@ -185,4 +200,336 @@ impl InputMessage {
...
+
+    fn encode(&mut self, item: &[u8], buf: &mut BytesMut) -> Result<(), Self::Error> {
+        let mut codec = InputMessageCodec;
+        let input_message = InputMessage::simple(item, self.0);
+        codec.encode(input_message, buf)?;
+        Ok(())
+    }
+}
+
+pub struct InputMessageCodec;
```
jstuczyn: are we sure `InputMessage` is the right abstraction for the AsyncRead/AsyncWrite impl? if you're concerned about reading/writing bytes, why should you care about internal `TransmissionLane` (that's only used for backpressure)? 

wouldn't it be sufficient to just use the inbuilt `LengthDelimitedCodec` wrapped around simple `Vec<u8>` and then call normal "send" within the client on that message?

common/nymsphinx/forwarding/src/packet.rs
```diff
@@ -183,4 +188,31 @@ impl MixPacket {
     }
 }
 
+// MAX TODO implement for v1 as well for back compat? - this was added in the original asyncread/write work when we only had one v
+impl Serialize for MixPacket {
```
jstuczyn: given we have to implement `serde` for internal `MixPacket` makes me think we chose wrong abstraction. client really shouldn't be concerned about this guy. it's just a wrapper for data that's eventually going to get encoded in websocket messages for the gateway

common/nymsphinx/src/lib.rs
```diff
@@ -22,3 +22,10 @@ pub use nym_sphinx_framing as framing;
 
 // TEMP UNTIL FURTHER REFACTORING
 pub use preparer::payload::NymPayloadBuilder;
+
+fn make_bincode_serializer() -> impl bincode::Options {
```
jstuczyn: not sure how to feel about using `bincode` inside sphinx itself...

common/nymsphinx/src/receiver.rs
```diff
@@ -15,10 +19,13 @@ use nym_sphinx_chunking::reconstruction::MessageReconstructor;
...
     PacketEncryptionAlgorithm, PacketHkdfAlgorithm, ReplySurbEncryptionAlgorithm,
 };
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use tokio_util::bytes::{Buf, BytesMut};
+use tokio_util::codec::{Decoder, Encoder};
 
 // TODO: should this live in this file?
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
```
jstuczyn: given that this struct is simple enough - do we really need to use bincode for it with whole serde pipeline?
looking at it, we just want to call `encode` / `decode`, so how about the following:
first byte would be either 0 or 1 to annotate whether `sender_tag` is set. if it's `1`, it's followed by `16` bytes of the tag and then the payload. if it's `0`, it's followed by payload directly

nym-outfox/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+// MAX: temp ignore deprecated, can be dealt with in its own PR
+#![allow(deprecated)] // silences clippy warning: deprecated associated function `chacha20::cipher::generic_array::GenericArray::<T, N>::from_slice`: please upgrade to generic-array 1.x - TODO
```
jstuczyn, line 2: given this repeats in multiple places - can you make a ticket for it?

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -60,6 +65,24 @@ pub struct MixnetClient {
     _buffered: Vec<ReconstructedMessage>,
     pub(crate) forget_me: ForgetMe,
     pub(crate) remember_me: RememberMe,
+
+    // internal state used for the `AsyncRead` implementation
+    _read: ReadBuffer,
```
jstuczyn: apart from the buffer, I think we need an additional state flag/mode within the struct itself, so that if somebody is using `AsyncRead`/`AsyncWrite` they'd be unable to call "normal" send/receive

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -280,6 +304,25 @@ impl MixnetClient {
...
+        cx: &mut Context<'_>,
+    ) -> Poll<tokio::io::Result<()>> {
+        if self._read.buffer.len() < buf.capacity() {
+            // let written = self._read.buffer.len();
+            buf.put_slice(&self._read.buffer);
+            self._read.clear();
+            Poll::Ready(Ok(()))
+        } else {
+            let written = buf.capacity();
+            buf.put_slice(&self._read.buffer[..written]);
```
jstuczyn: use `split` method on the `BytesMut`

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +331,193 @@ pub struct MixnetClientSender {
...
+                    "failed to read message from input",
+                )))
+            }
+            Poll::Pending => return Poll::Pending,
+            Poll::Ready(None) => return Poll::Ready(Ok(0)),
+        };
+
+        let msg_size = msg.serialized_size();
+
+        let mut fut = pin!(self.client_input.send(msg));
```
jstuczyn: you're creating a new send future every single time you call `poll_write` without letting the previous one resolve. you will have to actually store the future on the `MixnetClient` itself and check for its existence in subsequent calls

jstuczyn: just a random thought (don't know if it's a good idea!): what if `AsyncRead` / `AsyncWrite` was changed to the simplest possible variant where they only read/write bytes to internal client buffers without any encoding. but there would be a dedicated internal client task/thread that periodically would check those buffers and send/recv from the mixnet on a timer based on their contents

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +331,193 @@ pub struct MixnetClientSender {
...
+            ))),
+            Poll::Pending => Poll::Pending,
+        }
+    }
+
+    fn poll_flush(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<std::prelude::v1::Result<(), std::io::Error>> {
+        Sink::poll_flush(self, cx).map_err(|_| std::io::Error::other("failed to flush the sink"))
```
jstuczyn: I'm not sure if calling the internal Sink impl for this guy is a good call

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +331,193 @@ pub struct MixnetClientSender {
...
+    }
+
+    fn poll_shutdown(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<std::prelude::v1::Result<(), std::io::Error>> {
+        AsyncWrite::poll_flush(self, cx)
```
jstuczyn: if we don't have anything substantial to do here, it's fine to just return an `Ok(())` straightaway

sdk/rust/nym-sdk/src/mixnet/native_client.rs
```diff
@@ -288,6 +331,193 @@ pub struct MixnetClientSender {
...
+                    "failed to read message from input",
+                )))
+            }
+            Poll::Pending => return Poll::Pending,
+            Poll::Ready(None) => return Poll::Ready(Ok(0)),
+        };
+
+        let msg_size = msg.serialized_size();
+
+        let mut fut = pin!(self.client_input.send(msg));
```
jstuczyn: same problem as with `AsyncRead` - this future will never have time to resolve unless it was already ready and won't wake the waker

jstuczyn (changes requested):

common/nymsphinx/forwarding/src/packet.rs
mfahampshire: Removed need for `serde`

sdk/rust/nym-sdk/src/mixnet/native_client.rs
mfahampshire: > you're creating a new send future every single time you call `poll_write` without letting the previous one resolve. you will have to actually store the future on the `MixnetClient` itself and check for its existence in subsequent calls

This should be fixed now :)

common/nymsphinx/src/lib.rs
mfahampshire: Moved it out of the `nym-sphinx` crate

sdk/rust/nym-sdk/src/mixnet/native_client.rs
mfahampshire: Added a mode which is triggered when the AsyncRead/Write is used to stop people then mixing the functions later

common/nym-connection-monitor/src/sync_self_ping.rs
mfahampshire: Oops - fixed this!

sdk/rust/nym-sdk/src/mixnet/native_client.rs
mfahampshire: Done

common/client-core/src/client/inbound_messages.rs
mfahampshire: Its used in `sdk/rust/nym-sdk/src/mixnet/native_client.rs` - removed `expect_used`.

common/nymsphinx/src/receiver.rs
mfahampshire: Pushed a change which does this, let me know what you think. Removes the reliance on serde + bincode..

common/client-core/src/client/inbound_messages.rs
mfahampshire: If we remove the `InputMessage` format, don't we lose info (e.g. if its a reply)? I can have a go at it, but been thinking it through, and feels like we lose something there?

nym-outfox/src/lib.rs
mfahampshire: https://nymtech.atlassian.net/browse/NET-781?atlOrigin=eyJpIjoiNTcxZTdkNDQyOGRjNGZjNWIwNWM4YzhkMGU2MWEzNWIiLCJwIjoiaiJ9

done

sdk/rust/nym-sdk/src/mixnet/native_client.rs
mfahampshire: At the moment it's just checking the write buffer doesn't have anything to send, otherwise just returning `Ok(())` (after fixing the pending future problem this code has changed a bit)

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+use nym_task::connections::TransmissionLane;
+
+use protocol::{decode_stream_message, encode_stream_message, StreamMessageType};
+
+use crate::mixnet::native_client::MixnetClient;
+use crate::{Error, Result};
+
+/// The shared stream routing table. The router task reads it to dispatch
+/// incoming data; streams insert/remove themselves directly.
+pub(crate) type StreamMap = Arc<Mutex<HashMap<StreamId, mpsc::UnboundedSender<Vec<u8>>>>>;
```
jstuczyn: this is more of a nit, but I think this type is complx enough to warrant creation of an explicit struct rather than a type alias (plus it will make any future changes easier), e.g.

```rust
pub(crate) type StreamSendingHandle = mpsc::UnboundedSender<Vec<u8>>;

struct StreamMap {
    inner: Arc<Mutex<HashMap<StreamId, StreamSendingHandle>>>,
}
```

jstuczyn: furthermore, given the async context of the router, tokio's async `Mutex` would imo be a better fit than stdlib `Mutex`, so that if we have to wait, we wouldn't block the entire thread

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+        )
+    }
+}
+
+impl fmt::Display for StreamId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(self, f)
+    }
+}
+
```
jstuczyn, line 52: I think it would be definitely useful to have an explicit `Header` struct defined, something along the lines of 
```rust
struct MixStreamHeader {
    version: u8,
    stream_id: StreamId,
    message_type: StreamMessageType,
}
```

it's important when encoding that the version field is first so we would be able to read it without having to parse other fields and adjust furher parsing based on the version (if it changes)

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+        &self.0
+    }
+}
+
+impl fmt::Debug for StreamId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "StreamId({:02x}{:02x}..{:02x}{:02x})",
+            self.0[0], self.0[1], self.0[14], self.0[15]
```
jstuczyn: do we really need 16 bytes here? : ) I think a `u64` (or even `u32`) would be more than enough for a client

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+    buf.extend_from_slice(id.as_bytes());
+    buf.push(msg_type as u8);
+    buf.extend_from_slice(payload);
+    buf
+}
+
+/// Decode a stream message, returning `(StreamId, MessageType, payload)`.
+///
+/// Returns `None` if the buffer is too short or the message type byte is unknown.
+pub fn decode_stream_message(bytes: &[u8]) -> Option<(StreamId, StreamMessageType, &[u8])> {
```
jstuczyn: I'm really not a fan of returning tuples with more than 2 fields. so with my suggestion of `MixStreamHeader` how about creating an explicit Message/Frame, something like we do in LP (you can take a look at how we handle parsing there too), i.e.
```rust
struct MixStreamFrame {
  header: MixStreamHeader,
  data: Bytes // or Vec<u8>, depending on what's more convenient
}
```

sdk/rust/nym-sdk/src/mixnet/stream/mixnet_stream.rs
jstuczyn, line 1: maybe as suggested by clippy we should rename it? because `stream::stream::MixnetStream` is indeed confusing : D

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    client_input: nym_client_core::client::base_client::ClientInput,
+    packet_type: Option<nym_sphinx::params::PacketType>,
+    streams: StreamMap,
+}
+
+impl MixnetListener {
+    /// Wait for a remote peer to open a stream.
+    ///
+    /// Returns `None` if the router has shut down.
+    pub async fn accept(&mut self) -> Option<MixnetStream> {
```
jstuczyn, line 169: I think this is something to discuss with @mmsinclair in the future. Once we have mixnet-v2 (™️ ) with LP, we should incorporate the stream packets into the framing there so that we would be able to infer from its framing that we got a `MixnetStream` packet so we would open it automatically. because the current flow is that we call `listener.accept()` hoping what we received into the router is a valid packet.

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                        req.stream_id
+                    );
+                    continue;
+                }
+            };
+
+            let (tx, rx) = mpsc::unbounded_channel();
+            self.streams
+                .lock()
+                .expect("stream map poisoned")
```
jstuczyn: please no panics : )

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                        map.remove(&stream_id);
+                    }
+                } // MAX TODO if we decide we need close logic add another enum member
+            }
+        }
+    }
+}
+
+/// Lazily initialise the stream subsystem and router on first use.
+fn ensure_init(client: &mut MixnetClient) -> &mut StreamState {
```
jstuczyn: I don't really see a way to disable the "stream mode" once enabled. if this is by design, then it's fine

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    if client.streams.is_none() {
+        client.stream_mode.store(true, Ordering::SeqCst);
+
+        let (_, dummy_rx) = futures::channel::mpsc::unbounded();
+        let real_rx = std::mem::replace(&mut client.reconstructed_receiver, dummy_rx);
```
jstuczyn: rather than putting dummy values here, could we either replace it with an explicit `Option` or refactor the uses of that channel? because if something goes wrong, it will be very annoying to debug trying to figure out what actual channel is being used in there....

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+pub(crate) async fn open_stream(
+    client: &mut MixnetClient,
+    recipient: Recipient,
+    reply_surbs: u32,
+) -> Result<MixnetStream> {
+    let streams = ensure_init(client).streams.clone();
+
+    let stream_id = StreamId::random();
+    let (tx, rx) = mpsc::unbounded_channel();
+    streams
```
jstuczyn: once we move `StreamMap` into a dedicated struct, we could have a simple fn there that streamlines new stream creation, i.e. something along the lines of 
```rust
async fn naming_things_is_difficult(&self, stream_id) -> mpsc::UnboundedReceiver<Vec<u8>> {
    let (tx, rx) = mpsc::unbounded_channel();
    self.inner.lock().await.insert(stream_id, tx);
    rx
}
```

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                StreamMessageType::Data => {
+                    let mut map = streams.lock().expect("stream map poisoned");
+                    let should_remove = map
+                        .get(&stream_id)
+                        .map(|tx| tx.send(payload.to_vec()).is_err())
+                        .unwrap_or(false);
+                    if should_remove {
+                        map.remove(&stream_id);
+                    }
+                } // MAX TODO if we decide we need close logic add another enum member
```
jstuczyn: I think we definitely need some sort of closing logic. how about for each stream, apart from the sending channel, we'd also have an `Instant` indicating time the channel last send/received anything. then in the router have an `interval` that periodically clears all streams with no activity in say last 30min or whatever

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Added

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
mfahampshire: Good call - done

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
mfahampshire: Yeah good point, lowered it

sdk/rust/nym-sdk/src/mixnet/stream/protocol.rs
mfahampshire: Done

sdk/rust/nym-sdk/src/mixnet/stream/mixnet_stream.rs
mfahampshire: Yeah I think this was left in, changed the name :)

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Removed!

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: IMO once you've dropped into stream-mode, I don't see why you'd drop out of it.. I assumed that you'd sort of just select one depending on your usecase

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Yeah good point, replaced it with something less crazy

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Implemented a timeout check on the streams to clean them up

jstuczyn (approved):

PR #6320, opened 2026-01-12
Max/sdk stream wrapper

mfahampshire:
Part 3 of breaking #6129 into smaller chunks. 

This is the third (following `AsyncRead/Write` streams in the SDK + adding the LP framing to the SDK). 

~~This contains the nym client -> nym client Rust SDK stream wrapper, and the stream wrapper for the Rust SDK which uses gateway IPRs to send IP packets through the mixnet.~~

This PR contains the necessary prework for `smolmix` - having the Rust SDK client use the `stream` functionality to interact with IPRs (sending IP packets), unlocking client-side-only Mixnet integration possibilities. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6320)


mfahampshire:
Not sure why the `test monitor::geodata::api_regression::should_parse_response` test is failing on the runner, its working locally.

sdk/rust/nym-sdk/Cargo.toml
```diff
@@ -94,6 +96,8 @@ tokio-stream = { workspace = true }
 tokio-util = { workspace = true, features = ["codec"] }
 parking_lot = { workspace = true }
 hex = { workspace = true }
+pnet_packet = "0.35"
```
jstuczyn: try to keep all dep versions in the workspace for the ease of upgrades

sdk/rust/nym-sdk/src/ipr_wrapper/architecture.md
```diff
@@ -0,0 +1,72 @@
...
+- **IPR -> Client**: Both inline responses (connect handshake, pong) and async
+  TUN responses are wrapped in LP Stream frames with the same stream_id. The
+  client's stream router dispatches by stream_id to the correct `MixnetStream`.
+
+## IP Allocation
+
+The IPR owns a subnet (`10.0.0.0/16` for IPv4, `fc00::/112` for IPv6) and
+allocates addresses to clients on connect. Two modes are supported:
+
+- **Dynamic** (used by `IpMixStream`): The IPR picks a random unused `IpPair`
```
jstuczyn, line 48: wait. what is this pool? is it the same one as the one used by the authenticator (and LP) dvpn? what if they overlap?

sdk/rust/nym-sdk/src/ipr_wrapper/network_env.rs
```diff
@@ -0,0 +1,45 @@
+use std::fs;
+use std::path::PathBuf;
+
+use crate::Error;
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NetworkEnvironment {
```
jstuczyn: I'm not sure that whole thing is going to be of any use, because it only make sense when the binary is r un from somewhere within our monorepo. and consumers of rust-sdk will never do that. plus imo it adds needless noise to already overly complex env setup we have

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -172,11 +173,19 @@ impl MixnetListener {
...
+            match result {
+                Ok(packet) => {
+                    let result = self
+                        .handle_packet(packet.as_bytes(), data_request.version)
+                        .await;
+                    responses.push(result);
+                }
+                Err(e) => {
+                    log::warn!("Failed to decode bundled IP packet: {e}");
+                    break;
```
jstuczyn: we shouldn't stop our listener if we fail to decode a packet, because it means anyone could send a garbage data to the IPR to kill it

sdk/rust/nym-sdk/examples/ipr_tunnel.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+
+    tunnel.disconnect().await;
+    Ok(())
+}
+
+// --- IPv4 ICMP helpers ---
+
+fn build_icmp_ping(src: Ipv4Addr, dst: Ipv4Addr, seq: u16) -> anyhow::Result<Vec<u8>> {
```
jstuczyn: I wonder if it would be useful to extract those functions to some shared crate for testing or something. I could imagine people wanting to easily construct icmp packets to test things around

sdk/rust/nym-sdk/src/ip_packet_client.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2023-2024 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn: nit: well, it's not 2023-2024 anymore, is it? : D
plus library code should be licensed under apache

sdk/rust/nym-sdk/src/ip_packet_client.rs
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: GPL-3.0-only
+
+//! Minimal fork of `nym-ip-packet-client` for use by [`IpMixStream`](crate::ipr_wrapper::IpMixStream).
```
jstuczyn: if it's a 'fork' would it be possible instead to share the code between those two rather than copy it around?

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Clone)]
+pub struct IprWithPerformance {
+    pub address: Recipient,
+    pub identity: ed25519::PublicKey,
+    pub performance: u8,
+}
+
+#[allow(clippy::result_large_err)]
+pub fn create_nym_api_client(
```
jstuczyn, line 27: give the IPR client is created as part of the existing SDK's mixnet client, why can't we just use its own nym-api client, or better than that, its list of gateways?

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    Ok(described)
+}
+
+/// Select the highest-performance IPR gateway from the directory.
+pub async fn get_best_ipr(client: nym_http_api_client::Client) -> Result<Recipient, Error> {
+    let nodes = retrieve_exit_nodes_with_performance(client).await?;
+    info!("Found {} Exit Gateways", nodes.len());
+
+    let selected_gateway = nodes
```
jstuczyn: imo you should include a little bit of randomness here, like when we choose "best" gateway by latency - we select them randomly, but bias the selection by the latency itself.

why? say there's 100 IPRs with performance of 99 and 1 IPR with performance of 100. is it much "better" than the rest of them? probably not. but it means **EVERY** client will now choose it and the remaining 100 will have nobody connecting to them

sdk/rust/nym-sdk/src/ip_packet_client/listener.rs
```diff
@@ -0,0 +1,83 @@
...
+        )));
+    }
+    Ok(())
+}
+
+/// Parse raw IPR response bytes into an outcome.
+pub fn handle_ipr_response(data: &[u8]) -> Result<Option<MixnetMessageOutcome>, crate::Error> {
+    check_ipr_message_version(data)?;
+
+    match IpPacketResponse::from_bytes(data) {
```
jstuczyn: can you add an explicit prefix that this is `v8::IpPacketResponse`? because once we add another version, this bit will be incredibly easy to miss

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -0,0 +1,190 @@
...
+    client: MixnetClient,
+    allocated_ips: IpPair,
+    connected: bool,
+}
+
+impl IpMixStream {
+    /// Discover the best IPR, connect through the mixnet, and establish the IP tunnel.
+    ///
+    /// Returns a ready-to-use tunnel with allocated IP addresses.
+    pub async fn new(env: NetworkEnvironment) -> Result<Self, Error> {
```
jstuczyn: I don't think it makes sense to make the default "new" method  require env setup. our users wouldn't and shouldn't care about it. they should only use mainnet.

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -95,17 +103,35 @@ impl StreamMap {
...
+            if seq < entry.next_seq {
+                warn!(
+                    "Stream {stream_id}: dropping old seq {seq} (expected >= {})",
+                    entry.next_seq
+                );
             } else {
+                entry.pending.insert(seq, data);
+            }
+
+            // Drain contiguous messages
```
jstuczyn, line 146: do we have any mechanism of draining stale data?
say we have the following:
next_seq = 1
map: = [2,3,4,5, .... 10000]
so we have A LOT of data in it, but never got 1 because sender was malicious. 
will our memory keep on growing?

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -436,13 +485,117 @@ impl MixnetListener {
                 .unwrap_or("missing".to_owned())
         );
 
+        // Check if this is an LP Stream frame
```
jstuczyn, line 446: that's an incredibly nasty and error prone check.  since we already had to modify the IPR code, why not just bump up the protocol version number and then on v8 (and below) do whatever we were doing before and on v9 **always** parse out an LP header and put the old data in some other LP variant? then on stream, we do streamy stuff and on non-stream we do old stuff

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Added a buffer + check to force moving onto the next highest seq to mitigate this

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: Good point, modified this to use `.choose_weighted()` with the performance as a passed param

sdk/rust/nym-sdk/src/ip_packet_client.rs
mfahampshire: That comment was old / a bit misleading :grimacing: I've ended up extracting the shared code between the pair of them into `/common/ip-packet-requests/src/response_helpers.rs`

sdk/rust/nym-sdk/examples/ipr_tunnel.rs
mfahampshire: Extracted them to `/common/ip-packet-requests/src/icmp_utils.rs` and put behind a flag

sdk/rust/nym-sdk/src/ipr_wrapper/network_env.rs
mfahampshire: Killed it

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: From having a look, the nym-api client isn't really accessible after init / on `MixnetClient`. I think I could make it accessible by just cloning it and adding it as a field to `BaseClient` but this is then touching some core client stuff, not sure of the ripple effects. Happy to try it in this PR and happy to make a ticket to do this in its own PR to stop this ballooning - what do you think?

sdk/rust/nym-sdk/src/ip_packet_client/listener.rs
mfahampshire: This has been covered in the refactored `response_helpers.rs` file

sdk/rust/nym-sdk/src/ipr_wrapper/architecture.md
mfahampshire: Added a comment below wrt this in the file - from my understanding they're separated

service-providers/ip-packet-router/src/mixnet_listener.rs
mfahampshire: Implemented your suggestion - let me know if there are any other modifications you think I should make

nym-ip-packet-client/src/helpers.rs
```diff
@@ -1,30 +1,24 @@
 // Copyright 2023-2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::cmp::Ordering;
-
+use nym_ip_packet_requests::response_helpers::IprResponseError;
 use nym_sdk::mixnet::ReconstructedMessage;
 
-use crate::{Error, current::VERSION as CURRENT_VERSION, error::Result};
+use crate::{current::VERSION as CURRENT_VERSION, error::Result};
```
jstuczyn, line 7: looking through the code, `current::VERSION` is still defined as v8 rather than v9 (`nym-ip-packet-client/src/lib.rs`)

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
jstuczyn: definitely no point in ballooning this PR. if it's simple to do (like less than 1-2h of work), maybe try to introduce it in another PR - otherwise imo don't waste time on it

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
```diff
@@ -0,0 +1,100 @@
...
+    let exit_gateways = client.get_all_basic_nodes_with_metadata().await?.nodes;
+
+    let mut described = Vec::new();
+
+    for exit in exit_gateways {
+        if let Some(ipr_info) = all_nodes
+            .get(&exit.ed25519_identity_pubkey)
+            .and_then(|n| n.description.ip_packet_router.clone())
+        {
+            if let Ok(parsed_address) = ipr_info.address.parse() {
```
jstuczyn, line 91: now, for the purposes of sdk IPR client, we might have to do additional filtering here. since we're going to be sending LP-wrapped packets, we have to only use nodes that support v9+ protocol. though how to do it... not sure if there's a query for the supported protocol. maybe we could do it based on the semver? but can we guarantee when this PR is going to get released? sorry. it's just me rambling aloud here : D

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -0,0 +1,193 @@
...
+    connected: bool,
+}
+
+impl IpMixStream {
+    /// Discover the best IPR, connect through the mixnet, and establish the IP tunnel.
+    ///
+    /// Returns a ready-to-use tunnel with allocated IP addresses.
+    pub async fn new() -> Result<Self, Error> {
+        let network_defaults = NymNetworkDetails::new_mainnet();
+        let api_client = create_nym_api_client(network_defaults.nym_api_urls.unwrap_or_default())?;
```
jstuczyn: well, there's no real point in doing `unwrap_or_default`, because the default here is going to be an empty `vec`

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -0,0 +1,193 @@
...
+        if self.connected {
+            Ok(())
+        } else {
+            Err(Error::IprStreamClientNotConnected)
+        }
+    }
+
+    async fn connect_tunnel(stream: &mut MixnetStream) -> Result<IpPair, Error> {
+        let (mut request, request_id) = IpPacketRequest::new_connect_request(None);
+        request.protocol.version = nym_ip_packet_requests::v9::VERSION;
```
jstuczyn: I think the fact that you re-exported v8 structs rather than defined some new variants for v9 is making things a little bit confusing. like the fact you have to explicitly overwrite the version number here. it could lead to some weird errors where we forget to do it in some location : /
but it's annoying. can't see an obvious way around it without duplicating a lot of code or making an inconsistent API

jstuczyn: just thinking aloud here, not sure how unwieldy this is going to be. redefine `IpPacketRequest` (and I guess response?) and when creating the request you'd call the v8 variant followed by the conversion into current, i.e. something long the lines of 

```rust
impl IpPacketRequestV9 {
    pub fn new_connect_request(buffer_timeout: Option<u64>) -> (Self, u64) {
       let (v8, request_id) = IpPacketRequestV8::new_connect_request(buffer_timeout);
       (v8.into(), request_id)
    }
}
```

service-providers/ip-packet-router/src/clients/connected_client_handler.rs
```diff
@@ -189,28 +217,50 @@ fn create_ip_packet_response(
         ClientVersion::V6 => IpPacketResponseV6::new_ip_packet(packets).to_bytes(),
         ClientVersion::V7 => IpPacketResponseV7::new_ip_packet(packets).to_bytes(),
         ClientVersion::V8 => IpPacketResponseV8::new_ip_packet(packets).to_bytes(),
+        ClientVersion::V9 => {
+            let mut resp = IpPacketResponseV8::new_ip_packet(packets);
```
jstuczyn: don't you mean V9?

service-providers/ip-packet-router/src/mixnet_listener.rs
jstuczyn: I'm confused, it does not appear to have changed. isn't it the case that now every packet is going to have LP header? oh. right. you won't know the protocol version until you have actually parsed the header which might or might not be. that's quite tricky. but at the very least you must do the following:
- if you parsed out LP header (or something looking like LP header), the protocol MUST be 9 or higher
- if you tried parsing out the LP header, but failed, you should continue with the lower protocol (in case it was a legit v8 or lower message)
- and if you didn't get an LP header, the protocol MUST BE 8 or lower

service-providers/ip-packet-router/src/mixnet_listener.rs
```diff
@@ -436,13 +486,117 @@ impl MixnetListener {
...
+        if reconstructed.message.len() >= LpFrameHeader::SIZE
+            && let Ok(header) = LpFrameHeader::parse(&reconstructed.message)
+            && header.kind == LpFrameKind::SphinxStream
+        {
+            return self.on_stream_frame(reconstructed, header).await;
+        }
+
         // Check if this is a KCP-wrapped message from an LP client
         if is_kcp_message(&reconstructed.message) {
```
jstuczyn: and remove that KCP. it makes no sense for it to be there

mfahampshire: :fire:-ed it

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
mfahampshire: Sorted this

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: Added the filtering we discussed in Zulip

nym-ip-packet-client/src/helpers.rs
mfahampshire: fixed

jstuczyn (approved):

PR #6321, opened 2026-01-12
Max/mixtcp

mfahampshire:
Part 3 of breaking https://github.com/nymtech/nym/pull/6129 into smaller chunks.

~~**Should only be looked at after https://github.com/nymtech/nym/pull/6320 is merged and this branch is rebased** removing most of the commits).~~ DONE 

This contains the `mixtcp` proof of concept - the `-core` crate that we can build connectors from for the `Tokio` async ecosystem, exposing a `tunnel` abstraction which gives developers something that looks and acts like either a TCP or UDP socket, for client-side-only integrations. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6321)


Cargo.toml
```diff
@@ -333,7 +335,7 @@ rayon = "1.5.1"
 regex = "1.10.6"
 reqwest = { version = "0.13.1", default-features = false }
 rs_merkle = "1.5.0"
-rustls = { version = "0.23.37", default-features = false }
+rustls = { version = "0.23.37", default-features = false, features = ["std", "ring"] }
```
jstuczyn: I'm not sure we want to be requiring `ring` feature everywhere, all the time, because afaik it is not wasm-compatible. unless it's only pulled in in situation where it doesn't matter

documentation/docs/pages/developers/native.mdx
```diff
@@ -61,3 +61,27 @@ Applications that support SOCKS4, 4a, or 5 can use the Socks Client exposed by t
 <Callout type="info">
 Development is in progress to allow for this proxy method from native Rust, C, and Go without requiring a separate SOCKS client. Stay tuned.
 </Callout>
+<<<<<<< HEAD
```
jstuczyn: it seems something got messed up during a merge?

smolmix/core/src/tunnel.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl Tunnel {
+    /// Create a new tunnel, automatically discovering the best IPR exit node.
+    ///
+    /// This is the simplest entry point — one line gets you a working tunnel:
+    /// ```ignore
+    /// let tunnel = Tunnel::new().await?;
+    /// let tcp = tunnel.tcp_connect("1.1.1.1:443".parse()?).await?;
+    /// ```
+    pub async fn new() -> Result<Self, SmolmixError> {
```
jstuczyn, line 167: for futureproofness (if that's even a word 😅 ), we'd probably also want to expose a builder for the tunnel, because eventually we'll want to, for example, setup bunch of things to do with credentials or maybe configure sending rates, etc.

smolmix/core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[dependencies]
+smoltcp = { workspace = true, features = [
+    "std",
+    "medium-ip",
+    "proto-ipv4",
+    "socket-tcp",
+    "socket-udp",
+] }
+tokio = { workspace = true }
+tokio-smoltcp = "0.5"
```
jstuczyn: move it please to workspace deps : )

smolmix/core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+thiserror.workspace = true
+nym-bin-common = { workspace = true, features = ["basic_tracing"] }
+
+[dev-dependencies]
+futures = { workspace = true }
+tokio = { workspace = true, features = ["io-util", "macros", "rt-multi-thread", "time", "net"] }
+tokio-tungstenite.workspace = true
+webpki-roots.workspace = true
+rustls = { workspace = true }
+tokio-rustls = "0.26"
```
jstuczyn: same

smolmix/core/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+    "socket-udp",
+] }
+tokio = { workspace = true }
+tokio-smoltcp = "0.5"
+futures = { workspace = true }
+tracing = { workspace = true }
+nym-sdk = { workspace = true }
+nym-ip-packet-requests = { workspace = true }
+thiserror.workspace = true
+nym-bin-common = { workspace = true, features = ["basic_tracing"] }
```
jstuczyn: why would we need `bin-common` in a library? isn't this only used in examples?

smolmix/core/src/device.rs
```diff
@@ -0,0 +1,88 @@
...
+// When smoltcp produces a packet (e.g. TCP SYN, data segment, UDP datagram),
+// tokio-smoltcp sends it here and we forward it to the bridge for mixnet delivery.
+//
+// All methods are trivial — the unbounded channel is always ready and never blocks.
+// Backpressure is handled at the mixnet layer, not here.
+impl Sink<Vec<u8>> for NymAsyncDevice {
+    type Error = io::Error;
+
+    fn poll_ready(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
```
jstuczyn: depending on the internal state of the channel (like whether it's closed), `poll_ready()` can fail. just delegate the call to the internal channel, i.e. `self.tx.poll_ready(cx)` (you might have to use `Pin::new()`)

smolmix/core/src/device.rs
```diff
@@ -0,0 +1,88 @@
...
+            .send(item)
+            .map_err(|_| io::Error::new(io::ErrorKind::BrokenPipe, "bridge channel closed"))
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
```
jstuczyn: similar issue to `poll_ready`, but in here it's actually even more important to delegate the call, because if look at the implementation of `Sink` for `UnboundedSender`, it calls `self.disconnect();` in the implementation and it's important we preserve this behaviour to ensure the receiver also correctly disconnects

jstuczyn (commented):
you will also have to definitely resolve the compilation issue. it looks like one of the new dependencies imports different version of one of the crypto libraries - not entirely sure which one, I haven't looked too deeply into that. if we have no choice but try to update something on our end - let's talk about it.
and after that is done, I reckon, CI is going to scream more about different things, but let's take things one at a time : )

smolmix/core/src/tunnel.rs
mfahampshire: Added a builder in, but I think there's a conversation to be had about how we easily expose the full config capabilities in a nice way in the future (e.g. if someone wanted to configure the base `MixnetClient`, right now it'd be a little clunky). But this builder suffices for the mo I think :)

documentation/docs/pages/developers/native.mdx
mfahampshire: Fixed

smolmix/core/Cargo.toml
mfahampshire: was just for an example, moved to dev-deps

smolmix/core/src/device.rs
mfahampshire: should be fixed now

smolmix/core/src/device.rs
mfahampshire: should be fixed now

smolmix/core/src/bridge.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl BridgeShutdownHandle {
+    /// Signal the bridge to shut down gracefully.
+    ///
+    /// Sends a one-shot signal that breaks the bridge event loop. The bridge
+    /// then calls `IpMixStream::disconnect()` before returning. Consumes
+    /// `self` — can only be called once.
+    ///
+    /// If this handle is dropped without calling `shutdown()`, the oneshot
+    /// channel closes and the bridge detects the hangup, triggering the same
```
jstuczyn: that's more of a nit, but would it make sense to just explicitly implenent `Drop for BridgeShutdownHandle`, where we'd call `self.shutdown()` if the receiver is still alive?

jstuczyn (approved):
looks good!

smolmix/core/src/bridge.rs
mfahampshire: Good point, impl-ed `Drop` for it and updated comment!

PR #6335, opened 2026-01-16
LP: x25519/ed22519 cleanup round

jstuczyn:
- use more strict typing for x25519/ed25519 keys
- remove usage of nymsphinx key types in favour of nym_crypto types
- adjust `LpStateMachine` constructor to take explicit remote x25519 key rather than derive it locally
- introduce additional convenience methods for `nym_crypto` types

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6335)


common/nym-lp/src/keypair.rs
```diff
@@ -1,200 +1,9 @@
...
-    }
-
-    pub fn public_key(&self) -> Result<PublicKey, LpError> {
-        PublicKey::from_base58_string(&self.public)
-    }
-}
+// declare LP keys as type aliases to x25519 keys
+pub type PrivateKey = x25519::PrivateKey;
+pub type PublicKey = x25519::PublicKey;
+pub type KeyPair = x25519::KeyPair;
```
simonwicky: Do we even need that anymore?

simonwicky (approved):

common/nym-lp/src/keypair.rs
jstuczyn: nope - removed.

PR #6337, opened 2026-01-16
Apply configured api urls via env

jmwample:
**Problem**

Some of the expected configuration applied in the mainnet files are not being propogated to the Nym API and by proxy not being applied in the client through the NetworkDetails.

**Solution**

Create working defaults for applying and pasing the newer `nym_api_urls` and `nym_vpn_api_urls` to the env settings. 

Tested as working in a local run of the  `nym_api` binary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6337)


jstuczyn:
it seems you need to apply a bit more feature locking to make sure unexpected things don't get pulled into the contracts (and prevent compilation)

tommyv1987 (approved):

PR #6356, opened 2026-01-22
LP: mixnet reg fixes

jstuczyn:
this PR does the following:
- removes the need for an x25519 key conversion within LP mixnet registration
- splits LP registration handling for VPN and mixnet - you no longer need to send a wireguard peer key for mixnet registration
- added explicit kind byte to each raw LP transport data packet sent to remove exhausting deserialisations on every possible type of return data (https://github.com/nymtech/nym/compare/lp/mixnet-reg-x25519?expand=1#diff-c17520a2342498875242769f8da62615586d9f7464d9fe981b67225c26fe4db5L604-L631)
- as the consequence of above, unfortunately, there is now an additional 1 byte overhead of sending data on the data channel. however, we might be able to resolve it by possibly introducing additional `LpInput` variant to ignore the encoding

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6356)


common/registration/src/lp_messages.rs
```diff
@@ -3,17 +3,49 @@
...
-
 /// Registration request sent by client after LP handshake
 /// Aligned with existing authenticator registration flow
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct LpRegistrationRequest {
+    /// Mode specific registration data
+    pub registration_data: LpRegistrationData,
+
+    /// Client's IP address (for tracking/metrics)
+    pub client_ip: IpAddr,
```
simonwicky: This is unused btw ;)
And from the top, we're feeding in the gateway's IP so we must consider removing that

gateway/src/node/lp_listener/messages.rs
```diff
@@ -6,5 +6,6 @@
 //! Re-exports shared message types from nym-registration-common.
 
 pub use nym_registration_common::{
-    LpGatewayData, LpRegistrationRequest, LpRegistrationResponse, RegistrationMode,
+    LpDvpnRegistrationRequest, LpMixnetGatewayData, LpMixnetRegistrationRequest,
```
simonwicky: Why the re-export in the first place? It's used in a place where `registration-common` is already imported anyway

gateway/src/node/lp_listener/registration.rs
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use super::messages::{
-    LpGatewayData, LpRegistrationRequest, LpRegistrationResponse, RegistrationMode,
+    LpDvpnRegistrationRequest, LpMixnetGatewayData, LpRegistrationData, LpRegistrationRequest,
```
simonwicky: We have a mix of imports between the re-export and the original `nym_registration_common` below. 
Why not use everything from `nym_registration_common` and nuke that re-export?

gateway/src/node/lp_listener/registration.rs
```diff
@@ -195,12 +195,121 @@ async fn check_existing_registration(
...
 
+async fn process_dvpn_registration(
+    request: Box<LpDvpnRegistrationRequest>,
+    state: &LpHandlerState,
+) -> LpRegistrationResponse {
+    // Track dVPN registration attempts
+    inc!("lp_registration_dvpn_attempts");
+
+    // Check for idempotent re-registration (same WG key already registered)
+    // This allows clients to retry registration after network failures
```
simonwicky, line 205: It doesn't, at this point, `LPDvpnRegistrationRequest` already contains a credential.
A malicious could say "Oh you're already registered", keep the credential to create a double spending when the client tries to use the same credential somewhere else

simonwicky (approved):
Non of these comments are blocking on that PR per se, but are points that I think should be addressed before the release

PR #6372, opened 2026-01-26
Lp/use noise x25519

jstuczyn:
this PR makes nym-nodes re-use their existing [noise] x25519 key for KTT rather than deriving one from their long-term ed25519 identity.

similarly clients will also pull that one instead


common/registration/src/lib.rs
```diff
@@ -1,47 +1,49 @@
...
-    pub public_key: PublicKey,
+    pub public_key: x25519::PublicKey,
     pub endpoint: SocketAddr,
     pub private_ipv4: Ipv4Addr,
     pub private_ipv6: Ipv6Addr,
 }
 
 #[derive(Clone, Debug)]
-pub struct LpData {
+pub struct GatewayLpData {
```
simonwicky: I'm seeing this now, but the name here is gonna be confusing. 
`NymNode` and this are information about the gateway that we need **before** registration.
`GatewayData` and `AssignedAddresses` are return data that we feed the vpn-client for its tunnel **after**  registration. 

I reckon the naming is too close for struct that aren't close.
Can we rename this into `LpNode` and maybe `GatewayData` into WirguardData` or something?

nym-api/nym-api-requests/src/models/described/type_translation.rs
```diff
@@ -424,3 +425,25 @@ impl From<nym_node_requests::api::v1::lewes_protocol::models::LPHashFunction> fo
...
+            LPKEM::McEliece => KEM::McEliece,
+        }
+    }
+}
+
+impl From<LPHashFunction> for HashFunction {
+    fn from(value: LPHashFunction) -> Self {
+        match value {
+            LPHashFunction::Blake3 => HashFunction::Blake3,
+            LPHashFunction::Shake128 => HashFunction::SHAKE128,
```
simonwicky: nit : Do we want to keep this all caps in the enum name? Or adapt it?

nym-api/nym-api-requests/Cargo.toml
```diff
@@ -49,7 +49,7 @@ nym-noise-keys = { workspace = true }
 nym-network-defaults = { workspace = true }
 nym-ticketbooks-merkle = { workspace = true }
 nym-ecash-signer-check-types = { workspace = true }
-
+nym-kkt-ciphersuite = { path = "../../common/nym-kkt-ciphersuite" }
```
simonwicky: nit Can we set that as a workspace dependency?  and import it from there?

nym-gateway-probe/Cargo.toml
```diff
@@ -65,6 +65,8 @@ nym-node-status-client = { path = "../nym-node-status-api/nym-node-status-client
 nym-node-requests = { path = "../nym-node/nym-node-requests" }
 nym-registration-client = { path = "../nym-registration-client" }
 nym-lp = { path = "../common/nym-lp" }
+nym-kkt-ciphersuite = { path = "../common/nym-kkt-ciphersuite" }
```
simonwicky: Same as above

nym-registration-client/src/lp_client/helpers.rs
```diff
@@ -81,3 +83,7 @@ pub(crate) fn try_convert_forward_response(action: LpAction) -> Result<Vec<u8>,
 
     Ok(response_data.content.into())
 }
+
+pub(crate) fn to_lp_remote_peer(identity: ed25519::PublicKey, data: GatewayLpData) -> LpRemotePeer {
```
simonwicky: Do we even need `GatewayLpData` anymore? Could we use `LpRemotePeer` directly?

simonwicky (approved):

nym-registration-client/src/lp_client/helpers.rs
jstuczyn: not yet, because even without LP you need to have gateway's identity key

PR #6378, opened 2026-01-27
Lp/dvpn psk injection

jstuczyn:
this PR makes clients generate pseudorandom PSK after LP registration (i.e. when post-quantum secure channel has been established)


common/registration/src/lp_messages.rs
```diff
@@ -50,6 +51,9 @@ pub struct LpDvpnRegistrationRequest {
 
     /// Ticket type for bandwidth allocation
     pub ticket_type: TicketType,
+
+    /// Preshared key to be used for the connection
+    pub psk: [u8; 32],
```
simonwicky, line 56: nit : feels like this should be strong typed somehow

simonwicky (approved):

neacsu (approved):

PR #6386, opened 2026-01-28
Lp/two step dvpn reg

jstuczyn:
this PR changes dVPN registration within LP to more closely resemble the pattern happening within the Authenticator, where it's a 2-step process.

Firstly the client requests registration **without** sending a credential, so that if it had already registered before, it would not need to retrieve one.
Only if it's a fresh request, a bandwidth claim is required.

There's some cleanup that would be needed later on, such as sharing the code with the authenticator, in particular the IpPool allocation

### DIFF
--- a/common/bandwidth-controller/src/traits.rs
+++ b/common/bandwidth-controller/src/traits.rs
@@ -5,7 +5,6 @@ use async_trait::async_trait;
 use nym_credential_storage::storage::Storage;
 use nym_credentials_interface::TicketType;
 use nym_crypto::asymmetric::ed25519;
-use nym_crypto::asymmetric::ed25519::PublicKey;
 use nym_validator_client::nyxd::contract_traits::DkgQueryClient;
 
 use crate::{error::BandwidthControllerError, BandwidthController, PreparedCredential};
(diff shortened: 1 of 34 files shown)


common/bandwidth-controller/src/traits.rs
```diff
@@ -57,3 +58,22 @@ where
...
     }
 }
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl<T: BandwidthTicketProvider + ?Sized + Send> BandwidthTicketProvider for Box<T> {
+    async fn get_ecash_ticket(
+        &self,
+        ticket_type: TicketType,
+        gateway_id: PublicKey,
```
simonwicky: nit : `ed25519::PublicKey` so we can remove the above import

common/registration/src/lp_messages.rs
```diff
@@ -207,16 +289,172 @@ impl LpRegistrationResponse {
...
+    #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+    pub struct RequiresCredentialResponse;
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct RegistrationFailureResponse {
+        pub error: String,
+    }
+}
+
+use crate::WireguardConfiguration;
```
simonwicky: I think that import got lost

gateway/src/node/lp_listener/mod.rs
```diff
@@ -328,12 +341,13 @@ impl<T> TimestampedState<T> {
     }
 }
 
+pub type ReceiverIndex = u32;
```
simonwicky: nit : maybe define that at the top of the file?

simonwicky (approved):

PR #6400, opened 2026-01-30
[LP Gateway Probe] CLI and behavior improvements 

simonwicky:
Its goal is to improve the CLI and the behavior of the probe which were a bit of a mess, without changing the output (at least for now)
```
Usage: nym-gateway-probe [OPTIONS] <COMMAND>

Commands:
  run-local  Run the probe on an unannounced gateway. IP must be provided. Bypasses directory lookup
  run        Run the probe on a bonded gateway. Uses directory lookup
  run-agent  Run the probe by NS agents
  help       Print this message or the help of the given subcommand(s)

Options:
  -c, --config-env-file <CONFIG_ENV_FILE>  Path pointing to an env file describing the network
      --no-log                             Disable logging during probe
  -h, --help                               Print help
  -V, --version                            Print version
```

### DIFF
--- a/nym-gateway-probe/src/common/nodes.rs
+++ b/nym-gateway-probe/src/common/nodes.rs
@@ -4,9 +4,8 @@
 use anyhow::{Context, anyhow, bail};
 use nym_api_requests::models::{
     AuthenticatorDetailsV2, DeclaredRolesV2, DescribedNodeTypeV2, HostInformationV2,
-    IpPacketRouterDetailsV2, LewesProtocolDetailsV1, NetworkRequesterDetailsV1,
-    NetworkRequesterDetailsV2, NymNodeDataV2, OffsetDateTimeJsonSchemaWrapper, WebSocketsV2,
-    WireguardDetailsV2,
+    IpPacketRouterDetailsV2, NetworkRequesterDetailsV2, NymNodeDataV2,
+    OffsetDateTimeJsonSchemaWrapper, WebSocketsV2, WireguardDetailsV2,
 };
 use nym_authenticator_requests::AuthenticatorVersion;
 use nym_bin_common::build_information::BinaryBuildInformationOwned;
@@ -27,7 +26,7 @@ use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
 use std::time::Duration;
 use time::OffsetDateTime;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 use url::Url;
 // in the old behaviour we were getting all skimmed nodes to retrieve performance
 // that was ultimately unused
@@ -108,6 +107,11 @@ impl DirectoryNode {
             .as_ref()
             .map(|ipr| ipr.address.parse().context("malformed ipr address"))
             .transpose()?;
+        let network_requester_address = description
+            .network_requester
+            .as_ref()
+            .map(|nr| nr.address.parse().context("malformed nr address"))
+            .transpose()?;
         let authenticator_address = description
             .authenticator
             .as_ref()
@@ -140,12 +144,11 @@ impl DirectoryNode {
             }),
             _ => None,
         };
-        let network_requester_details = self.described.description.network_requester.clone();
 
         Ok(TestedNodeDetails {
             identity: self.identity(),
             exit_router_address,
-            network_requester_details,
+            network_requester_address,
             authenticator_address,
             authenticator_version,
             ip_address: Some(ip_address),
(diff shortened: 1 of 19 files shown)


nym-gateway-probe/src/common/nodes.rs
```diff
@@ -431,45 +457,27 @@ impl NymApiDirectory {
...
         };
-        if !maybe_entry.described.description.declared_role.exit_nr {
-            bail!("{identity} doesn't support exit NR mode")
+        if !maybe_entry
+            .described
+            .description
+            .declared_role
+            .can_operate_exit_gateway()
+        {
+            bail!("{identity} is not an entry node")
```
jstuczyn: exit

nym-gateway-probe/src/config/credentials.rs
```diff
@@ -1,29 +1,94 @@
...
+    pub async fn acquire(
+        &self,
+        disconnected_mixnet_client: &DisconnectedMixnetClient<OnDiskPersistent>,
+        storage: &OnDiskPersistent,
+    ) -> anyhow::Result<()> {
+        // Return immediately as there is nothing to do
+        if self.use_mock_ecash {
+            return Ok(());
+        }
+        let ticketbook_count = storage
```
jstuczyn, line 71: you probably want to check per ticketbook type. because what if you have 1 ticketbook of type `V1WireguardExit`? you'd still fail to test anything in entry or mixnet mode

nym-gateway-probe/src/config/test_mode.rs
```diff
@@ -4,87 +4,72 @@
...
-    /// Traditional mixnet testing - connects via mixnet, tests entry/exit pings + WireGuard via authenticator
+    /// Mixnet tests + WireGuard via authenticator
     #[default]
-    Mixnet,
-    /// LP registration + WireGuard on single gateway (no mixnet, no forwarding)
-    SingleHop,
-    /// Entry LP + Exit LP (nested session forwarding) + WireGuard tunnel
-    TwoHop,
-    /// LP registration only - test handshake and registration, skip WireGuard
+    Default,
```
jstuczyn: rather than calling a variant `Default`, let's just explicitly say it's `MixnetWithWg` or something as the name is not very descriptive (compared to the alternatives)

nym-gateway-probe/src/lib.rs
```diff
@@ -1,897 +1,449 @@
...
-        test_lp_wg: bool,
-        min_mixnet_performance: Option<u8>,
-        network_details: NymNetworkDetails,
+    /// Run a probe as an NS agent, i.e. a bonded node on a known network
+    pub async fn probe_run_agent(
+        mut self,
+        credential_args: CredentialArgs,
     ) -> anyhow::Result<ProbeResult> {
-        let tickets_materials = self.credentials_args.decode_attached_ticket_materials()?;
+        let storage = Ephemeral::default();
```
jstuczyn, line 70: hehe, sneak peek: this will change in the future

nym-gateway-probe/src/lib.rs
```diff
@@ -1,897 +1,449 @@
...
-        let (mixnet_entry_gateway_id, node_info) = self.lookup_gateway(&directory).await?;
+        let mixnet_debug_config = helpers::mixnet_debug_config(
+            self.config.min_gateway_mixnet_performance,
+            self.config.ignore_egress_epoch_role,
+        );
 
-        let storage = Ephemeral::default();
+        // If we need to run at least one mixnet client, prefetch topology
+        if self.config.test_mode.needs_mixnet() || self.config.test_mode.socks5_tests() {
+            self.topology = helpers::hardcoded_topology(&self.network, &mixnet_debug_config)
```
jstuczyn: the name `hardcoded_topology` is quite deceiving as it's not returning some "hardcoded" value. it's just fetching one from the nym-api.

jstuczyn: plus, I wonder, if we're already providing topology, would it make sense to always include there the target entry and exit gateways? so that we could, for example, tests nodes that are not neccessarily included as part of target networks

nym-gateway-probe/src/lib.rs
```diff
@@ -1,897 +1,449 @@
...
-        }
-
         let storage_paths = StoragePaths::new_from_dir(config_dir)?;
         let storage = storage_paths
             .initialise_default_persistent_storage()
             .await?;
 
-        // Connect to the mixnet via the entry gateway, without forget-me flag so that gateway remembers client
-        // and keeps its bandwidth between probe runs
+        // We cannot run mixnet tests on unannounced gateway, but we still need one to import credential if not using mock ecash
```
jstuczyn, line 136: what if we could

nym-gateway-probe/src/lib.rs
```diff
@@ -1,897 +1,449 @@
...
-        )
-        .await
+        self.do_probe_test(None, bandwidth_provider).await
     }
 
-    /// Probe LP registration only, skipping all mixnet tests
-    /// This is useful for testing local dev gateways that aren't registered in nym-api
-    pub async fn probe_lp_only(
-        self,
+    pub async fn probe_run(
```
jstuczyn, line 163: it's nit if anything, but it appears there's a lot of duplication from `pub async fn probe_run_agent(` => could we maybe somehow share the topology and client setup?

jstuczyn (commented):
very nice work. those improvements were quite overdue. it will definitely help in future work on the probe

nym-gateway-probe/src/lib.rs
simonwicky: Point taken on the name. 
We can't include our nodes though, because other nodes won't find them when routing back to them

nym-gateway-probe/src/lib.rs
simonwicky: We cannot, because mixnodes will not have their noise keys (and will refuse to route to them as they are not in the network)

nym-gateway-probe/src/common/nodes.rs
simonwicky: Fixed

nym-gateway-probe/src/lib.rs
simonwicky: Client setup is slightly different, and the underlying storage being different makes it a bit of a hassle to unify them.
Given this, I don't think it's worth it, even though I agree with you

nym-gateway-probe/src/config/credentials.rs
simonwicky: After side-channel discussion, this is a future improvement as function to test if we still have valid credentials are not available

nym-gateway-probe/src/config/test_mode.rs
simonwicky: Fixed, it's now Core

jstuczyn (approved):

PR #6411, opened 2026-02-03
[LP-fix] expose wg psk for the vpn-client

simonwicky:
This PR :
- exposes the client generated Wireguard PSK in `WireguardConfiguration` so it can be used by the vpn client
- stores the exit lp address for nested session as `SocketAddr` and not `String`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6411)

### DIFF
--- a/common/nym-lp/src/message.rs
+++ b/common/nym-lp/src/message.rs
@@ -9,6 +9,7 @@ use nym_crypto::asymmetric::{ed25519, x25519};
 use rand::RngCore;
 use serde::{Deserialize, Serialize};
 use std::fmt::{self, Display};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 
 /// Data structure for the ClientHello message
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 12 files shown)


common/nym-lp/src/message.rs
```diff
@@ -289,32 +301,45 @@ impl ForwardPacketData {
...
-        }
+        let target_lp_address_is_ipv6 = bytes[32] != 0;
+
+        let (target_lp_address, next_index) = if target_lp_address_is_ipv6 {
+            // IPv6, first check we have actually enough bytes
+            // smallest possible packet with ipv6 and empty data
+            if bytes.len() < 55 {
+                // 32 + 1 + 16 + 2 + 4 + 0
+                return Err(LpError::DeserializationError(format!(
+                    "Too few bytes to deserialise ipv6 ForwardPacketData[1]. got {}",
```
jstuczyn: nit: the `[1]` bit here doesn't make sense. in the previous code it was there since the rest of the error message was duplicated in two locations

gateway/src/node/lp_listener/registration.rs
```diff
@@ -486,6 +487,7 @@ impl LpHandlerState {
                 ticket_type,
                 wireguard_config: WireguardConfiguration {
                     public_key: gateway_pubkey,
+                    psk: None, // @JS do we still need to store that entire struct?
```
jstuczyn: no - I've changed that

jstuczyn (approved):

gateway/src/node/lp_listener/registration.rs
simonwicky: That's what I thought

PR #6412, opened 2026-02-03
Lp/ip pool fixes

jstuczyn:
this PR fixes the `IpPool` and more importantly unifies the process of Peer registration between the Authenticator and LP through the now shared `LpRegistrator`.

In general the registration has the following structure:
1. Initial request message is received
1.1. We check if the peer has already registered before -> if so, we returned the past information
1.2. We check if the peer already has a pending registration -> if so, we return the past information
1.3. We pre-allocated [`nym_wireguard::ip_pool::IpPair`] and save time-sensitive pending registration. If it does not complete within specified time interval, the information is going to get removed.
2. Finalisation request message is received, where credential has to be attached is verified.  Upon successful completion, pending registration is transformed into a properly inserted peer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6412)

### DIFF
--- a/common/authenticator-requests/src/util.rs
+++ b/common/authenticator-requests/src/util.rs
@@ -5,8 +5,8 @@ use nym_network_defaults::{WG_TUN_DEVICE_IP_ADDRESS_V4, WG_TUN_DEVICE_IP_ADDRESS
 use std::net::{Ipv4Addr, Ipv6Addr};
 
 pub fn authenticator_ipv6_to_ipv4(addr: Ipv6Addr) -> Ipv4Addr {
-    let before_last_byte = addr.octets()[2];
-    let last_byte = addr.octets()[3];
+    let before_last_byte = addr.octets()[14];
+    let last_byte = addr.octets()[15];
 
     Ipv4Addr::new(
         WG_TUN_DEVICE_IP_ADDRESS_V4.octets()[0],
--- a/common/authenticator-requests/src/util.rs
+++ b/common/authenticator-requests/src/util.rs
@@ -1,6 +1,38 @@
 // Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+use nym_network_defaults::{WG_TUN_DEVICE_IP_ADDRESS_V4, WG_TUN_DEVICE_IP_ADDRESS_V6};
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+pub fn authenticator_ipv6_to_ipv4(addr: Ipv6Addr) -> Ipv4Addr {
+    let before_last_byte = addr.octets()[2];
+    let last_byte = addr.octets()[3];
+
+    Ipv4Addr::new(
+        WG_TUN_DEVICE_IP_ADDRESS_V4.octets()[0],
+        WG_TUN_DEVICE_IP_ADDRESS_V4.octets()[1],
+        before_last_byte,
+        last_byte,
+    )
+}
+
+pub fn authenticator_ipv4_to_ipv6(addr: Ipv4Addr) -> Ipv6Addr {
+    let before_last_byte = addr.octets()[2];
+    let last_byte = addr.octets()[3];
+
+    let last_bytes = ((before_last_byte as u16) << 8) | last_byte as u16;
+    Ipv6Addr::new(
+        WG_TUN_DEVICE_IP_ADDRESS_V6.segments()[0],
+        WG_TUN_DEVICE_IP_ADDRESS_V6.segments()[1],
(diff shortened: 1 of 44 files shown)


common/authenticator-requests/src/util.rs
```diff
@@ -1,6 +1,38 @@
 // Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+use nym_network_defaults::{WG_TUN_DEVICE_IP_ADDRESS_V4, WG_TUN_DEVICE_IP_ADDRESS_V6};
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+pub fn authenticator_ipv6_to_ipv4(addr: Ipv6Addr) -> Ipv4Addr {
+    let before_last_byte = addr.octets()[2];
+    let last_byte = addr.octets()[3];
```
simonwicky: This is incorrect.
It should be 14 and 15.
I reckon for once a unit test for both fn could be real useful
E.g.
```
    let ipv6: Ipv6Addr = "fc01::1818".parse().unwrap();
    let ipv4: Ipv4Addr = "10.1.24.24".parse().unwrap();

    assert_eq!(authenticator_ipv4_to_ipv6(ipv4), ipv6);
    assert_eq!(authenticator_ipv6_to_ipv4(ipv6), ipv4);
```

common/credential-verification/src/upgrade_mode.rs
```diff
@@ -291,3 +291,40 @@ struct UpgradeModeStateInner {
...
+            let private_key =
+                ed25519::PrivateKey::from_bytes(&DUMMY_ATTESTER_ED25519_PRIVATE_KEY).unwrap();
+            private_key.public_key()
+        }
+
+        let upgrade_mode_state = UpgradeModeState::new(dummy_attester_public_key());
+        let upgrade_mode_details = UpgradeModeDetails::new(
+            UpgradeModeCheckConfig {
+                // essentially we never want to trigger this in our tests
+                min_staleness_recheck: Duration::from_nanos(1),
```
simonwicky, line 323: Wouldn't this always trigger a recheck? 
(Disclaimer I don't know exactly how it works, but I'd expect like 1 year for no recheck)
(There's another one similar somewhere else)

gateway/src/node/internal_service_providers/authenticator/mixnet_listener.rs
```diff
@@ -36,50 +22,28 @@ use nym_task::ShutdownToken;
...
-            registration_in_progress: Default::default(),
-            taken_private_network_ips: Default::default(),
-        }
-    }
-}
-
 pub(crate) struct MixnetListener {
     // The configuration for the mixnet listener
-    pub(crate) config: Config,
+    pub(crate) _config: Config,
```
simonwicky, line 37: nit : too hard to straight up remove the config?

gateway/src/node/lp_listener/handler.rs
```diff
@@ -1265,23 +1260,6 @@ mod tests {
         use nym_crypto::asymmetric::ed25519;
         use rand::rngs::OsRng;
```
simonwicky, line 1258: Lots of unused imports here according to my clippy

nym-gateway-probe/src/common/probe_tests.rs
```diff
@@ -252,7 +252,6 @@ pub async fn lp_registration_probe(
     );
     info!("  - Private IPv4: {}", gateway_data.private_ipv4);
     info!("  - Private IPv6: {}", gateway_data.private_ipv6);
-    info!("  - Endpoint: {}", gateway_data.endpoint);
     lp_outcome.can_register = true;
```
simonwicky, line 256: Why the `endpoint` deletion? 
`gateway_data` is what clients will see, so endpoint here should be correct

simonwicky (changes requested):

common/credential-verification/src/upgrade_mode.rs
jstuczyn: discussed on zulip. the name is just a bit misleading

gateway/src/node/internal_service_providers/authenticator/mixnet_listener.rs
jstuczyn: followed the pattern from other service providers in case values from the config become useful later on. so we might as well have it threaded through.

nym-gateway-probe/src/common/probe_tests.rs
jstuczyn: probably got lost in the rebase noise

simonwicky (approved):

PR #6419, opened 2026-02-04
[LP fix] Registration client with fallback

simonwicky:
Adapt registration client to LP :
- Don't start mixnet client unless needed. 
- Fallback to mixnet based registration if available
- `RegistrationMode::Lp` isn't a mode anymore, never should have been one

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6419)

### DIFF
--- a/nym-registration-client/src/builder/mod.rs
+++ b/nym-registration-client/src/builder/mod.rs
@@ -13,7 +13,12 @@ use nym_validator_client::{
     nyxd::{Config as NyxdClientConfig, NyxdClient},
 };
 
-use crate::{RegistrationClient, config::RegistrationClientConfig, error::RegistrationClientError};
+use crate::{
+    RegistrationClient,
+    clients::{LpBasedRegistrationClient, MixnetBasedRegistrationClient},
+    config::{RegistrationClientConfig, RegistrationMode},
+    error::RegistrationClientError,
+};
 use config::BuilderConfig;
 
 pub(crate) mod config;
@@ -27,7 +32,27 @@ impl RegistrationClientBuilder {
         Self { config }
     }
 
+    pub fn use_lp(&self) -> bool {
+        self.config.enable_lp_regitration
+            && self.config.entry_node.node.lp_data.is_some()
+            && self.config.exit_node.node.lp_data.is_some()
+            // To remove when LP supports Mixnet registration
+            && self.config.mode == RegistrationMode::Wireguard
+    }
+
     pub async fn build(self) -> Result<RegistrationClient, RegistrationClientError> {
+        if self.use_lp() {
+            Ok(RegistrationClient::Lp(Box::new(self.build_lp().await?)))
+        } else {
+            Ok(RegistrationClient::Mixnet(Box::new(
+                self.build_mixnet().await?,
+            )))
+        }
+    }
+
+    pub(crate) async fn build_mixnet(
+        self,
+    ) -> Result<MixnetBasedRegistrationClient, RegistrationClientError> {
         let storage = self.config.setup_storage().await?;
         let config = RegistrationClientConfig {
             entry: self.config.entry_node.clone(),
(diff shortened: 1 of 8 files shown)


nym-registration-client/src/builder/mod.rs
```diff
@@ -27,14 +32,29 @@ impl RegistrationClientBuilder {
         Self { config }
     }
 
+    pub fn use_lp(&self) -> bool {
+        self.config.enable_lp_regitration
```
jstuczyn: I think it'd be useful to have a log there if `enable_lp_regitration` is set to true, i.e. you expect to use LP reg, but any of the following requirements fail. so you can't use LP, even though you wanted to because, say, you don't have entry information for LP. I can already see debugging issues with "why is LP not being used??"

nym-registration-client/src/builder/mod.rs
```diff
@@ -93,6 +113,30 @@ impl RegistrationClientBuilder {
...
+
+        let nyxd_client = get_nyxd_client(&self.config.network_env)?;
+
+        let bandwidth_controller: Box<dyn BandwidthTicketProvider> =
+            if let Some(credential_storage) = storage {
+                Box::new(BandwidthController::new(credential_storage, nyxd_client))
+            } else {
+                Box::new(BandwidthController::new(
+                    EphemeralCredentialStorage::default(),
+                    nyxd_client,
```
jstuczyn, line 138: this is just a thought, but we'll have to do something about it because that `nyxd_client` is so deceiving there. it kinda implies we need to be querying the nyx chain, while if we use it in the vpn, this will never happen. maybe somehow keep it in mind when you eventually get time to look into the `BandwidthController`

nym-registration-client/src/clients/mixnet.rs
```diff
@@ -0,0 +1,148 @@
...
+        self.cancel_token
+            .clone()
+            .run_until_cancelled(async {
+                match self.config.mode {
+                    RegistrationMode::Mixnet => self.register_mix_exit().await,
+                    RegistrationMode::Wireguard => self.register_wg().await,
+                }
+            })
+            .await
+            .ok_or(RegistrationClientError::Cancelled)?
```
jstuczyn, line 146: it seems like we're missing some sort of cleanup on cancellation? i.e. if we get cancelled, shouldn't we also call disconnect on the client or is it done elsewhere?

nym-registration-client/src/clients/mixnet.rs
```diff
@@ -0,0 +1,148 @@
...
+            self.config.exit.keys,
+            self.config.exit.node.ip_address,
+        );
+
+        let entry_fut = entry_auth_client
+            .register_wireguard(&*self.bandwidth_controller, TicketType::V1WireguardEntry);
+        let exit_fut = exit_auth_client
+            .register_wireguard(&*self.bandwidth_controller, TicketType::V1WireguardExit);
+
+        let (entry, exit) = Box::pin(async { tokio::join!(entry_fut, exit_fut) }).await;
```
jstuczyn, line 105: do we want to be doing both registrations at the same time? 
if one of them fails, we end up losing two tickets.
but then, I guess, the process is quite lengthy so maybe this is is a sacrifice user is forced to make

jstuczyn (approved):

nym-registration-client/src/builder/mod.rs
simonwicky: I will add some logging

nym-registration-client/src/clients/mixnet.rs
simonwicky: Will double check

nym-registration-client/src/clients/mixnet.rs
simonwicky: Discussed in DMs, this was done to speed up registration

PR #6468, opened 2026-02-18
Add a port check option to the probe

benedettadavico:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6468)


nym-gateway-probe/src/config/netstack.rs
```diff
@@ -37,6 +42,18 @@ pub struct NetstackArgs {
...
+    #[arg(long = "use-target", default_value = "portquiz.net")]
+    pub port_check_target: String,
+
+    /// List ports to check, separated by a comma.
+    #[arg(long = "check-ports", value_delimiter = ',', default_values_t = Vec::<u16>::new())]
+    pub port_check_ports: Vec<u16>,
+
+    /// Timeout in seconds for each individual port check attempt
+    #[arg(long, default_value_t = 5)]
+    pub port_check_timeout_sec: u64,
```
dynco-nym, line 56: you need to add the fields to Default implementation a couple lines below
```
error[E0063]: missing fields `port_check_ports`, `port_check_target` and `port_check_timeout_sec` in initializer of `NetstackArgs`
  --> nym-gateway-probe/src/config/netstack.rs:61:9
   |
61 |         Self {
   |         ^^^^ missing `port_check_ports`, `port_check_target` and `port_check_timeout_sec`

For more information about this error, try `rustc --explain E0063`.
```

nym-gateway-probe/src/common/wireguard.rs
```diff
@@ -12,6 +12,7 @@ use tracing::{error, info};
 use crate::NetstackArgs;
 use crate::common::netstack::{NetstackRequest, NetstackRequestGo, NetstackResult};
 use crate::common::types::WgProbeResults;
+use crate::config::NetstackArgs;
```
dynco-nym: double import of the same thing

did you run `cargo check` before committing?

nym-gateway-probe/src/config/netstack.rs
```diff
@@ -37,6 +42,18 @@ pub struct NetstackArgs {
...
+    /// Target host for exit policy port checks (must listen on all tested ports)
+    #[arg(long = "use-target", default_value = "portquiz.net")]
+    pub port_check_target: String,
+
+    /// List ports to check, separated by a comma.
+    #[arg(long = "check-ports", value_delimiter = ',', default_values_t = Vec::<u16>::new())]
+    pub port_check_ports: Vec<u16>,
+
+    /// Timeout in seconds for each individual port check attempt
+    #[arg(long, default_value_t = 5)]
```
dynco-nym: please don't hardcode default only in `default_value_t` because that only makes it available to `clap`

rather, add the default value to the `Default` implementation of `NetStackArgs`, and use it here as `default_values_t = NetstackArgs::default().<field name>` as other fields do

dynco-nym (changes requested):

PR #6487, opened 2026-02-25
attempt at fixing ci

benedettadavico:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6487)


Cargo.toml
mfahampshire, line 438: Not sure if we need to define the version since we're not publishing it

.github/workflows/ci-crates-publish-dry-run.yml
```diff
@@ -15,6 +15,9 @@ env:
 jobs:
   publish-dry-run:
     runs-on: arc-linux-latest
+    timeout-minutes: 35
```
mfahampshire, line 18: where did the 35 minute timeout come from? out of interest. thinking if its related to the # of crates we're publishing we'll have to keep upping this as we publish more - not a problem, just something to probably comment / document somewhere

mfahampshire (approved):
approving with comments

PR #6503, opened 2026-03-01
fix: gate mix-fetch debug logs behind a build-time flag

### LINKED ISSUE #6397: console.log in mixFetch production build
Found this `console.log` that had slipped into the production code for mixFetch (TypeScript SDK):

https://github.com/nymtech/nym/blob/d3648f13c5e5e6c1dd3895e03c03443dddee7edd/sdk/typescript/packages/mix-fetch/src/create-mix-fetch.ts#L56

Because mixFetch is a library, it shouldn't be logging to the console, but instead logging to some log utility configured by the user. This way the user can opt-in/out of logging.

shuv-amp:
Keeps the existing mix-fetch debug logs available for local development, but removes them from production builds.

As raised in #6397, mix-fetch is a library and should not write debug output to the consumer's console in production. At the same time, the existing logs are useful when working on the package locally. This change gates those logs behind a build-time `MIX_FETCH_DEBUG` flag.

`build:dev`, `build:dev:esm`, and `build:dev:esm-no-inline` now enable debug logging. The production build path explicitly disables it, so release artifacts stay quiet even if `MIX_FETCH_DEBUG` is set in the shell.

Changes:
- add a small debug helper and test coverage for the flag behavior
- gate the existing logs in `create-mix-fetch.ts`, `worker/main.ts`, and `worker/wasm-loading.ts`
- thread `MIX_FETCH_DEBUG` through the ESM, CJS, and worker Rollup configs
- force `build-prod.sh` to compile with debug logging disabled

No API changes. The `console.error` for unhandled worker exceptions in `worker/index.ts` is unchanged.

Closes #6397

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6503)

### DIFF
--- a/sdk/typescript/packages/mix-fetch/package.json
+++ b/sdk/typescript/packages/mix-fetch/package.json
@@ -13,9 +13,9 @@
   "browser": "dist/esm/index.js",
   "scripts": {
     "build": "scripts/build-prod.sh",
-    "build:dev": "scripts/build.sh",
-    "build:dev:esm": "MIX_FETCH_DEV_MODE=true scripts/build-dev-esm.sh",
-    "build:dev:esm-no-inline": "scripts/build-dev-esm.sh",
+    "build:dev": "MIX_FETCH_DEBUG=true scripts/build.sh",
+    "build:dev:esm": "MIX_FETCH_DEV_MODE=true MIX_FETCH_DEBUG=true scripts/build-dev-esm.sh",
+    "build:dev:esm-no-inline": "MIX_FETCH_DEBUG=true scripts/build-dev-esm.sh",
     "build:worker": "rollup -c rollup-worker.config.mjs",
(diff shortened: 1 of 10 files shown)


mfahampshire:
Thanks for this @shuv-amp

mfahampshire:
I would prefer we don't just remove the `console.log`s here, ideally we'd actually do some proper debug logging which can be set by some kind of env var at build time, so we have them for when we're developing and need the debug logs. 

Is this something you'd want to have a go at doing?

shuv-amp:
Thanks, that makes sense. I’ll update this to gate the logs behind a build-time debug flag so they’re available when developing, but not in the production build.

mfahampshire:
Closing this as we're now moving to a new V2 of `mixFetch` with a new codebase - thanks for the PR though

PR #6507, opened 2026-03-03
chore: split up lp listener

jstuczyn:
this PR splits up the LP listener into explicit tasks for easier indepepdent work on control and data parts

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6507)


nym-node/src/node/mod.rs
```diff
@@ -487,28 +487,23 @@ impl NymNode {
         config.save()
     }
 
-    pub async fn build_lp_listener(
-        &mut self,
+    pub async fn build_lp(
```
simonwicky: nit: `build_lp_tasks`?

simonwicky (approved):

PR #6522, opened 2026-03-05
Remove dep leak of strum iterator

neacsu:
strum iterator over an enum leaks the version needed to iterate over it, which can cause problems to dependent crates that use a different strum version.

While at it, bump the strum crates as well

This is made clear by not being able to bump version in https://github.com/nymtech/nym-vpn-client/pull/4766 because the crate here is not yet bumped.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6522)


common/credentials-interface/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
+use strum::IntoEnumIterator as _;
```
pronebird, line 7: this `as _` seems to be superficial?

jstuczyn (approved):

common/credentials-interface/src/lib.rs
neacsu: It's not superficial, it's just a trait import which is not needed by name, only by implementation. If it were to be explicitly used by name somewhere then `as _` should be removed.

PR #6552, opened 2026-03-11
Channel based mixnode packet Ingress

jmwample:
## Description

Adjust the handling of mix packet ingress.

Changes:
- reduce the (per node in next layer) outbound mix packet buffer size to 128 (prev 2000)
- Add ingress buffer after framing, but before crypto operations (size 2000)
  - Add metric for packets dropped by ingress buffer
  - Add configuration for ingress buffer size   
- Add worker pool that takes from ingress buffer. 
  - MPMC channel
    - receives from _M_ framing decoder threads (one per conneciton from a previous layer node)
    - sends to _N_ sphinx ingress processing workers (worker count = available parallelism)
- remove batching for bloomfilter lookups. worker threads just block on lock 
  - bouned # of workers means finite wait time -- negligible next to sphinx ops
  - remove config related to bloom batch lookups 

## Goals

**Limit spurious drops** -- Network traffic comes in bursts and the sphinx operations being a bottleneck puts an upper bound on the rate at which traffic can be processed which will (usually) be lower than the inbound network traffic. If for short periods of time the network ingress rate exceeds the processing rate for sphinx packets we want to be able to buffer some extras that we will be able to handle in a reasonable amount of time. Previous to this PR this is done with egress buffering as the sphinx processing is done serially with frame decoding.

**Handle CPU saturation gracefully** -- If the network receive rate exceeds the rate at which packets can be processed for a continuous period of time or if the rate is so large in a short burst that we will not be able to process packets in time we need to have a way to drop packets and ensure that the packets that do get processed are still relevant

**Handle Network saturation gracefully** -- Nodes may also read from network in bursts, and if a node is busy working on somthing else or is network limited we want to give them some grace and keep a short buffer of packets to be sent to them. 

**Bound packet delays** -- buffering packets at node egress means that in the worst case there are `N = # or nodes` full buffers of packets that a saturated node must process. The receiving node has no control over these buffers and the sending nodes have no information about the processing capabilities of the receiver. This results in severly delayed packets, but the reciever has no information about how delayed they are and must process them and pass them on.  

**DO NOT PROCESS OVERLY DELAYED PACKETS** -- Receiving and parsing the framing of packets should be fast, so egress buffers should see very few drops and add very little delay. Ingress buffers allow a node to tune how delayed a packet could be. if the node can process 5000 packets per second, then a buffer of size 5000 should limit the maximum delay to around 1 second. This helps to ensure that packets that likely would not be processed in time are dropped. So that downstream nodes do not waste sphinx operations on packets that are already too delayed to be useful. 

### Before 
ingress_direct(1)

### After
ingress_channeled(1)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6552)


jmwample:
I want to note that I have not tested this because I am not the most handy with node deployments and test environments. Accordingly I don't know that the changes here improve the processing towards the goals I talk about in the PR. Or even that I didn't introduce a bug and miss something in the sphinx processing because I haven't run the built node yet. 

That is something that I could use help with.

PR #6553, opened 2026-03-11
Version bump

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6553)


mfahampshire:
we can ignore the docs build failure - its because it can't resolve the new version of mixfetch. it will build when i run the docs build CI after we've published it.

mmsinclair (approved):

PR #6558, opened 2026-03-12
critical bugfix

simonwicky:
Debug bytes should use uppercase for hexadecimal digits

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6558)


simonwicky:
Closed because some people don't like uppercases

PR #6555, opened 2026-03-12
Feat/nmv3 agents contract

jstuczyn:
NYM-760.

this PR introduces the smart contract storing information about ip addresses of the authorised NM agents

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6555)

PR #6567, opened 2026-03-17
Feat/nmv3 agents subscription

jstuczyn:
NR-759

# Nym node chain watcher and network monitors agents bypass

## Summary

This PR introduces real-time blockchain monitoring for network monitor agent authorizations and implements replay protection bypass for authorized network monitor (NM) agents. This allows NM agents to perform node stress testing, while maintaining security for regular traffic.

### 1. Real-Time Blockchain Subscription for Network Monitor Agents
- **New Module**: `nym-node/src/node/nyxd_watcher/network_monitor_agents.rs`
  - Implements `MsgModule` trait to watch for Network Monitors smart contract messages
  - Automatically updates authorized NM agent list when `AuthoriseNetworkMonitor`, `RevokeNetworkMonitor`, or `RevokeAllNetworkMonitors` messages are executed on-chain

- **Integration**: Websocket subscription to nyx blockchain using the enhanced `nyxd-scraper` framework
  - Real-time updates without polling overhead
  - Survives transient network failures with automatic reconnection

### 2. Replay Protection Bypass for Authorized Network Monitors
- **Implementation Location**: `nym-node/src/node/mixnet/handler.rs`
  - Checks if replayed packet originates from an authorized NM agent IP
  - Allows NM agents to bypass replay detection while enforcing it for all other traffic
  - Maintains per-connection authorization check via `is_from_authorised_network_monitor_agent()`

### 3. Enhanced Prometheus Metrics
- **New/Fixed Metrics**:
  - `ingress_replayed_packet()` - Now properly tracked when replay detection triggers
  - `ingress_network_monitor_packet()` - Counts packets from authorized NM agents

### Packet Journey for Replay Protection
```
handle_received_nym_packet()
  ↓
handle_received_packet_with_replay_detection()
  ↓
try_partially_unwrap_packet() [derive secrets, check MAC]
  ↓
batch_check_and_set() [bloomfilter replay detection]
  ↓
handle_post_replay_detection_packets()
  ↓
if replayed && !is_from_authorised_network_monitor_agent():
    → drop packet + increment metrics
else:
    → finalise_unwrapping() + forward packet
```

### Authorization Flow
```
Contract ExecuteMsg::AuthoriseNetworkMonitor { address: IpAddr }
  ↓
Blockchain transaction committed
  ↓
NetworkMonitorAgentsModule receives message via MsgModule
  ↓
DeclaredNetworkMonitors.add_known(address)
  ↓
ArcSwap updates live filter (lock-free)
  ↓
ConnectionHandler checks is_from_authorised_network_monitor_agent()
```

## Testing Considerations
- Network monitor agents can now intentionally send replayed packets for testing
- Operators can monitor NM activity via new Prometheus metrics
- Real-time authorization updates without node restarts

## Breaking Changes
None - fully backward compatible with existing configurations.

## Security Implications
- Authorization is on-chain and requires execute privileges
- Only specific IP addresses can bypass replay protection
- All bypass activity is logged and metrics-tracked
- Operators should monitor `ingress_network_monitor_packet` metric for unexpected activity

## Related Issues
- Depends on: Network Monitors smart contract deployment (https://github.com/nymtech/nym/pull/6555)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6567)


nym-node/src/config/mod.rs
```diff
@@ -540,6 +566,56 @@ impl Default for Http {
...
+        #[allow(clippy::expect_used)]
+        let nyxd_urls = if let Ok(env_value) = env::var(var_names::NYXD) {
+            parse_urls(&env_value)
+        } else {
+            vec![mainnet::NYXD_URL.parse().expect("invalid default nyxd URL")]
+        };
+
+        #[allow(clippy::expect_used)]
+        let nyxd_websocket_url = if let Ok(env_value) = env::var(var_names::NYXD_WEBSOCKET) {
+            read_parsed_var(&env_value).expect("malformed default nyxd websocket URL")
```
simonwicky, line 606: This will actually crash, because `read_parsed_var` expect the name of the var (it does `std::env:var`) and not the value of the var

nym-node/src/node/routing_filter/network_filter.rs
```diff
@@ -84,17 +117,122 @@ impl UnknownNodes {
...
+    pub(crate) fn new(known: HashSet<IpAddr>) -> Self {
+        Self {
+            inner: Arc::new(DeclaredNetworkMonitorsInner {
+                known: ArcSwap::from_pointee(known),
+            }),
+        }
+    }
+
+    fn swap(&self, new: HashSet<IpAddr>) {
+        self.inner.known.store(Arc::new(new))
```
simonwicky, line 179: That's not a compare and swap, if by any chance two threads arrive at the same time, the first update will be lost

nym-node/src/node/shared_network/refresher.rs
simonwicky, line 269: We're not retrying on that error?

simonwicky, line 139: If the pending addresses are already resolve, we might want to clear them no?

nym-node/src/node/nym_apis_client.rs
```diff
@@ -7,15 +7,19 @@ use futures::{StreamExt, stream};
...
+use nym_validator_client::nym_api::{NodesByAddressesResponse, SemiSkimmedNodesWithMetadata};
 use rand::prelude::SliceRandom;
 use rand::thread_rng;
+use std::net::IpAddr;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, RwLockReadGuard};
 use tokio::time::sleep;
+use tracing::log::error;
```
simonwicky: `tracing::error` not `tracing::log::error`

nym-node/src/node/shared_network/refresher.rs
simonwicky, line 155: What if some addresses were added during the execution of that? They will get cleared without being checked

common/network-defaults/src/mainnet.rs
```diff
@@ -141,6 +141,11 @@ pub fn read_parsed_var_if_not_default<T: std::str::FromStr>(
         .map(std::str::FromStr::from_str)
 }
 
+#[cfg(feature = "env")]
+pub fn read_parsed_var<T: std::str::FromStr>(var: &str) -> Result<T, T::Err> {
```
simonwicky, line 145: The callsite of that function is incorrect, giving a var value and calling expect on it

common/nyxd-scraper-shared/src/block_processor/types.rs
```diff
@@ -27,11 +32,26 @@ pub struct ParsedTransactionResponse {
...
 
-    pub block: Block,
+// just get all everything out of tx::Response, but parse raw `tx` bytes
+#[derive(Clone, Debug)]
+pub struct ParsedTransactionResponse {
+    pub tx_details: ParsedTransactionDetails,
+
+    pub decoded_messages: BTreeMap<usize, DecodedMessage>,
+    /*
+        pub parsed_messages: BTreeMap<usize, serde_json::Value>,
```
simonwicky: Any reason to keep that commented?

nyx-chain-watcher/src/chain_scraper/mod.rs
```diff
@@ -165,7 +153,7 @@ impl MsgModule for BankScraperModule {
             return Ok(());
         }
 
-        let msg = self.recover_bank_msg(tx.hash, index, msg)?;
+        let msg: MsgSend = parse_msg(msg)?;
```
simonwicky, line 156: Can't use `DecodedMessage` here?

nym-node/src/config/mod.rs
simonwicky: Also, is the `expect` fine here? It can fail since it's gonna be a custom url and not a default one

nym-node/src/config/mod.rs
```diff
@@ -540,6 +566,56 @@ impl Default for Http {
...
+}
+
+impl Default for Nyx {
+    fn default() -> Self {
+        // SAFETY:
+        // our hardcoded values should always be valid
+        // is if there's anything set in the environment, otherwise fallback to mainnet
+        #[allow(clippy::expect_used)]
+        let nyxd_urls = if let Ok(env_value) = env::var(var_names::NYXD) {
+            parse_urls(&env_value)
```
simonwicky, line 599: Using `parse_urls`, which has an `expect` in it, with a non default value, going against the comment on the fn

nym-node/src/node/mixnet/handler.rs
```diff
@@ -446,19 +474,38 @@ impl ConnectionHandler {
...
+                // CRITICAL SECURITY DECISION POINT: Replay Protection Bypass for Network Monitors
+                //
+                // This is where we decide whether to enforce replay protection for this packet.
+                // The decision tree is:
+                //
+                // 1. Is packet replayed? (bloomfilter check already completed)
+                //    NO  → Process normally (finalise_unwrapping)
+                //    YES → Go to step 2
+                //
+                // 2. Is source IP an authorised network monitor?
```
simonwicky, line 510: This might not be enough, IP addresses can be spoofed

nym-node/src/node/nyxd_watcher/network_monitor_agents.rs
```diff
@@ -0,0 +1,118 @@
...
+        tx: &ParsedTransactionDetails,
+    ) -> Result<(), ScraperError> {
+        // don't process failed transactions
+        if !tx.tx_result.code.is_ok() {
+            return Ok(());
+        }
+
+        // propagate error as this is a critical failure indicating our code is incompatible with
+        // the current CometBFT schema so parsing can't proceed
+        let execute_msg: MsgExecuteContract = parse_msg(msg)?;
```
simonwicky, line 78: Can't use `DecodedMessage`?

nym-node/src/node/nyxd_watcher/network_monitor_agents.rs
```diff
@@ -0,0 +1,118 @@
...
+use tracing::error;
+
+/// Blockchain message handler for Network Monitor agent authorisation events.
+///
+/// Watches for `MsgExecuteContract` messages targeting the Network Monitors smart contract
+/// and updates the runtime list of authorised agents accordingly.
+///
+/// # Thread Safety
+///
+/// Safe to use across threads - updates to `network_monitors` use lock-free ArcSwap internally.
```
simonwicky: Actually not, as the update is not a compare and swap. Badly timed updates will overwrite each other. See comment in `network_filter.rs`

simonwicky (changes requested):

nym-node/src/config/mod.rs
jstuczyn: indeed - I have run into this issue and fixed it as part of https://github.com/nymtech/nym/pull/6582

nym-node/src/node/routing_filter/network_filter.rs
jstuczyn: which is not a problem because there is only ever one thread performing writes to the list of known addresses. the arcswap exists here for the purposes of lock-free reads and eventual consistency

nym-node/src/node/shared_network/refresher.rs
jstuczyn: nope. if there are network failures - we're out of luck. it means API is down. we only want to retry if the network is still bootstrapping and nodes themselves are not yet alive

jstuczyn: the likelihood of that is quite low as there's only a very limited number of addresses that could be added to the filter ("known" set of nym-nodes). but if that happens, then it will have to wait for resolution until the next cycle

common/network-defaults/src/mainnet.rs
jstuczyn: I'm not sure I understand? we read provided environmental variable and attempt to parse it into desired type

common/nyxd-scraper-shared/src/block_processor/types.rs
jstuczyn: artifact of refactoring - removed

nyx-chain-watcher/src/chain_scraper/mod.rs
jstuczyn: unfortunately no, because `DecodedMessage` is not a `MsgSend`. it's just an opaque json. so some additional parsing would have been needed anyway

common/network-defaults/src/mainnet.rs
simonwicky: It was related to the only use of that function, which was incorrect but fixed by you in a later PR

nym-node/src/node/routing_filter/network_filter.rs
simonwicky: Can we then update the comment about it being thread-safe?

nym-node/src/config/mod.rs
jstuczyn: yes. that looks like an issue. I've copied that pattern from other instances where it's used. it should be fixed, but I'd say it's out of scope for this PR

nym-node/src/node/nyxd_watcher/network_monitor_agents.rs
jstuczyn: same reasoning as `BankMsg` - it's not the right type

nym-node/src/node/nyxd_watcher/network_monitor_agents.rs
jstuczyn: as I mentioned there's only a single writer thread, so it's fine. the module wasn't designed with multiple writers in mind

simonwicky (approved):

PR #6569, opened 2026-03-17
nym-data-observatory: add CLI command to process blocks from a range

mmsinclair:
This feature adds a command to process (or reprocess) block ranges:

You can run the scraper to operate on a range of blocks, without listening for new blocks. This is useful for:

- processing historical data
- reprocessing block ranges

Run this as follows to process blocks starting at 1000 and ending at 2000:

```bash
./nym_data_observatory process --start 1000 --end 2000
```

Or to process a single block:

```bash
./nym_data_observatory process --start 1000
```

Or to process 10 blocks starting at block 1000:

```bash
./nym_data_observatory process --start 1000 --blocks 10
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6569)

### DIFF
--- a/common/nyxd-scraper-shared/src/scraper.rs
+++ b/common/nyxd-scraper-shared/src/scraper.rs
@@ -119,6 +119,22 @@ where
         Ok(scraper)
     }
 
+    pub async fn build_unsafe(self) -> Result<NyxdScraper<S>, ScraperError> {
+        self.config.pruning_options.validate()?;
+        let storage =
+            S::initialise(&self.config.database_storage, &self.config.run_migrations).await?;
+        let rpc_client = RpcClient::new(&self.config.rpc_url)?;
+
+        Ok(NyxdScraper {
+            config: self.config,
+            task_tracker: TaskTracker::new(),
+            cancel_token: CancellationToken::new(),
+            startup_sync: Arc::new(Default::default()),
+            storage,
+            rpc_client,
+        })
+    }
+
     pub fn new(config: Config) -> Self {
         NyxdScraperBuilder {
             _storage: PhantomData,
--- a/nym-data-observatory/src/chain_scraper/mod.rs
+++ b/nym-data-observatory/src/chain_scraper/mod.rs
@@ -1,4 +1,3 @@
-use crate::cli::commands::run::Args;
 use crate::db::DbPool;
 use nyxd_scraper_psql::{PostgresNyxdScraper, PruningOptions};
 use tracing::info;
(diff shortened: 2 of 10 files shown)


common/nyxd-scraper-shared/src/scraper.rs
```diff
@@ -119,6 +119,22 @@ where
         Ok(scraper)
     }
 
+    pub async fn build_unsafe(self) -> Result<NyxdScraper<S>, ScraperError> {
```
jstuczyn, line 122: why redefine the function that already exists? i.e. `NyxdScraper::new(config)`

jstuczyn: also, I'd perhaps rename it and remove the `unsafe` suffix. it just creates an instance in a non-started state

nym-data-observatory/src/cli/commands/process/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+                &config,
+                scraper_pool,
+                start_block_height,
+                end_block_height,
+            )
+            .await?;
+
+            info!("⏰ shutting down...");
+            cancel_after_processing.cancel();
+
```
jstuczyn, line 70: to ensure graceful shutdown we probably want to call `scraper.stop().await` here to wait until its internal task tracker returns

nym-data-observatory/src/chain_scraper/mod.rs
```diff
@@ -40,3 +39,41 @@ pub(crate) async fn run_chain_scraper(
...
+        pruning_options: PruningOptions::nothing(),
+        store_precommits: false,
+        start_block: nyxd_scraper_psql::StartingBlockOpts {
+            start_block_height: Some(start_block_height),
+            use_best_effort_start_height: false,
+        },
+        run_migrations: false, // ignore the base migrations
+    })
+    .with_msg_module(crate::modules::wasm::WasmModule::new(connection_pool))
+    .with_tx_module(webhook::WebhookModule::new(config.clone())?);
```
jstuczyn, line 69: are we sure we want to be calling the webhook on reprocessing transactions? wouldn't this accidentally create duplicate subscriptions or something along those lines?

mmsinclair: the webhook is optional, so if you don't want it to trigger, don't have it set or configured

you're right, it will duplicate things if you do have it set - it is the responsibility of the webhook receiver to handle duplicates properly (a big ask)

PR #6570, opened 2026-03-17
HTTP domain rotation conditions

jmwample:
This is a relatively small changeto add more explicit handling for df enable and domain rotation.  Mostly I wanted to make sure that requests that fail because the network is unreachable are not triggering domain rotation as it will not help and might confuse things. In general though I think specificity in the handling here is a good thing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)

### DIFF
--- a/common/http-api-client/src/lib.rs
+++ b/common/http-api-client/src/lib.rs
@@ -1214,6 +1214,8 @@ impl ApiClientCore for Client {
     }
 }
 
+const MAX_ERR_SOURCE_ITERATIONS: usize = 4;
+
 /// only if there was a network issue should we consider updating the host info
 pub(crate) fn is_network_error(err: &reqwest::Error) -> bool {
     if err.is_timeout() {
@@ -1225,31 +1227,32 @@ pub(crate) fn is_network_error(err: &reqwest::Error) -> bool {
         return false;
     }
 
-    // reqwest::Error -> hyper_util::Error
-    if let Some(inner1) = err.source() {
-        // hyper_util::Error -> hyper_util::ClientError
-        if let Some(inner2) = inner1.source() {
-            // hyper_util::ClientError -> io::Error
-            if let Some(inner3) = inner2.source() {
-                // downcase from dyn
-                if let Some(io_err) = inner3.downcast_ref::<std::io::Error>() {
-                    match io_err.kind() {
-                        // device not connected to the internet
-                        ErrorKind::NetworkUnreachable | ErrorKind::NetworkDown => return false,
-                        // custom errors may be case by case, but in general not DF related
-                        // -- includes DNS errors for hyper_util
-                        ErrorKind::Other => return false,
-                        // timeouts can indicate packet drops or adress blocklisting
-                        // -- this should hit the timeout conditional above
-                        ErrorKind::TimedOut => return true,
-                        // connection errors can indicate connection interference
-                        ErrorKind::ConnectionReset
-                        | ErrorKind::HostUnreachable
-                        | ErrorKind::ConnectionRefused => return true,
-                        _ => return false,
-                    }
+    // The io::Error source is several layers deep, for clarity this is done as a loop
+    // * reqwest::Error -> hyper_util::Error
+    // * hyper_util::Error -> hyper_util::ClientError
+    // * hyper_util::ClientError -> io::Error
+    let mut inner = err.source();
+    for _ in 0..MAX_ERR_SOURCE_ITERATIONS {
+        if let Some(e) = inner {
+            // try downcast to io::Error from <dyn std::error:Error>
+            if let Some(io_err) = e.downcast_ref::<std::io::Error>() {
+                match io_err.kind() {
+                    // device not connected to the internet
+                    ErrorKind::NetworkUnreachable | ErrorKind::NetworkDown => return false,
+                    // custom errors may be case by case, but in general not DF related
+                    // -- includes DNS errors for hyper_util
+                    ErrorKind::Other => return false,
+                    // connection errors can indicate connection interference
+                    ErrorKind::ConnectionReset
+                    | ErrorKind::HostUnreachable
+                    | ErrorKind::ConnectionRefused => return true,
+                    _ => return false,
                 }
+            } else {
+                inner = e.source();
             }
+        } else {
+            break;
         }
     }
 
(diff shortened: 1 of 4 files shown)


pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on jmwample and jstuczyn).

___
*[`common/http-api-client/src/lib.rs` line 1229 at r1](https://reviewable.io/reviews/nymtech/nym/6570#-Onx2NWZ84MZ42N78ivN:-Onx2NWZ84MZ42N78ivO:b-1hgrx7) ([raw file](https://github.com/nymtech/nym/blob/05211828c3bb0daadbb6bdfdaa0a829879fc4808/common/http-api-client/src/lib.rs#L1229)):*
> ```XML
> 
>     // reqwest::Error -> hyper_util::Error
>     if let Some(inner1) = err.source() {
> ```

This is a very fine cascade. :-)  Maybe this could be a loop looking for `std::io::Error`?

pronebird (changes requested):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on jmwample and jstuczyn).

___
*[`common/http-api-client/src/lib.rs` line 1219 at r1](https://reviewable.io/reviews/nymtech/nym/6570#-Onx3HZ38DenvIqxl18Y:-Onx3HZ38DenvIqxl18Z:b-4r9rw9) ([raw file](https://github.com/nymtech/nym/blob/05211828c3bb0daadbb6bdfdaa0a829879fc4808/common/http-api-client/src/lib.rs#L1219)):*
> ```XML
> /// only if there was a network issue should we consider updating the host info
> pub(crate) fn is_network_error(err: &reqwest::Error) -> bool {
>     if err.is_timeout() {
> ```

You already handle `ErrorKind::TimedOut` below. Perhaps you could remove that statement then?

jmwample (commented):
> @jmwample made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on jstuczyn and pronebird).

___
*[`common/http-api-client/src/lib.rs` line 1219 at r1](https://reviewable.io/reviews/nymtech/nym/6570#-Onx3HZ38DenvIqxl18Y:-OnxDVrdDD-BkXzvbGF3:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/05211828c3bb0daadbb6bdfdaa0a829879fc4808/common/http-api-client/src/lib.rs#L1219)):*

You already handle `ErrorKind::TimedOut` below. Perhaps you could remove that statement then?
</blockquote></details>

Done.
___
*[`common/http-api-client/src/lib.rs` line 1229 at r1](https://reviewable.io/reviews/nymtech/nym/6570#-Onx2NWZ84MZ42N78ivN:-OnxDWOw49i3R2piJ2J1:b-896fix) ([raw file](https://github.com/nymtech/nym/blob/05211828c3bb0daadbb6bdfdaa0a829879fc4808/common/http-api-client/src/lib.rs#L1229)):*

This is a very fine cascade. :-)  Maybe this could be a loop looking for `std::io::Error`?
</blockquote></details>

Done.

common/http-api-client/src/lib.rs
```diff
@@ -1222,6 +1220,52 @@ impl ApiClientCore for Client {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+const MAX_ERR_SOURCE_ITERATIONS: usize = 4;
+
+/// only if there was a network issue should we consider updating the host info
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) fn is_network_error(err: &reqwest::Error) -> bool {
```
jstuczyn: just a tiny nit. I'm not entirely sure "is network error" is the most appropriate name given that technically network being unreachable or down are network errors, which we actually want to avoid.

how about something along the lines of "is peer error"? (naming things is difficult...)

jstuczyn (approved):

common/http-api-client/src/lib.rs
pronebird: Peer error isn't less ambiguous. On the other side, I wonder if TLS/cert errors should be considered as network errors. For example in case of MITM.

pronebird (commented):
> @pronebird made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on jstuczyn).

jmwample:
I fixed the failing test and added handling for more error types. Also if you want a simpler place to play with this I made a PoC gist https://gist.github.com/jmwample/f6827a9c24c620a597a7d4118a64bbb8. 

WRT naming -- this is an internal function. I can just make sure that the doc comment is good and specific as to what this is checking.

common/http-api-client/src/lib.rs
jmwample: TLS Errrors are now considered. DNS NXDomain as well. There may be more errors we want here, the nice thing is that this lets us be as specific as we need to be.

WRT Naming - this is internal so as long as docs are clear this shouldn't cause too much confusion.

jmwample (commented):
> @jmwample made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on jstuczyn and pronebird).

pronebird (commented):
> @pronebird reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on jstuczyn).

pronebird (approved):
:lgtm:

> @pronebird reviewed 4 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6570#-:-Oo4dwNlALcEVaEYnmad:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on jstuczyn).

PR #6573, opened 2026-03-17
Max/lp stream framing

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6573)

### DIFF
--- a/common/nym-lp/src/packet/frame.rs
+++ b/common/nym-lp/src/packet/frame.rs
@@ -9,13 +9,13 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 #[derive(Debug, Clone, PartialEq)]
 pub struct LpFrameHeader {
     pub kind: LpFrameKind,
-    pub frame_attributes: [u8; 14],
+    pub frame_attributes: LpFrameAttributes,
 }
 
 impl LpFrameHeader {
     pub const SIZE: usize = 16; // message_kind(2) + message_attributes(14)
 
-    pub fn new(kind: LpFrameKind, frame_attributes: [u8; 14]) -> Self {
+    pub fn new(kind: LpFrameKind, frame_attributes: LpFrameAttributes) -> Self {
         Self {
             kind,
             frame_attributes,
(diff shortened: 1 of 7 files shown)


sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
```diff
@@ -295,7 +296,7 @@ pub(crate) async fn open_stream(
     let rx = streams.register_stream(stream_id).await;
 
     // Send Open to the peer
-    let wire = encode_stream_message(&stream_id, StreamMessageType::Open, &[]);
+    let wire = encode_stream_message(&stream_id, StreamMsgType::Open, 0, &[]);
```
simonwicky: Random question : What if we don't send 0 when we open? 
Does the receiver adapt its SEQ, kinda like TCP?

simonwicky (approved):

nym-node/src/node/lp/control/ingress/client_handler.rs
```diff
@@ -261,7 +261,7 @@ where
                 self.handle_forwarding_request(receiver_idx, forward_data)
                     .await
             }
-            typ @ LpFrameKind::Opaque => {
+            typ @ (LpFrameKind::Opaque | LpFrameKind::Stream) => {
```
jstuczyn: given the possibly growing number of types, perhaps remove the `@ ...` binding

common/nym-lp/src/packet/frame.rs
```diff
@@ -115,6 +122,62 @@ pub enum LpFrameKind {
     Opaque = 0,
     Registration = 1,
     Forward = 2,
+    Stream = 3,
```
jstuczyn: Perhaps rename it to `SphinxStream` or something, because I can imagine a generic stream would have more than 2 possible message types : )

common/nym-lp/src/packet/frame.rs
```diff
@@ -115,6 +122,62 @@ pub enum LpFrameKind {
...
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct StreamFrameAttributes {
+    pub stream_id: u64,
+    pub msg_type: StreamMsgType,
+    pub sequence_num: u32,
+}
+
+impl StreamFrameAttributes {
+    pub fn encode(&self) -> [u8; 14] {
```
jstuczyn: that's my fault it doesn't already exist, but since you're modifying this code, I will  politely ask you to fix it : D
could you add a type alias for those `[u8; 14]` `i.e. pub type LpFrameAttributes = [u8; 14]` so it'd be a bit more explicit what those represent?

jstuczyn (approved):

sdk/rust/nym-sdk/src/mixnet/stream/mod.rs
mfahampshire: Currently hardcoded as we don't have message ordering in place *yet* in the SDK streams - when it *is* added then it will set the receiver's expected starting seq number. Gives us the ability down the road to e.g. pick up a dropped stream from where it left off, was my thinking.

PR #6576, opened 2026-03-18
Fix socks5 GW probe regression

dynco-nym:
- initial PR for socks5 tests had some [socks5 specific config](https://github.com/nymtech/nym/pull/6393/changes#diff-e61ca4cee8cec1346daf63820fbbd5fead61f6590b86f750b9546a84093bc38a) that made it work
- some of it was removed [here](https://github.com/nymtech/nym/pull/6400/changes#diff-78489a6999fab434ab161e688e0717f8cbcabf34e8114a6e9d8659c35f8935e1L680) so we had a regression

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6576)

### DIFF
--- a/nym-gateway-probe/src/common/probe_tests.rs
+++ b/nym-gateway-probe/src/common/probe_tests.rs
@@ -447,7 +447,6 @@ pub(crate) async fn do_socks5_connectivity_test(
     entry_gateway_id: NodeIdentity,
     network_details: NymNetworkDetails,
     socks5_args: Socks5Args,
-    maybe_topology: Option<NymTopology>,
 ) -> anyhow::Result<Socks5ProbeResults> {
     info!(
         "Starting SOCKS5 test through Network Requester: {}",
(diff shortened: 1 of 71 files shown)


nym-gateway-probe/src/common/probe_tests.rs
```diff
@@ -463,23 +461,31 @@ pub(crate) async fn do_socks5_connectivity_test(
...
+
     // create ephemeral SOCKS5 client
     let socks5_config = Socks5::new(nr_recipient.to_string());
 
-    // debug config similar to main probe
     let debug_config = mixnet_debug_config(min_gw_performance, true);
 
-    let mut socks5_client_builder = MixnetClientBuilder::new_ephemeral()
+    // don't reuse topology: our gateway might be filtered out in it
+    info!("Fetching topology for SOCKS5 client...");
```
simonwicky, line 475: Not reusing the topology we had for the mixnet test (if any)?
Was this causing problem?

nym-gateway-probe/src/common/socks5_test/json_rpc_client.rs
```diff
@@ -98,15 +98,23 @@ impl JsonRpcClient {
         {
             Ok(response) => {
                 let status = response.status();
+                let response_text = response
+                    .text()
+                    .await
+                    .context("Failed to extract response text")
+                    .unwrap_or_else(ToString::to_string);
```
simonwicky: `rustc` doesn't like that part on my side. Is it working for you?

simonwicky (approved):

nym-gateway-probe/src/common/socks5_test/json_rpc_client.rs
jstuczyn: yep. this is quite a weird fallback. ideally we should avoid `unwrap_or_else` (or `unwrap_or_default`) as it's been causing us a lot of issues. but in this case just change the call

jstuczyn (approved):

nym-gateway-probe/src/common/socks5_test/json_rpc_client.rs
dynco-nym: while I agree in general, I think it's fine here. I don't want an early return (.map_err)

instead, I use unwrap_or_else. If that code ever executes, it would mean response body was unreadable, which would definitely fail serde deserialization in the subsequent lines, and return a proper error there

nym-node-status-api/nym-node-status-agent/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "nym-node-status-agent"
-version = "1.1.3"
+version = "1.1.4-rc.2"
```
dynco-nym: TODO: remove rc suffix before merge

PR #6577, opened 2026-03-18
[feat] propagate NM agent noise keys to nym-node routing

jstuczyn:
Network Monitor agents need to perform Noise protocol handshakes with nym-nodes, just like any other nym-node on the network. Previously the contract stored only the agent's IP address, and the noise key map in nym-node only tracked keys for nym-api-sourced nodes. This PR closes the gap end-to-end: from on-chain registration through to live packet routing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6577)

### DIFF
--- a/contracts/network-monitors/src/storage.rs
+++ b/contracts/network-monitors/src/storage.rs
@@ -153,6 +153,8 @@ impl NetworkMonitorsStorage {
         env: &Env,
         sender: &Addr,
         monitor_address: IpAddr,
+        bs58_x25519_noise: String,
+        noise_version: u8,
     ) -> Result<(), NetworkMonitorsContractError> {
         // only orchestrators can authorise new monitors
         self.ensure_is_orchestrator(deps.as_ref(), sender)?;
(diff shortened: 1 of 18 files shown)


contracts/network-monitors/src/storage.rs
```diff
@@ -135,18 +184,22 @@ impl NetworkMonitorsStorage {
...
         self.authorised_agents.save(
             deps.storage,
-            monitor_address.to_string(),
+            monitor_address.into(),
             &AuthorisedNetworkMonitor {
-                address: monitor_address,
+                mixnet_address: monitor_address,
                 authorised_by: sender.clone(),
                 authorised_at: env.block.time,
+                bs58_x25519_noise,
```
simonwicky, line 218: Could we have some check on that? 
At best check it's an actual pub key, at worst that it's an actual base58 string of length 32?
Right now we can submit the entirety of "War and Peace" if we want to

simonwicky (approved):

PR #6582, opened 2026-03-19
feat: network monitor agent - standalone node stress-testing

jstuczyn:
# Summary

Introduces the nym-network-monitor-agent binary. The agent can connect to a single mixnode and run a controlled stress test against it without requiring an orchestrator, making it useful for manual diagnostics as well as the foundation for the future automated pipeline.

Also scaffolds the nym-network-monitor-orchestrator crate (empty for now) to be added in the future PR(s).

# What the agent does

The `test-node` subcommand executes a five-step protocol against a target node:

1. Egress connection - opens an outbound Noise-encrypted TCP connection to the node
2. Ingress listener - binds a local TCP listener to receive sphinx packets the node sends back
3. Connectivity probe - sends a single packet in isolation to confirm the node responds and records the baseline latency
4. Bloomfilter probe (when `--reuse-header`) - replays the same sphinx header to verify the node's bloomfilter bypass (i.e. the chain watcher) is correctly configured and the node is not outdated
5. Load test - sends packets at `--target-rate` packets/s (dispatched in batches of `--sending-batch-size`) for `--sending-duration`, then waits up to `--waiting-duration` for stragglers and computes RTT statistics

The result includes noise handshake durations, packets sent/received, RTT distribution, a duplicate-packet flag, and an error message if any step aborted early.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6582)


nym-node/src/config/mod.rs
```diff
@@ -603,7 +602,9 @@ impl Default for Nyx {
 
         #[allow(clippy::expect_used)]
         let nyxd_websocket_url = if let Ok(env_value) = env::var(var_names::NYXD_WEBSOCKET) {
-            read_parsed_var(&env_value).expect("malformed default nyxd websocket URL")
+            env_value
+                .parse()
+                .expect("malformed default nyxd websocket URL")
```
simonwicky, line 607: That except can still fire because it can be a custom url. Are we okay with that?

nym-node/src/node/routing_filter/network_filter.rs
```diff
@@ -188,7 +188,14 @@ impl RoutableNetworkMonitors {
     }
 
     pub(crate) fn is_known(&self, address: &IpAddr) -> bool {
-        self.inner.known.load().contains(address)
+        let guard = self.inner.known.load();
+        if guard.contains(address) {
+            return true;
+        }
+        if address.is_ipv6() {
```
simonwicky: IMO we should store canonical addresses. I feel like something weird can happen here, just can't work exactly what. What do you think?

nym-network-monitor-v3/nym-network-monitor-agent/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# this will only work with VPN, otherwise remove the harbor part
+FROM harbor.nymte.ch/dockerhub/rust:latest AS builder
+
+RUN apt update && apt install -yy libdbus-1-dev pkg-config libclang-dev
+
+COPY ./ /usr/src/nym
+WORKDIR /usr/src/nym
+
+RUN cargo build --bin nym-node-network-monitor-agent --release
```
simonwicky: That Dockerfile won't build properly. The binary name is wrong

nym-network-monitor-v3/nym-network-monitor-agent/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+FROM harbor.nymte.ch/dockerhub/ubuntu:24.04
+
+RUN apt-get update && apt-get install -y ca-certificates
+
+WORKDIR /nym
+
+COPY --from=builder /usr/src/nym/target/release/nym-node-network-monitor-agent ./
+COPY --from=builder /usr/src/nym/nym-node-status-api/nym-node-status-agent/entrypoint.sh ./
```
simonwicky: Wrong directory (here and above)

nym-network-monitor-v3/nym-network-monitor-agent/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+nym-noise = { workspace = true }
+time = { workspace = true }
+tokio = { workspace = true, features = ["macros", "sync", "rt-multi-thread"] }
+tokio-util = { workspace = true }
+tracing = { workspace = true }
+url = { workspace = true }
+
+# methods to recreate lioness
+# we don't care about particular versions - just pull whatever is used by sphinx
+lioness = "*"
```
simonwicky: Can we consider putting it in the workspace? It's also in `Nym-node`

nym-network-monitor-v3/nym-network-monitor-agent/src/sphinx_helpers.rs
```diff
@@ -0,0 +1,264 @@
...
+    )
+}
+
+/// Builds a single test sphinx packet along `route` with the given per-hop `delay`.
+///
+/// The packet uses [`PacketSize::AckPacket`] to keep its size as small as possible. If `initial_secret`
+/// is provided it is used as the sender's ephemeral key, allowing the resulting header to
+/// be deterministically reproduced (needed for `create_test_sphinx_packet_header`).
+pub(crate) fn build_test_sphinx_packet(
+    route: &[Node],
```
simonwicky: Can we enforce the size of the route? Since we're generating two delays anyway

nym-network-monitor-v3/nym-network-monitor-agent/src/sphinx_helpers.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// Internally this builds one full sphinx packet to capture the header, then manually
+/// re-derives the per-hop payload keys by replaying the DH key-agreement steps along the
+/// route. This is necessary because the sphinx library does not expose the payload keys
+/// after packet construction.
+///
+/// The derived `payload_key` vec has one entry per hop; the last entry (index 1) is the
+/// key held by this agent as the final recipient and is used by [`TestPacketHeader::recover_payload`].
+pub(crate) fn create_test_sphinx_packet_header(
+    route: Vec<Node>,
```
simonwicky: Here as well? 
There is an assert later so anything else will fail anyway

nym-network-monitor-v3/nym-network-monitor-agent/src/egress_connection.rs
```diff
@@ -0,0 +1,123 @@
...
+    ) -> anyhow::Result<Self> {
+        info!("attempting to establish connection to {address}");
+        let stream = timeout(timeout_duration, TcpStream::connect(address)).await??;
+
+        info!("beginning the noise handshake (initiator)");
+
+        let noise_handshake_start = Instant::now();
+        let noise_stream = upgrade_noise_initiator(stream, noise_config).await?;
+
+        if !noise_stream.is_noise() {
```
simonwicky, line 59: Can't we just remove that TCP fallback altogether ?

nym-network-monitor-v3/nym-network-monitor-agent/src/listener/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+            return;
+        }
+        self.last_noise_handshake_duration = Some(noise_handshake_duration);
+
+        self.handle_stream(Framed::new(noise_stream, NymCodec))
+            .await
+    }
+
+    /// Processes one connection at a time until the shutdown token is cancelled.
+    /// Returns `self` so that the caller can inspect fields such as
```
simonwicky, line 142: We're keeping everything alive just for one field? Can't we return it/return a result struct instead?

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/config.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Config {
+    /// Total number of packets the agent intends to send: `floor(target_rate * sending_duration)`.
+    pub(crate) fn expected_packets(&self) -> usize {
+        (self.target_rate as f32 * self.sending_duration.as_secs_f32()).floor() as usize
+    }
+
+    /// Time between consecutive batch dispatches needed to sustain `target_rate`:
+    /// `sending_batch_size / target_rate` seconds.
+    pub(crate) fn batch_interval(&self) -> Duration {
+        Duration::from_secs_f64(self.sending_batch_size as f64 / self.target_rate as f64)
```
simonwicky, line 48: This will panic if one of them is 0
(If `sending_batch_size` is 0, it panics later when fed into an `interval`)

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/mod.rs
```diff
@@ -0,0 +1,495 @@
...
+            .build_mixnet_listener(processor.sender(), shutdown_token.clone())
+            .await?;
+        let listener_join = tokio::spawn(async move { listener.run().await });
+
+        // 3. probe: send a single packet to confirm the node responds
+        if !self
+            .send_connectivity_probe(&mut egress, &mut processor, &mut result)
+            .await?
+        {
+            return Ok(result);
```
simonwicky, line 458: We don't shut down the mixnet_listener if we hit that branch. 
The cleanup is a bit leaky

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/tested_node.rs
```diff
@@ -0,0 +1,37 @@
...
+impl TestedNodeDetails {
+    /// Returns a sphinx [`Node`](nym_sphinx_types::Node) representation of this node,
+    /// suitable for use as a hop in a sphinx route.
+    pub(crate) fn as_sphinx_node(&self) -> nym_sphinx_types::Node {
+        as_sphinx_node(self.address, self.sphinx_key)
+    }
+
+    /// Returns a [`NoiseNode`] representation of this node for use in the Noise network view.
+    pub(crate) fn as_noise_node(&self) -> NoiseNode {
+        NoiseNode::new_from_inner_key(self.noise_key, 1, true)
```
simonwicky: Do we really need that new constructor just for that? Isn't `new_nym_node` enough?

simonwicky (changes requested):

nym-node/src/config/mod.rs
jstuczyn: i'd say so. because it means the user  has set their environment in an invalid way and it's on them

nym-node/src/node/routing_filter/network_filter.rs
jstuczyn: I like it. ensure the value is always sanitised. and then we have one fewer look ups in a degenerative case.

nym-network-monitor-v3/nym-network-monitor-agent/src/egress_connection.rs
jstuczyn: what if node doesn't support noise? (or if we want to test it on a localnet )

simonwicky: That node will not support stress testing. (And it should be kicked because it's too old anyway)

nym-network-monitor-v3/nym-network-monitor-agent/src/listener/mod.rs
jstuczyn: because more fields might be needed in the future and I feel it's easier to just return self (and so that it could potentially be reused)

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/config.rs
jstuczyn: fixed

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/mod.rs
jstuczyn: fixed

simonwicky (approved):

PR #6597, opened 2026-03-23
Feat/nmv3 orchestrator queue

jstuczyn:
part of NYM-742 - initial database, config, CLI, and storage layer

This PR bootstraps the orchestrator side of the v3 network monitor. It introduces the persistent storage layer, configuration, CLI wiring, and the initial NetworkMonitorOrchestrator struct - everything needed before the actual scheduling and HTTP server logic lands in follow-up PRs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6597)


nym-network-monitor-v3/nym-network-monitor-orchestrator/src/storage/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Inserts a new node record. If a node with the same `node_id` already exists,
+    /// all fields except `identity_key` are updated — `identity_key` is intentionally left
+    /// unchanged because a given `node_id` always corresponds to exactly one identity key
+    /// and is never reassigned.
+    pub(crate) async fn insert_or_update_nym_node(&self, node: &NewNymNode) -> anyhow::Result<()> {
+        self.storage_manager.insert_or_update_nym_node(node).await
+    }
+
+    /// Inserts a completed test run.
+    pub(crate) async fn insert_test_run(
```
simonwicky, line 62: This is not atomic. We end up in a incoherent state if the second or third call fails
Maybe wrap everything in a transaction?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/storage/manager.rs
```diff
@@ -0,0 +1,771 @@
...
+    ///
+    /// Nodes with a row in `testrun_in_progress` are excluded entirely.
+    ///
+    /// Returns `None` if no eligible idle node exists.
+    pub(crate) async fn assign_next_testrun(
+        &self,
+        now: OffsetDateTime,
+        last_tested_before: OffsetDateTime,
+    ) -> anyhow::Result<Option<NymNode>> {
+        let mut tx = self.connection_pool.begin().await?;
```
simonwicky: nit : This does not start an "immediate" transaction, and the operations are read and then write. 
If two threads are calling this, they can both start a read transaction and fail to escalate it to a write transaction (because they both see an open read transaction). 

This is a no issue if there is only one thread managing the database, but a landmine if we were to add more. 
( We fixed a similar issue in #6299 )

nym-network-monitor-v3/nym-network-monitor-orchestrator/Cargo.toml
```diff
@@ -8,10 +8,40 @@ edition.workspace = true
...
 rust-version.workspace = true
 readme.workspace = true
-version.workspace = true
+version = "1.0.0"
 publish = false
 
 [dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true, features = ["cargo", "env"] }
+futures = { workspace = true }
```
simonwicky, line 17: This looks unused? (For now maybe)

nym-network-monitor-v3/nym-network-monitor-orchestrator/Cargo.toml
```diff
@@ -8,10 +8,40 @@ edition.workspace = true
...
+zeroize = { workspace = true }
+
+nym-validator-client = { workspace = true, features = ["http-client"] }
+nym-bin-common = { workspace = true, features = ["basic_tracing", "output_format"] }
+nym-crypto = { workspace = true, features = ["asymmetric"] }
+nym-network-defaults = { workspace = true }
+nym-task = { workspace = true }
+
+[dev-dependencies]
+nym-test-utils = { workspace = true }
```
simonwicky, line 33: This looks unused? (For now maybe)

nym-network-monitor-v3/nym-network-monitor-orchestrator/Cargo.toml
jstuczyn: yes. just for now - those will be needed in subsequent PRs as the orchestrator is expanded

nym-network-monitor-v3/nym-network-monitor-orchestrator/Cargo.toml
jstuczyn: yes. just for now - those will be needed in subsequent PRs as the orchestrator is expanded

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/storage/manager.rs
jstuczyn: very nice catch. fixed with the same approach

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/storage/mod.rs
jstuczyn: I think that's fine. the likelihood of failure is minimal and results are frequently persisted to nym-api. even if the entire database is purged on the orchestrator, it's not an issue. it's designed to be ephemeral-ish (k8s)

simonwicky (approved):

PR #6608, opened 2026-03-25
feat: multiple deposit prices

jstuczyn:
NYM-907

## Summary

Introduces tiered deposit pricing for the ecash contract, allowing whitelisted accounts to deposit ticket book funds at a reduced price instead of the default. Also simplifies the gateway redemption flow by removing the multisig proposal path since all of the tokens were going to the holding account anyway.

### Core mechanism

Selected addresses can be whitelisted by the contract admin to deposit at a custom (reduced) price. Non-whitelisted addresses continue paying the default deposit amount. The whitelisted price for each address is stored in a `Map<Addr, Coin>` and looked up at deposit time - if the sender is in the map, their custom price is used; otherwise the default applies.

Removing an address from the whitelist only affects future deposits. Historical statistics and past deposits remain unchanged, so it is safe to revoke reduced pricing mid-session.

### Validation rules

All deposit prices (both default and reduced) are subject to the following constraints:
- **Denomination**: must match the default deposit denom (e.g. `unym`)
- **Upper bound**: reduced price must be strictly less than the current default
- **Lower bound**: no price can be set below `ticket_book_size` (currently 50), which is the minimum viable amount to avoid zero-value ticket redemptions

These rules are enforced on both admin operations (`SetReducedDepositPrice`, `UpdateDefaultDepositValue`) and during migration whitelist seeding.

### Admin operations

Two new execute messages for managing the whitelist:
- **`SetReducedDepositPrice { address, deposit }`** — set or overwrite a reduced deposit price for a specific address
- **`RemoveReducedDepositPrice { address }`** — remove the custom price, reverting the address to the default. Returns an error if no custom price has been set.

### New queries

- **`GetDefaultDepositAmount`** — renamed from `GetRequiredDepositAmount` with serde aliases (`get_required_deposit_amount`, `GetRequiredDepositAmount`) for backwards compatibility
- **`GetReducedDepositAmount { address }`** — returns `Option<Coin>` for a specific address's custom price
- **`GetAllWhitelistedAccounts`** — returns all whitelisted addresses and their prices (no pagination — we literally just expect 2 entries here - the NS API and credential proxy)
- **`GetDepositsStatistics`** — returns a comprehensive breakdown of deposits by tier:
- Total deposits (count and value) across all tiers
- Default-tier deposits (count and value)
- Custom-tier deposits (count, value, and per-account breakdowns)

### Deposit statistics tracking

New `DepositStatsStorage` module tracks deposit activity separated by price tier:
- Default-price deposits: global counter and accumulated amount
- Reduced-price deposits: per-account counters and accumulated amounts (`Map<Addr, u32>` / `Map<Addr, Coin>`)

Each deposit call updates both the existing pool counters (total across all tiers) and the tier-specific stats. An invariant - default-tier + custom-tier counts/amounts must equal the totals — is enforced in tests.

### Migration

The `migrate` entrypoint takes an `initial_whitelist: Vec<WhitelistedDeposit>` via `MigrateMsg`. The `add_tiered_pricing` migration function:
1. **Backfills statistics**: assigns all pre-migration deposits to the default price tier using the existing deposit ID counter and pool counters
2. **Seeds the whitelist**: validates and stores each entry through the same validation path as the admin endpoint

### Removed multisig proposal flow

Gateways no longer create multisig proposals for ticket redemption. The `RedeemTickets` execute message is now called directly.

### Removed holding account transfer

Tokens are no longer transferred to the holding account after redemption.

## Credential proxy (and NS API)

The `required_deposit_cache` in `credential-proxy` now queries both the reduced and default deposit amounts:
1. First tries `get_reduced_deposit_amount(address)` for the depositing address
2. Falls back to `get_default_deposit_amount()` if the address is not whitelisted

The result is cached with a 1-minute validity window to avoid excessive chain queries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6608)


common/cosmwasm-smart-contracts/ecash-contract/src/msg.rs
```diff
@@ -81,7 +108,22 @@ pub enum QueryMsg {
...
+    /// Initial set of whitelisted accounts with their reduced deposit prices.
+    /// Each entry is validated and stored during migration.
+    pub initial_whitelist: Vec<WhitelistedDeposit>,
 }
 
+/// An address and its reduced deposit price, used when seeding the whitelist
+/// via migration.
 #[cw_serde]
-pub struct MigrateMsg {}
+pub struct WhitelistedDeposit {
```
simonwicky, line 126: That feels out of place no?

neacsu (approved):
Really great description of the work done in the PR. It makes following the code much easier, thank you!

> @neacsu reviewed 38 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6608#-:-OodmFFVDYj5A7BAAN0k:b4matdn)* status: all files reviewed, 1 unresolved discussion (waiting on jstuczyn and simonwicky).

___
*[`common/cosmwasm-smart-contracts/ecash-contract/src/reduced_deposit.rs` line 1 at r1](https://reviewable.io/reviews/nymtech/nym/6608#-OodnWMxE1S6hKD0GgO7:-OodnWMxE1S6hKD0GgO8:bxqvv30) ([raw file](https://github.com/nymtech/nym/blob/378b3e9154c07a75c7665032b15c1ad7accf9b3d/common/cosmwasm-smart-contracts/ecash-contract/src/reduced_deposit.rs#L1)):*
> ```XML
> // Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
> ```

nit: 2026

simonwicky (approved):

common/cosmwasm-smart-contracts/ecash-contract/src/msg.rs
jstuczyn: I wouldn't say so, that struct is only defined to be used within the `MigrateMsg` created just above

PR #6626, opened 2026-03-27
add node refresher for periodic scraping of bonded nym-node details

jstuczyn:
NYM-914

- NodeRefresher periodically queries the mixnet contract for all bonded nodes and probes each node's HTTP API for host information, sphinx keys, noise keys, and key rotation IDs
- Extract NymNodeApiClientRetriever into nym-node-requests with port probing, identity verification, and host information signature checking
- Add clone_query_client on NyxdClient so the refresher can hold its own query client without locking the signing client
- Batch upsert for nym_node rows (single transaction instead of per-row)
- Reuse the new helpers in nym-api's node_describe_cache

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6626)


nym-node/nym-node-requests/src/api/helpers.rs
```diff
@@ -0,0 +1,259 @@
+// Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+/// The default HTTP API port exposed by nym-nodes that are not behind a reverse proxy.
+pub const STANDARD_NYM_NODE_HTTP_PORT: u16 = 8080;
```
simonwicky: That constant exists already as `DEFAULT_NYM_NODE_HTTP_PORT` in the network-defaults

nym-node/nym-node-requests/src/error.rs
```diff
@@ -16,4 +16,45 @@ pub enum Error {
         #[from]
         source: nym_wireguard_types::Error,
     },
+
+    #[cfg(feature = "client")]
+    #[error("node {node_id} has provided malformed host information ({host}: {source}")]
```
simonwicky:
```suggestion
    #[error("node {node_id} has provided malformed host information ({host}: {source})")]
```

nym-node/nym-node-requests/src/api/helpers.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// `/health` endpoint reports an "up" status:
+    ///
+    /// 1. `http://<host>:<custom_port>` (only when `custom_port` is `Some`)
+    /// 2. `http://<host>:8080` — the standard nym-node API port
+    /// 3. `https://<host>` — node behind an HTTPS reverse proxy (port 443)
+    /// 4. `http://<host>` — node behind an HTTP reverse proxy (port 80)
+    ///
+    /// This function is intended for infrastructure binaries (nym-api, network monitor, etc.),
+    /// not regular clients, which is why hickory DNS is explicitly disabled.
+    pub async fn try_get_valid_nym_node_api_client(
```
simonwicky, line 204: Nym node status API still has its own copy of that (look for `try_get_client`).
Can we make it use this one instead maybe?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/node_refresher.rs
```diff
@@ -0,0 +1,190 @@
...
+        node_update
+    }
+
+    async fn refresh_bonded_nodes(&self) -> anyhow::Result<()> {
+        // 1. retrieve all nodes from the contract
+        let nodes = self.client.get_all_nymnode_bonds().await?;
+        info!("retrieved {} bonded nodes from the contract", nodes.len());
+
+        // 2. retrieve detailed information from the self-described endpoints
+        let refreshed_nodes = Mutex::new(Vec::new());
```
simonwicky: LLM suggestion ( that looks correct to me) : 
```
let refreshed_nodes: Vec<_> = stream::iter(nodes)
    .map(|b| self.get_node_details(b))
    .buffer_unordered(self.number_of_concurrent_node_queries)
    .collect()
    .await;

``` 
Removes the need for the mutex in the first place

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/node_refresher.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    async fn get_node_details(&self, bond: NymNodeBond) -> NewNymNode {
+        let mut node_update = NewNymNode::from_bond(&bond);
+
+        let node_id = bond.node_id;
+        let self_described = match self.get_node_details_inner(bond).await {
+            Err(err) => {
+                error!("failed to retrieve self-described node details for node {node_id}: {err}");
+                return node_update;
```
simonwicky, line 130: If the `get_node_details_inner` query fails, you still return a "partial" node. 
That node will get updated in the storage, overwriting the existing information like `noise_key`, `sphinx_key` etc. 
I'm not sure this is the behavior we want?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/node_refresher.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        // 3. update the storage
+        self.storage
+            .batch_insert_or_update_nym_nodes(&refreshed_nodes)
+            .await?;
+        Ok(())
+    }
+
+    pub(crate) async fn run(&self) {
+        let mut interval = interval(self.node_refresh_rate);
```
simonwicky, line 168: How about setting the missed tick behavior to Delay?
Just te be safe if by any chance an update is long or the delay too short

simonwicky (approved):

nym-node/nym-node-requests/src/api/helpers.rs
jstuczyn: absolutely - good catch

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/node_refresher.rs
jstuczyn: nice one

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/node_refresher.rs
jstuczyn: yes, that's the behaviour we want. we want to overwrite node information in the storage with NULLs so we'd known the node is not reachable and thus not eligible for testing

PR #6633, opened 2026-03-31
nym-node-setup: plumb HOST_SSH_PORT through tunnel manager, CLI, and env setup

p17o:
## Summary

This PR makes the host management SSH port configurable across the `nym-node-setup` flow instead of hardcoding port `22` in the tunnel manager.

It also removes host firewall drift in `network-tunnel-manager.sh` by using shared host firewall port definitions for configuration, status, and test output.

### `scripts/nym-node-setup/network-tunnel-manager.sh`

- added support for `HOST_SSH_PORT` with a default of `22`
- validate `HOST_SSH_PORT` as an integer in the range `1..65535`
- load values from `ENV_FILE` (or local `env.sh`) so the script can consume operator-provided configuration
- replaced hardcoded host SSH port `22` in host firewall handling
- centralized managed host firewall TCP/UDP port definitions
- fixed the existing drift where UDP `4443` was opened during configuration but not shown in status output or checked in test output

### `scripts/nym-node-setup/nym-node-cli.py`

- added CLI support for `--host-ssh-port`
- plumbed `HOST_SSH_PORT` into the generated `env.sh`
- ensured the CLI validates and persists the configured host SSH port alongside the other setup values

### `scripts/nym-node-setup/setup-env-vars.sh`

- added interactive prompting for `HOST_SSH_PORT`
- validate the entered port before writing `env.sh`
- include `HOST_SSH_PORT` in the confirmation summary
- persist `HOST_SSH_PORT` in the generated env file

## Why

Before this change:

- the host management SSH port was hardcoded to `22`
- the tunnel manager did not cleanly consume a configured host SSH port from the broader setup flow
- the managed host firewall ports were defined separately in multiple places
- UDP `4443` was configured but omitted from status and test output
- the SSH-related naming was ambiguous

Using `HOST_SSH_PORT` makes it explicit that this setting controls the host's management SSH port, not any SSH-related traffic that might be relevant to exit-policy handling.

## Backwards compatibility

- default behavior remains unchanged when `HOST_SSH_PORT` is not set
- the host management SSH port still defaults to `22`
- existing setup flows continue to work without additional input
- operators can now override the host SSH port either interactively, via env, or via CLI

## Examples

Interactive setup:
```bash
./setup-env-vars.sh

<!-- Reviewable:start -->
- - -
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/nymtech/nym/6633)
<!-- Reviewable:end -->

### DIFF
--- a/scripts/nym-node-setup/network-tunnel-manager.sh
+++ b/scripts/nym-node-setup/network-tunnel-manager.sh
@@ -556,6 +556,34 @@ apply_smtps_465_rate_limit() {
 ###############################################################################
 
 NETWORK_FIREWALL_COMMENT="NYM-NETWORK-FW"
+SSH_PORT="${SSH_PORT:-22}"
+NETWORK_FIREWALL_TCP_PORTS=("$SSH_PORT" 80 443 1789 1790 8080 9000 9001 41264)
+NETWORK_FIREWALL_UDP_PORTS=(4443 51822 51264)
+NETWORK_FIREWALL_WG_TCP_PORT=51830
+
+validate_port_value() {
+  local name="$1"
+  local value="$2"
+
+  if ! [[ "$value" =~ ^[0-9]+$ ]] || (( 10#$value < 1 || 10#$value > 65535 )); then
+    error "invalid ${name}='${value}'. expected integer 1-65535"
+    exit 1
+  fi
+}
+
+validate_port_value "SSH_PORT" "$SSH_PORT"
+
+build_network_firewall_status_pattern() {
+  local patterns=("$NETWORK_FIREWALL_COMMENT")
+  local port
+
+  for port in "${NETWORK_FIREWALL_TCP_PORTS[@]}" "${NETWORK_FIREWALL_UDP_PORTS[@]}" "$NETWORK_FIREWALL_WG_TCP_PORT"; do
+    patterns+=("dpt:${port}")
+  done
+
+  local IFS='|'
+  printf '%s' "${patterns[*]}"
+}
 
 delete_managed_input_rules() {
   local cmd="$1"
@@ -593,22 +621,19 @@ configure_network_firewall() {
   delete_managed_input_rules iptables
   delete_managed_input_rules ip6tables
 
-  local tcp_ports=(22 80 443 1789 1790 8080 9000 9001 41264)
-  local udp_ports=(4443 51822 51264)
-
   local port
-  for port in "${tcp_ports[@]}"; do
+  for port in "${NETWORK_FIREWALL_TCP_PORTS[@]}"; do
     add_input_port_rule iptables "$port" tcp
     add_input_port_rule ip6tables "$port" tcp
   done
 
-  for port in "${udp_ports[@]}"; do
+  for port in "${NETWORK_FIREWALL_UDP_PORTS[@]}"; do
     add_input_port_rule iptables "$port" udp
     add_input_port_rule ip6tables "$port" udp
   done
 
-  add_input_port_rule iptables 51830 tcp "$WG_INTERFACE"
-  add_input_port_rule ip6tables 51830 tcp "$WG_INTERFACE"
+  add_input_port_rule iptables "$NETWORK_FIREWALL_WG_TCP_PORT" tcp "$WG_INTERFACE"
+  add_input_port_rule ip6tables "$NETWORK_FIREWALL_WG_TCP_PORT" tcp "$WG_INTERFACE"
 
   save_iptables_rules
   ok "host network firewall configuration completed"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by p17o** [comment]
@serinko next improvement will be to allow specifying the host ssh interface (e.g. for folks using tailscale). Happy to add to this one, let me know once you're done reviewing

**Comment by serinko** [comment]
Hey, I went through it quickly and it looks very good. Definitely want to do a proper review, testing and merge it before EOW.

Really appreciate your contribution.

File: scripts/nym-node-setup/network-tunnel-manager.sh [hunk 1]
```diff
@@ -517,6 +517,34 @@ apply_smtps_465_rate_limit() {
...
+  local name="$1"
+  local value="$2"
+
+  if ! [[ "$value" =~ ^[0-9]+$ ]] || (( 10#$value < 1 || 10#$value > 65535 )); then
+    error "invalid ${name}='${value}'. expected integer 1-65535"
+    exit 1
+  fi
+}
+
+validate_port_value "HOST_SSH_PORT" "$HOST_SSH_PORT"
```
serinko, line 535: I would consider to move the validation to the global config part simply for future maintenance readability, but this works perfectly fine as well, not a merge blocker.

scripts/nym-node-setup/network-tunnel-manager.sh
```diff
@@ -1575,6 +1602,7 @@ environment overrides:
   NETWORK_DEVICE                    Auto-detected uplink (e.g., eth0). Set manually if detection fails.
   TUNNEL_INTERFACE                  Default: nymtun0. Requires root privileges (sudo) to manage.
   WG_INTERFACE                      Default: nymwg - Must match your WireGuard interface name.
+  HOST_SSH_PORT                     Default: 22. Set manually if you connect to your host's SSH daemon through another port.
 
 EOF
     status=0
```
serinko, line 1608: Looks great.

This improves the host firewall section in the right way: `HOST_SSH_PORT` is now configurable with a sane default, validated early, and the managed `TCP/UDP/WG` port definitions are in sync w configuration, status, and test outputs.

Nice fix for the existing drift where `UDP 4443` was configured but omitted from the status and test flows.

No blockers fr the NTM, going for the other pars now.

scripts/nym-node-setup/nym-node-cli.py
```diff
@@ -68,23 +68,60 @@ def print_welcome_message(self):
...
+        if not sval.isdigit():
+            print(f"Invalid SSH port: {sval!r}. Expected integer 1..65535.")
+            raise SystemExit(1)
+        port = int(sval, 10)
+        if not 1 <= port <= 65535:
+            print(f"Invalid SSH port: {port}. Expected integer 1..65535.")
+            raise SystemExit(1)
+        return str(port)
+
+    def _resolve_field(args, existing, arg_name, env_key, prompt, *, default=None, validator=None):
```
serinko, line 84: this fn is defined inside the class without a self parameter and without being marked `@staticmethod`. It still works when called as `self._resolve_field(...)` thanks to the way Py binds `self` into the first position and your first parameter is `args`, but it can be a a bit misleading . I would prefer t have explicit naming here, if you don't mind to add `self`.

Though I can see that I will need to do th same homework myslef on the rest of the program, so I would say we merge it as it is now and do a styling work later on.

scripts/nym-node-setup/setup-env-vars.sh
```diff
@@ -28,6 +30,7 @@ while true; do
   printf "%-20s %s\n" "EMAIL:"       "$EMAIL"
   printf "%-20s %s\n" "MONIKER:"     "$MONIKER"
   printf "%-20s %s\n" "DESCRIPTION:" "$DESCRIPTION"
+  printf "%-20s %s\n" "HOST_SSH_PORT:" "$HOST_SSH_PORT"
   echo "---------------------------------------"
 
   read -rp "Are these correct? (y/n): " CONFIRM
```
serinko, line 36: Is there a reason to not keep the persistent nature of the env.sh and include this inthe writing block:
```sh
echo "export HOST_SSH_PORT=\"${HOST_SSH_PORT}\""
```

This is the only comment which needs answer - resoution and then the PR is ready to merge

serinko (changes requested):
Thank you again for this work. this PR is a great improvement making the flow much more flexible. 

Please check the latst comment about setup-env-vars.sh and then we merge it to production.

scripts/nym-node-setup/setup-env-vars.sh
p17o: Good catch — there is no reason not to persist it there as well.

The intent is for `HOST_SSH_PORT` to behave like the other operator-provided values in `env.sh`, so omitting it from the write block was an oversight on my side. I’ll add:

`echo "export HOST_SSH_PORT=\"${HOST_SSH_PORT}\""`

to the generated `env.sh` block and push an update to this PR.

mfahampshire (approved):

PR #6636, opened 2026-04-01
Optimize GW probe in NS agent

dynco-nym:
Changes are mostly in the node status agent and probe. Agent was previously invoking already compiled binary of the gateway probe with a `Command`, as if running it from a command line. This required
- compiling probe beforehand
- passing any additional gateway probe CLI parameters non-transparently to the agent via stringified interface
- yanky log capture on the ns agent side of the probe subprocess and parsing its output
- separate versioning of the probe and agent

I have refactored that to agent invoking the probe as a library. This means
- any code changes in the probe are instantly propagated to the agent
- backward-incompatible changes in the probe are immediately visible in the agent, thus forcing to reconcile them at compile time
- agent-only versioning: as long as you run the latest agent, you're guaranteed to run the latest probe
- much cleaner interface between probe and agent for data transfer: instead of passing string and using grep to parse it, we have compile-time guarantees from Rust's type system

Changes to gateway probe are minimal and mostly amount to
- moving code around to make it possible to use probe as a library, as opposed to only a CLI tool
- ensure "agent" flow is neatly encapsulated in the probe, so the agent only has a single entry point
- this is part of an effort to keep the probe logic inside the probe, and agent being just an orchestrator with configuration ability
  - configuration of the probe can now be transparently done (via strong typing) through agent's CLI interface
  - previously this was done by passing a huge env variable with all CLI flags concatenated 
- as part of this, refactored a much safer "default" system
  - moved it from clap (dependency) to Rust's Default
  - reusable outside of clap interface

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6636)


nym-gateway-probe/src/config/mod.rs
```diff
@@ -29,7 +29,7 @@ pub struct ProbeConfig {
     ///   socks5-only - Socks5 network requester test
     ///   all         - Mixnet, wireguard over authenticator and LP registration
     ///
-    #[arg(long, default_value_t = TestMode::default(), verbatim_doc_comment)]
+    #[arg(long, default_value_t = ProbeConfig::default().test_mode, verbatim_doc_comment)]
```
simonwicky: nit : that looks a bit convoluted to me to ultimately get `TestMode::default`

nym-gateway-probe/src/config/netstack.rs
```diff
@@ -5,40 +5,62 @@ use clap::Args;
 
 #[derive(Args, Clone, Debug)]
 pub struct NetstackArgs {
-    #[arg(long, hide = true, default_value_t = 180)]
+    #[arg(long, env = "PROBE_NETSTACK_DOWNLOAD_TIMEOUT_SEC", default_value_t = NetstackArgs::default().netstack_download_timeout_sec)]
```
simonwicky: nit: IIRC I added `hide= true` here because it was making the CLI a bit messy, and it was realistically not used on standalone runs. Any reason to remove it?

nym-gateway-probe/src/config/socks5.rs
```diff
@@ -2,21 +2,19 @@ use clap::Args;
 
 use crate::common::socks5_test::JsonRpcClient;
 
-const DEFAULT_RPC_ENDPOINT: &str = "https://cloudflare-eth.com";
-
-#[derive(Args, Debug)]
+#[derive(Args, Clone, Debug)]
 pub struct Socks5Args {
-    #[arg(long, hide = true, value_delimiter = ';', default_value = DEFAULT_RPC_ENDPOINT)]
+    #[arg(long, env = "PROBE_SOCKS5_JSON_RPC_URL_LIST", default_values_t = Socks5Args::default().socks5_json_rpc_url_list, value_delimiter = ';')]
```
simonwicky: nit: same comment about `hide = true`

simonwicky (approved):
Way cleaner to run it like that, nice one 🚀

nym-gateway-probe/src/config/mod.rs
simonwicky: edit : I understand the similarity to other structure like the NetstackConfig, but the fact that here the default is derived makes it kinda clunky imo

nym-gateway-probe/src/config/netstack.rs
dynco-nym: didn't know if it was there intentionally or not, restored for your convenience :)

nym-gateway-probe/src/config/mod.rs
dynco-nym: set to `TestMode::default()`

PR #6255, opened 2025-11-29
Support different ipv6 network device in network-tunnel-manager.sh

heri16:
On many VPS providers, the IPv4 uplink network interface is eth0, while the IPv6 uplink may be on a different interface (e.g., eth1 or another device).

This PR:
- [x] Improves automatic detection of the uplink network interface for both IPv4 and IPv6.
- [x] Adds support for configurations where the IPv4 and IPv6 uplinks are on different interfaces, while remaining fully compatible with setups where they share the same interface.
- [x] Change the expected environment variable name from `NETWORK_DEVICE` to `IPV4_UPLINK_DEV` and `IPV6_UPLINK_DEV`
- [x] Add CLI flag for `--uplink-dev-v6`

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6255)

PR #6640, opened 2026-04-03
Handle split IPv4/IPv6 uplinks in network-tunnel-manager

p17o:
## Summary

This PR fixes `network-tunnel-manager.sh` for hosts where IPv4 and IPv6 use different uplink interfaces.

Before this change, the script auto-detected a single `NETWORK_DEVICE` from the IPv4 default route and reused it for both IPv4 and IPv6 iptables/ip6tables setup. On hosts where IPv4 egress is on one interface and IPv6 egress is on another, this caused IPv6 tunnel connectivity checks and forwarding/NAT setup to fail.

## What changed

- added separate uplink detection for:
  - `NETWORK_DEVICE_V4`
  - `NETWORK_DEVICE_V6`
- kept `NETWORK_DEVICE` as a backward-compatible override for both families
- updated IPv4 NAT/FORWARD rules to use the IPv4 uplink
- updated IPv6 NAT/FORWARD rules to use the IPv6 uplink
- updated IPv6 display/status paths to use the IPv6 uplink
- updated exit-policy setup and checks so IPv4 and IPv6 hooks use the correct uplink

## Why this is needed

Some deployments use:
- one interface for public IPv4 connectivity
- another interface for global IPv6 connectivity

In that setup, reusing the IPv4 uplink for IPv6 breaks the IPv6 path even when the host itself has working IPv6 on a different interface.

## Backward compatibility

Existing setups that use a single uplink remain unchanged:
- `NETWORK_DEVICE` still works as before
- the new variables are optional
- auto-detection now prefers per-family defaults when not explicitly set

## Validation

Tested on a host with:
- IPv4 uplink on `eth0`
- IPv6 uplink on `eth2`

Results:
- IPv6 connectivity via `nymwg`  and `nymtun0` succeeded

## Notes

This change is intentionally scoped to uplink selection and rule application only.
It does not change the existing behavior for single-uplink hosts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6640)


serinko (changes requested):
Amazing work @p17o !!!!

I have two points where I would like to ask for attention:

1. `check_forward_chain()` still validates the `IPv4 FORWARD` table only and only checks that rule `1` is `NYM-EXIT`  (plus the presence of a RELATED,ESTABLISHED accept somewhere). I will admit that it had been already quite a bit loose before, and with split uplinks the ordering check is questionable.  It is not wrong, just not as specific as the rest of the script .

2. Do you reckon that we should we should make an exception for the NTM to not hard fail if it doesn't determine IPv6 uplink? Not as that we want nodes to be just IPv4, but in case there is some temp issues and the operators then can still setup everything else and rather see the outcome in their tests (jokes etc).

Last but not least, if we first merge the previous one, I assume some merge conflict solving will nned to happen on this PR prior to merge it to production.

serinko:
Additionally would you say we should add the flag to node cli as well as to QUIC setup?

This patch was proposing that: https://github.com/nymtech/nym/pull/6255/changes, but after merging your PR's there is too much overlap and the #6255 is reduntant to merge.

p17o:
Thanks @serinko! My take and proposed path forward are:

1. **`check_forward_chain()`**
   Agreed. It is still too tied to the old single-uplink / positional assumption. I plan to update it so it validates the expected IPv4 and IPv6 hooks semantically rather than relying on “rule 1 = NYM-EXIT”.

2. **Hard fail on missing IPv6 uplink**
   I agree this should be softened. My plan is to make NTM warn loudly, continue with the IPv4 parts, skip IPv6-specific setup when the IPv6 uplink cannot be determined, and let the outcome show up in the connectivity tests instead of failing early.

3. **Merge conflict after the previous PR**
   Agreed. Once the earlier PR lands, I will rebase this branch on top of the updated target branch, resolve conflicts, and rerun validation before merge.

4. **`nym-node-cli` and QUIC setup**
   After reviewing `nym-node-cli.py` and `quic_bridge_deployment.sh`, my view is:

   * `nym-node-cli.py` should expose the split-uplink overrides, since it is the operator-facing entry point and currently only models a single uplink.
   * For `quic_bridge_deployment.sh`, I think the cleaner approach is *not* to add parallel split-uplink logic there. Since it no longer manages iptables and already defers routing/NAT to `network-tunnel-manager.sh`, I would prefer to avoid duplicating uplink/NAT validation in that script. I would either remove the uplink-dependent NAT/firewall inspection there or make it best-effort / informational only.

**Proposed path forward for this PR**

* update `check_forward_chain()` to validate the actual expected IPv4/IPv6 hooks;
* change missing IPv6 uplink from hard-fail to loud warning + partial setup;
* rebase after the earlier PR merges and resolve the overlap.

**Follow-up scope**

* add split-uplink overrides to `nym-node-cli.py`;
* trim or soften the uplink/NAT checks in `quic_bridge_deployment.sh` rather than duplicating tunnel-manager logic.

The only thing I’d need from you before I update the PR is confirmation on scope:

* whether you want the `nym-node-cli.py` and QUIC-script changes included in this PR as well,
* or whether you’d prefer those as a small follow-up PR after this one.

serinko:
Thank you for an in depth answer. 

The way how you propose it is sound and I agree that the QUIC will need more love after the drop of ufw for good etc. 

I would ask you then to do the final touches here as you propose them, and we look into the cli and quic configurator in a separate PR.

mfahampshire (approved):

PR #6670, opened 2026-04-14
Block non-public IPR/NR checks

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6670)


jstuczyn (commented):
nice set of changes, but I wonder if there's a usecase, where you'd want to allow routing to local addresses. I can imagine it could be useful for some local testnets maybe? perhaps this check, if it's not implicitly done already, could be ignored if the node is running in `open-proxy` mode? or maybe there could be some specific `unsafe-route-local` flag

dynco-nym:
> nice set of changes, but I wonder if there's a usecase, where you'd want to allow routing to local addresses. I can imagine it could be useful for some local testnets maybe? perhaps this check, if it's not implicitly done already, could be ignored if the node is running in `open-proxy` mode? or maybe there could be some specific `unsafe-route-local` flag

added a CLI flag, separate for NR and IPR, in case it's ever needed

jstuczyn (approved):

PR #6683, opened 2026-04-17
Fix invalid ticket spend

dynco-nym:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6683)


nym-node-status-api/nym-node-status-api/src/ticketbook_manager/storage/mod.rs
```diff
@@ -89,7 +91,7 @@ impl TicketbookManagerStorage {
         &self,
         ticket_type: TicketType,
         testrun_id: i32,
-    ) -> anyhow::Result<Option<RetrievedTicketbook>> {
+    ) -> anyhow::Result<Option<(AttachedTicket, EpochId, Date)>> {
```
jstuczyn: I'm not a huge fan of returning very complex types, especially if it involves tuples with multiple elements. couldn't we either retrieve the epochid/date from the tickets or alternatively bundle the data in some wrapper?

nym-node-status-api/nym-node-status-api/src/ticketbook_manager/storage/mod.rs
```diff
@@ -113,16 +115,45 @@ impl TicketbookManagerStorage {
...
             ticketbook: deserialised,
-        }))
+        };
+        let epoch_id = retrieved_ticketbook.ticketbook.epoch_id();
+        let expiration_date = retrieved_ticketbook.ticketbook.expiration_date();
+
+        let attached_ticket = retrieved_ticketbook
+            .try_into()
+            .map(|t| (t, epoch_id, expiration_date))
+            .inspect_err(|e| warn!("Failed to convert retrieved ticketbook: {e}"))
```
jstuczyn, line 140: this should probably be an error rather than a warning, because assuming we have fixed the bug, it should NEVER happen. similarly, maybe, in that case we shouldn't allow the function to continue?

jstuczyn (approved):

nym-node-status-api/nym-node-status-api/src/ticketbook_manager/storage/mod.rs
dynco-nym: returned an error instead

nym-node-status-api/nym-node-status-api/src/ticketbook_manager/storage/mod.rs
dynco-nym: fixed

PR #6685, opened 2026-04-17
feat: NMv3: Eviction of stale testrun data

jstuczyn:
NYM-756

## Summary
- Adds a background task that periodically evicts stale data: test runs stuck "in progress" past `test_timeout` (freed for reassignment) and finalised test runs older than `testrun_eviction_age` (dropped to bound the results table).
- Runs one synchronous sweep at orchestrator start-up so crash-left state is cleared before agents poll for work, then loops on a cadence derived from the smaller of the two timeouts (floored at 60s, `MissedTickBehavior::Delay`).
- Cooperative shutdown: cancellation is observed between sweeps, so an in-flight sweep is allowed to finish.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6685)

### DIFF
--- a/nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
+++ b/nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
@@ -1,0 +1,50 @@
+// Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: GPL-3.0-only
+
+use crate::storage::NetworkMonitorStorage;
+use std::time::Duration;
+use tokio::time::{Instant, interval_at};
+use tracing::error;
+
+pub(crate) struct StaleResultsEviction {
+    storage: NetworkMonitorStorage,
+    max_result_age: Duration,
+    max_testrun_timeout: Duration,
+    check_interval: Duration,
+}
+
+impl StaleResultsEviction {
+    pub(crate) fn new(
+        storage: NetworkMonitorStorage,
+        max_result_age: Duration,
+        max_testrun_timeout: Duration,
+    ) -> Self {
+        // let check interval be half of the minimum of the two timeouts
+        let check_interval = Duration::min(max_result_age, max_testrun_timeout) / 2;
+
+        Self {
+            storage,
+            max_result_age,
+            max_testrun_timeout,
+            check_interval,
+        }
+    }
+
+    pub(crate) async fn evict_stale_results(&self) -> anyhow::Result<()> {
+        self.storage
+            .clear_timed_out_testruns_in_progress(self.max_testrun_timeout)
+            .await?;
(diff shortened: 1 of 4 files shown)


nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// orchestrator invokes [`Self::evict_stale_results`] once during start-up
+    /// (to reap anything left behind by a prior crash or restart), so an
+    /// immediate tick here would redo that work.
+    ///
+    /// `MissedTickBehavior::Delay` prevents burst catch-up ticks when a sweep
+    /// runs long under DB load — otherwise a slow sweep would queue multiple
+    /// back-to-back ticks and amplify the pressure that made it slow in the
+    /// first place.
+    pub(crate) async fn run(&self) {
+        let mut interval = interval_at(Instant::now() + self.check_interval, self.check_interval);
```
simonwicky, line 111: nit: How about just creating a normal interval and tick it once on the spot? IIRC this is what we do in other places where it's needed

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
```diff
@@ -0,0 +1,130 @@
...
+        interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+        loop {
+            tokio::select! {
+                biased;
+                _ = self.shutdown_token.cancelled() => break,
+                _ = interval.tick() => {}
+            }
+
+            if let Err(err) = self.evict_stale_results().await {
```
simonwicky: nit: Why is that not in the `interval.tick()` arm? The behavior is identical but it better match the code style of every other places

simonwicky (approved):

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
jstuczyn: does it really matter? we use both patterns in the code base : )

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
jstuczyn: because we get the benefit of rust format, but that's fair. let me put it into the tick branch to make it more obvious and explicit

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/stale_results_eviction.rs
simonwicky: Keyword is "nit" ;)

PR #6688, opened 2026-04-19
Title: fix(dns): shuffle resolved IPs

rachyandco:
## Problem

  Resolved IPs were always returned in the same fixed order,
  so connection retries would repeatedly target the same dead address even
  when multiple IPs were available.

## Changes

  - Shuffle resolved IPs on each lookup so successive connection attempts
    cycle through all available edge nodes

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6688)

### DIFF
--- a/common/http-api-client/src/dns.rs
+++ b/common/http-api-client/src/dns.rs
@@ -376,24 +376,6 @@ impl HickoryDnsResolver {
         }
     }
 
-    /// Remove the preresolve cache entry for a single host, forcing the next
-    /// lookup for that host to go through the network resolver again.
-    ///
-    /// Call this after a hard connection failure (RST, ECONNREFUSED, timeout)
-    /// to ensure the next attempt gets a fresh DNS answer rather than a cached
-    /// IP that may no longer be reachable.
-    pub fn invalidate_preresolve_for(&self, name: &str) {
-        debug!("invalidating pre-resolve for {name}");
-        if let Some(cell) = &self.static_base
-            && let Some(static_base) = cell.get()
-        {
-            static_base.invalidate_preresolve_entry(name)
-        }
-        if self.use_shared {
-            SHARED_RESOLVER.invalidate_preresolve_for(name);
-        }
-    }
-
     /// Get the current map of hostnames to addresses used in the fallback static lookup stage if one
     /// exists.
     pub fn get_static_fallbacks(&self) -> Option<HashMap<String, Vec<IpAddr>>> {
--- a/common/http-api-client/src/dns.rs
+++ b/common/http-api-client/src/dns.rs
@@ -56,7 +56,6 @@ use std::{
 use hickory_resolver::{
     TokioResolver,
     config::{NameServerConfig, NameServerConfigGroup, ResolverConfig, ResolverOpts},
-    lookup_ip::LookupIpIntoIter,
     name_server::TokioConnectionProvider,
 };
 use once_cell::sync::OnceCell;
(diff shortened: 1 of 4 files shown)


common/http-api-client/src/dns.rs
```diff
@@ -227,9 +231,13 @@ async fn resolve(
     let primary_err = match resolve_fut.await {
         Err(_) => ResolveError::Timeout,
         Ok(Ok(lookup)) => {
-            let addrs: Addrs = Box::new(SocketAddrs {
-                iter: lookup.into_iter(),
-            });
+            // Shuffle so that successive connection attempts cycle through all
+            // returned IPs rather than always hitting the same first address.
+            // This distributes retries across the full set of CDN edge nodes
+            // even within the cache TTL window.
```
jmwample:
```suggestion
```

I think the extra specifics could be misleading. Not all of the endpoints that are looked up involve CDNs. Otherwise this change seems fine to me -- though it is possible that `reqwest` does this internally with resolved addrs already.

common/http-api-client/src/dns.rs
```diff
@@ -380,6 +376,24 @@ impl HickoryDnsResolver {
...
+        if let Some(cell) = &self.static_base
+            && let Some(static_base) = cell.get()
+        {
+            static_base.invalidate_preresolve_entry(name)
+        }
+        if self.use_shared {
+            SHARED_RESOLVER.invalidate_preresolve_for(name);
+        }
+    }
+
```
jmwample:
```suggestion
```

I don't think this should be included.

This is unused and I don't think this is some thing that we want to use lightly. The whole model of DNS right now is that we do DNS lookups UP FRONT and then fix those addresses so that they are static for firewall exceptions during connection establishment, and random resolutions cannot pop up at the wrong time causing a timeout. If we allow evictions from DNS pre-resolve for the shared resolver then random subsystems could clear entries that then get re-resolved to a different address (breaking firewall) or cause timeouts during tunnel establishment which would be very hard to debug.  Once the tunnel is connected it is swapped over to using the system resolver (over the tunnel).

In my mind the presumed state of things is that the fallback IPs for critical endpoints will be used. If DNS resolution succeeds and we are able to use more specific addresses than we have from the configuration then that is extra performance.

common/http-api-client/src/dns/static_resolver.rs
```diff
@@ -107,6 +107,12 @@ impl StaticResolver {
         self.fallback_addr_map.lock().unwrap().extend(addrs);
     }
 
+    /// Remove a single host from the preresolve table, forcing the next lookup
+    /// for that host to go through the network resolver again.
+    pub fn invalidate_preresolve_entry(&self, name: &str) {
+        self.preresolve_addr_map.lock().unwrap().remove(name);
+    }
+
```
jmwample:
```suggestion
```

Same as `invalidate_preresolve_for`.

common/http-api-client/src/dns.rs
```diff
@@ -67,7 +66,12 @@ mod constants;
 mod static_resolver;
 pub(crate) use static_resolver::*;
 
-pub(crate) const DEFAULT_POSITIVE_LOOKUP_CACHE_TTL: Duration = Duration::from_secs(1800);
+// Fastly (and similar CDNs) deliberately publish short A-record TTLs (30–60 s)
+// so that clients re-resolve when an edge node is removed. Pinning to 1800 s
+// defeats that mechanism and leaves the client hitting a dead IP for up to
+// 30 minutes after a Fastly failover. 60 s is long enough to amortise DNS
+// query overhead while still following CDN-signalled failovers within a minute.
+pub(crate) const DEFAULT_POSITIVE_LOOKUP_CACHE_TTL: Duration = Duration::from_secs(60);
```
jmwample:
```suggestion
pub(crate) const DEFAULT_POSITIVE_LOOKUP_CACHE_TTL: Duration = Duration::from_secs(1800);
```

The concern here is not about amortizing DNS query overhead, it is about controlling when DNS queries happen and preventing them from happening at the wrong moment, causing a timeout and connection failure. We do DNS in a specific way for critical endpoints (using specially configured DoH and DoT resolvers) which can make DNS resolution expensive in the wrong conditions (high blocking of nameservers). If a record times out at a the wrong moment (e.g. just before doing a mixnet establish) and we are forced to wait for a DNS resolution the (mixnet) handshake will timeout. 

This also doesn't take into account the fact that this resolver is only used for critical API endpoint urls over which we have control, and we set the timeouts for them high so we should not need to worry about CDN-signalled failover.

jmwample (changes requested):

jmwample:
Thank you for the contribution! The DNS system is very specific and has a serious impact on the function of the system as a whole so changes require a lot of consideration and context.

jmwample (approved):
Looks good. The failing CI lint is unrelated to this PR and the builds are working for me.

PR #6689, opened 2026-04-20
feat: NMv3: add read-only results API to orchestrator

jstuczyn:
NYM-752

## Summary

- Add six read-only endpoints under `/v1/results/*` that expose the orchestrator's  local SQLite database to external consumers (grafana, harbourmaster, etc.)
- All routes are protected by the shared `metrics_and_results` bearer token
- Endpoints: paginated node list, node by id (with latest test run), paginated test run list, per-node test run history, single test run by id, and paginated in-progress test run list
- Storage layer extended with snapshot-consistent paginated queries (page + total in a single transaction) for all four relevant tables
- Replaced the previous silent 200-row cap on in-progress runs with standard `Pagination` / `PagedResult<T>` contract, matching the other endpoints
- Added `idx_testrun_test_timestamp` migration index to support efficient global test-run pagination without a full-table scan
- OpenAPI `responses` blocks on all agent and results handlers now document 4xx/5xx error shapes alongside the happy-path 200

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6689)


nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
```diff
@@ -290,13 +296,196 @@ impl AppState {
...
-                storage,
                 testrun_staleness_age,
             },
             validator_client,
         }
     }
+
+    /// Selects the most stale idle node and atomically marks it as having a test
+    /// in progress. Returns `None` if no node is currently eligible.
+    pub(crate) async fn assign_next_testrun(&self) -> Result<Option<TestRunAssignment>, ApiError> {
```
simonwicky, line 309: We seem to have a mix of functions being defined here and some directly in the endpoint function. 
Am I missing something? Any reason for that?

simonwicky (approved):

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
jstuczyn: just a stylistic choice to not have to deal with storage on the endpoint level, especially if calls interact with multiple things on the state

PR #6694, opened 2026-04-21
Testing port checks in NS Agents

benedettadavico:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6694)

## Summary by CodeRabbit

* **New Features**
  * TCP port reachability checks (per-port open/closed) and a port-only mode.
  * New run-ports CLI/agent flow to request, execute, and submit ports-check jobs; HTTP endpoints and background scheduler to enqueue checks.
  * Configurable port-check target, ports list, and timeout; built-in exit-policy port list generated at build time; results exposed via gateway APIs.

* **Tests**
  * Added/updated unit tests for port-list generation and ports-check normalization.

* **Chores**
  * Bumped agent and API package versions.


nym-node-status-api/nym-node-status-api/migrations_pg/20260414000000_ports_check_testrun_kind.sql
jstuczyn, line 1: what's the point of having 4 different migrations, especially if the modify the same table, if all of them would be applied during the same binary update?

nym-node-status-api/nym-node-status-client/src/api.rs
```diff
@@ -12,6 +12,10 @@ impl ApiPaths {
         format!("{}/internal/testruns", self.server_address)
     }
 
+    pub(super) fn request_ports_check_testrun(&self) -> String {
```
jstuczyn, line 15: I don't see those routes existing on NS api

nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -569,6 +588,92 @@ func ping(req NetstackRequestGo) (NetstackResponse, error) {
 	return response, nil
 }
 
+// checkPorts tests TCP connectivity to a target on each port through the WG
+// tunnel. blocked ports (NYM-EXIT iptables) will timeout or get reset.
+// runs concurrently.
+func checkPorts(target string, ports []uint16, timeoutSec uint64, tnet *netstack.Net) map[string]bool {
+	if target == "" {
+		target = "portquiz.net"
```
jstuczyn, line 598: what if `portquiz.net` is unavailable? then all nodes will report as having no open ports as opposed to the endpoint being unreachable. is there no way to make the tests without this external dep?

nym-node-status-api/nym-node-status-client/src/lib.rs
```diff
@@ -76,6 +79,46 @@ impl NsApiClient {
...
+            }
+        }
+
+        serde_json::from_str(&response_text)
+            .map(|testrun| {
+                tracing::info!("Received ports-check testrun assignment: {:?}", testrun);
+                testrun
+            })
+            .map_err(|err| {
+                tracing::error!("err");
```
jstuczyn: you probably meant `tracing::error!("{err}")`

nym-node-status-api/nym-node-status-agent/src/cli/mod.rs
```diff
@@ -71,28 +74,53 @@ pub(crate) struct RunProbeArgs {
...
+    #[arg(short, long, required = true)]
+    pub server: Vec<String>,
+
+    /// Only choose gateway with that minimum performance
+    #[arg(long)]
+    pub min_gateway_mixnet_performance: Option<u8>,
+
+    /// Ignore egress epoch role constraints
+    #[arg(long)]
+    pub ignore_egress_epoch_role: bool,
```
jstuczyn, line 94: why do we even care about epoch roles if we're not sending mixnet traffic... are we?

nym-gateway-probe/src/lib.rs
```diff
@@ -20,18 +22,40 @@ use nym_sdk::mixnet::{
...
+        None => true,
+        Some(ts) => now.saturating_sub(ts) >= FOURTEEN_DAYS_SECS,
+    }
+}
+
+fn unix_timestamp_secs() -> i64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_secs() as i64)
+        .unwrap_or(0)
```
jstuczyn:
```suggestion
    OffsetDateTime::now_utc().unix_timestamp()
```

nym-gateway-probe/src/config/netstack.rs
```diff
@@ -3,6 +3,11 @@
 
 use clap::Args;
 
+// EXIT_POLICY_PORTS is generated at build time by parsing PORT_MAPPINGS
+// from scripts/nym-node-setup/network-tunnel-manager.sh.
+// To add or remove ports, update PORT_MAPPINGS in the shell script and rebuild.
+include!(concat!(env!("OUT_DIR"), "/exit_policy_ports.rs"));
```
jstuczyn, line 9: wouldn't it be safer for NS to provide list of ports (which it would determine by querying the exit policy itself?) then we wouldn't need to worry about possibly stale data

nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -569,6 +588,92 @@ func ping(req NetstackRequestGo) (NetstackResponse, error) {
...
+	}
+	if timeoutSec == 0 {
+		timeoutSec = 5
+	}
+
+	// resolve target once via host DNS (we only care about TCP reachability
+	// through the tunnel, not DNS-through-tunnel behaviour)
+	targetIP := target
+	if net.ParseIP(target) == nil {
+		addrs, err := net.LookupHost(target)
```
jstuczyn, line 606: what would happen if DNS was broken, as it usually is? could we perhaps have some fallback IP or do we not care?

nym-gateway-probe/netstack_ping/lib.go
```diff
@@ -569,6 +588,92 @@ func ping(req NetstackRequestGo) (NetstackResponse, error) {
...
+			log.Printf("Port check: resolved %s -> %s", target, targetIP)
+		}
+	}
+
+	timeout := time.Duration(timeoutSec) * time.Second
+
+	// batching: portquiz.net enforces a per-source-IP connection limit.
+	// testing all 114 ports in one burst trips it after ~29 connections.
+	// small batches with a cooldown in between keep us under the limit.
+	const batchSize = 20
```
jstuczyn: that seems... fragile

jstuczyn (approved):
I'm going to approve it, but please, do double check the timeouts for running the port scans and what happens if they take more than the specified limit

PR #6709, opened 2026-04-24
feat: NMv3: submission of stress testing result into nym-api

jstuczyn:
NYM-749

## Summary

Allow the network monitor orchestrator to submit stress-testing results to the nym-api over a signed, authenticated channel.

- **Contract**: orchestrators can now announce an ed25519 identity key on-chain; shape-only validation (base58 + 32 bytes), no curve-point check by design.
- **Orchestrator startup**: verifies on-chain authorisation (with retry/backoff), reconciles its announced identity key against the local one, and self-checks that the nym-api recognises it before
spawning background tasks.
- **Result submission**: new `ResultSubmitter` task periodically drains freshly-completed `testrun` rows, wraps them in a signed `StressTestBatchSubmission`, and POSTs them to nym-api. Watermark is
advanced only on POST success (at-least-once delivery — losing measurements is worse than occasional duplicates).
- **nym-api**:
  - `POST /v3/nym-nodes/stress-testing/batch-submit` — staleness check, authorised-set membership, monotonic-timestamp replay protection, signature verify, then per-entry insert.
  - `GET /v3/nym-nodes/stress-testing/known-monitors/{identity_key}` — lets orchestrators self-check authorisation after announcing their key.
  - Authorised set is TTL-cached with single-flight refresh to avoid thundering the chain.
- **Dedup**: each result carries its orchestrator-local `testrun_id`, stored alongside the submitter's pubkey as a composite primary key. Retried submissions drop duplicates via `INSERT OR IGNORE`.

Results are persisted but not yet consumed — downstream usage lands in a follow-up PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6709)


nym-api/migrations/20260424120000_stress_testing.sql
```diff
@@ -0,0 +1,29 @@
...
+    -- it uniquely identifies a measurement and lets us dedupe retried submissions (the
+    -- orchestrator uses at-least-once delivery and may re-POST the same row after a crash
+    -- between a successful POST and its watermark update).
+    testrun_id       INTEGER                     NOT NULL,
+
+    -- Base58-encoded ed25519 identity key of the submitting orchestrator. Part of the primary
+    -- key so distinct orchestrators can coincidentally share a `testrun_id` without colliding.
+    submitter_pubkey TEXT                        NOT NULL,
+
+    -- unfortunately, due to legacy reasons we have separate tables for mixnodes and gateways
```
simonwicky, line 18: Huh?! Either I don't see it, or that comment is outdated?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/orchestrator/mod.rs
```diff
@@ -90,20 +98,141 @@ impl NetworkMonitorOrchestrator {
...
+    /// `chain_authorisation_check_max_attempts` attempts spaced by
+    /// `chain_authorisation_check_retry_delay`; once exhausted the function returns an error and
+    /// `new()` aborts before any background tasks are spawned.
+    async fn verify_orchestrator_chain_authorisation(&self) -> anyhow::Result<Option<String>> {
+        let query_client = self.client.read().await.nyxd.clone_query_client();
+        let address = self.address().await;
+        let max_attempts = self.config.chain_authorisation_check_max_attempts;
+        let retry_delay = self.config.chain_authorisation_check_retry_delay;
+
+        if max_attempts == 0 {
```
simonwicky: nit: Use `NonZeroUsize` to remove that check?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/storage/models.rs
```diff
@@ -239,6 +240,37 @@ impl From<TestRun> for TestRunData {
...
+///   that case apart from a genuine zero score.
+/// - `was_reachable` is `error.is_none()` — i.e. the test completed without an abort error. A run
+///   that aborted before the node responded sets `error` to the first failure, so the inverse is
+///   an accurate "did we reach the node at all" signal.
+impl From<TestRun> for StressTestResult {
+    fn from(run: TestRun) -> Self {
+        let id = run.id;
+        let inner = run.inner;
+        let test_performance = if inner.packets_sent > 0 {
+            (inner.packets_received as f64 / inner.packets_sent as f64).clamp(0.0, 1.0)
```
simonwicky: I'm not fond of that clamp. It should never trigger, and it triggering is actually an error or a sign that something went wrong. Maybe the test run should be discarded?
We're also losing the `duplicates` column. What if we received a whole bunch of duplicates? It would inflate those performances.

nym-api/migrations/20260424120000_stress_testing.sql
jstuczyn: discussed on zulip - not outdated : )

simonwicky (approved):

PR #6712, opened 2026-04-27
[LP-DATA] Mix simulator, pipeline traits

simonwicky:
## Summary

Adds two new workspace crates that together form an abstract pipeline framework for the upcoming Lewes Protocol (LP) plus a runnable simulator that consumes it.

- **`common/nym-lp-data`** — a trait-only "vocabulary" crate that defines the shape of client and mix-node packet-processing pipelines. No concrete cryptography, transport, or networking code.
- **`nym-mix-sim`** — a discrete-time mixnet simulator running on localhost UDP. Doubles as the reference implementation of every trait in `nym-lp-data`.

> ⚠️ **Disclaimer :** the traits introduced in `nym-lp-data` are not stable yet, they might be subject to changes and tweaks in the future

## `nym-lp-data` — the vocabulary

Three modules:

- **`common`** — wire-layer traits shared by clients and nodes (`Framing` / `Transport` and their unwrap counterparts) plus the supertraits `WireWrappingPipeline` / `WireUnwrappingPipeline`. `common/helpers.rs` adds `NoOpWireWrapper` / `NoOpWireUnwrapper` markers for packet types that are already self-contained on the wire (e.g. a Sphinx packet).
- **`clients`** — the six-stage outbound pipeline `Chunking → Reliability → Obfuscation → RoutingSecurity → Framing → Transport`, the supertrait `ClientWrappingPipeline` that ties them together with a default `process()`, an inbound counterpart, no-op marker traits per stage, a `Pipeline` composition struct, and a tick-driven `ClientWrappingPipelineDriver`.
- **`mixnodes`** — `MixnodeProcessingPipeline` (unwrap → `mix()` → re-wrap).

A few generic wrappers thread per-packet state through every stage: `TimedData`, `AddressedTimedData`, and `PipelineData` (which adds per-message `InputOptions`).

### A couple of design points worth flagging

- **`process()` is called every tick, with or without input.** `Reliability` and `Obfuscation` are explicitly invoked twice per tick (once with the real input, once with `None`) so retransmissions and cover-traffic loops can fire on idle ticks without the caller knowing what's in flight.
- **`InputOptions` are per-message, not per-pipeline.** They toggle whether reliability / obfuscation / routing-security run for a given payload and carry the next-hop id, which is what lets a single pipeline interleave real and cover packets with different routing rules.
- **The `Frame` type is an associated item** cross-constrained by the wire-pipeline supertraits, so the parameter list stays short at the cost of forcing implementors to spell out the cross-constraint when composing. Open to feedback.

A small synthetic integration test wires a mock pipeline against the `nym-lp` packet types as a self-contained example of every trait being implemented.

## `nym-mix-sim` — the simulator

Discrete-time mixnet simulator that runs a configurable number of mix nodes and clients on localhost UDP. Time advances in **ticks**; each tick runs the client phase, then drains sockets, mixes buffered packets, and dispatches outgoing ones.

Two binaries: `nym-mix-sim` (generates `topology.json` and runs the loop) and `mix-client` (stdin → app socket of a running client).

Three driver flavours, controlled by `--driver`:

| Driver | Timestamp | Encryption | Cover traffic | Reliability | Manual mode |
|--------|-----------|------------|---------------|-------------|-------------|
| `simple` | u32 tick counter | none | no | no | yes |
| `sphinx` | wall-clock `Instant` | full Sphinx | Poisson | SURB ACKs | no |
| `discrete-sphinx` (default) | u32 tick counter (1 tick = 1 ms) | full Sphinx | Poisson | SURB ACKs | yes |

`--manual` pauses for ENTER between ticks and pretty-prints each node's buffer state per phase, which is useful for stepping a Sphinx packet hop-by-hop. The Sphinx driver implements 3-hop onion encryption, two-loop Poisson cover traffic, a SURB-ACK reliability layer, and fragment reconstruction — exercising every trait in `nym-lp-data`.

See [`nym-mix-sim/README.md`](nym-mix-sim/README.md) and [`common/nym-lp-data/README.md`](common/nym-lp-data/README.md) for full docs and quick-start examples.

## Small workspace touches

- `Cargo.toml`: register both new crates and expose `nym-lp-data` as a workspace dependency.
- `common/crypto`: new `bs58_x25519_private_key` serde helper used by `topology.json` to round-trip per-node Sphinx private keys.
- `common/nym-lp`: expose `LpHeader::SIZE` as a `pub const` so the integration test can size frames against it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6712)

## Summary by CodeRabbit

* **New Features**
  * Added a discrete-time mixnet simulator with multiple modes (simple, Sphinx, discrete Sphinx), a runnable CLI (init-topology, run) and a standalone mix-client for sending messages.
  * Added a reusable LP data crate providing timestamped pipeline types, client/mixnode pipeline interfaces, and no-op helpers.

* **Documentation**
  * Added comprehensive READMEs for the simulator and LP data crate.

* **Tests**
  * Added integration tests and mocks exercising the LP pipeline.

* **Chores**
  * Workspace updated to include the new crates and binaries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym/pull/6712?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cargo.toml
```diff
@@ -459,6 +462,7 @@ nym-id = { version = "1.20.4", path = "common/nym-id" }
 nym-ip-packet-client = { version = "1.20.4", path = "nym-ip-packet-client" }
 nym-ip-packet-requests = { version = "1.20.4", path = "common/ip-packet-requests" }
 nym-lp = { version = "1.20.4", path = "common/nym-lp" }
+nym-lp-data = { version = "1.20.4", path = "common/nym-lp-data" }
```
jstuczyn: does it actually work if it's not published to crates.io?

common/nym-lp-data/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Helpers
+
+- **Client-stage no-op marker traits** ([`NoOpReliability`], [`NoOpRoutingSecurity`], [`NoOpObfuscation`] in [`clients/helpers.rs`](src/clients/helpers.rs)) — implement these to opt out of a pipeline stage with zero overhead. Useful for stub or testing pipelines.
+- **Wire-layer no-op marker traits** ([`NoOpWireWrapper`], [`NoOpWireUnwrapper`] in [`common/helpers.rs`](src/common/helpers.rs)) — collapse the entire wire layer (framing + transport, or their inverses) to a pass-through. Use these when your packet type is already self-contained on the wire (e.g. a Sphinx packet) and needs no extra framing or transport header. `NoOpWireWrapper` requires `Pkt: From<Vec<u8>>`; `NoOpWireUnwrapper` requires `Pkt: Into<Vec<u8>>` and `Mk: Default`.
+- **`Pipeline` composition structs** (in [`clients/types.rs`](src/clients/types.rs) and [`mixnodes/types.rs`](src/mixnodes/types.rs)) — generic structs that aggregate one component per pipeline stage and provide blanket impls of the relevant supertraits, so you can build a working pipeline by plugging in any combination of stage implementations.
```
jstuczyn: probably a stale comment - `mixnodes/types.rs` does not exist (anymore (?))

common/nym-lp-data/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+    Ts: Debug,
+    NdId: Debug,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "PipelineData {{")?;
+        writeln!(f, "    dst: {:#?}", &self.dst)?;
+        writeln!(f, "    data: {:#?}", &self.data)?;
+        write!(f, "}}")
+    }
+}
```
jstuczyn: this comment is going to apply to all of manual implementations of `Debug`, there's no point in me repeating myself. why not use the `debug_struct` rather than implementing the whole thing manually and having to deal with braces, spacing, etc.? i.e.
```rust
 f.debug_struct("PipelineData")
              .field("dst", &self.dst)
              .field("data", &self.data)
              .finish_non_exhaustive()   // or .finish(), depending on what you want
```

this will also automagically deal with the `{:?} vs {:#?}` cases

jstuczyn: and any weird inner formatting (does not apply to `PipelineData`) could be dealt with per-field wrappers. 
I just don't see a benefit in reinventing the wheel

common/nym-lp-data/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// A value of type `D` tagged with a timestamp of type `Ts`.
+///
+/// `TimedData` threads timing information through every stage of the LP
+/// pipeline.  It is produced by [`clients::traits::Chunking`] and propagated
+/// unchanged (or with the timestamp transformed) through every subsequent
+/// pipeline stage until the packet is sent on the wire.
+#[derive(Clone)]
+pub struct TimedData<Ts, D> {
+    pub timestamp: Ts,
```
jstuczyn, line 43: after looking around, I'm not really convinced it's useful to have a generic timestamp. 
1. it's producing a lot of noise in an already generic-heavy pipeline
2. forces existence of additional two traits to hide `Ts` (`GenerateDelay` + `AddDelay`)
3. the only case where `Instant` is not being used is just the mix-sim

so because of the last point, I understand why you didn't want to make it an `Instant`, but how about create an explicit `Tick(u64)` type instead? then both of those helper traits could be simply collapsed onto `Tick` impl. internally it would just hold nanos since the pipeline begun with two helpers on it to allow use in both production (`Instant`) and the mix sim (the +1 thing):
```rust
  impl Add<Duration> for Tick { ... }
  impl Add<u64> for Tick { ... }
```

common/nym-lp-data/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+/// [`Obfuscation`]: crate::clients::traits::Obfuscation
+/// [`RoutingSecurity`]: crate::clients::traits::RoutingSecurity
+/// [`Framing`]: crate::common::traits::Framing
+/// [`Transport`]: crate::common::traits::Transport
+/// [`InputOptions`]: crate::clients::InputOptions
+#[derive(Clone)]
+pub struct PipelineData<Ts, D, Opts, NdId> {
+    pub data: TimedData<Ts, D>,
+    pub options: Opts,
+    pub dst: NdId,
```
jstuczyn, line 99: another generic that I think could be dropped. didn't we already discuss to define destination as a concrete enum to have either client's address or node's socket address?

common/nym-lp-data/src/clients/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Each pipeline stage (reliability, routing security, obfuscation) is optional
+/// and toggled per-message by the corresponding accessor.  The next-hop
+/// destination is also resolved from the options so that addressing is decided
+/// before the payload reaches [`Framing`].
+///
+/// # Type Parameters
+/// - `NdId`: addressing type used to identify the next-hop destination.
+///
+/// [`Framing`]: crate::common::traits::Framing
+pub trait InputOptions<NdId>: Clone {
```
jstuczyn, line 20: this honestly ~~could have been an email~~ could have been a struct. literally as simple as 
```rust
  pub struct InputOptions {
      pub reliability: bool,
      pub routing_security: bool,
      pub obfuscation: bool,
      pub next_hop: NextHop,
  }
```
what does making it into generic give us? `SphinxInputOptions` and `SimpleInputOptions` are just constant toggles with no real interesting variation between them

common/nym-lp-data/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+//! [`TimedPayload`] is a convenience alias for `TimedData<Ts, Vec<u8>>`.
+
+use std::fmt::Debug;
+
+pub mod clients;
+pub mod common;
+pub mod mixnodes;
+
+/// Convenience alias for [`TimedData`] when the payload is a raw byte buffer.
+pub type TimedPayload<Ts> = TimedData<Ts, Vec<u8>>;
```
jstuczyn, line 29: investigate whether perhaps using `Bytes` rather than `Vec<u8>` would be more efficient when processing -> we might be able to avoid extra allocations

common/nym-lp-data/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+            options: self.options,
+            dst: self.dst,
+        }
+    }
+
+    /// Drop the pipeline options, producing a plain addressed payload.
+    pub fn into_addressed(self) -> AddressedTimedData<Ts, D, NdId> {
+        AddressedTimedData {
+            data: self.data,
+            options: (),
```
jstuczyn, line 143: the empty opts make it look quite weird - another argument for perhaps just getting rid of that generic : )

common/nym-lp-data/src/common/traits.rs
```diff
@@ -0,0 +1,182 @@
...
+pub trait WireWrappingPipeline<Ts, Pkt, Opts, NdId>:
+    Transport<Ts, Pkt, NdId>
+    + Framing<Ts, Opts, NdId, Frame = <Self as Transport<Ts, Pkt, NdId>>::Frame>
+where
+    Ts: Clone,
+    NdId: Clone,
+{
+    // IMPORTANT NOTE : This fn can be not constant to allow e.g. flexible MTU
+    // However, every possible value must be able to accomodate the different overhead.
+    // If it doesn't, the pipeline becomes unusable
```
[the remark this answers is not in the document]
jstuczyn: I guess our AI overlords are right with this one lol

common/nym-lp-data/src/common/traits.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn packet_size(&self) -> usize;
+
+    fn frame_size(&self) -> usize {
+        // SAFETY : While this CAN technically fail, it means that something is wrong in the code and it's pointless to continue anyway
```
jstuczyn, line 130: could we perhaps have an explicit check for this when constructing given pipeline (in `new` or something) to make sure no packets could even be inserted into the pipeline?

common/nym-lp-data/src/clients/traits.rs
```diff
@@ -0,0 +1,304 @@
...
+        timestamp: Ts,
+    ) -> Vec<PipelinePayload<Ts, Opts, NdId>>;
+}
+
+/// Trait for applying reliability encoding (e.g. SURB ACKs, retransmissions) to
+/// a timed payload.
+///
+/// # Type Parameters
+/// - `Ts`: Timestamp type carried by the [`PipelinePayload`].
+/// - `Opts`: Per-message pipeline options.
```
jstuczyn, line 39: just thinking aloud: under what condition would we want a message to have different pipeline options. shouldn't the options be property of the pipeline itself or perhaps of the stage itself? (i.e. part of `Reliability` in this instance)?

common/nym-lp-data/src/clients/traits.rs
```diff
@@ -0,0 +1,304 @@
...
+/// - `encrypt`: Encrypt the given payload, returning a new [`PipelinePayload`].
+///
+/// # Provided Methods
+/// - `nb_frames`: Number of transport frames that one encrypted payload expands
+///   into; defaults to `1`.  Override when the encryption scheme (e.g. Sphinx)
+///   produces multiple frames per input chunk.
+pub trait RoutingSecurity<Ts, Opts, NdId> {
+    const OVERHEAD_SIZE: usize;
+    fn nb_frames(&self) -> usize {
+        1
```
jstuczyn, line 106: I feel that could be a landmine and it probably shouldn't provide e default implementation. also, wouldn't number of frames depend on available  size?

common/nym-lp-data/src/clients/traits.rs
```diff
@@ -0,0 +1,304 @@
...
+    Ts: Clone,
+    NdId: Clone,
+    Opts: InputOptions<NdId>,
+{
+    fn chunk_size(&self, input_options: Opts) -> usize {
+        // Frame size comes from WireWrappingPipeline
+        let mut chunk_size = self.frame_size();
+
+        if input_options.routing_security() {
+            // SAFETY : While this CAN technically fail, it means that something is wrong in the code and it's pointless to continue anyway
```
jstuczyn, line 148: let's add a check for it elsewhere imo

common/nym-lp-data/src/clients/traits.rs
```diff
@@ -0,0 +1,304 @@
...
+                .expect("not enough room in a packet for reliability overhead");
+            chunk_size = pre_reliability_chunk_size;
+        }
+
+        chunk_size
+    }
+
+    fn process(
+        &mut self,
+        input: Option<(Vec<u8>, Opts)>, // Optional to be able to tick the pipeline without input
```
jstuczyn, line 170: I think I already added a comment about per-packet options. I think they should be property of the pipeline, especially since they're going to be copied **A LOT**

common/nym-lp-data/src/clients/traits.rs
```diff
@@ -0,0 +1,304 @@
...
+        timestamp: Ts,
+    ) -> Vec<AddressedTimedData<Ts, Pkt, NdId>> {
+        let mut chunks = if let Some((input_data, input_options)) = input {
```
jstuczyn, line 173: this is just a nit because we'd need proper benchmarks for the thing, but it seems you're doing a lot of intermediate allocations via `.collect();` -> I wonder if we could have chained the iterators instead

common/nym-lp-data/src/nymnodes/traits.rs
jstuczyn, line 1: nit: i'd perhaps rename it to either `nymnodes` or just `nodes` as gateways would also have to apply that transformation

common/nym-lp-data/src/nymnodes/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+///   [`WireWrappingPipeline::wire_wrap`].
+///
+/// [`mix`]: MixnodeProcessingPipeline::mix
+pub trait MixnodeProcessingPipeline<Ts, Pkt, Opts, Mk, NdId>:
+    WireUnwrappingPipeline<Ts, Pkt, Mk> + WireWrappingPipeline<Ts, Pkt, Opts, NdId>
+where
+    Ts: Clone,
+    NdId: Clone,
+{
+    fn mix(
```
jstuczyn, line 39: the name of this function is quite confusing, doesn't it kinda unwrap the sphinx layer?

nym-mix-sim/src/node/sphinx.rs
```diff
@@ -0,0 +1,178 @@
...
+            Ok(packet) => match packet.data {
+                nym_sphinx::ProcessedPacketData::ForwardHop {
+                    next_hop_packet,
+                    next_hop_address,
+                    delay,
+                } => {
+                    let timed_sphinx = AddressedTimedData::new_addressed(
+                        timestamp.add_delay(delay),
+                        next_hop_packet.to_bytes(),
+                        next_hop_address.as_bytes()[0],
```
jstuczyn, line 131: huh. why are we taking the first byte of the next hop?

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::mpsc;
```
jstuczyn, line 4: I'm not sure if stdlib mpsc is the best choice. for non-async crossbeam is usually the go-to

nym-mix-sim/src/node/sphinx.rs
```diff
@@ -0,0 +1,178 @@
...
+                } => {
+                    if let Ok(plaintext) = payload
+                        .recover_plaintext()
+                        .inspect_err(|e| tracing::warn!("Impossible to recover plaintext : {e}"))
+                    {
+                        let (surb_ack_bytes, message) = SurbAck::extract_ack_and_message(plaintext);
+                        let mut packets_to_forward = vec![AddressedTimedData::new_addressed(
+                            timestamp.clone(),
+                            message,
+                            destination.as_bytes()[0],
```
jstuczyn, line 148: same thing - what's with this first byte check? I don't understand it

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
...
+///
+/// `Ts` must implement `Clone + PartialOrd` so that timestamps can be compared
+/// to decide which packets are due.
+///
+pub struct ClientWrappingPipelineDriver<Ts, Pkt, Opts, NdId>
+where
+    Ts: Clone + PartialOrd,
+    Opts: InputOptions<NdId>,
+{
+    pipeline: Box<dyn DynClientWrappingPipeline<Ts, Pkt, Opts, NdId>>,
```
jstuczyn, line 32: does it really have to be called with a dynamic dispatch? we're already full of generics as it is anyway

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    packet_buffer: Vec<AddressedTimedData<Ts, Pkt, NdId>>,
+
+    input: mpsc::Receiver<(Vec<u8>, Opts)>,
```
jstuczyn, line 36: I already suggested removing per-packet `Opts`, but **IF** we don't, can we at the very least wrap that tuple into a struct with named fields? it is not obvious, looking at the type signature, what each field is meant to represent

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl<Ts, Pkt, Opts, NdId> ClientWrappingPipelineDriver<Ts, Pkt, Opts, NdId>
+where
+    Ts: Clone + PartialOrd,
+    Opts: InputOptions<NdId>,
+{
+    /// Create a new driver wrapping `pipeline`.
+    ///
+    /// Internally allocates a zero-capacity `sync_channel` for input payloads.
```
jstuczyn, line 49: in that case, do we even need a channel?

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Reads a pending input payload (if both the packet buffer and the
+    /// obfuscation buffer are empty), runs it through the pipeline, then
+    /// returns all packets whose `timestamp ≤ now`.
+    pub fn tick(&mut self, timestamp: Ts) -> Vec<(Pkt, NdId)> {
+        // We're reading a message only if our buffer is empty
+        // Otherwise, we will have buffers adding latencies to data
+        let next_message = if self.packet_buffer.is_empty() {
+            self.input
+                .try_recv()
+                .inspect_err(|_| tracing::trace!("No message in the queue"))
```
jstuczyn, line 80: I feel like our input channel could have non-zero buffer and here we'd just attempt to process all stored messages at once (like say 8 or 32)

common/nym-lp-data/src/clients/driver.rs
```diff
@@ -0,0 +1,93 @@
...
+                .ok()
+        } else {
+            None
+        };
+        self.packet_buffer
+            .extend(self.pipeline.process(next_message, timestamp.clone()));
+
+        self.packet_buffer
+            .extract_if(.., |p| p.data.timestamp <= timestamp)
```
jstuczyn, line 89: how many items do we expect to see in the buffer? because each tick we have to go through the **entire** vec and allocate a new one with extracted elements. what if we stored everything already sorted by timestamp?

common/nym-lp-data/src/clients/helpers.rs
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::PipelinePayload;
+use crate::clients::traits::{Obfuscation, Reliability, RoutingSecurity};
+
+/// Marker trait for a no-op [`Reliability`] implementation.
+///
+/// Implement this for your pipeline type to get a [`Reliability`] impl that
+/// passes the payload through unchanged with zero byte overhead.
+pub trait NoOpReliability {}
```
jstuczyn, line 11: nit: (also for other `NoOp` steps), define it as a unit struct instead, i.e. `pub trait NoOpReliability`, this way it will be more convenient to instantiate

common/nym-lp-data/src/common/helpers.rs
```diff
@@ -0,0 +1,105 @@
...
+        WireWrappingPipeline,
+    },
+};
+
+/// Marker trait for a no-op [`WireWrappingPipeline`] implementation.
+///
+/// Implement this for your pipeline type to get a [`WireWrappingPipeline`] impl that
+/// passes the payload through unchanged with zero byte overhead.
+pub trait NoOpWireWrapper {
+    const PACKET_SIZE: usize = 1500;
```
jstuczyn, line 17: should it really be hardcoded? I could imagine people wanted to use different values, especially in non-standard MTU settings or in tests, etc.

common/nym-lp-data/tests/integration/main.rs
```diff
@@ -0,0 +1,34 @@
...
+        transport: LpTransport,
+        packet_size,
+    };
+
+    let output = mock_pipeline.process(None, 1);
+
+    assert!(output.is_empty());
+}
+
+// TODO More test to come later
```
jstuczyn, line 35: sure : )

common/nym-lp-data/tests/integration/common/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+                    chunk.to_vec(),
+                    input_options,
+                    input_options.next_hop(),
+                )
+            })
+            .collect()
+    }
+}
+
+pub struct KcpReliability;
```
jstuczyn: another nit: I'd perhaps add a `Mock` prefix  to all below types to indicate they're not the real implementation (especially when we do implement to production variant and you're going to be doing a global search in your IDE searching for it)

nym-mix-sim/src/bin/mix_client.rs
```diff
@@ -0,0 +1,102 @@
...
+    println!("(Ctrl-C to quit)");
+
+    let mut line = String::new();
+    loop {
+        line.clear();
+        if std::io::stdin().read_line(&mut line)? == 0 {
+            break; // EOF
+        }
+
+        let text = line.trim_end_matches('\n').trim_end_matches('\r');
```
jstuczyn: you could just do `line.trim()`

nym-mix-sim/src/node/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+pub struct BaseNode<Ts, Pkt, Pn> {
+    /// Identifier of this node within the topology.
+    pub(crate) id: NodeId,
+    /// Notional reliability percentage; not yet used by the simulator but kept
+    /// so future tests can drive the reliability layer.
+    _reliability: u8,
+    /// UDP address this node is bound to.
+    pub(crate) socket_address: SocketAddr,
+    /// Non-blocking UDP socket used for both receive and send.
+    socket: UdpSocket,
```
jstuczyn, line 89: it'd be probably too much hassle changing it now, but I wish the simulator was purely in memory without any networking between nodes, i.e. instead they'd be writing to some shared memory between themselves

nym-mix-sim/src/node/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /// Attempt to receive one UDP datagram and deserialise it as `Pkt`.
+    ///
+    /// Returns `None` when the socket would block (no datagram waiting).
+    pub fn recv_packet(&self) -> Option<anyhow::Result<Pkt>>
+    where
+        Pkt: WirePacketFormat,
+    {
+        let mut buf = [0; 1500];
```
jstuczyn, line 176: this is a simulator so that's fine, but what if the datagram is bigger than the buf? then we'll never read a valid Pkt. maybe the buf should always be set to the size of the expected packet (based on the pipeline)?

nym-mix-sim/src/packet/simple.rs
```diff
@@ -0,0 +1,281 @@
...
+    ///
+    /// Layout: 16 bytes UUID (LE) + 48 bytes payload = 64 bytes total.
+    const SIZE: usize = 64;
+
+    /// Create a new [`SimplePacket`] with a freshly generated UUID v4 and the
+    /// provided 48-byte payload.
+    ///
+    /// The payload array is exactly `SIZE - 16 = 48` bytes so that the packet
+    /// serialises to exactly [`SimplePacket::SIZE`] bytes.
+    pub fn new(data: [u8; Self::SIZE - 16]) -> Self {
```
jstuczyn: rather than having magic constants, could we definie explicit constant like `UUID_SIZE: usize = 16` and use it here instead?

nym-mix-sim/src/packet/sphinx.rs
```diff
@@ -0,0 +1,309 @@
...
+        rng: &mut R,
+        recipient: ClientId,
+        packet_id: u64,
+        directory: &Directory,
+    ) -> Self
+    where
+        R: Rng,
+    {
+        let route = directory
+            .random_route(3, rng)
```
jstuczyn, line 95: nit: I think it could be potentially a nice feature to simulate different number of hops

nym-mix-sim/src/client/sphinx/poisson_cover_traffic.rs
```diff
@@ -0,0 +1,152 @@
...
+        address: ClientId,
+        directory: Arc<Directory>,
+        current_timestamp: Ts,
+        rng: R,
+    ) -> Self {
+        Self {
+            address,
+            directory,
+            main_loop_next_timestamp: current_timestamp.clone(),
+            secondary_loop_next_timestamp: current_timestamp,
```
jstuczyn, line 77: with this our first packet will ALWAYS be a cover packet - is that what we want?

jstuczyn (commented):
first of all, good job, it's definitely a non-trivial piece of work. apart from few nits, I have just two comments:
1. it seems you're reinventing the wheel with hand-crafted `Debug` implementation for multiple structs. we could just use the inbuilt helpers with perhaps additional wrappers for say the output of `format_debug_bytes`
2. I think we've leaned too much into generics and some of them are quite redundant. they don't bring any value apart from reducing readaibility and increasing complexity. In my opinion things like `Ts`, `NdId` or `InputOptions` could be converted into concrete types instead

Cargo.toml
simonwicky: No idea, I copied it from the rest. I don't see why it wouldn't though, the path is there

common/nym-lp-data/src/lib.rs
simonwicky: I will double check those debug implementation indeed. 
Some of them are necessary to keep the nice alignment of `format_debug_bytes`

simonwicky: Removing custom impl from `nym-lp-data` crate. 
Mix-sim ones are staying, otherwise `format_debug_bytes` isn't displayed nicely

common/nym-lp-data/src/lib.rs
simonwicky: 1. Agreed on the noise, although every concrete use in prod will fix it and can hide it with an alias.
2. Only because `sphinx::Delay` is rigid, and I don't want to implement two mixing stage. (Also happens only in the mix-sim
3. For now. I can see us not using `Instant` in some tests as well. 

`Tick(u64)` does sound nice, but I reckon it will just make prod more complex at the benefit  of slightly simple mix-sim which makes little sense don't you think?

common/nym-lp-data/src/lib.rs
simonwicky: In production, yes. 
Mix-sim uses simple IDs, I reckon tests could use it too. 
This is debatable though

common/nym-lp-data/src/clients/mod.rs
simonwicky: I disagree.
Sure, for now the two concrete types are uninteresting for now, but it could be more. 
E.g. is how `InputMessage`s have options on them already, number of surbs, anonymous tags to use, use the sender address in plain. That would be options required only by the specific pipeline.

common/nym-lp-data/src/lib.rs
simonwicky: Will investigate

common/nym-lp-data/src/lib.rs
simonwicky: Initially, `AddressedTimedData` and `PipelineData` were distinct type, but it was creating a lot of duplication. 
Still, I think options are needed. This might look a bit weird, but it keeps code to a minimum no?

common/nym-lp-data/src/common/traits.rs
simonwicky: `process` could have a `debug_assert` maybe?
This is a tricky one, because `packet_size` isn't a constant on purpose, to allow flexible MTUs. 
Given that, I don't really know how we can have compile time checks

common/nym-lp-data/src/clients/traits.rs
simonwicky: What if we want to have a stream of obfuscated packets and one where we don't really care. 
Then we would need two pipelines instead of just indicating at the packet level

common/nym-lp-data/src/clients/traits.rs
simonwicky: Agreed, I don't really know how to not make it a runtime check

common/nym-lp-data/src/clients/traits.rs
simonwicky: I'm not sure you understood it correctly. It is a way of saying "A Sphinx/Outfox packet should span X final frames", then having a default of 1 kinda makes sense. 
It is a mean of reducing the overhead by fragmenting a secure packet over multiple frames

common/nym-lp-data/src/clients/traits.rs
simonwicky: Subject to discuss

common/nym-lp-data/src/clients/traits.rs
simonwicky: I will look at it

common/nym-lp-data/src/nymnodes/traits.rs
simonwicky: In mixing nodes, yes. In other places, it can do a lot of different things. 
The name is a bit confusing, I didn't want to use `process`

nym-mix-sim/src/node/sphinx.rs
simonwicky: Mix-sim addressing is using `u8` but sphinx packets want more. I just repeated the `u8` in all the slots, so I can just take the first one when unwrapping.

nym-mix-sim/src/node/sphinx.rs
simonwicky: Mix-sim addressing is using `u8` but sphinx packets want more. I just repeated the `u8` in all the slots, so I can just take the first one when unwrapping.

common/nym-lp-data/src/clients/driver.rs
simonwicky: Now that I'm a bit further, I think we might even not need it at all.

common/nym-lp-data/src/clients/driver.rs
simonwicky: Yes of course

common/nym-lp-data/src/clients/driver.rs
simonwicky: Yes, why? 
It is a mean for process using the pipeline to be aware of the back pressure. 
We've had many issues in the past with buffer adding latencies everywhere that I wanted to avoid that. 
Again though, we might not even use that driver at all in the end

common/nym-lp-data/src/clients/driver.rs
simonwicky: Indeed. In the spirit of back pressure though, this might not be ideal. 
Let's keep the driver code for a future time when we will be sure we need it

common/nym-lp-data/src/clients/driver.rs
simonwicky: I don't expect a lot, but it might be a good idea indeed

common/nym-lp-data/src/clients/helpers.rs
simonwicky: They are convenient marker traits, not structs

common/nym-lp-data/src/common/helpers.rs
simonwicky: They are not obliged to used that marker traits then. I can make it generic though ;)

common/nym-lp-data/tests/integration/main.rs
simonwicky: Let's see if the driver is actually useful

common/nym-lp-data/tests/integration/common/mod.rs
simonwicky: Yes indeed

nym-mix-sim/src/bin/mix_client.rs
simonwicky: Indeed

nym-mix-sim/src/node/mod.rs
simonwicky: That is correct

nym-mix-sim/src/packet/sphinx.rs
simonwicky: There you go, something you could put in an per-packet option struct that would implement `InputOptions`

nym-mix-sim/src/client/sphinx/poisson_cover_traffic.rs
simonwicky: First of all disclaimer : This is an attempt at creating Poisson cover traffic and not too much thought was given to that. 
That being said, unless you start sending traffic right away, first packets will always be cover traffic

nym-mix-sim/src/node/mod.rs
simonwicky: Probably not. A matter of changing the base nodes, which we can imagine a test version with.
For now, we need sockets as the client is a separate binary, and making it different would make inputs a bit weird

common/nym-lp-data/src/lib.rs
simonwicky: Not much cloning around and lots of transformation, I reckon Bytes wouldn't be that useful

PR #6714, opened 2026-04-28
Feat/nmv3 updated performance calculation

jstuczyn:
NYM-750

## Summary

Wire stress-testing scores submitted by the network-monitor orchestrator into nym-api's
node performance calculation, behind a config gate and an availability guard so an
orchestrator outage cannot silently slash every node's score.

### Performance pipeline
- Added a new `StressTestingScore { score, was_reachable }` carried alongside the existing routing and config scores. `was_reachable` distinguishes a real zero (tested and scored 0) from "no successful sample collected".
- Introduced `NodeAnnotationV2` / `DetailedNodePerformanceV2` (with `From<V2> for V1` for legacy callers). The cache now stores V2 internally; V1 is reconstructed at the API edge.
- Split the `NodePerformanceProvider` trait into `get_node_routing_score` and `get_node_stress_testing_score` (plus batched variants).
- The `LegacyStoragePerformanceProvider` aggregates stress-testing rows from `nym_node_stress_testing_result` over a configurable rolling window. SQL uses `AVG(result)` and `MAX(was_reachable)` (so `was_reachable=false` only when *every* sample for that window was unreachable), with `GROUP BY` + `fetch_optional` so "no rows" surfaces as `None` rather than an all-NULL aggregate row.
- The refresher now blends the new score in via a weighted arithmetic mean (`sw·stress + (1-sw)·routing·config`) — weighted, not multiplicative, so a single zero doesn't collapse the whole score.

### Orchestrator-outage safety guard
The most important behaviour in this branch: if the orchestrator goes down or hasn't submitted enough data yet, stress-testing data is **not** applied at all. Three defences in depth:
1. `NodesStressTestingScores::available_count` only counts nodes with at least one  reachable sample (`Ok(score) && score.was_reachable`). Tested-but-unreachable nodes intentionally don't count, so an orchestrator that ran but produced only failures  is treated the same as one that was down.
2. The refresher gates inclusion on `available_ratio >= minimum_available_stress_testing_results`  (default 0.5). Below the threshold, performance falls back to `routing × config`.
3. Once the threshold passes, individual nodes without data still score 0 in the blended formula — by design, on the assumption the orchestrator tried to test  them and failed. Per-node misses are real signal once we trust the orchestrator.

### API
- New route `GET /v2/nym-nodes/annotation/:node_id` returning the new `AnnotationResponseV2` (with the full `DetailedNodePerformanceV2` payload).

## Configuration notes for operators

- The defaults ship with `use_stress_testing_data = false`. Stress-testing data is collected and stored regardless, but does not influence rewarding until the flag is flipped
- The 0.5 availability threshold means at least 50% of bonded nodes must have a reachable sample in the configured window for the metric to be applied. If the orchestrator misses even one refresh window for too many nodes, the API
  transparently falls back; an `info!` log records the ratio and threshold. This % might have to get adjusted

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6714)


nym-api/src/support/config/mod.rs
```diff
@@ -680,7 +721,7 @@ impl Default for DescribeCacheDebug {
     }
 }
 
-#[derive(Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[derive(Debug, Deserialize, PartialEq, Serialize)]
```
simonwicky, line 724: That change is unnecessary no?

simonwicky (approved):

nym-api/src/support/config/mod.rs
jstuczyn: it is necessary. you can't derive `Eq` with an f32 or f64 inside of it

PR #6717, opened 2026-04-29
feat: Node Families: initial contract storage

jstuczyn:
NYM-1200

## Summary

Bootstraps the `node-families` CosmWasm contract with its full storage layer. No execute/query handlers yet - those will be done in subsequent PRs according to their tickets.

## What's in here

**Storage shape** (`storage/mod.rs`, `storage/storage_indexes.rs`)
- `families: IndexedMap<NodeFamilyId, NodeFamily, _>` with `UniqueIndex`es on `owner` (one-family-per-owner) and on `name` (globally-unique family names).
- `family_members: Map<NodeId, NodeFamilyId>` - current membership lookup.
- `pending_family_invitations`: `IndexedMap` keyed by `(family_id, node_id)` with multi-indexes on `family` and `node`.
- Two append-only archives - `past_family_members` and `past_family_invitations` - keyed by `((family_id, node_id), counter)` with multi-indexes on `family` and `node`. Per-`(family, node)` counters live in their own `Map`s and are bumped via `next_past_member_counter` / `next_past_invitation_counter` to keep archival writes O(1) and disambiguate repeat events that share a block timestamp.
- `node_family_id_counter` issues monotonically increasing ids starting from `1`

**Storage operations** (all `pub(crate)`, callers in transaction layer responsible for auth/validation)
- `register_new_family`
- `add_pending_invitation`
- `accept_invitation` (checks expiry, removes pending, records membership, bumps `members`, archives as `Accepted`)
- `reject_pending_invitation` / `revoke_pending_invitation` (mirror methods, archive as `Rejected` / `Revoked`; work on expired entries)
- `remove_family_member` (shared kick/leave path; archives as `PastFamilyMember`)
- `disband_family` (only on empty families; sweeps remaining pending invitations as `Revoked`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6717)

### DIFF
--- a/common/cosmwasm-smart-contracts/node-families-contract/src/types.rs
+++ b/common/cosmwasm-smart-contracts/node-families-contract/src/types.rs
@@ -1,17 +1,95 @@
 // Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
-use cw_controllers::AdminError;
-use thiserror::Error;
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, Coin};
+use nym_mixnet_contract_common::NodeId;
 
-#[derive(Error, Debug, PartialEq)]
-pub enum NodeFamiliesContractError {
-    #[error("could not perform contract migration: {comment}")]
-    FailedMigration { comment: String },
+/// Identifier of a node family.
+///
+/// Issued sequentially by the contract on family creation; never reused even if the
+/// family is later disbanded.
+pub type NodeFamilyId = u32;
 
-    #[error(transparent)]
-    Admin(#[from] AdminError),
+/// Runtime configuration of the node families contract.
+pub struct Config {
+    /// Fee charged on each successful `create_family` execution.
+    pub create_family_fee: Coin,
+}
 
-    #[error(transparent)]
-    StdErr(#[from] cosmwasm_std::StdError),
+/// On-chain representation of a node family.
+#[cw_serde]
+pub struct NodeFamily {
+    /// The id of the node family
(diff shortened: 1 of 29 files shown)


common/cosmwasm-smart-contracts/node-families-contract/src/types.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// The optional description of the node family
+    pub description: String,
+
+    /// The owner of the node family
+    pub owner: Addr,
+
+    /// Memoized value of the current number of members in the node family
+    /// Used to detect if the family is empty
+    pub members: u64,
```
simonwicky, line 37: nit : I guess that could be `u32`

contracts/node-families/src/storage/mod.rs
```diff
@@ -0,0 +1,1143 @@
...
+    /// sentinel value and must never be assigned to a real family.
+    pub(crate) fn next_family_id(
+        &self,
+        store: &mut dyn Storage,
+    ) -> Result<NodeFamilyId, NodeFamiliesContractError> {
+        let next_id = self
+            .node_family_id_counter
+            .may_load(store)?
+            .unwrap_or_default()
+            + 1;
```
simonwicky, line 134: I don't expect us to arrive to that point, but if it overflows, it will return 0 which is the no family sentinel

simonwicky (approved):

contracts/node-families/src/storage/mod.rs
jstuczyn: if we're ever even remotely close to that we could just switch to u64

PR #6731, opened 2026-05-05
feat: Node Families: contract queries

jstuczyn:
NYM-1201

## Summary

Exposes the read-only query surface for the node-families contract so the Nym API, wallet, and other consumers can read family state without modifying it. Backs every unbounded list (i.e. all families, members, etc.) with a paginated query (exclusive `start_after` cursor + `limit` clamped per query)

Storage layer was already in place on the parent branch (#6717) - this PR is queries-only. `instantiate` / `execute` remain stubs and are out of scope.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6731)

### DIFF
--- a/contracts/node-families/src/queries.rs
+++ b/contracts/node-families/src/queries.rs
@@ -1,13 +1,15 @@
 // Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+use crate::helpers::normalise_family_name;
 use crate::storage::{retrieval_limits, NodeFamiliesStorage};
 use cosmwasm_std::{Deps, Env, Order, StdResult};
 use cw_storage_plus::Bound;
 use node_families_contract_common::{
     AllPastFamilyInvitationsPagedResponse, FamiliesPagedResponse, FamilyMemberRecord,
     FamilyMembersPagedResponse, GlobalPastFamilyInvitationCursor, NodeFamiliesContractError,
-    NodeFamilyId, NodeFamilyMembershipResponse, NodeFamilyResponse, PastFamilyInvitationCursor,
+    NodeFamilyByNameResponse, NodeFamilyByOwnerResponse, NodeFamilyId,
+    NodeFamilyMembershipResponse, NodeFamilyResponse, PastFamilyInvitationCursor,
     PastFamilyInvitationForNodeCursor, PastFamilyInvitationsForNodePagedResponse,
     PastFamilyInvitationsPagedResponse, PastFamilyMemberCursor, PastFamilyMemberForNodeCursor,
     PastFamilyMembersForNodePagedResponse, PastFamilyMembersPagedResponse,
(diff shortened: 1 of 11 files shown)


contracts/node-families/src/queries.rs
```diff
@@ -1,2 +1,2201 @@
...
+    family_id: NodeFamilyId,
+    start_after: Option<PastFamilyInvitationCursor>,
+    limit: Option<u32>,
+) -> Result<PastFamilyInvitationsPagedResponse, NodeFamiliesContractError> {
+    let limit = limit
+        .unwrap_or(retrieval_limits::PAST_INVITATIONS_DEFAULT_LIMIT)
+        .min(retrieval_limits::PAST_INVITATIONS_MAX_LIMIT) as usize;
+
+    let lower =
+        start_after.map(|(node_id, counter)| Bound::exclusive(((family_id, node_id), counter)));
```
simonwicky: Just like `upper` is always the first key of the next family, shouldn't lower be `Bound::inclusive(((family_id, 0), 0))` if `start_after` is `None`?

contracts/node-families/src/queries.rs
```diff
@@ -1,2 +1,2201 @@
...
+    deps: Deps,
+    family_id: NodeFamilyId,
+    start_after: Option<PastFamilyMemberCursor>,
+    limit: Option<u32>,
+) -> Result<PastFamilyMembersPagedResponse, NodeFamiliesContractError> {
+    let limit = limit
+        .unwrap_or(retrieval_limits::PAST_MEMBERS_DEFAULT_LIMIT)
+        .min(retrieval_limits::PAST_MEMBERS_MAX_LIMIT) as usize;
+
+    let lower =
```
simonwicky: Same comment as line 326. 
No need to answer "same response" if there is nothing to do :P

contracts/node-families/src/queries.rs
```diff
@@ -1,2 +1,2201 @@
...
+            let tester = init_contract_tester();
+
+            let res =
+                query_past_invitations_for_family_paged(tester.deps(), 99, None, None).unwrap();
+            assert_eq!(res.family_id, 99);
+            assert!(res.invitations.is_empty());
+        }
+
+        #[test]
+        fn returns_only_archived_invitations_from_requested_family() {
```
simonwicky, line 1408: If my comments up there are correct, then this test does not catch the bug, since you're querying the family with the lowest id (hence the first in the map). 
Regardless, how about making three families and querying the middle one. That will make sure we're not overstepping before AND after

contracts/node-families/src/queries.rs
```diff
@@ -1,2 +1,2201 @@
...
+            let tester = init_contract_tester();
+
+            let res = query_past_members_for_family_paged(tester.deps(), 99, None, None).unwrap();
+            assert_eq!(res.family_id, 99);
+            assert!(res.members.is_empty());
+            assert!(res.start_next_after.is_none());
+        }
+
+        #[test]
+        fn returns_only_archives_from_requested_family() {
```
simonwicky, line 2005: Same comment as the above one

simonwicky (approved):

PR #6774, opened 2026-05-13
feat: Node Families: cache and expose family data within nym API

jstuczyn:
Closes:
[NYM-1203](https://nymtech.atlassian.net/browse/NYM-1203)
[NYM-1204](https://nymtech.atlassian.net/browse/NYM-1204)
[NYM-1205](https://nymtech.atlassian.net/browse/NYM-1205)

## Summary

Adds a cached node-families view to nym-api, exposing read-only HTTP endpoints that combine on-chain families-contract state with mixnet-contract stake/bond data. Refresh is periodic and persistent across restarts.

## Endpoints (all under `/v1/node-families`)

| Method | Path | Returns |
|---|---|---|
| `GET` | `/` | `PaginatedResponse<NodeFamily>` (page/per_page) |
| `GET` | `/{family_id}` | `NodeFamilyResponse { family: Option<NodeFamily> }` |
| `GET` | `/by-node/{node_id}` | `NodeFamilyForNodeResponse { node_id, family: Option<NodeFamily> }` |

Each `NodeFamily` payload includes `id`, `name`, `description`, `owner`, `created_at`, `members` (with per-member stake info), `pending_invitations`, `total_stake`, and a time-weighted `average_node_age`. Lookup endpoints return
a wrapper with `Option<NodeFamily>` instead of 404 so callers can treat "not found" as a normal response.

## Notable internals

- **Cache shape**: `BTreeMap<NodeFamilyId, CachedFamily>` + side index  `HashMap<NodeId, NodeFamilyId>`, giving O(log n) family lookup and O(1) member-to-family lookup. BTreeMap ordering keeps the list endpoint stable across refreshes.
- **Refresher**: pulls families/members/pending invitations from the node-families contract in one tick, joins per-member stake from the mixnet contract cache, and persists to disk via the existing `CacheRefresher` pipeline (bincode dump, rehydrated on startup).
- **Block-timestamp cache**: `HashMap<u64, OffsetDateTime>` stored alongside the families. The refresher parallel-fetches only heights it hasn't seen before, so `average_node_age` becomes pure arithmetic on cached timestamps; survives restarts.
- **Validator-client traits**: `NodeFamiliesQueryClient`, `PagedNodeFamiliesQueryClient`, `NodeFamiliesSigningClient` - mirrors the performance-contract pattern.
- **New contract query**: `GetAllFamilyMembersPaged` (global cross-family scan, used by the refresher).

[NYM-1203]: https://nymtech.atlassian.net/browse/NYM-1203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NYM-1204]: https://nymtech.atlassian.net/browse/NYM-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NYM-1205]: https://nymtech.atlassian.net/browse/NYM-1205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6774)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -473,6 +473,7 @@ nym-noise-keys = { version = "1.20.4", path = "common/nymnoise/keys" }
 nym-nonexhaustive-delayqueue = { version = "1.20.4", path = "common/nonexhaustive-delayqueue" }
 nym-node-requests = { version = "1.20.4", path = "nym-node/nym-node-requests", default-features = false }
 nym-node-metrics = { version = "1.20.4", path = "nym-node/nym-node-metrics" }
+node-families-contract-common = { version = "1.20.5", path = "common/cosmwasm-smart-contracts/node-families-contract" }
 nym-ordered-buffer = { version = "1.20.4", path = "common/socks5/ordered-buffer" }
 nym-outfox = { version = "1.20.4", path = "nym-outfox" }
 nym-registration-common = { version = "1.20.4", path = "common/registration" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -473,7 +473,7 @@ nym-noise-keys = { version = "1.20.4", path = "common/nymnoise/keys" }
 nym-nonexhaustive-delayqueue = { version = "1.20.4", path = "common/nonexhaustive-delayqueue" }
 nym-node-requests = { version = "1.20.4", path = "nym-node/nym-node-requests", default-features = false }
 nym-node-metrics = { version = "1.20.4", path = "nym-node/nym-node-metrics" }
-node-families-contract-common = { version = "1.20.5", path = "common/cosmwasm-smart-contracts/node-families-contract" }
+nym-node-families-contract-common = { version = "1.20.5", path = "common/cosmwasm-smart-contracts/node-families-contract" }
 nym-ordered-buffer = { version = "1.20.4", path = "common/socks5/ordered-buffer" }
 nym-outfox = { version = "1.20.4", path = "nym-outfox" }
 nym-registration-common = { version = "1.20.4", path = "common/registration" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -473,7 +473,7 @@ nym-noise-keys = { version = "1.20.4", path = "common/nymnoise/keys" }
 nym-nonexhaustive-delayqueue = { version = "1.20.4", path = "common/nonexhaustive-delayqueue" }
 nym-node-requests = { version = "1.20.4", path = "nym-node/nym-node-requests", default-features = false }
 nym-node-metrics = { version = "1.20.4", path = "nym-node/nym-node-metrics" }
-nym-node-families-contract-common = { version = "1.20.5", path = "common/cosmwasm-smart-contracts/node-families-contract" }
+nym-node-families-contract-common = { version = "1.20.4", path = "common/cosmwasm-smart-contracts/node-families-contract" }
 nym-ordered-buffer = { version = "1.20.4", path = "common/socks5/ordered-buffer" }
 nym-outfox = { version = "1.20.4", path = "nym-outfox" }
 nym-registration-common = { version = "1.20.4", path = "common/registration" }
--- a/nym-api/nym-api-requests/src/models/node_families.rs
+++ b/nym-api/nym-api-requests/src/models/node_families.rs
@@ -2,49 +2,100 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use cosmwasm_std::Coin;
-use nym_mixnet_contract_common::NodeId;
+use nym_mixnet_contract_common::{NodeId, NodeRewarding};
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 use time::OffsetDateTime;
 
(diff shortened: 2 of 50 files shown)


Cargo.toml
```diff
@@ -473,6 +473,7 @@ nym-noise-keys = { version = "1.20.4", path = "common/nymnoise/keys" }
 nym-nonexhaustive-delayqueue = { version = "1.20.4", path = "common/nonexhaustive-delayqueue" }
 nym-node-requests = { version = "1.20.4", path = "nym-node/nym-node-requests", default-features = false }
 nym-node-metrics = { version = "1.20.4", path = "nym-node/nym-node-metrics" }
+nym-node-families-contract-common = { version = "1.20.5", path = "common/cosmwasm-smart-contracts/node-families-contract" }
```
simonwicky: How are we dealing with versions? Don't we want everything aligned?

nym-api/nym-api-requests/src/models/node_families.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub node_id: NodeId,
+
+    /// Block-time at which the node joined the family.
+    #[serde(with = "time::serde::rfc3339")]
+    #[schema(value_type = String)]
+    pub joined_at: OffsetDateTime,
+
+    /// Stake/bond/delegation snapshot; `None` if the node was not in the
+    /// mixnet-contract cache at refresh time.
+    pub stake_information: Option<NodeStakeInformation>,
```
simonwicky, line 73: This seems a bit out of place, why do we care about that here? 
I mean it's to have it somewhere but why here?

simonwicky: Also, seems like it would nice to just have it on its own without the node being in a family no?

nym-api/src/node_families/cache/refresher.rs
```diff
@@ -0,0 +1,209 @@
...
+            .filter_map(|x| async move { x })
+            .collect()
+            .await;
+
+        block_timestamps.extend(fetched);
+        block_timestamps
+    }
+}
+
+/// Time-weighted average member age: for each member with a known bonding
```
simonwicky: I don't see how this is time-weighted. It's a normal average of the age 
(it shows up in a couple of other places as well I reckon)

nym-api/src/node_families/cache/refresher.rs
```diff
@@ -0,0 +1,209 @@
...
+        let Some(ts) = block_timestamps.get(&height) else {
+            continue;
+        };
+        total_secs += (now - *ts).whole_seconds();
+        count += 1;
+    }
+    if count == 0 {
+        return Duration::ZERO;
+    }
+    Duration::from_secs((total_secs / count).max(0) as u64)
```
simonwicky: That line assume that somehow, you can get an overall negative age, so it's clamped to 0. 
But if only some of the ages are negative but not the overall one, the average will make no sense. 
I would remove that `max(0)` call, but check and discard any age where `whole_seconds()` is `0` instead

nym-api/nym-api-requests/src/models/node_families.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Display name (canonical form — see `normalise_family_name`).
+    pub name: String,
+
+    /// Free-form family description.
+    pub description: String,
+
+    /// Owner address (cosmos `Addr` rendered as a string).
+    pub owner: String,
+
+    /// Time-weighted average age of members.
```
simonwicky: As said in another comment, that's not time-weighted, it's a normal average

nym-api/src/node_families/cache/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Display name (canonical form — see `normalise_family_name`).
+    pub(crate) name: String,
+
+    /// Free-form family description.
+    pub(crate) description: String,
+
+    /// Owner address (cosmos `Addr` rendered as a string).
+    pub(crate) owner: String,
+
+    /// Time-weighted average age of members, computed from cached per-height
```
simonwicky: Same here, not time-weighted

nym-api/src/node_families/cache/refresher.rs
jstuczyn: artifact of previous implementation where it was height-weighted. fixed comments.

nym-api/nym-api-requests/src/models/node_families.rs
jstuczyn: that information is already stored as part of the mixnet contract cache, so should we need it, we could expose it in a dedicated endpoint (or it could be retrieved as part of bonded endpoint)

simonwicky (approved):

PR #6779, opened 2026-05-14
Switch from yarn to pnpm

mmsinclair:
Switch from yarn to pnpm. Move all dependencies into the pnpm workspace catalog.

Replaces https://github.com/nymtech/nym/pull/4683.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6779)


package.json
```diff
@@ -23,8 +24,8 @@
     "lint": "lerna run lint --stream",
     "lint:fix": "lerna run lint:fix --stream",
     "nuke": "npx rimraf **/node_modules node_modules",
```
mfahampshire: wondering if we could have this also nuke all wasm package `pkg/`s  for a squeaky clean repo

package.json
```diff
@@ -23,8 +24,8 @@
...
-    "postbuild:ci": "yarn dev:off",
-    "prebuild:ci": "yarn dev:on && yarn",
+    "postbuild:ci": "pnpm dev:off",
+    "prebuild:ci": "pnpm dev:on && pnpm install",
```
mfahampshire: I'm getting a lot of deprecated dependency warnings on this - not a blocker but just wanted to highlight, suppose updating all of that might be out of scope of this pr though

wasm/client/Makefile
```diff
@@ -1,18 +1,25 @@
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
```
mfahampshire, line 2: nit / random thought: how might we tell we're actually on a Github runner, and not a local dev machine? Since I only introduced taskset to stop wiping out the CI machines

jstuczyn (approved):

package.json
mmsinclair: We are stuck with this for a bit, then we can change it once this is merged, to use local dependency overrides.

wasm/client/Makefile
mmsinclair: we have some Windows and Mac runners, so this is still needed

PR #6784, opened 2026-05-15
Max/smolmix wasm

mfahampshire:
Initial version of a browser-friendly version of `smolmix`, containing a pure-rust version of `mixFetch` and also `mixWebSocket` + `mixResolve` (DNS) features. 
- Leverages all of the experiments we've been running over the past year+ and smushes them into a new WASM swiss-army-knife-thing :) 
- Had to do some feature-gating to get some of the dependencies to compile. 
- Also added in `playwright` headless testing so we can hook this up to the CI (useful format to take forward to smoketest for breaking changes to WASM targets) + I think we can do cool stuff with this in the future, like deploy a modified version of the `internal-dev` UI to a webpage with data on CI runs and timing-averages etc. 
- Very unoptimised build, not sure whether we want to work on that in this PR or in a future one. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6784)

## Summary by CodeRabbit

* **New Features**
  * Added WebAssembly-based browser networking tunnel with support for DNS resolution, HTTPS fetch, and WebSocket connections over the mixnet.

* **Documentation**
  * Added comprehensive guides for the new browser tunnel module, including build instructions, architecture overview, and test suite documentation.

* **Chores**
  * Updated workspace dependencies (async-tungstenite, rustls, hyper, simple-dns, and others).
  * Added build configuration, Makefile automation, and Playwright test infrastructure for the new browser module.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym/pull/6784?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Cargo.toml
```diff
@@ -336,6 +341,8 @@ regex = "1.10.6"
 reqwest = { version = "0.13.1", default-features = false }
 rs_merkle = "1.5.0"
 rustls = { version = "0.23.37", default-features = false }
+rustls-pki-types = "1"
+rustls-rustcrypto = "0.0.2-alpha"
```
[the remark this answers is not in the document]
mfahampshire: For reviewers: from the [readme itself:](https://github.com/RustCrypto/rustls-rustcrypto) 

> Not only that this is incomplete that only few selected TLS suites implemented (it should be well enough to cover 70% of the usage), but the elephant in the room is that neither did rustls nor RustCrypto packages were formally verified and certified with FIPS compliance.

> Note that RustCrypto performance is generally inferior than ring, but in exchange you got a pure Rust implementation that theoretically compiles everywhere Rust was ported to. In our case, we need to have std but foundational support for future no_std expansion is already here.

IMO this is the only option I can see, and I think the CodeRabbit warning is maybe a little OTT but good to flag. In `wasm/smolmix/tls.rs` I'm made sure to restrict the ciphersuites to only what is said to be implemented in the crate readme. 

I think we *could* make ring work though, but it might be a bit messy. Will leave it up to reviewers :)

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,145 @@
...
+                crate::util::debug_error!("[bridge] incoming error: {e}");
+            }
+        }
+    }
+
+    if pushed > 0 {
+        crate::util::debug_log!("[bridge] ▼ rx");
+        let _ = notify_reactor.unbounded_send(());
+    }
+}
```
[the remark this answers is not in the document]
mfahampshire: Tunnel should never panic on parsing, feels like an attack vector

wasm/smolmix/src/dns.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    // Anti-spoof: drop responses whose transaction ID doesn't match the
+    // query we sent. Without this an attacker who guesses our open queries
+    // can inject forged A records.
+    if response.id != query_id {
+        return Err(FetchError::Dns(format!(
+            "DNS response ID mismatch: expected {query_id:#06x}, got {:#06x}",
+            response.id
+        )));
+    }
```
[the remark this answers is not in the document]
mfahampshire: made changes, review

wasm/smolmix/src/bridge.rs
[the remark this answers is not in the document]
jstuczyn: that's an interesting problem to consider in the future - how do we ensure the message we got is actually from the IPR we think we're talking with? what if an attacker discovered you opened an IPR session and started feeding you malicious data?

wasm/smolmix/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    nym_wasm_utils::set_panic_hook();
+}
+
+/// Initialise the mixnet tunnel. See [`SetupOpts`] for the JS-side shape.
+#[wasm_bindgen(js_name = "setupMixTunnel")]
+#[cfg(target_arch = "wasm32")]
+pub fn setup_mix_tunnel(opts: SetupOpts) -> js_sys::Promise {
+    future_to_promise(async move {
+        let result: Result<JsValue, FetchError> = async move {
+            let ipr_address: nym_wasm_client_core::Recipient = opts
```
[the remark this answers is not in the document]
jstuczyn: maybe rather than using `set`, we could use `get_or_init`? unless we want the library consumers to be able to re-initialise the tunnel

wasm/smolmix/src/reactor.rs
```diff
@@ -0,0 +1,106 @@
...
+
+            futures::select! {
+                _ = wasmtimer::tokio::sleep(sleep_for).fuse() => {},
+                _ = notify.next().fuse() => {},
+            }
+
+            // Coalesce: drain queued notifications so one poll() handles all
+            // state changes. Without this, rapid-fire notifications from TLS
+            // writes monopolise the single-threaded WASM event loop.
+            while notify.next().now_or_never().flatten().is_some() {}
```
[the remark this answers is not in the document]
jstuczyn: yep. agreed with mr AI here. probably at this point we should even set the shutdown flag to ensure the rest of the threads also stop

wasm/smolmix/src/reactor.rs
```diff
@@ -0,0 +1,106 @@
...
+/// 2. Ask smoltcp how long it can wait before the next poll (`poll_delay`).
+/// 3. Sleep for that duration (capped at [`MAX_IDLE`]) or until a notification.
+/// 4. Coalesce any further notifications that arrived during the sleep.
+///
+/// Notifications come from the bridge (new rx packets in the device, needing
+/// `iface.poll()` to ingest them) and from socket writes (data queued in
+/// smoltcp's tx buffer, needing `iface.poll()` to dispatch it to the device).
+pub fn start_reactor(
+    stack: Arc<Mutex<SmoltcpStack>>,
+    mut notify: mpsc::UnboundedReceiver<()>,
```
jstuczyn: is there any point of this being an unbonded mpsc receiver? it doesn't appear there's any difference in sending one notification vs 10 of them. how about we just use tokio's native `Notify` instead?

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,370 @@
...
+impl Unpin for WasmTcpStream {}
+
+impl Drop for WasmTcpStream {
+    fn drop(&mut self) {
+        let mut s = self.stack.lock().unwrap();
+        s.sockets
+            .get_mut::<smoltcp_tcp::Socket>(self.handle)
+            .abort();
+        s.sockets.remove(self.handle);
+    }
```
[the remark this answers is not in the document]
jstuczyn: another good observation from our clanker overlord - we might need some state machine here and return Poll::Pending after the initial call to socket.close() until it actually closes

wasm/smolmix/src/tunnel.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        iface.update_ip_addrs(|addrs| {
+            addrs
+                .push(IpCidr::new(IpAddress::from(allocated_ips.ipv4), 32))
+                .unwrap();
+        });
+        iface
+            .routes_mut()
+            .add_default_ipv4_route(Ipv4Address::UNSPECIFIED)
+            .unwrap();
```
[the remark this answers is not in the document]
jstuczyn: given it's simple enough we might as well do it.

Cargo.toml
```diff
@@ -220,6 +221,7 @@ anyhow = "1.0.98"
 arc-swap = "1.7.1"
 argon2 = "0.5.0"
 async-trait = "0.1.88"
+async-tungstenite = { version = "0.24", default-features = false }
```
jstuczyn, line 229: how does it differ from tokio-tungstenite we already use? do we need both of them? could we maybe remove the tokio variant if this one is better?

common/nym-lp/src/lib.rs
```diff
@@ -1,26 +1,43 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
+// Always available (light deps: bytes, num_enum, thiserror, nym-common)
+pub mod packet;
```
jstuczyn: given we ONLY need the packet here, maybe it's a sign we should refactor it to another crate? @simonwicky ?

wasm/smolmix/src/mixdns.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn: nit: you copy-pasted old copyright notices - you might want to do a global search and replace

wasm/smolmix/src/mixdns.rs
```diff
@@ -0,0 +1,22 @@
...
+use wasm_bindgen_futures::future_to_promise;
+
+use crate::error::FetchError;
+use crate::{dns, TUNNEL};
+
+/// Resolve a hostname to an IP address through the mixnet tunnel.
+///
+/// Returns the IP as a string (e.g. `"93.184.216.34"`).
+#[wasm_bindgen(js_name = "mixResolve")]
+pub fn mix_resolve(hostname: String) -> js_sys::Promise {
```
jstuczyn, line 15: just want to say that this is a pretty damn cool feature : )

wasm/smolmix/src/util.rs
```diff
@@ -0,0 +1,38 @@
...
+        .join(" ");
+    if buf.len() > max_bytes {
+        format!("{hex} ...")
+    } else {
+        hex
+    }
+}
+
+/// `console.log` gated behind the `debug` feature (`()` in release).
+macro_rules! debug_log {
```
jstuczyn: didn't we already have those helpers in wasm-common utils? can't we import them instead?

wasm/smolmix/Cargo.toml
```diff
@@ -0,0 +1,76 @@
+[package]
+name = "smolmix-wasm"
+version = "0.0.1"
+authors = { workspace = true }
+edition = "2021"
```
jstuczyn: shouldn't we inherit other things from the workspace too? e.g. edition, rust-version, etc.

wasm/smolmix/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+futures = { workspace = true }
+js-sys = { workspace = true }
+thiserror = { workspace = true }
+url = { workspace = true }
+wasm-bindgen = { workspace = true }
+wasm-bindgen-futures = { workspace = true }
+wasmtimer = { workspace = true }
+
+# DNS wire-format types (tokio-free with default-features = false).
+hickory-proto = { workspace = true, default-features = false, features = ["std"] }
```
jstuczyn: I wonder if maybe different bits of smolmix should be feature-locked. like explicit `dns` feature, explicit `ws` feature, etc. so if somebody only cares about `fetch`, they wouldn't need to compile and ship the rest of the baggage

wasm/smolmix/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+bytes = { workspace = true }
+rand = { workspace = true }
+nym-bin-common = { workspace = true }
+
+# JS interop: typed `SetupOpts` deserialisation.
+serde = { workspace = true, features = ["derive"] }
+serde-wasm-bindgen = { workspace = true }
+tsify = { workspace = true, features = ["js"] }
+
+# crypto.getRandomValues() backend for the 0.2 line.
```
jstuczyn, line 70: I think I know the answer to this question, but can we not get rid of the 0.2 dep?

wasm/smolmix/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+/// `serde-wasm-bindgen` + `tsify`, which gives us typed access without manual
+/// `Reflect::get` plumbing and emits a matching `.d.ts` for the TS side.
+#[derive(Tsify, Deserialize)]
+#[tsify(from_wasm_abi)]
+#[serde(rename_all = "camelCase")]
+#[cfg(target_arch = "wasm32")]
+pub struct SetupOpts {
+    /// Nym address of the IPR exit node (required).
+    pub preferred_ipr: String,
+    /// Client storage namespace; randomise per session for clean state.
```
jstuczyn, line 97: should we also allow clients to configure whether their storage should be encrypted (and thus require a password)?

wasm/smolmix/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    })
+}
+
+/// Disconnect from the mixnet. The tunnel becomes unusable until page reload.
+#[wasm_bindgen(js_name = "disconnectMixTunnel")]
+#[cfg(target_arch = "wasm32")]
+pub fn disconnect_mix_tunnel() -> js_sys::Promise {
+    future_to_promise(async {
+        if let Some(tunnel) = TUNNEL.get() {
+            tunnel.shutdown().await;
```
jstuczyn, line 264: at this point, should we maybe clear the OnceLock so library consumers were able to set it again? or maybe we just retain the existing behaviour of having to reload the page - I don't mind either. just pointing it out.

wasm/smolmix/src/reactor.rs
```diff
@@ -0,0 +1,106 @@
...
+/// TCP retransmit and keepalive timers if `poll_delay` ever returns `None`; on
+/// an active connection the wake source is the bridge or a socket write.
+const MAX_IDLE: Duration = Duration::from_secs(60);
+
+/// Shared smoltcp network stack, accessed by the reactor, bridge, and sockets.
+///
+/// Wrapped in `Arc<Mutex<>>` (not `Rc<RefCell<>>`) so that `WasmTunnel` can
+/// live in a `OnceLock` which requires `Send + Sync`. On wasm32 (single-threaded),
+/// `Mutex` is essentially a no-op lock, zero overhead vs `RefCell`.
+pub struct SmoltcpStack {
```
jstuczyn, line 41: nit: maybe to prevent misuse (and make it easier to pass around) we could use the inner pattern, i.e.
```rust
pub struct SmoltcpStack {
    inner: Arc<Mutex<SmolTcpStackInner>>,
}

struct SmolTcpStackInner {
   iface: Interface,
   sockets: SocketSet<'static>,
   device: WasmDevice,
}
```

wasm/smolmix/src/reactor.rs
```diff
@@ -0,0 +1,106 @@
...
+/// Get the current smoltcp timestamp from `Date.now()`.
+///
+/// smoltcp's `Instant` is just `i64` microseconds, not `std::time::Instant`.
+/// We convert JS milliseconds (f64) to microseconds (i64).
+pub fn smoltcp_now() -> Instant {
+    Instant::from_micros((js_sys::Date::now() * 1000.0) as i64)
+}
+
+/// Type alias for the channel that notifies the reactor to re-poll.
+pub type ReactorNotify = mpsc::UnboundedSender<()>;
```
jstuczyn: as mentioned below, I think this could just be an `Arc<Notify>`

wasm/smolmix/src/reactor.rs
```diff
@@ -0,0 +1,106 @@
...
+                } = *s;
+                iface.poll(now, device, sockets);
+                iface.poll_delay(now, sockets)
+            };
+
+            // Translate smoltcp's deadline into a sleep duration. A zero delay
+            // means "poll again immediately"; treat it as 1 ms so the task
+            // yields back to the JS event loop before re-entering the loop.
+            let sleep_for = match delay {
+                Some(d) if d.total_micros() == 0 => Duration::from_millis(1),
```
jstuczyn: wasn't there some fancy `yield` you discovered the other day we could have used instead? 1ms is an eternity in CPU time : D

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,145 @@
...
+            if shutdown.load(Ordering::Relaxed) {
+                break;
+            }
+
+            // Block until something happens (incoming message or timer tick).
+            // `futures::select!` polls pseudo-randomly when both branches are
+            // ready, so textual order is not a priority guarantee.
+            // TODO: consider `futures::select_biased!` if a sustained timer
+            // burst ever shows up as incoming-side latency.
+            futures::select! {
```
jstuczyn: there's a tiny issue here. if we're stuck waiting for message or tick (and nothing arrives) we might never get the shutdown signal and just hang forever

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            // Drain outgoing packets (capped to avoid starving incoming).
+            let packets: Vec<Vec<u8>> = {
+                let mut s = stack.lock().unwrap();
+                s.device.drain_tx().take(MAX_OUTGOING_PER_TICK).collect()
+            };
+
+            if !packets.is_empty() {
+                crate::util::debug_log!("[bridge] ▲ tx");
```
jstuczyn: since we're already logging we might as well the number of packets being sent

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            for packet in packets {
+                let current_seq = seq.fetch_add(1, Ordering::Relaxed);
```
jstuczyn: since  that sequence number is used for outbound packets does it have to be atomic?

wasm/smolmix/src/tunnel.rs
```diff
@@ -0,0 +1,396 @@
...
+
+        let stack = Arc::new(Mutex::new(SmoltcpStack {
+            iface,
+            sockets: SocketSet::new(Vec::new()),
+            device,
+        }));
+
+        // Bridge starts at seq=1 (ConnectRequest was Data seq=0).
```
jstuczyn: should it start at seq=1 though? I think there was a reason TCP randomises its starting sequence

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            Ok(None) => {}
+            Err(e) => {
+                crate::util::debug_error!("[bridge] incoming error: {e}");
+            }
+        }
+    }
+
+    if pushed > 0 {
+        crate::util::debug_log!("[bridge] ▼ rx");
```
jstuczyn: again, we could log the count

wasm/smolmix/src/ipr.rs
```diff
@@ -0,0 +1,231 @@
...
+use nym_wasm_client_core::Recipient;
+use nym_wasm_client_core::ReconstructedMessage;
+
+use crate::error::FetchError;
+
+/// Reply-SURB counts for Open and Data frames. Defaults: open=5, data=2.
+#[derive(Clone, Copy)]
+pub struct SurbsConfig {
+    pub open: u32,
+    pub data: u32,
```
jstuczyn, line 44: I think it should be more intelligent than that. for example say you're building a mix dns resolver. sending 2 surbs with every request will be an overkill and your recipient will have a huge stack of unused surbs. I think client's background task handles sending extra surbs

wasm/smolmix/src/ipr.rs
```diff
@@ -0,0 +1,231 @@
...
+    let ip_pair = wasmtimer::tokio::timeout(CONNECT_TIMEOUT, async {
+        loop {
+            let batch = receiver
+                .next()
+                .await
+                .ok_or_else(|| FetchError::Tunnel("message channel closed".into()))?;
+
+            for msg in batch {
+                let Some((attrs, content)) = decode_lp_stream(&msg.message) else {
+                    continue;
```
jstuczyn, line 103: if we fail to decode it, shouldn't we bail because we got garbage?

wasm/smolmix/src/device.rs
```diff
@@ -0,0 +1,166 @@
...
+    rx_queue: VecDeque<Vec<u8>>,
+    tx_queue: VecDeque<Vec<u8>>,
+    capabilities: DeviceCapabilities,
+}
+
+impl WasmDevice {
+    pub fn new() -> Self {
+        let mut capabilities = DeviceCapabilities::default();
+        capabilities.medium = Medium::Ip;
+        capabilities.max_transmission_unit = 1500;
```
jstuczyn: wouldn't we actually want to set it to the available size of the sphinx packet payload so that we wouldn't fragment the data?

wasm/smolmix/src/dns.rs
```diff
@@ -0,0 +1,191 @@
...
+use hickory_proto::rr::{Name, RData, RecordType};
+
+use crate::error::FetchError;
+use crate::stream::WasmUdpSocket;
+use crate::tunnel::WasmTunnel;
+
+/// Maximum number of CNAME hops before giving up.
+const MAX_CNAME_HOPS: usize = 8;
+
+const PRIMARY_DNS: SocketAddr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8)), 53);
```
jstuczyn: I feel like users would definitely appreciate if this was configurable. given the audience, some would definitely hate 8.8.8.8 being the default

wasm/smolmix/src/dns.rs
```diff
@@ -0,0 +1,191 @@
...
+        return parse_response(&response, hostname);
+    }
+}
+
+/// Build a DNS query and return its bytes plus transaction ID.
+fn build_query(hostname: &str, record_type: RecordType) -> Result<(Vec<u8>, u16), FetchError> {
+    let mut msg = Message::query();
+    msg.metadata.recursion_desired = true;
+    // Override the auto-generated ID with a CSPRNG value (getrandom/js).
+    let id: u16 = rand::random();
```
jstuczyn: what's the point?

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// TCP stream over the WASM tunnel. Implements `futures::io::{AsyncRead, AsyncWrite}`.
+pub struct WasmTcpStream {
+    pub(crate) stack: Arc<Mutex<SmoltcpStack>>,
+    pub(crate) handle: SocketHandle,
+    pub(crate) notify: ReactorNotify,
+}
+
+/// UDP socket over the WASM tunnel. Used for DNS queries.
+pub struct WasmUdpSocket {
```
jstuczyn, line 52: could it also theoretically allow us to send "normal" UDP traffic that normally is impossible to do from a browser?

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,411 @@
...
+        // Nudge the reactor so any queued tx data dispatches promptly rather
+        // than waiting for the next `poll_delay` deadline.
+        let _ = self.notify.unbounded_send(());
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        let mut s = self.stack.lock().unwrap();
+        let socket = s.sockets.get_mut::<smoltcp_tcp::Socket>(self.handle);
+        socket.close();
```
jstuczyn: as mentioned in the AI thread we might need to hold some small state here to ensure graceful shutdown

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,411 @@
...
+    fn drop(&mut self) {
+        let mut s = self.stack.lock().unwrap();
+        s.sockets.remove(self.handle);
+    }
+}
+
+/// Construct a fresh `Arc<AtomicU16>` ephemeral port counter, seeded at
+/// [`EPHEMERAL_PORT_START`]. Shared via `Arc` so any caller that needs to
+/// allocate a socket can draw from the same range as the tunnel itself.
+pub(crate) fn new_port_counter() -> Arc<AtomicU16> {
```
jstuczyn: shouldn't this be a static then?

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,411 @@
...
+/// collide.
+pub(crate) async fn tcp_connect(
+    stack: Arc<Mutex<SmoltcpStack>>,
+    notify: ReactorNotify,
+    next_port: &Arc<AtomicU16>,
+    addr: SocketAddr,
+) -> io::Result<WasmTcpStream> {
+    let remote = to_smoltcp_endpoint(addr);
+    let local_port = allocate_port(next_port);
+    let tcp_rx = smoltcp_tcp::SocketBuffer::new(vec![0; 65536]);
```
jstuczyn: the smallest nit. if you're already going with u16 max value, you might want to set it to `65535` instead : )

wasm/smolmix/src/stream.rs
```diff
@@ -0,0 +1,411 @@
...
+    let tcp_tx = smoltcp_tcp::SocketBuffer::new(vec![0; 65536]);
+    let mut socket = smoltcp_tcp::Socket::new(tcp_rx, tcp_tx);
+    socket.set_keep_alive(Some(smoltcp::time::Duration::from_millis(10_000)));
```
jstuczyn: i feel it would be nice if all those random timeouts all over the place were configurable (eventually)

wasm/smolmix/src/mixsocket.rs
```diff
@@ -0,0 +1,287 @@
...
+            let conn = fetch::new_connection(tunnel, host, port, is_tls).await?;
+
+            let mut request = url.into_client_request()?;
+            if !protocol_list.is_empty() {
+                let header_value = protocol_list.join(", ").parse().map_err(|e| {
+                    FetchError::Http(format!("invalid Sec-WebSocket-Protocol value: {e}"))
+                })?;
+                request
+                    .headers_mut()
+                    .insert("Sec-WebSocket-Protocol", header_value);
```
jstuczyn, line 65: shouldn't that header only be included if user explicitly asked for wss rather than ws?

wasm/smolmix/src/http.rs
```diff
@@ -0,0 +1,255 @@
...
+    body: Option<&[u8]>,
+) -> Result<(HttpResponse, bool, S), FetchError>
+where
+    S: AsyncRead + AsyncWrite + Unpin + 'static,
+{
+    crate::util::debug_log!("[http] sending {method} request via hyper...");
+
+    // Build the HTTP request
+    let path = match url.query() {
+        Some(q) => format!("{}?{q}", url.path()),
```
jstuczyn: that feels a bit junky : ) surely there must exist a library dealing with that formatting

wasm/smolmix/src/http.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    crate::util::debug_log!(
+        "[http] body complete: {} bytes, reusable={reusable}",
+        body_data.len()
+    );
+
+    // Content preview: text for UTF-8-valid bodies, hex for binary
+    if !body_data.is_empty() {
```
jstuczyn: I think this could be just a needless noise

wasm/smolmix/src/fetch.rs
```diff
@@ -0,0 +1,357 @@
...
+        );
+
+        // 3. Return connection to pool if reusable
+        if reusable {
+            crate::util::debug_log!("[fetch] returning connection to pool for {host}:{port}");
+            tunnel.return_to_pool(host, port, conn);
+        }
+
+        // 4. Follow redirects (3xx with Location header)
+        if (300..400).contains(&response.status) {
```
jstuczyn: nit: just do an early return if the response is not within 300-400 so that we could collapse that if a bit

wasm/smolmix/src/fetch.rs
```diff
@@ -0,0 +1,357 @@
...
+    method: String,
+    headers: Vec<(String, String)>,
+    body: Option<Vec<u8>>,
+}
+
+/// Maximum number of HTTP redirects to follow before giving up.
+const MAX_REDIRECTS: u8 = 5;
+
+/// Execute a fetch request through the mixnet tunnel.
+pub async fn fetch(
```
jstuczyn, line 37: should we also imitate other bits of the spec like cors and whatnot?

jstuczyn (commented):
I wish I had a bit more time to go through it more thoroughly, but in general I think this is in a pretty good shape. I haven't gone super in-depth for detailed logic checks - I assume our AI friends have dealt with it. speaking of which, I've left some comments in the existing conversations. we might have to fix some additional things there

also I think it might be worth revisiting all the loops where we get data from IPR - should we continue on errors or just bail? we should have some way of protecting ourselves from garbage

Also I think it would be nice if there was some equivalent of a task tracker that makes sure all tasks required by the tunnel are still healthy because if one of them fails, the rest becomes useless

wasm/smolmix/Cargo.toml
mfahampshire: Yeah this is a good point - then when we turn it into typescript SDK packages those can just be pulling in the relevant feature stuff.

wasm/smolmix/src/ipr.rs
mfahampshire: I thought we sent a set amount of surbs with outgoing requests because the client's background task only dealt with handling extra surbs once the receiver's surb bucket was empty; wanted to reduce the # of round trips for sending a request for more surbs, then the sender sending them back to the receiver to use. Maybe I'm working on a misconception though.

wasm/smolmix/src/device.rs
mfahampshire: I did this for a reason but I can't remember what it was. Will double check my notes, and if I can't find it will do that and see how it affects behaviour

wasm/smolmix/src/dns.rs
mfahampshire: Well caught, I meant to make this configurable.

wasm/smolmix/src/stream.rs
mfahampshire: Yes - I don't see a use for it off the bat, but yes

wasm/smolmix/src/stream.rs
mfahampshire: Yeah, some kinda builder makes sense

wasm/smolmix/src/bridge.rs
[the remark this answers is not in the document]
mfahampshire: I can make a ticket for this, think it deserves a little spike to look into / is out of scope for this PR?

wasm/smolmix/src/lib.rs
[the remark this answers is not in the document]
mfahampshire: Gone with having an err if the tunnel is already initialised, re-init after disconnect looks like it might be kinda out of scope for this without a bunch of refactoring + CR's suggestion.

wasm/smolmix/Cargo.toml
mfahampshire: Good catch - one thing I'm trying to think through is whether we want to do `version = { workspace = true}` and then include it in the workspace crates / publication flow. That would stop stuff breaking if we make a breaking change to some dependency / change e.g. an API schema somewhere that isn't backwards compatible, but means we have to then do a release of the TS SDK modules that will rely on this (`mixResolve`, `mixFetch`, `mixSocket`) every release as well. Only downside of this, aside from adding more to the release process, is that we'd also be putting this on crates.io, unless I make an exception in the version-bump CI script to bump this crate even though it would have `Publish = false` in the crate `Cargo.toml` file.

wasm/smolmix/src/lib.rs
mfahampshire: For the moment I think we retain the behaviour of having to reset the page / reload the worker, but I could make a ticket for a future version to make it so you can disconnect it and then create a new tunnel? This PR is already pretty big / old, would rather avoid it rotting more and having to merge `develop` again if possible.

wasm/smolmix/Cargo.toml
mfahampshire: Did the feature-gating

wasm/smolmix/Cargo.toml
mfahampshire: Some of our deps pull it in - I'd be tempted to see if we can change it in a separate PR?

wasm/smolmix/src/reactor.rs
mfahampshire: Done

wasm/smolmix/src/ipr.rs
mfahampshire: One thought here, which relates to the IPR auth comment; if we just bail on a malformed msg *without* having the IPR auth functionality, wouldn't that open up a DoS possibility? For the moment adding a debug log so we can see when we get them, but not bail.

wasm/smolmix/src/reactor.rs
mfahampshire: Done

common/nym-lp/src/lib.rs
mfahampshire: Was reworked by Simon into a new crate anyway

wasm/smolmix/src/fetch.rs
mfahampshire: I think we could do some rubber ducking and decide what to imitate then do in another PR? Would be good to get others' opinions as well.

wasm/smolmix/src/dns.rs
mfahampshire: Was in there based on my misconception about how the autogenerated IDs were randomised - removed

Cargo.toml
mfahampshire: I used it because (iirc) if I used tokio-tungstenite I would have had to use a shim to get it to talk to all the `futures::io` stuff in here. 

Interestingly it does have tokio-flavoured features (https://github.com/sdroege/async-tungstenite/blob/e978c4006b2b06bef3f0519b9ec07a286db98682/Cargo.toml#L17) so we could maybe look into replacing the tokio variant, yeah.

wasm/smolmix/Cargo.toml
jstuczyn: I can imagine this crate would have a different release cadence to the other bits of the repo so perhaps leaving `version` independent is the right call

wasm/smolmix/Cargo.toml
jstuczyn: let's try it in the separate PR then

wasm/smolmix/src/ipr.rs
jstuczyn: > once the receiver's surb bucket was empty

I'm like 80% sure the receiver will requests more surbs pre-emptively once it reaches a lower threshold. I guess there's an easy way to check it?

common/wasm/utils/src/lib.rs
```diff
@@ -41,6 +41,44 @@ macro_rules! console_error {
...
 }
 
+/// `console.log` gated behind calling crate's `debug` feature flag.
+///
+/// Compiles to `()` when the caller's `debug` feature is off, so format-args
+/// allocation is elided in release builds.
+#[macro_export]
+macro_rules! debug_log {
+    ($($t:tt)*) => {{
+        #[cfg(feature = "debug")]
```
jstuczyn: that might backfire in weird ways because this relies on consuming library to have a `debug` feature which would be specific to wasm.

just thinking aloud here. what if there was a static `DEBUG_ENABLED` in this crate that would control those logs (and would be enabled by consumers on THEIR features being activated?)

pnpm-workspace.yaml
```diff
@@ -8,22 +8,17 @@ packages:
...
-  fsevents: true
-  lefthook: true
-  lmdb: true
-  msgpackr-extract: true
-  nx: true
-  protobufjs: true
-  sharp: true
-  tiny-secp256k1: true
-  unrs-resolver: true
+  '@biomejs/biome': set this to true or false
```
jstuczyn: is this something we have to set?

service-providers/ip-packet-router/Cargo.toml
```diff
@@ -25,8 +25,8 @@ nym-config = { workspace = true }
 nym-crypto = { workspace = true }
 nym-exit-policy = { workspace = true }
 nym-id = { workspace = true }
-nym-ip-packet-requests = { workspace = true }
-nym-lp-data = { workspace = true }
+nym-ip-packet-requests = { workspace = true, default-features = true  }
+nym-lp-data = { workspace = true, default-features = true  }
```
jstuczyn: since those are the default-features, do we have to explicitly enable them?

wasm/smolmix/src/device.rs
```diff
@@ -28,7 +28,31 @@ impl WasmDevice {
...
-        capabilities.max_transmission_unit = 1500;
+        // Sized so one IP packet fits in one sphinx packet payload, no LP
+        // fragmentation.
+        //
+        // ┌─────────────────────────────────────────────────────────────────────┐
+        // │ Sphinx packet (2352 B total)                                        │
+        // │ ┌─────────────────────────────────────────────────────────────────┐ │
+        // │ │ Sphinx payload (2048 B usable)                                  │ │
+        // │ │ ┌──────────────────────┬──────────────────────────────────────┐ │ │
+        // │ │ │ LP OuterHeader (12B) │ Encrypted LP blob (2036 B)           │ │ │
```
jstuczyn: that's untrue, there's also other non-LP overhead in this current revision of the packet format. for example the surb-ack takes some considerable amount of space in the payload

common/wasm/utils/src/lib.rs
mfahampshire: Yeah true... that's more robust than relying on the build flags. Good point.

wasm/smolmix/src/device.rs
mfahampshire: :facepalm: will sort

common/wasm/utils/src/lib.rs
mfahampshire: This is a good catch actually, I forgot to do this after we had that community PR on mixfetchV1 asking for the same thing

wasm/smolmix/src/bridge.rs
```diff
@@ -0,0 +1,152 @@
...
+    data_surbs: u32,
+) {
+    // Cloned so finalise_task can check is_cancelled() on the way out.
+    let token = tracker.clone_shutdown_token();
+    tracker.try_spawn_named_with_shutdown(
+        async move {
+            let mut tx_interval = wasmtimer::tokio::interval(Duration::from_millis(5));
+            // Outbound seq starts at 1; seq=0 is reserved for handshake
+            // frames (see `ipr::open_and_connect`).
+            let mut seq: u32 = 1;
```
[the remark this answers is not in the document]
mfahampshire: Fixing

wasm/smolmix/Cargo.toml
mfahampshire: tend to agree after mulling on it

wasm/smolmix/Cargo.toml
mfahampshire: cool, will make a ticket

wasm/smolmix/src/ipr.rs
mfahampshire: *thinking how we might make this more configurable* - will also double check re the SURBs. From what I remember of writing the IpMixStream module though they're static, so touching this might be a bit complicated. Will report back.

mfahampshire: With the current setup, esp. in the case of something like the dns resolver, you could always just set 0 'data' surbs, and send a bunch upfront, which works? Saves having to do a rewrite, and is more of a 'here are some sane defaults' in documentation?

pnpm-workspace.yaml
mfahampshire: I fixed this in develop earlier today, pulling in fix from there

wasm/smolmix/src/device.rs
mfahampshire: Ok I think my maths on this should be correct now..

service-providers/ip-packet-router/Cargo.toml
mfahampshire: I think these were artefacts from a while ago in this PR when I was having to do more on feature-gating dependencies, removed

jstuczyn (approved):

PR #6815, opened 2026-05-22
feat: expose node's chain address on self-described API

jstuczyn:
## Summary

This PR exposes nym-nodes on-chain address on its self-described endpoint so that it could be queried by nym-api in order to determine whether the address has any tokens (or a feegrant) and thus could send transactions to the chain

## Why a v2 endpoint

The v1 `auxiliary-details` payload can be served over bincode, whose wire format is positional and ignores `#[serde(default)]`. Inserting a new field into `AuxiliaryDetails` would silently break every bincode-deserialising client. A new `AuxiliaryDetailsV2` served from a v2 endpoint that supports JSON/YAML only (`OutputParamsV2`) lets the schema evolve without breaking existing consumers.

## Changes

- `nym-node-requests`: new `api::v2::node::models::AuxiliaryDetailsV2` (adds `address: String`); existing type renamed `AuxiliaryDetails` → `AuxiliaryDetailsV1`; `From<V2> for V1` drops the new field so the v1 endpoint still returns the original shape.
- `nym-node`: address is derived from the gateway mnemonic at startup and served via a new `/api/v2/auxiliary-details` handler. Static node data (build info, roles, description, etc.) consolidated into `AppState::static_information` and pulled in by handlers via `State(...)` instead of closure capture.
- `nym-node` http: single unified `ApiDoc` covering v1 + v2, mounted at `/api/swagger` (was `/api/v1/swagger`). 307 redirect preserved at the old v1 path for back-compat. Swagger UI tags prefixed with `v1 / ...` / `v2 / ...` so groups don't collide.

## Test plan

- [ ] Hit `/api/v1/auxiliary-details` on an upgraded node - shape unchanged.
- [ ] Hit `/api/v2/auxiliary-details` - `address` matches the on-chain account derived from the node's mnemonic.
- [ ] Load `/api/swagger`, confirm both v1 and v2 groups are present and "Try it out" on each routes to the correct path.
- [ ] Load `/api/v1/swagger/` - redirects to `/api/swagger/`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6815)

## Summary by CodeRabbit

* **New Features**
  * Introduced API v2 with a new /auxiliary-details endpoint and expanded auxiliary details (includes on-chain address and additional metadata).
  * Added shared‑wallet signing support for multi‑instance wallet use.

* **Documentation**
  * Versioned OpenAPI/Swagger tags, improved Swagger UI routing, and landing page now links to the version‑neutral API docs.

* **Refactoring**
  * Simplified HTTP server configuration, centralized static node state, and reworked API routing to serve v1 and v2 side‑by‑side.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym/pull/6815?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/common/client-libs/validator-client/src/signing/direct_wallet.rs
+++ b/common/client-libs/validator-client/src/signing/direct_wallet.rs
@@ -11,6 +11,8 @@ use cosmrs::tx;
 use cosmrs::tx::SignDoc;
 use nym_config::defaults;
 use std::borrow::Cow;
+use std::ops::Deref;
+use std::sync::Arc;
 use thiserror::Error;
 use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
@@ -45,6 +47,22 @@ pub struct DirectSecp256k1HdWallet {
     #[zeroize(skip)]
     // unfortunately `dyn EcdsaSigner` does not guarantee Zeroize
     accounts: Vec<AccountData>,
+}
+
+impl OfflineSigner for Arc<DirectSecp256k1HdWallet> {
+    type Error = DirectSecp256k1HdWalletError;
+
+    fn get_accounts(&self) -> &[AccountData] {
+        self.deref().get_accounts()
+    }
+
+    fn sign_direct_with_account(
+        &self,
+        signer: &AccountData,
+        sign_doc: SignDoc,
+    ) -> Result<tx::Raw, Self::Error> {
+        self.deref().sign_direct_with_account(signer, sign_doc)
+    }
 }
 
 impl OfflineSigner for DirectSecp256k1HdWallet {
--- a/common/client-libs/validator-client/src/signing/direct_wallet.rs
+++ b/common/client-libs/validator-client/src/signing/direct_wallet.rs
@@ -49,22 +49,6 @@ pub struct DirectSecp256k1HdWallet {
     accounts: Vec<AccountData>,
 }
 
-impl OfflineSigner for Arc<DirectSecp256k1HdWallet> {
-    type Error = DirectSecp256k1HdWalletError;
-
-    fn get_accounts(&self) -> &[AccountData] {
-        self.deref().get_accounts()
-    }
-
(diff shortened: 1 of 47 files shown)


nym-api/src/support/http/openapi.rs
```diff
@@ -41,7 +41,7 @@ use utoipauto::utoipauto;
         nym_config::defaults::NymContracts,
         ContractVersionSchemaResponse,
         nym_bin_common::build_information::BinaryBuildInformationOwned,
-        nym_node_requests::api::v1::node::models::AuxiliaryDetails,
+        nym_node_requests::api::v1::node::models::AuxiliaryDetailsV1,
```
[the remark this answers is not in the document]
jstuczyn: sure, but it's not yet consumed by nym-api.

common/client-libs/validator-client/src/signing/direct_wallet.rs
```diff
@@ -47,6 +49,22 @@ pub struct DirectSecp256k1HdWallet {
     accounts: Vec<AccountData>,
 }
 
+impl OfflineSigner for Arc<DirectSecp256k1HdWallet> {
```
simonwicky: Shower though : While you're at it, why not making it a blanket impl with something like `impl OfflineSigner for Arc<T> where T : OfflineSigner`

simonwicky: Actually, it looks like it still compiles without it, do we need it?

simonwicky (approved):

common/client-libs/validator-client/src/signing/direct_wallet.rs
jstuczyn: good catch. yes, let me remove it. it's an artifact from an earlier refactor attempt to try to create a client instance with clone-able wallet that would not be cloning the underlying keys

PR #6834, opened 2026-05-27
feat: implement UpdateFamily for the node families contract

jstuczyn:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6834)

## Summary by CodeRabbit

* **New Features**
  * Family owners can now update their family's name and/or description after creation. Updates are independent—each field is optional. Validation rules match those used during family creation. Update events are emitted only for fields that actually changed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nymtech/nym/pull/6834?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contracts/node-families/src/helpers.rs
+++ b/contracts/node-families/src/helpers.rs
@@ -4,7 +4,7 @@
 use crate::storage::NodeFamiliesStorage;
 use cosmwasm_std::{Addr, Deps};
 use nym_mixnet_contract_common::{MixnetContractQuerier, NodeId};
-use nym_node_families_contract_common::NodeFamiliesContractError;
+use nym_node_families_contract_common::{Config, NodeFamiliesContractError, NodeFamilyId};
 
 /// Normalise a family name into the canonical form used as the unique-index key.
 ///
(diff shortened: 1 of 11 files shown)


contracts/node-families/src/helpers.rs
```diff
@@ -18,6 +18,80 @@ pub fn normalise_family_name(name: &str) -> String {
         .collect()
 }
 
+/// A new `(display, normalised)` family name pair that has passed
+/// length and non-empty-after-normalisation validation. Constructed by
+/// [`validate_family_name`]; consumed by storage write paths that update
+/// the family's `name` + `normalised_name` columns together.
+pub(crate) struct NewFamilyName {
```
simonwicky, line 25: Shouldn't this be in `common/types` or something like that?

simonwicky (approved):

contracts/node-families/src/helpers.rs
jstuczyn: no because it's not exposed outside the contract

PR #6846, opened 2026-06-03
Smolmix RTT storm fix 

mfahampshire:
Note this branch relies on the [Nym fork](https://github.com/nymtech/smoltcp/tree/rtt-patch) of `smoltcp` to quickly unblock the TS SDK stuff - need to experiment with whether native `smolmix` has the same issue (probably does), which is a larger crates juggling issue I'm leaving for another day (TL;DR - we're not publishing `smolmix-wasm` on crates.io, so having a `git` dependency import doesn't matter, but it would block native `smolmix`; I'd rather not have to publish 2 forked crates though (`smoltcp` and its `tokio` bindings) so not sure how to proceed unless we reach out to the `smoltcp` maintainers). 

The core idea of this branch / its fix: we were accidentally spamming endpoints with retransmitted SYN packets because of the default `smoltcp` RTT assumptions, which was leading to flaky handshakes. Toned this down via the fork + added a BUNCH of more detailed logging which is helpful for seeing how our connections are moving through `smolmix`. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6846)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection reliability with automatic retry logic (3 attempts) and exponential backoff for failed TCP connections.

* **Chores**
  * Updated dependency configuration to use a patched version for better compatibility.
  * Adjusted network device MTU from 1600 to 1500 bytes for standards compliance.

* **Refactor**
  * Enhanced default protocol configuration for improved return path capacity during downloads.

### DIFF
--- a/wasm/smolmix/src/device.rs
+++ b/wasm/smolmix/src/device.rs
@@ -29,11 +29,9 @@ impl WasmDevice {
     pub fn new() -> Self {
         let mut capabilities = DeviceCapabilities::default();
         capabilities.medium = Medium::Ip;
-        // Match the standard Ethernet MTU (1500). As well as giving us room
-        // for SURB headers w/out fragmenting our payload, most remote hosts
-        // and the path to them are tuned around 1500, so emitting 1500 B packets
-        // keeps us within what the wider internet expects and sidesteps path-MTU
-        // surprises upstream.
+        // Match the standard Ethernet MTU (1500). 1500 keeps a
+        // packet (plus its LP/IPR framing and SURB headers) within a single
+        // Sphinx packet's payload, avoiding LP-layer fragmentation.
         capabilities.max_transmission_unit = 1500;
         // Native smolmix also uses Some(1) in the device, but tokio-smoltcp
         // compensates with a burst loop that calls Interface::poll() up to 100
--- a/wasm/smolmix/src/device.rs
+++ b/wasm/smolmix/src/device.rs
@@ -29,17 +29,11 @@ impl WasmDevice {
     pub fn new() -> Self {
         let mut capabilities = DeviceCapabilities::default();
         capabilities.medium = Medium::Ip;
-        // Match the standard Ethernet MTU (1500). Two independent reasons
-        // both point at the same value:
-        //   1. One IP packet must fit in one sphinx packet payload, else the
-        //      LP layer fragments it. The usable budget from the 2048 B
-        //      sphinx plaintext, after SURB-ack (344), x25519 ephemeral key
-        //      (32), frag header (7), padding (1) and LP+IPR framing + AEAD
-        //      (53), is ~1611 B. 1500 sits inside that with ~110 B headroom
-        //      for framing variability.
-        //   2. Remote hosts and the path to them are tuned around 1500, so
-        //      emitting 1500 B packets keeps us within what the wider
-        //      internet expects and sidesteps path-MTU surprises upstream.
+        // Match the standard Ethernet MTU (1500). As well as giving us room
+        // for SURB headers w/out fragmenting our payload, most remote hosts
+        // and the path to them are tuned around 1500, so emitting 1500 B packets
+        // keeps us within what the wider internet expects and sidesteps path-MTU
+        // surprises upstream.
         capabilities.max_transmission_unit = 1500;
         // Native smolmix also uses Some(1) in the device, but tokio-smoltcp
         // compensates with a burst loop that calls Interface::poll() up to 100
--- a/wasm/smolmix/src/device.rs
+++ b/wasm/smolmix/src/device.rs
@@ -8,11 +8,12 @@
 //! `Device` impl instead. The bridge pushes incoming IP packets into `rx_queue`
 //! and pops outgoing packets from `tx_queue`.
 
-use std::collections::VecDeque;
+use std::collections::{BTreeSet, VecDeque};
 
 use smoltcp::phy::{Device, DeviceCapabilities, Medium, RxToken, TxToken};
 use smoltcp::time::Instant;
-use smoltcp::wire::{IpAddress, IpProtocol, Ipv4Packet, TcpPacket};
+use smoltcp::wire::{IpAddress, IpProtocol, Ipv4Address, Ipv4Packet, TcpPacket};
+use std::collections::BTreeMap;
 
 /// smoltcp device backed by in-memory packet queues.
 ///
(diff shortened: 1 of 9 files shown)


mfahampshire:
> 🧹 Nitpick comments (1)
> 
> > wasm/smolmix/Cargo.toml (1)> `52-56`: _⚡ Quick win_
> > > **Make `smoltcp` pin deterministic despite tag drift**
> > > `Cargo.lock` already resolves the git `tag = "v0.12.0-nym-rtt"` to a specific commit (`...#62ac5b8b3287d4773694f19a3b55e4c004354a0b`), and this git-sourced `smoltcp` appears only in `wasm/smolmix/Cargo.toml`. Adding `rev = "<commit>"` alongside the tag would still make `Cargo.toml` deterministic even if the lockfile is regenerated later.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@wasm/smolmix/Cargo.toml` around lines 52 - 56, The smoltcp dependency in
> > > wasm/smolmix/Cargo.toml is pinned only by tag ("v0.12.0-nym-rtt") which can
> > > drift; update the smoltcp entry to include the exact commit hash by adding rev =
> > > "<commit>" (the commit resolved in Cargo.lock, e.g.,
> > > 62ac5b8b3287d4773694f19a3b55e4c004354a0b) alongside tag so the dependency
> > > declaration for smoltcp is deterministic and matches Cargo.lock even if the
> > > lockfile is regenerated.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@wasm/smolmix/Cargo.toml`:
> - Around line 52-56: The smoltcp dependency in wasm/smolmix/Cargo.toml is pinned
> only by tag ("v0.12.0-nym-rtt") which can drift; update the smoltcp entry to
> include the exact commit hash by adding rev = "<commit>" (the commit resolved in
> Cargo.lock, e.g., 62ac5b8b3287d4773694f19a3b55e4c004354a0b) alongside tag so the
> dependency declaration for smoltcp is deterministic and matches Cargo.lock even
> if the lockfile is regenerated.
> ```
> 
> ℹ️ Review info

Done, rereview

wasm/smolmix/src/device.rs
```diff
@@ -95,10 +93,112 @@ impl RxToken for WasmRxToken {
     where
         F: FnOnce(&[u8]) -> R,
     {
+        // Diagnostic only: the smoltcp receive path is invisible from outside
+        // (it validates, reassembles and drops silently). When runtime debug
+        // logging is on, log every inbound TCP packet's seq / length / flags /
+        // checksum so a stalled inbound flight can be read straight from the
+        // log. The `f()` call below is unchanged; smoltcp owns all processing.
+        if nym_wasm_utils::debug_logging_enabled() {
+            log_inbound_tcp(&self.buffer);
```
jstuczyn: are we sure this wouldn't spam the console beyond belief given it's done once per TCP packet? perhaps we'd want to  have some aggregate instead?

wasm/smolmix/src/device.rs
```diff
@@ -28,13 +29,10 @@ impl WasmDevice {
...
-        // chunking-layer fragmentation). Budget in bytes from the 2048 B
-        // sphinx plaintext: − 344 (SURB-ack) − 32 (x25519 ephemeral key,
-        // Repliable msgs) − 7 (frag header) − 1 (padding) − 53 (LP+IPR
-        // framing + AEAD) ≈ 1611. 1600 leaves ~11 B headroom for IPR
-        // overhead variability.
-        capabilities.max_transmission_unit = 1600;
+        // Match the standard Ethernet MTU (1500). 1500 keeps a
+        // packet (plus its LP/IPR framing and SURB headers) within a single
+        // Sphinx packet's payload, avoiding LP-layer fragmentation.
+        capabilities.max_transmission_unit = 1500;
```
jstuczyn, line 38: what's the reason for the decrease?

wasm/smolmix/src/fetch.rs
```diff
@@ -191,16 +191,61 @@ pub async fn fetch(
...
     is_https: bool,
 ) -> Result<PooledConn, FetchError> {
+    // Resolve once: a DNS failure is not transient the way a lost handshake
+    // segment is, and re-resolving per attempt would just repeat the lookup.
     let ip = dns::resolve(tunnel, host).await?;
     let addr = SocketAddr::new(ip, port);
 
+    let mut last_err = None;
+    for attempt in 1..=CONNECT_ATTEMPTS {
+        match connect_once(tunnel, addr, host, is_https).await {
```
jstuczyn, line 229: I think we might want to have some backoff between subsequent retries

jstuczyn (approved):

wasm/smolmix/src/device.rs
mfahampshire: It does spam it but its been helpful for finding this problem - I could do an aggregate though, or something like that. Or alternatively make a `trace` style debug which this would be?

wasm/smolmix/src/device.rs
mfahampshire: Two things, both kinda 'hunches' but until I build a proper test harness to test this + some other IPR stuff I've discussed with @mmsinclair I can't validate beyond 'the connection is more stable when I repeatedly run the internal-dev tool with different IPRs'. 

- I was worried about fragmentation beyond the IPR, path discovery, etc, and wanted to match standard Ethernet MTU to make the packets as 'normal' as I could. 
- I wanted to leave more room for SURBs w/out fragmentation of the IP packet payload.

wasm/smolmix/src/fetch.rs
mfahampshire: I was thinking about this, but my reasoning for not was that you're only doing a few connection attempts, and from a UX perspective its better to retry quickly, rather than have someone go 'oh something is wrong' and reload the page/extension/app. However, if you think that logic is spotty happy to implement a backoff, just explaining what my thinking was here (rather than not having thought about it)

wasm/smolmix/src/device.rs
mfahampshire: Actually, aggregate will give us what we need from this when debugging imo, so going to go with that.

wasm/smolmix/src/fetch.rs
mfahampshire: Going for smol backoff as discussed in DM

PR #6856, opened 2026-06-05
Keep peer in wg table when updating psk

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6856)

### DIFF
--- a/common/wireguard/src/peer_controller/mod.rs
+++ b/common/wireguard/src/peer_controller/mod.rs
@@ -76,6 +76,12 @@ pub enum PeerControlRequest {
         peer: Peer,
         response_tx: oneshot::Sender<AddPeerControlResponse>,
     },
+    /// Update PSK for an existing peer, without changing its IP allocation
+    UpdatePeerPsk {
+        peer_key: Key,
+        psk: Key,
+        response_tx: oneshot::Sender<UpdatePeerPskControlResponse>,
+    },
     /// Attempt to allocate an IP pair from the pool
     PreAllocateIpPair {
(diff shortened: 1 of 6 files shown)


common/wireguard/src/peer_controller/mod.rs
```diff
@@ -317,6 +324,34 @@ impl PeerController {
...
+
+        // Try to update WireGuard peer
+        if let Err(e) = self.wg_api.configure_peer(&peer) {
+            nym_metrics::inc!("wg_peer_update_psk_failed");
+            nym_metrics::inc!("wg_config_errors_total");
+            return Err(e.into());
+        };
+
+        // try to immediately update the host information, to eliminate races
+        if let Ok(host_information) = self.wg_api.read_interface_data() {
```
jstuczyn: shouldn't we get the lock before trying to configure the peer to avoid the races?

neacsu: From what I remember, the `host_information` is just trying to get the most up-to-date information about what's in the kernel table, to be able to reflect that in the userspace. `read_interface_data` would get that latest information, being after a `configure_peer` call, so to prevent the race would mean to get it before calling `read_interface_data`. But being an external and potentially big sys call, I would be in favour of a possibility of race rather than a possibility of deadlock

jstuczyn: looking at the code in more detail, I no longer think this is an issue

jstuczyn (approved):

PR #6861, opened 2026-06-05
Bugfix: set connection fd callback on LP registration socket

rachyandco:
The LP registration client opened its TCP connection to the gateway
  control port with a plain TcpStream::connect, without invoking the
  connection_fd_callback used by the websocket registration path. On
  Linux the callback sets SO_MARK (fwmark) so the daemon's own firewall
  allows the connection during the connecting state; without it the
  unmarked SYN matched no allow rule and was rejected locally with
  ECONNREFUSED. The daemon blamed the gateway, blacklisted it, and
  cycled through every gateway the same way, making it impossible to
  connect when LP registration is enabled.

  Plumb the existing connection_fd_callback from the builder config
  through RegistrationClientConfig into the LP client via an optional
  dialer, which creates the socket, applies the callback before
  connect(), and only then dials. The fd callback must run before
  connect so the SYN itself carries the mark.

  The exit gateway needs no dialer: its registration is forwarded
  through the entry gateway's nested session and never opens its own
  TCP connection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6861)

### DIFF
--- a/nym-registration-client/src/builder/config.rs
+++ b/nym-registration-client/src/builder/config.rs
@@ -97,6 +97,8 @@ impl BuilderConfig {
             exit: self.exit_node.clone(),
             mode: self.mode,
             lp_registration_config: self.lp_registration_config,
+            #[cfg(unix)]
+            connection_fd_callback: self.connection_fd_callback.clone(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


pronebird (approved):

pronebird:
It seems that we solved that in different way in https://github.com/nymtech/nym-vpn-client/pull/5516

PR #6878, opened 2026-06-15
Safe account removal in password wallets and more reliable network switching

tommyv1987:
- Users with a password-protected wallet and more than one saved account can remove an individual account from the accounts list, with clear warnings and a password confirmation step.
- Removal is blocked when it would be unsafe: you cannot remove the account you are currently using (switch first), and you cannot remove your only saved account from this screen (full wallet reset still requires signing out and removing the wallet file after a backup).
- Before removal, the app shows where the wallet file lives on disk and asks you to confirm you have backed up, so accidental data loss is less likely.
- The wallet backend now checks these rules before deleting anything, and shows clearer error messages if something goes wrong.

Bonus fix 🕺 
Fixed a bug where changing network (e.g. mainnet to sandbox) while logged in could leave the app on the wrong network or show stale balances

## Release Notes

* **New Features**
  * Added account deletion UI with safeguards to prevent removing the last stored account or the currently active account.
  * Added a full “Delete Account” flow with backup confirmation, backup reminder, and password confirmation.
  * Added wallet storage location lookup to power the backup reminder during deletion.

* **Changes**
  * Improved user-facing account removal error messages and delete-flow recovery behavior.
  * Refined account removal handling to reload accounts after successful deletion.
  * Improved network switching refresh/commit behavior and simplified the Network Selector to Mainnet and Sandbox.

* **Tests**
  * Added/expanded coverage for deletion, error mapping, and network-switch logic.

### DIFF
--- a/nym-wallet/src-tauri/src/error.rs
+++ b/nym-wallet/src-tauri/src/error.rs
@@ -110,6 +110,12 @@ pub enum BackendError {
     WalletDifferentPasswordDetected,
     #[error("Unexpected mnemonic account for login")]
     WalletUnexpectedMnemonicAccount,
+    #[error(
+        "Cannot remove the only stored account. Back up and remove the wallet file to reset entirely."
+    )]
+    WalletCannotRemoveLastAccount,
+    #[error("Switch to another account before removing the active account")]
+    WalletCannotRemoveActiveAccount,
     #[error("Failed to derive address from mnemonic")]
     FailedToDeriveAddress,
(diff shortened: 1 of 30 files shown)


tommyv1987:
##Testing

- Displaying the account you wish to delete

Pasted Graphic

- Switch to another account then on thr drop down menu bar for that account select the delete icon for the account you wish to remove (a prompt for backing up your saving wallet appears)

image

- Enter your password to confirm removal of account
image

- Account is succesfully removed from the saved wallet
image

benedettadavico (approved):

PR #6895, opened 2026-06-20
Add npm notice

mfahampshire:
## Summary by CodeRabbit

* **Documentation**
  * Updated network node and gateway statistics to reflect current mesh topology metrics
  * Refreshed timestamp reference data
  * Added version pinning setup instructions across developer tool documentation, noting that manual package configuration is required as the release has not yet been published to the default distribution channel


mfahampshire:
No longer necessary, released `mix-fetch` 2.0.0

PR #6898, opened 2026-06-22
Bandwidth controller step 1 : Standalone

simonwicky:
# Standalone, actor-based bandwidth controller

## Summary

Step 1 of the bandwidth-controller rework. This makes the `BandwidthController` a self-contained, actor-style component that runs on its own task and is addressed through a cloneable handle, instead of being a generic struct embedded (and cloned) into every gateway client.

The behavior is intended to be **equivalent** to the previous implementation — this is a structural refactor that removes generics from the consumer side and decouples the controller from the DKG query client.

## Key changes

**Controller becomes an actor** (`common/bandwidth-controller/src/controller.rs`)
- `BandwidthController<St>` — generic only over storage now.
- A `run(shutdown_token)` event loop processes `BandwidthControllerRequest`s off an mpsc channel until cancelled, then closes storage on shutdown.
- `get_request_sender()` hands out a cloneable `BandwidthControllerRequestSender` (`requests/sender.rs`) that implements `BandwidthTicketProvider` and talks to the controller over the channel (oneshot reply via `ReturnSender`).

**Public data fetching abstracted behind a trait** (`traits.rs`, `fetcher/`)
- New `CredentialPublicDataFetcher` trait for fetching the global ecash signing materials (master VK, coin-index sigs, expiration-date sigs).
- `NyxdGlobalDataFetcher` is the in-repo, nyxd-backed implementation, attached via `with_credential_public_data_fetcher(...)`. The previous `ApiClientsWrapper` logic moved here, now with a lock-free `ArcSwapOption` cache so the trait methods can stay `&self`.
- The controller's `ensure_*` helpers fetch-and-persist any missing global data on demand, removing the `DkgQueryClient` generic from the controller itself.

**Consumers de-generified**
- `GatewayClient<C, St>` → `GatewayClient`; it now holds a `Box<dyn BandwidthTicketProvider>` (`common/client-libs/gateway-client/src/client/mod.rs`).
- Base client wires up the controller in `start_bandwidth_controller(...)`, spawns it as a named `"BandwidthController"` task, and passes the request sender down (`common/client-core/src/client/base_client/mod.rs`). The standalone `CredentialStorage::close_on_shutdown` task is gone — the controller owns storage close now.
- `with_custom_bandwidth_provider(...)` added to both the base client builder and the SDK `MixnetClientBuilder` (`sdk/rust/nym-sdk/src/mixnet/client.rs`) for advanced use cases that manage their own controller.

## Notable API / behavior details
- `get_ecash_ticket` now returns `Option<PreparedCredential>` — `None` instead of the old `NoCredentialsAvailable` error when storage has nothing usable. Gateway client maps `None` to `NoMoreBandwidthCredentials`.
- `get_ecash_ticket` / `prepare_for_spending` take an explicit `spend_time: OffsetDateTime`.
- `attempt_revert_spending` and `get_upgrade_mode_token` are now part of the `BandwidthTicketProvider` trait.
- `BandwidthControllerError` slimmed down: removed now-unused variants, added `FetcherError`, `Missing{VerificationKey,CoinIndexSignatures,ExpirationDateSignatures}`, and `Internal` (for channel/comms failures).

## Diffstat
39 files changed, ~1130 insertions, ~600 deletions. Core changes in `common/bandwidth-controller`; the rest is propagating the de-generified types through gateway-client, client-core, the SDK, registration-client, nym-api, and probe/authenticator clients.

## Release Notes

* **New Features**
  * Added a dedicated bandwidth controller with support for fetching ticketbooks, upgrade-mode tokens, global signing materials, and reverting failed spending.
  * Introduced a unified bandwidth-provider interface that enables custom bandwidth sourcing for gateway and mixnet clients.

* **Improvements**
  * Ticket/payment generation now uses explicit timestamps for spending.
  * “No tickets available” errors are clearer and include the relevant ticket type.

* **Updates**
  * Refactored network monitoring and registration flows to use a request-sender pattern for bandwidth operations.

### DIFF
--- a/common/bandwidth-controller/src/controller.rs
+++ b/common/bandwidth-controller/src/controller.rs
@@ -1,4 +1,4 @@
-// Copyright 2021-2024 - Nym Technologies SA <contact@nymtech.net>
+// Copyright 2026 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::error::BandwidthControllerError;
@@ -78,8 +78,8 @@ impl<St: Storage> BandwidthController<St> {
         self.storage
             .attempt_revert_ticketbook_withdrawal(
                 info.ticketbook_id,
-                info.used_tickets,
                 info.tickets_withdrawn,
+                info.used_tickets,
             )
             .await
             .map_err(BandwidthControllerError::credential_storage_error)
--- a/common/bandwidth-controller/src/controller.rs
+++ b/common/bandwidth-controller/src/controller.rs
@@ -1,0 +1,378 @@
+// Copyright 2021-2024 - Nym Technologies SA <contact@nymtech.net>
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::error::BandwidthControllerError;
+use crate::requests::{BandwidthControllerRequest, BandwidthControllerRequestSender};
+use crate::traits::CredentialPublicDataFetcher;
+use crate::{
+    BandwidthTicketProvider, PreparedCredential, PreparedCredentialMetadata, UPGRADE_MODE_JWT_TYPE,
+};
+use async_trait::async_trait;
+use log::error;
+use nym_credential_storage::models::RetrievedTicketbook;
+use nym_credential_storage::storage::Storage;
+use nym_credentials::ecash::bandwidth::CredentialSpendingData;
+use nym_credentials_interface::{
+    AnnotatedCoinIndexSignature, AnnotatedExpirationDateSignature, TicketType, VerificationKeyAuth,
+};
+use nym_crypto::asymmetric::ed25519;
+use nym_ecash_time::{Date, OffsetDateTime};
+use nym_task::ShutdownToken;
+use nym_validator_client::nym_api::EpochId;
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+
+pub struct BandwidthController<St> {
+    storage: St,
+
+    // Channels used to receive commands from the outside.
+    request_channel: (
+        UnboundedSender<BandwidthControllerRequest>,
+        UnboundedReceiver<BandwidthControllerRequest>,
+    ),
+
+    // fetches the global ecash signing materials when they are missing locally
+    public_data_fetcher: Option<Box<dyn CredentialPublicDataFetcher>>,
+}
+
+impl<St: Storage> BandwidthController<St> {
+    pub fn new(storage: St) -> Self {
+        let request_channel = mpsc::unbounded_channel();
+        BandwidthController {
+            storage,
+            request_channel,
+            public_data_fetcher: None,
+        }
+    }
+
+    #[must_use]
+    pub fn with_credential_public_data_fetcher(
+        mut self,
+        fetcher: impl CredentialPublicDataFetcher + 'static,
+    ) -> Self {
+        self.public_data_fetcher = Some(Box::new(fetcher));
+        self
+    }
+
+    /// Get the request channel used to send request to the controller.
+    /// Request are handled only if the `BandwidthController` is running using `run`
+    pub fn get_request_sender(&self) -> BandwidthControllerRequestSender {
+        BandwidthControllerRequestSender::new(self.request_channel.0.clone())
+    }
+
+    /// Tries to retrieve one of the stored, unused credentials for the given type that hasn't yet expired.
+    pub async fn get_next_usable_ticketbook(
+        &self,
+        ticketbook_type: TicketType,
+        tickets: u32,
+    ) -> Result<Option<RetrievedTicketbook>, BandwidthControllerError> {
+        self.storage
+            .get_next_unspent_usable_ticketbook(ticketbook_type.to_string(), tickets)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)
+    }
+
+    pub async fn attempt_revert_ticket_usage(
+        &self,
+        info: PreparedCredentialMetadata,
+    ) -> Result<bool, BandwidthControllerError> {
+        self.storage
+            .attempt_revert_ticketbook_withdrawal(
+                info.ticketbook_id,
+                info.used_tickets,
+                info.tickets_withdrawn,
+            )
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)
+    }
+
+    pub async fn get_aggregate_verification_key(
+        &self,
+        epoch_id: EpochId,
+    ) -> Result<Option<VerificationKeyAuth>, BandwidthControllerError> {
+        self.storage
+            .get_master_verification_key(epoch_id)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)
+    }
+
+    pub async fn get_coin_index_signatures(
+        &self,
+        epoch_id: EpochId,
+    ) -> Result<Option<Vec<AnnotatedCoinIndexSignature>>, BandwidthControllerError> {
+        self.storage
+            .get_coin_index_signatures(epoch_id)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)
+    }
+
+    pub async fn get_expiration_date_signatures(
+        &self,
+        epoch_id: EpochId,
+        expiration_date: Date,
+    ) -> Result<Option<Vec<AnnotatedExpirationDateSignature>>, BandwidthControllerError> {
+        self.storage
+            .get_expiration_date_signatures(expiration_date, epoch_id)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)
+    }
+
+    /// Returns the master verification key for the epoch, fetching it via the configured public
+    /// data fetcher and persisting it if it isn't already in local storage.
+    async fn ensure_master_verification_key(
+        &self,
+        epoch_id: EpochId,
+    ) -> Result<VerificationKeyAuth, BandwidthControllerError> {
+        if let Some(key) = self.get_aggregate_verification_key(epoch_id).await? {
+            return Ok(key);
+        }
+        let Some(fetcher) = &self.public_data_fetcher else {
+            return Err(BandwidthControllerError::MissingVerificationKey { epoch_id });
+        };
+        let key = fetcher
+            .fetch_master_verification_key(epoch_id)
+            .await
+            .map_err(BandwidthControllerError::fetcher_error)?;
+        self.storage
+            .insert_master_verification_key(&key)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)?;
+        Ok(key.key)
+    }
+
+    /// Returns the coin index signatures for the epoch, fetching them via the configured public
+    /// data fetcher and persisting them if they aren't already in local storage.
+    async fn ensure_coin_index_signatures(
+        &self,
+        epoch_id: EpochId,
+    ) -> Result<Vec<AnnotatedCoinIndexSignature>, BandwidthControllerError> {
+        if let Some(signatures) = self.get_coin_index_signatures(epoch_id).await? {
+            return Ok(signatures);
+        }
+        let Some(fetcher) = &self.public_data_fetcher else {
+            return Err(BandwidthControllerError::MissingCoinIndexSignatures { epoch_id });
+        };
+        let signatures = fetcher
+            .fetch_coin_index_signatures(epoch_id)
+            .await
+            .map_err(BandwidthControllerError::fetcher_error)?;
+        self.storage
+            .insert_coin_index_signatures(&signatures)
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error)?;
+        Ok(signatures.signatures)
+    }
+
+    /// Returns the expiration date signatures for the epoch and expiration date, fetching them via
+    /// the configured public data fetcher and persisting them if they aren't already in local storage.
+    async fn ensure_expiration_date_signatures(
+        &self,
+        epoch_id: EpochId,
+        expiration_date: Date,
+    ) -> Result<Vec<AnnotatedExpirationDateSignature>, BandwidthControllerError> {
+        if let Some(signatures) = self
+            .get_expiration_date_signatures(epoch_id, expiration_date)
+            .await?
+        {
+            return Ok(signatures);
+        }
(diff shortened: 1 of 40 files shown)


common/bandwidth-controller/src/lib.rs
```diff
@@ -1,41 +1,30 @@
...
-#[derive(Debug)]
-pub struct BandwidthController<C, St> {
-    storage: St,
-    client: C,
-}
+pub const DEFAULT_TICKETS_TO_SPEND: u32 = 1;
+
+pub const UPGRADE_MODE_JWT_TYPE: &str = "UPGRADE_MODE_JWT";
 
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
simonwicky: PreparedCredential is not secret and has a custom debug impl to avoid spamming the log

common/bandwidth-controller/src/traits.rs
```diff
@@ -21,66 +22,74 @@ pub trait BandwidthTicketProvider: Send + Sync {
...
+        (**self).close().await;
     }
 }
 
+// This isn't an associated type because
+// a) it would make it dyn-incompatible and we want it
+// b) BandwidthController will pack everything its own variant anyway
+
+/// Error any fetcher implementation may return; the controller wraps it with context.
+pub type FetcherError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
jstuczyn, line 75: this might make it quite cumbersome to match on specific errors though...

common/bandwidth-controller/src/traits.rs
```diff
@@ -21,66 +22,74 @@ pub trait BandwidthTicketProvider: Send + Sync {
...
+
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-impl<T: BandwidthTicketProvider + ?Sized + Send> BandwidthTicketProvider for Box<T> {
-    async fn get_ecash_ticket(
+pub trait CredentialPublicDataFetcher: Send + Sync {
```
jstuczyn, line 79: given it has a single implementation, does it need to be a trait? do we expect to create mocks for testing or something?

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::traits::{CredentialPublicDataFetcher, FetcherError};
+
+/// In-repo [`CredentialPublicDataFetcher`] that retrieves the global ecash signing materials from the
+/// nym-apis using a nyxd query client.
+pub struct NyxdGlobalDataFetcher<C: DkgQueryClient + Send + Sync> {
+    client: C,
+
+    // Lock-free cache for the ecash api clients, lazily instantiated. `ArcSwapOption` lets us
+    // refresh it through `&self` (so the trait methods can stay `&self`) without a mutex.
+    ecash_api_clients: ArcSwapOption<EcashApiClientsCache>,
```
jstuczyn, line 33: honestly, an ArcSwap is an overkill here (how often is the data being accessed?), but I guess it works... though all consumers might not see the updates immediately

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub fn new(client: C) -> Self {
+        NyxdGlobalDataFetcher {
+            client,
+            ecash_api_clients: ArcSwapOption::empty(),
+        }
+    }
+
+    async fn ecash_api_clients(&self, epoch_id: EpochId) -> Result<Vec<EcashApiClient>, Error> {
+        // fast path: atomic load, return if cached for this epoch
+        if let Some(cache) = self.ecash_api_clients.load_full() {
```
jstuczyn, line 46: if that's not too difficult, I'd add a cache timeout, like even 30min or so. why? because we might need to update some signer details manually (if people decide to offboard) and we really want to pick up new addresses eventually

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
```diff
@@ -0,0 +1,168 @@
...
+            if cache.epoch_id == epoch_id {
+                return Ok(cache.clients.clone());
+            }
+        }
+
+        // empty or stale - refresh and atomically swap in the new cache, then return it.
+        // a concurrent miss may fetch redundantly and the later store wins; harmless since
+        // the fetch is idempotent.
+        let cache = Arc::new(EcashApiClientsCache::from_dkg_client(&self.client, epoch_id).await?);
```
jstuczyn, line 55: is there any scenario where we might need to have clients from multiple epochs? (I don't know - just asking). in that case we could keep overwriting the entries with each call.
maybe at epoch transition when we have credentials from the old epoch and start getting new ones for the following one?

common/bandwidth-controller/src/requests/sender.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Basic set of commands that can be sent to the bandwidth controller
+
+impl BandwidthControllerRequestSender {
+    pub fn new(command_tx: UnboundedSender<BandwidthControllerRequest>) -> Self {
+        Self { command_tx }
+    }
+
+    #[instrument(skip(self), level = "debug")]
```
jstuczyn, line 29: `gateway_id` will be logged as this nasty, unreadable, finite field point, you might want to manually re-add it with correct formatting via `fields`

common/bandwidth-controller/src/requests/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    AttemptRevertSpending(
+        ReturnSender<bool, BandwidthControllerError>,
+        PreparedCredentialMetadata,
+    ),
+}
+
+#[derive(Debug)]
+pub struct ReturnSender<T, E> {
+    sender: oneshot::Sender<Result<T, E>>,
```
jstuczyn: given it's always parameterised with `BandwidthControllerError` would it make sense to just drop the second generic? maybe even further simplify because the only possible error variant is channel being closed and the value being returned

common/bandwidth-controller/src/controller.rs
```diff
@@ -0,0 +1,378 @@
...
+    storage: St,
+
+    // Channels used to receive commands from the outside.
+    request_channel: (
+        UnboundedSender<BandwidthControllerRequest>,
+        UnboundedReceiver<BandwidthControllerRequest>,
+    ),
+
+    // fetches the global ecash signing materials when they are missing locally
+    public_data_fetcher: Option<Box<dyn CredentialPublicDataFetcher>>,
```
jstuczyn, line 34: under what circumstances would we NOT want to have it? as in, why is it even an option?

common/bandwidth-controller/src/controller.rs
```diff
@@ -0,0 +1,378 @@
...
+        loop {
+            tokio::select! {
+                biased;
+                _ = shutdown_token.cancelled() => {
+                    log::debug!("bandwidth controller received cancellation request; shutting down");
+                    break;
+                }
+                request = self.request_channel.1.recv() => match request {
+                    Some(request) => self.handle_request(request).await,
+                    None => {
```
jstuczyn, line 310: just `unwrap` it to get rid of this noise

common/client-core/src/client/base_client/mod.rs
```diff
@@ -784,6 +797,34 @@ where
...
+        shutdown_tracker: &ShutdownTracker,
+    ) -> Box<dyn BandwidthTicketProvider> {
+        // if an externally managed bandwidth controller was provided, use it as-is and
+        // don't spin up our own.
+        if let Some(provider) = custom_bandwidth_provider {
+            return provider;
+        }
+
+        let mut bandwidth_controller = BandwidthController::new(credential_store);
+        // if a dkg query client is available, attach a nyxd-backed public data fetcher so the
```
jstuczyn, line 813: why wouldn't it be available?

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -1076,14 +1055,11 @@ impl<C, St> GatewayClient<C, St> {
     }
 }
 
-// type alias for an ease of use
-pub type InitGatewayClient = GatewayClient<InitOnly>;
-
-#[derive(Debug)]
-pub struct InitOnly;
+// type alias to make it clear it's incomplete
+pub type InitGatewayClient = GatewayClient;
```
jstuczyn: I wonder if we still need it at that point...

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -1100,6 +1076,10 @@ impl GatewayClient<InitOnly, EphemeralCredentialStorage> {
         let shutdown_token = ShutdownToken::default();
         let packet_router = PacketRouter::new(ack_tx, mix_tx, shutdown_token.clone());
 
+        // same comment as above
+        let (bc_tx, _) = tokio::sync::mpsc::unbounded_channel();
+        let bandwidth_provider = Box::new(BandwidthControllerRequestSender::new(bc_tx));
```
jstuczyn, line 1081: wouldn't an Option, like we had before, make more sense than a provider that's broken? (rx channel gets immediately dropped)

common/bandwidth-controller/src/requests/sender.rs
```diff
@@ -0,0 +1,103 @@
...
+            ))
+            .map_err(BandwidthControllerError::internal)?;
+        rx.await.map_err(BandwidthControllerError::internal)?
+    }
+
+    #[instrument(skip(self), level = "debug")]
+    pub async fn get_upgrade_mode_token(&self) -> Result<Option<String>, BandwidthControllerError> {
+        let (tx, rx) = ReturnSender::new();
+        self.command_tx
+            .send(BandwidthControllerRequest::UpgradeModeToken(tx))
```
jstuczyn, line 57: we should introduce some timeouts for all those sends in case we do a shutdown. because if bandwidth controller goes out of scope, those channels might hang

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
simonwicky: It's only used locally and because I need to take `&self` and not `&mut self`

common/bandwidth-controller/src/requests/mod.rs
simonwicky: I can have a look, I just copied it from the AC

common/bandwidth-controller/src/controller.rs
simonwicky: Vpn-client might not have it, because those public data comes with the ticket books

simonwicky: Plus if you already have all the data, you don't need it

common/client-core/src/client/base_client/mod.rs
simonwicky: You tell me, why is it an option on `BaseClientBuilder`?

common/client-libs/gateway-client/src/client/mod.rs
simonwicky: That alias is a bit weird indeed. 
It does construct a gateway client that isn't functional so this allows to emphasise that where it's used

common/client-libs/gateway-client/src/client/mod.rs
simonwicky: 1) Having an option just for that is annoying
2) packet router is also not functional (see the above comment)

common/bandwidth-controller/src/requests/sender.rs
simonwicky: `self.command_tx.send` is non-blocking and returns an error if the `rx` end was dropped

common/bandwidth-controller/src/controller.rs
simonwicky: While I do agree, it's not that noisy is it?

common/bandwidth-controller/src/traits.rs
simonwicky: Agreed, but why would you do it really?
As written, I am not making it an associated type, and each fetcher will have it's different error types

common/bandwidth-controller/src/requests/sender.rs
simonwicky: The Debug impl of public key is already the bs58 representation

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
simonwicky: Current behavior spawns a cache each time we need to fetch something, so no regression here, we can deal with that later if needed

common/bandwidth-controller/src/traits.rs
simonwicky: I do expect some mocking capability and maybe later a VPN-API based one. 
I also don't want generics spilling in the BC

common/client-libs/gateway-client/src/client/mod.rs
simonwicky: Replaced by a wrapper struct

common/bandwidth-controller/src/fetcher/nyxd_public_data.rs
jstuczyn: I'd still imo opt for a Arc<Mutex<...>> instead

common/client-libs/gateway-client/src/client/mod.rs
```diff
@@ -1110,44 +1091,59 @@ impl GatewayClient<InitOnly, EphemeralCredentialStorage> {
...
+            negotiated_protocol: inner.negotiated_protocol,
             #[cfg(unix)]
-            connection_fd_callback: self.connection_fd_callback,
+            connection_fd_callback: inner.connection_fd_callback,
             shutdown_token,
         }
     }
 }
+
+impl Deref for InitGatewayClient {
```
jstuczyn: that's the thing we want to avoid : )
this will let somebody misuse the client

simonwicky: Replaced by a `inner_mut` helper, avoiding reimplementation but keeping guardrails

jstuczyn (approved):

PR #6901, opened 2026-06-22
Networking and Bootstrapping configuration changes

jmwample:
NYM-927

This PR distills the networking information needed to bootstrap connections into a shared structure that nym core applications and VPN/discovery use cases can both reuse.

This change:

* Introduces v2::NymNetworkDetails (common/network-defaults/src/network/v2.rs) as a new struct alongside the existing NymNetworkDetails, rather than modifying it in place:
  * Groups nym_api_urls and nym_vpn_api_urls into a new NetworkingSpecifics block
  * Adds placeholder fields for DNS fallback resolution, with room for future internal-nameserver / covert-channel config
  * Provides From conversions in both directions, so existing callers of the original struct are unaffected and can opt in via .into()
  * Note: nym_vpn_api_url is not removed from the original struct — it's left as-is; the v1/v2 split means we don't need to touch it there
* Reworks NymNetworkDetails::new_from_env to parse NYM_APIS/NYM_VPN_APIS as arrays, falling back to the legacy single NYM_API var, and fixes two setter bugs (with_nym_vpn_api_urls and set_nym_api_urls were mutating the wrong field)
* Moves sandbox network defaults from nym-wallet-types into nym-network-defaults (sandbox.rs), matching how mainnet defaults are already structured
* Adds a new nym-api endpoint, GET /v1/network/details-v2, that serves the v2 struct alongside the existing /v1/network/details — same API version, only the response struct differs

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6901)

## Summary by CodeRabbit

* **New Features**
  * Added sandbox network configuration, including API, VPN, RPC, contract, and DNS fallback settings.
  * Network settings now support multiple Nym API and VPN endpoints.

* **Bug Fixes**
  * Improved handling of missing, invalid, and malformed API URLs.
  * Enhanced fallback connectivity when configured endpoints are unavailable.
  * Updated network clients and discovery tools to consistently use configured API endpoints.


common/network-defaults/src/sandbox.rs
```diff
@@ -0,0 +1,135 @@
...
+    },
+];
+
+pub const EXIT_POLICY_URL: &str =
+    "https://nymtech.net/.wellknown/network-requester/exit-policy.txt";
+
+#[cfg(feature = "network")]
+pub fn validators() -> Vec<ValidatorDetails> {
+    vec![ValidatorDetails::new(
+        "https://rpc.sandbox.nymtech.net",
```
benedettadavico: No longer exists, the sandbox validator is now: validator-sandbox-1.nymtech.net

common/network-defaults/src/sandbox.rs
```diff
@@ -0,0 +1,135 @@
...
+        Some("https://sandbox-nym-api1.nymtech.net/api"),
+        Some("wss://rpc.sandbox.nymtech.net/websocket"),
```
benedettadavico: ditto

common/network-defaults/src/network.rs
```diff
@@ -58,9 +58,7 @@ pub struct NymNetworkDetails {
     pub chain_details: ChainDetails,
     pub endpoints: Vec<ValidatorDetails>,
     pub contracts: NymContracts,
-    pub nym_vpn_api_url: Option<String>,
```
jstuczyn, line 61: as mentioned during the call, do not modify the struct as it will break the compatibility with `/v1/network/details` nym-api endpoint

jmwample: Addressed by making the changes in this PR purely addative.

common/network-defaults/src/sandbox.rs
jmwample: If I understand the change expected This should be fixed.

common/client-libs/validator-client/src/client.rs
```diff
@@ -113,31 +113,20 @@ impl Config {
...
-                    .iter()
-                    .map(|url| url.url.parse())
-                    .collect::<Result<Vec<_>, _>>()?,
-            );
-        }
-
-        if api_url.is_empty() {
-            return Err(ValidatorClientError::NoAPIUrlAvailable);
-        }
+        let api_url = details.endpoints[0]
```
jstuczyn: this can can panic if we have no endpoints set. use `details.get(0)` instead

common/network-defaults/src/network.rs
jstuczyn: well. technically it doesn't resolve it as `bincode` can't handle any schema changes, even additive ones. however, while technically, you could ask for bincode response for that endpoint, I'm not sure if anything actually does. you'd have to go through the callstack within the vpn-client to make sure of that. 

if it doesn't involve too many changes, imo, wouldn't it be better to have a `V2` struct? then we also wouldn't have to deal with the baggage of deprecated fields. and we could simply have a `From<NymNetworkDetailsV2> for NymNetworkDetails` for any legacy endpoints/consumers

common/network-defaults/src/network.rs
```diff
@@ -244,6 +303,18 @@ impl NymNetworkDetails {
         }
     }
 
+    pub(crate) fn mainnet_specifics() -> NetworkingSpecifics {
```
jstuczyn: nit: since you added a new `sandbox.rs` a helper for constructing sandbox details would be useful

common/network-defaults/src/network.rs
```diff
@@ -150,12 +183,38 @@ impl NymNetworkDetails {
...
-            front_hosts: None,
-        }]);
-        let nym_vpn_api_urls = try_parse_api_urls(var_names::NYM_VPN_APIS);
+        let nym_api_urls = try_parse_api_urls(var_names::NYM_APIS).unwrap_or_else(|e| {
+            panic!(
+                "{} is set but could not be parsed: {e}",
+                var_names::NYM_APIS
+            )
+        });
+        // NYM_APIS was introduced to replace the singular NYM_API; fall back to it so
```
jstuczyn: but how could we fallback if we already panicked because we failed to parse it? we can't unwrap the `None`

jstuczyn: ah, I haven't got to that part - you updated the function signature

nym-gateway-probe/src/common/helpers.rs
```diff
@@ -64,17 +64,21 @@ pub async fn fetch_topology(
...
-                .and_then(|e| e.api_url())
-                .map(|url| vec![url])
-        })
-        .unwrap_or_default();
+        .nym_api_urls()
+        .iter()
+        .filter_map(|u| u.url.parse().ok())
+        .collect();
+
+    let nym_api_urls = if nym_api_urls.is_empty() {
```
jstuczyn: could we instead use a helper on `network_details`?

jstuczyn (changes requested):

nym-gateway-probe/src/common/helpers.rs
jmwample: Yes, I added this to the default behavior of the nym_api_urls helper.

This was a holdout of an older behavior. The `TopologyProvider` had domain rotation before the it was added to the HTTP client where it would try to rotate between validators API urls if multiple validators were defined (afaik there was never more than one in our config). However the http client now manages the rotation of API urls on failures, and does has better defined conditions for when it is and is not appropriate to try to rotate urls. So I removed the URL rotation from the TopologyProvider and added the behavior to the Network details that if no nym_api_urls are defined, but there are validator endpoints, then it will try to use them.

common/network-defaults/src/network.rs
jmwample: Added :+1:, though it is unused.

Just to note though, this was more of a migration to square the location for these types of details. This information for sandbox was just in ny-wallet before.

common/client-libs/validator-client/src/client.rs
jmwample: Fixed - this can no longer panic using `.first().ok_or(...)?`

common/network-defaults/src/network.rs
jmwample: I tried to make the functionality clearer here. If there is a potential issue lmk

common/network-defaults/src/network.rs
jmwample: I extracted the changes to a v2. this includes conversions and a second api endpoint. I explicitly left bincode support off of the new `network/details-v2` endpoint as this type is likely to have more details added -- specifically thinking about default dns nameserver details and covert channel default priorities. And I don't want to make a v3 for those potential changes.

jstuczyn (approved):

jstuczyn (changes requested):
final thing: can you change the `/v1/network/details-v2` into `/v2/network/details` for consistency sake (e.g. we change the type of allowed output)?

jstuczyn (approved):

PR #6936, opened 2026-07-03
feat: Verifiable Directory Contract retrieval

jstuczyn:
# Verifiable directory retrieval client (proven mode)

## Summary

Adds the read/verify side of the directory system: a `nym-directory-client` library that retrieves the on-chain directory and cryptographically proves it is the complete, untampered set committed by the contract at a specific block height, with no trust in the serving RPC beyond the block header. This PR is stacked on https://github.com/nymtech/nym/pull/6933 and should merge after it

This is the OpenSpec change `directory-retrieval-client` (archived at `openspec/changes/archive/2026-07-03-directory-retrieval-client/`, capability spec synced to `openspec/specs/directory-retrieval-client/`). It is scoped to the "proven" trust mode; the attested (nym-api quorum) mode, a full light client, and the producer side are deliberately out of scope and slot in later behind the same seam (see Future work).

## How verification works

- **Trust anchor** (`DirectoryTrustAnchor`): yields the chain state trusted at height `H` - `trusted_app_hash(H)` (the block app hash, the root every ICS23 proof checks against) and `trusted_digest(H)`. Proven mode reads the app hash from a configured RPC's `header[H+1]` (CometBFT off-by-one); a light-client or attested anchor can replace that source without touching the verifier.
- **Whole-directory read** (`verified_directory`): proves the single on-chain LtHash digest item via a two-layer ICS23 store proof against the trusted app hash, fetches all entries, recomputes the LtHash locally over the canonical leaves, and rejects on any mismatch. One digest proof plus a local recompute authenticates the entire set, so there are no per-entry proofs for the whole-directory path.
- **Single-block-height consistency**: the digest proof and every paginated `AllEntries` page are read at the same `H`, so a write committed mid-pagination can never interleave into a set that matches no committed digest.
- **Signature tiering**: each node entry's ed25519 signature is verified against the node's identity key from the mixnet bond; entries are classified node-authored vs admin-curated in the returned type.
- **Single-entry reads** (`verified_node_entry`, `verified_curated_entry`): prove one entry via its own ICS23 membership proof, or its absence via a full non-existence proof, so "absent" (`Ok(None)`) is distinct from a verification failure (`Err`). Presence is decided by the proof shape, not by whether the value is empty. The app hash comes from the anchor, not the RPC serving the proof, so a malicious RPC cannot substitute a self-consistent forged (app_hash, proof) pair.
- **Fail closed**: every fallible step propagates a typed `DirectoryClientError`; there is no path that returns unverified data as if verified.

## Future work (behind the same trust-anchor seam)

- Phase 1b: a real tendermint light client (validator-set verification from a pinned checkpoint) replacing the trusted-RPC app-hash source.
- Phase 2: the attested mode (K-of-N nym-api quorum) and the signed digest snapshot endpoint it needs.
- The producer side (nodes publishing entries) and the per-label payload taxonomy.
- Consumer integration (nym-api / node-status / clients) that adopts the library.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6936)


mmsinclair (approved):
nice!

PR #6938, opened 2026-07-06
[BandwidthController] Fetching Logic

simonwicky:
## Bandwidth controller: pluggable fetchers + auto-restock loop

First PR of the bandwidth-controller-step-2 stack. This reworks the
`BandwidthController` internals so credential acquisition and global-data
retrieval are delegated to **pluggable, storage-free fetchers**, and adds the
background loop that keeps ticket stocks topped up. No fetcher implementation
lives here yet (that's PR 2) — this is the controller-side contract and logic.

### What changes

**Fetcher traits (`traits.rs`)**
- `FetcherError` becomes a real trait (`std::error::Error + kind()`) instead of
  a bare boxed error. `CredentialFetcherError` is the boxed, dyn-compatible form,
  with a blanket `From` so `?` works on any concrete fetcher error.
- `FetcherErrorKind` (in `error.rs`) gives the controller a coarse category to
  branch on (`BandwidthDepleted`, `Api`, `Storage`, `Unexpected`, `Other`)
  without knowing the concrete error type.
- New `CredentialFetcher` trait: `fetch_ticketbooks`, `cleanup`, `reset`. It's a
  superset of `CredentialPublicDataFetcher`, so installing a credential fetcher
  also registers it for global-data retrieval.
- Fetchers are storage-free: they do only network/crypto work and hand results
  back for the controller to persist. The controller does **not** retry —
  transient-error handling is the fetcher's job.

**Controller (`controller.rs`)**
- The `BandwidthController` is now the **single writer** to credential storage;
  all acquisition/retrieval delegates to the two fetchers.
- Two ways to drive it:
  - `run()` — event loop. On native targets it proactively restocks low ticket
    types in the background (`JoinSet` of fetch tasks, `in_flight` dedup), keeps
    global data topped up, and resolves readiness waiters as fetches land.
  - direct one-shot calls (`fetch_ticketbook`, `prepare_ecash_ticket`) for use
    without spawning the loop.
- Builder-style config: `with_credential_fetcher`, `with_credential_public_data_fetcher`.

**Readiness (`readiness.rs`, new)**
- `wait_for_ticketbooks` support: callers park until every required ticket type
  is usable (stocked or covered by upgrade mode), or fail if a required type is
  neither stocked nor being fetched. Severity-ordered `ReadinessStatus` so a
  waiter learns *why* a type is unavailable (incl. the last fetch failure reason).

**Ticketbooks (`ticketbooks.rs`, new)**
- `AvailableTicketbook` / `AvailableTicketbooks` stock model with restock
  thresholds and SI-scaled reporting helpers.

**New request variants (`requests/mod.rs`)**
- `SetCredentialFetcher` / `SetPublicDataFetcher` (hot-swap; a new fetcher
  triggers an immediate restock), `Reset`, `ClearEmergencyCredentials`,
  `GetAvailableTicketbooks`, `WaitForTicketbooks`.

**`NymCredential`** — new enum: `Ticketbook` or `UpgradeModeToken { jwt, expiration }`.

### Notes
- The old inline `acquire`/`fetcher`/`utils` modules are still present here; they
  get removed in PR 3 once consumers are migrated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6938)

### DIFF
--- a/common/bandwidth-controller/src/controller.rs
+++ b/common/bandwidth-controller/src/controller.rs
@@ -2,16 +2,29 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::error::BandwidthControllerError;
+use crate::readiness::{FetchFailure, ReadinessRequest, ReadinessSnapshot, ReadinessStatus};
 use crate::requests::{BandwidthControllerRequest, BandwidthControllerRequestSender};
-use crate::traits::CredentialPublicDataFetcher;
+use crate::ticketbooks::AvailableTicketbooks;
+use crate::traits::{CredentialFetcher, CredentialPublicDataFetcher};
 use crate::{
     BandwidthTicketProvider, PreparedCredential, PreparedCredentialMetadata, UPGRADE_MODE_JWT_TYPE,
 };
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+
+#[cfg(not(target_arch = "wasm32"))]
+use crate::traits::CredentialFetcherError;
+use crate::NymCredential;
+use std::collections::HashSet;
+
 use async_trait::async_trait;
 use log::error;
+use nym_credential_storage::models::EmergencyCredentialContent;
 use nym_credential_storage::models::RetrievedTicketbook;
 use nym_credential_storage::storage::Storage;
 use nym_credentials::ecash::bandwidth::CredentialSpendingData;
+use nym_credentials::IssuedTicketBook;
 use nym_credentials_interface::{
     AnnotatedCoinIndexSignature, AnnotatedExpirationDateSignature, TicketType, VerificationKeyAuth,
 };
@@ -21,6 +34,36 @@ use nym_task::ShutdownToken;
 use nym_validator_client::nym_api::EpochId;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 
+/// How often the controller proactively checks whether any ticket type needs restocking.
+const TOPUP_INTERVAL: Duration = Duration::from_secs(3 * 3600); // 3 hours
+
+/// Result of a background ticketbook fetch, drained from `fetch_tasks` in the `run` loop.
+#[cfg(not(target_arch = "wasm32"))]
+struct FetchOutcome {
+    ticket_type: TicketType,
+    result: Result<Vec<NymCredential>, CredentialFetcherError>,
+}
+
+/// Owns all ecash credential state and is the **single writer** to the credential [`Storage`].
+///
+/// It serves ticket spending, credential acquisition, and global signing-data retrieval while
+/// keeping storage consistent by being the only component that writes to it. The acquisition and
+/// retrieval work is delegated to two pluggable, **storage-free** fetchers: they only do the
+/// network/cryptographic work and hand the results back for the controller to persist.
+///
+/// - [`CredentialPublicDataFetcher`] (`public_data_fetcher`): lazily retrieves the global ecash
+///   signing materials - master verification key, coin-index and expiration-date signatures - when
+///   they're missing locally.
+/// - [`CredentialFetcher`] (`credential_fetcher`): provisions usable ticketbooks (making deposits
+///   and aggregating wallet signatures). It is a superset of the public-data fetcher, so installing
+///   one via [`Self::with_credential_fetcher`] also registers it as the public-data fetcher.
+///
+/// It can be driven two ways:
+/// - [`Self::run`] — the event loop: handles requests from [`BandwidthControllerRequestSender`] and,
+///   on native targets, proactively restocks low ticket types in the background, keeps the global
+///   data topped up, and resolves `wait_for_ticketbooks` readiness waiters as fetches complete.
+/// - directly on a non-running instance — e.g. [`Self::fetch_ticketbook`] or
+///   [`Self::prepare_ecash_ticket`] — for one-shot use without spawning the loop.
 pub struct BandwidthController<St> {
     storage: St,
 
@@ -31,17 +74,51 @@ pub struct BandwidthController<St> {
     ),
 
     // fetches the global ecash signing materials when they are missing locally
-    public_data_fetcher: Option<Box<dyn CredentialPublicDataFetcher>>,
+    public_data_fetcher: Option<Arc<dyn CredentialPublicDataFetcher>>,
+
+    // provisions usable ticketbooks. Available on all targets for manual (inline) fetching;
+    // `Arc` so the native auto path can clone it into spawned fetch tasks.
+    credential_fetcher: Option<Arc<dyn CredentialFetcher>>,
+
+    // ticket types with an automatic fetch currently in flight, so we don't request the same
+    // type twice (native auto path only)
+    #[cfg(not(target_arch = "wasm32"))]
+    in_flight: HashSet<TicketType>,
+
+    // background ticketbook fetches; completions are drained in the `run` loop
+    #[cfg(not(target_arch = "wasm32"))]
+    fetch_tasks: tokio::task::JoinSet<FetchOutcome>,
+
+    // callers parked on `wait_for_ticketbooks`, re-evaluated whenever a fetch completes
+    pending_readiness: Vec<ReadinessRequest>,
 }
 
 impl<St: Storage> BandwidthController<St> {
+    // ---------------------------------------------------------------------
+    // Construction & configuration
+    // ---------------------------------------------------------------------
+
     pub fn new(storage: St) -> Self {
         let request_channel = mpsc::unbounded_channel();
         BandwidthController {
             storage,
             request_channel,
             public_data_fetcher: None,
+            credential_fetcher: None,
+            #[cfg(not(target_arch = "wasm32"))]
+            in_flight: HashSet::new(),
+            #[cfg(not(target_arch = "wasm32"))]
+            fetch_tasks: tokio::task::JoinSet::new(),
+            pending_readiness: Vec::new(),
         }
+    }
+
+    #[must_use]
+    pub fn with_credential_fetcher(mut self, fetcher: impl CredentialFetcher + 'static) -> Self {
+        let fetcher = Arc::new(fetcher);
+        self.credential_fetcher = Some(fetcher.clone());
+        self.public_data_fetcher = Some(fetcher);
+        self
     }
 
     #[must_use]
@@ -49,7 +126,7 @@ impl<St: Storage> BandwidthController<St> {
         mut self,
         fetcher: impl CredentialPublicDataFetcher + 'static,
     ) -> Self {
-        self.public_data_fetcher = Some(Box::new(fetcher));
+        self.public_data_fetcher = Some(Arc::new(fetcher));
         self
     }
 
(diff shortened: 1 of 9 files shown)


common/bandwidth-controller/src/traits.rs
```diff
@@ -67,29 +69,59 @@ impl<T: BandwidthTicketProvider + ?Sized + Send> BandwidthTicketProvider for Box
...
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait CredentialFetcher: CredentialPublicDataFetcher + Send + Sync {
+    /// Fetches (or recovers) ticketbooks of the given type. The controller does **not** retry on
+    /// failure - retrying transient errors is the fetcher's responsibility. Recovery can yield
+    /// several ticketbooks of the same type, hence the `Vec`.
+    async fn fetch_ticketbooks(
+        &self,
+        ticketbook_type: TicketType,
```
jstuczyn, line 98: should it rather take a Vec<TicketType> or maybe HashMap<TicketType, usize> for fetching multiple ticketbooks at once?

common/bandwidth-controller/src/traits.rs
```diff
@@ -67,29 +69,59 @@ impl<T: BandwidthTicketProvider + ?Sized + Send> BandwidthTicketProvider for Box
...
+    ) -> Result<Vec<NymCredential>, CredentialFetcherError>;
+
+    /// Persists any in-progress state (e.g. closing storage) before the fetcher is dropped, such
+    /// that it can be resumed later.
+    async fn cleanup(&self);
+
+    /// Wipes the fetcher's long-term data (e.g. removes its storage).
```
jstuczyn, line 105: why would we ever want to reset the long-term data? shouldn't it just happen automatically, without invoking trait-method, within some sort of controller on a timer?

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+
+use nym_credential_storage::models::BasicTicketbookInformation;
+use nym_credentials_interface::TicketType;
+use nym_ecash_time::{Date, EcashTime, OffsetDateTime};
+use strum::IntoEnumIterator;
+
+use crate::error::BandwidthControllerError;
+
+// If we go below this threshold, we should request more tickets
+const TICKET_NUMBER_THRESHOLD: u64 = 20;
```
jstuczyn, line 14: I think these constants should very much be configurable, because the numbers will heavily depend on particular usage case. I can imagine somebody downloading a file at 200Mbps would have a different requirement than a slow chat application through mixnet

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+const SOON_EXPIRY_THRESHOLD: Duration = Duration::from_secs(12 * 3600); // 12 hours
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct AvailableTicketbook {
+    pub id: i64,
+    pub typ: TicketType,
+    pub expiration: Date,
+    pub issued_tickets: u32,
+    pub claimed_tickets: u32,
+    pub ticket_size: u64,
```
jstuczyn, line 29: isn't the size implied from the type?

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub fn ticket_size_si(&self) -> String {
+        si_scale::helpers::bibytes2(self.ticket_size)
+    }
+
+    pub fn has_expired(&self) -> bool {
+        self.expiration <= nym_ecash_time::ecash_today().date()
+    }
+
+    // If that ticketbook will be expired in SOON_EXPIRY_THRESHOLD
+    pub fn expired_soon(&self) -> bool {
```
jstuczyn, line 54: nit: should perhaps the current time be an argument to the method so that we could potentially unit test the whole thing more easily?

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        let expiration = if self.expiration <= ecash_today {
+            format!("EXPIRED ON: {}", self.expiration)
+        } else {
+            format!("expires: {}", self.expiration)
+        };
+
+        write!(
+            f,
+            "{{ id: {}, type: {}, tickets: {}/{}, size: {}, remaining: {}/{}, {} }}",
```
jstuczyn, line 71: remember we're counting from 0

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    pub fn remaining_data(&self, typ: TicketType) -> u64 {
+        self.remaining_tickets(typ) * typ.to_repr().bandwidth_value()
+    }
+
+    pub fn remaining_data_si(&self, typ: TicketType) -> String {
+        si_scale::helpers::bibytes2(
+            self.remaining_tickets(typ) as f64 * typ.to_repr().bandwidth_value() as f64,
+        )
```
jstuczyn, line 123:
```suggestion
        si_scale::helpers::bybytes2(self.remaining_data(typ) as f64)
```

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+            issued_tickets: value.total_tickets,
+            claimed_tickets: value.used_tickets,
+            ticket_size: typ.to_repr().bandwidth_value(),
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct AvailableTicketbooks {
+    pub ticketbooks: Vec<AvailableTicketbook>,
```
jstuczyn, line 105: given all the filtering going around, would it perhaps be worth to instead store ticketbooks inside `HashMap<TicketType, AvailableTicketbook`?

common/bandwidth-controller/src/ticketbooks.rs
```diff
@@ -0,0 +1,210 @@
...
+        self.tickets_by_type(typ)
+            .filter(|ticketbook| !ticketbook.has_expired())
+            .map(|ticketbook| ticketbook.remaining_tickets())
+            .fold(0, |acc, remaining| acc.saturating_add(remaining.into()))
+    }
+
+    /// Whether `typ` should be proactively restocked
+    pub fn needs_restock(&self, typ: TicketType) -> bool {
+        let remaining = self.remaining_tickets_long_lasting(typ);
+        remaining <= TICKET_NUMBER_THRESHOLD
```
jstuczyn, line 149: as I said before, those imo should  be passed as arguments rather than hardcoded

common/bandwidth-controller/src/requests/sender.rs
```diff
@@ -74,6 +78,106 @@ impl BandwidthControllerRequestSender {
         rx.await
             .map_err(|_| BandwidthControllerError::ChannelClosed)?
     }
+
+    /// Installs the credential fetcher; the controller immediately restocks any low types.
+    #[instrument(skip(self, credential_fetcher))]
```
jstuczyn, line 83: 1. why are all these methods mapping to a different error variant than the previous ones? should we change the old ones?
2. should we have some sort of timeout?

common/bandwidth-controller/src/requests/mod.rs
```diff
@@ -18,6 +21,25 @@ pub enum BandwidthControllerRequest {
...
+    /// Sets (or clears, with `None`) the global-data fetcher.
+    SetPublicDataFetcher(
+        ReturnSender<()>,
+        Option<Arc<dyn CredentialPublicDataFetcher>>,
+    ),
+    /// Cancels in-flight fetches, drops the fetcher, clears all stored credentials, and fails every
+    /// parked readiness waiter.
+    Reset(ReturnSender<()>),
+    /// Removes the stored emergency (upgrade-mode) credentials only, leaving ticketbooks intact.
+    ClearEmergencyCredentials(ReturnSender<()>),
```
jstuczyn, line 36: why woulld we want that?

common/bandwidth-controller/src/requests/mod.rs
```diff
@@ -18,6 +21,25 @@ pub enum BandwidthControllerRequest {
...
+    /// Returns the currently stored ticketbooks (also logs a stock summary).
+    GetAvailableTicketbooks(ReturnSender<AvailableTicketbooks>),
+
+    /// Resolves once every required ticket type is usable (stocked or covered by upgrade mode),
+    /// or fails if a required type is neither stocked nor being fetched.
+    WaitForTicketbooks(ReturnSender<()>, Vec<TicketType>),
```
jstuczyn, line 42: this should definitely have a timeout : )
though it might be up to the caller 🤔

common/bandwidth-controller/src/readiness.rs
```diff
@@ -0,0 +1,103 @@
...
+    Ready,
+    InFlight,
+    Unavailable,
+    /// unavailable because the last fetch failed; carries the reason for the waiter
+    FetchFailed(String),
+}
+
+impl ReadinessStatus {
+    // higher = worse; `evaluate_readiness` reports the most severe status across required types
+    fn severity(&self) -> u8 {
```
jstuczyn, line 24: didn't you mean to put them in the reverse order?

common/bandwidth-controller/src/readiness.rs
```diff
@@ -0,0 +1,103 @@
...
+pub(crate) struct FetchFailure {
+    pub(crate) ticket_type: TicketType,
+    pub(crate) error: CredentialFetcherError,
+}
+
+/// The state readiness is judged against - built once and reused across a batch of waiters.
+#[derive(Debug)]
+pub(crate) struct ReadinessSnapshot {
+    pub(crate) upgrade_mode: bool,
+    pub(crate) tickets_readiness: HashMap<TicketType, ReadinessStatus>,
```
jstuczyn, line 45: what if you had multiple tickets of the same type?

common/bandwidth-controller/src/readiness.rs
```diff
@@ -0,0 +1,103 @@
...
+        required
+            .iter()
+            .map(|typ| {
+                self.tickets_readiness
+                    .get(typ)
+                    .cloned()
+                    .unwrap_or(ReadinessStatus::Unavailable)
+            })
+            .max_by_key(|status| status.severity())
+            .unwrap_or(ReadinessStatus::Ready)
```
jstuczyn, line 63: wait, if we don't have any tickets, we're defaulting to Ready??

common/bandwidth-controller/src/controller.rs
```diff
@@ -21,6 +34,36 @@ use nym_task::ShutdownToken;
 use nym_validator_client::nym_api::EpochId;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 
+/// How often the controller proactively checks whether any ticket type needs restocking.
+const TOPUP_INTERVAL: Duration = Duration::from_secs(3 * 3600); // 3 hours
```
jstuczyn, line 38: make it configurable?

common/bandwidth-controller/src/controller.rs
```diff
@@ -2,16 +2,29 @@
...
+use crate::traits::{CredentialFetcher, CredentialPublicDataFetcher};
 use crate::{
     BandwidthTicketProvider, PreparedCredential, PreparedCredentialMetadata, UPGRADE_MODE_JWT_TYPE,
 };
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+
+#[cfg(not(target_arch = "wasm32"))]
+use crate::traits::CredentialFetcherError;
```
jstuczyn, line 17: nit: move target-specific imports to the bottom

common/bandwidth-controller/src/controller.rs
```diff
@@ -21,6 +34,36 @@ use nym_task::ShutdownToken;
 use nym_validator_client::nym_api::EpochId;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 
+/// How often the controller proactively checks whether any ticket type needs restocking.
+const TOPUP_INTERVAL: Duration = Duration::from_secs(3 * 3600); // 3 hours
+
+/// Result of a background ticketbook fetch, drained from `fetch_tasks` in the `run` loop.
+#[cfg(not(target_arch = "wasm32"))]
+struct FetchOutcome {
```
jstuczyn, line 42: why are we not allowing background fetching in wasm?

common/bandwidth-controller/src/controller.rs
```diff
@@ -31,25 +74,59 @@ pub struct BandwidthController<St> {
...
+    credential_fetcher: Option<Arc<dyn CredentialFetcher>>,
+
+    // ticket types with an automatic fetch currently in flight, so we don't request the same
+    // type twice (native auto path only)
+    #[cfg(not(target_arch = "wasm32"))]
+    in_flight: HashSet<TicketType>,
+
+    // background ticketbook fetches; completions are drained in the `run` loop
+    #[cfg(not(target_arch = "wasm32"))]
+    fetch_tasks: tokio::task::JoinSet<FetchOutcome>,
```
jstuczyn, line 90: shouldn't these really be part of CredentialFetcher internals?

common/bandwidth-controller/src/controller.rs
```diff
@@ -59,6 +136,290 @@ impl<St: Storage> BandwidthController<St> {
...
+        tracing::info!("BandwidthController started successfully");
+        let mut topup_interval = {
+            let mut interval = tokio::time::interval(TOPUP_INTERVAL);
+            interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            interval
+        };
+
+        loop {
+            tokio::select! {
+                biased;
```
jstuczyn, line 157: given you're using biased selection, make sure all branches are in the order you expect, because right now we prefer topup_interval ticks over tasks completing or requests coming in

common/bandwidth-controller/src/controller.rs
```diff
@@ -59,6 +136,290 @@ impl<St: Storage> BandwidthController<St> {
...
+                self.check_and_restock(vec![ticket_type]).await;
+            }
+            BandwidthControllerRequest::UpgradeModeToken(return_sender) => {
+                return_sender.send(self.get_upgrade_mode_token().await)
+            }
+            BandwidthControllerRequest::AttemptRevertSpending(return_sender, metadata) => {
+                return_sender.send(self.attempt_revert_ticket_usage(metadata).await)
+            }
+            BandwidthControllerRequest::SetCredentialFetcher(return_sender, fetcher) => {
+                self.handle_set_credential_fetcher(fetcher).await;
```
jstuczyn, line 246: don't we have to reset the old one if there was one already set?

common/bandwidth-controller/src/controller.rs
```diff
@@ -59,6 +136,290 @@ impl<St: Storage> BandwidthController<St> {
...
+            fetcher.cleanup().await;
+        }
+        self.credential_fetcher = None;
+
+        let requests = std::mem::take(&mut self.pending_readiness);
+        requests.into_iter().for_each(|r| r.cancel());
+
+        // Clear credentials
+        self.storage
+            .clear_ticketbooks()
```
jstuczyn, line 310: under what circumstances would we want to be doing this?

common/bandwidth-controller/src/controller.rs
```diff
@@ -85,44 +446,253 @@ impl<St: Storage> BandwidthController<St> {
...
+            return;
+        }
+        let Some(fetcher) = &self.credential_fetcher else {
+            tracing::debug!("No credential fetcher set. No restock possible");
+            return;
+        };
+        let fetcher = Arc::clone(fetcher);
+        self.in_flight.insert(ticketbook_type);
+        tracing::debug!("requesting more {ticketbook_type} ticketbooks");
+        self.fetch_tasks.spawn(async move {
```
jstuczyn, line 507: I feel like this task should have some sort of upper bound timeout controlled by the bandwidth controller otherwise in some circumstances we might get stuck. for example a task that panics will never return

common/bandwidth-controller/src/traits.rs
simonwicky: That is debatable. 
I did this thinking about the vpn-client, where you can parallelise every claim, so here you would just call the same fn multiple times. 

In the case of the blockchain, it is suboptimal indeed. 
However, replacing this with a Vec or multiple fetching hampers the fetcher that can parallelise, because the BC doesn't know whether to call once with multiple types, or multiple times. 

For now, we keep the optimal way fo the vpn client, there is no regression for other clients, just no changes

common/bandwidth-controller/src/traits.rs
simonwicky: Vpn-client resets the pending credential storage e.g. on logout. Since it's goes through the BC, it must be on the trait

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: It is. Given how it is used to calculate totals, I reckon it's not a bad idea to derive it once from the type and call it a day

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: I'll have a look, but again, those structs are brought as-is from what was in AC before

common/bandwidth-controller/src/requests/mod.rs
simonwicky: You had that on the AC, when exiting upgrade mode

common/bandwidth-controller/src/requests/mod.rs
simonwicky: Up to the caller

common/bandwidth-controller/src/readiness.rs
simonwicky: No, I want to return "ready" if everything is "ready", "inflight" if everything is either ready or "in flight", etc. 
I pick the max and report it, that's the behavior I want

common/bandwidth-controller/src/readiness.rs
simonwicky: We want to know **if** we have tickets of that type, we don't care about how many

common/bandwidth-controller/src/readiness.rs
simonwicky: If we are waiting on no ticket type, we are ready by default yes

common/bandwidth-controller/src/controller.rs
simonwicky: I'd say it's future work

common/bandwidth-controller/src/controller.rs
simonwicky: Wasm doesn't play well with Tokio async and JoinSet. 
No regression though. Adding that to WASM is out of scope

common/bandwidth-controller/src/controller.rs
simonwicky: No
CredentialFetcher is about "how you get credentials". 
BC has a central fetching logic

common/bandwidth-controller/src/controller.rs
simonwicky: Vpn-client. on logout

common/bandwidth-controller/src/controller.rs
simonwicky: The fetcher mustn't panic. If it panics, we lose the info about what ticket type it was. I'll still have a look into it. 
As for timeouts, it's the fetcher's responsibility.

common/bandwidth-controller/src/controller.rs
simonwicky: That is correct. If a request and a completion arrives at the same time, it's pointless to handle the completion and then check again the types we already have. 
Handle the tick, see that stuff is in flight, not even a look into the storage. 
Then the completion completes and tickets are there

common/bandwidth-controller/src/controller.rs
simonwicky: Look at `handle_set_credential_fetcher` ;)
And we only want to cleanup, it can be the case that we set the same one not long after and want to keep some data

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: Added in a `BandwidthControllerConfig` on the top stack PR to avoid rebasing annoyance

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: Added on the top stack PR to avoid rebasing annoyance

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: Added in a `BandwidthControllerConfig` on the top stack PR to avoid rebasing annoyance

common/bandwidth-controller/src/controller.rs
simonwicky: Done on the top of the stack

common/bandwidth-controller/src/controller.rs
simonwicky: Done in the BandwidthControllerConfig on the top PR

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: We are reporting how much is left, so `total - used`. No off by one error here :
`Ticketbook: { id: 7, type: v1-wireguard-exit, tickets: 50/50, size: 476.84 MiB, remaining: 23.28 GiB/23.28 GiB, expires: 2026-07-08 }`

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: Nope, you can have multiple ticket books for one type. At best it could be `HashMap<TicketType, Vec<AvailableTicketbook>>`, but it's never big, so not worth it really

common/bandwidth-controller/src/requests/sender.rs
simonwicky: 1. my bad
2. Most of these don't need timeouts, and otherwise, up to the caller

simonwicky: 1 done on the top PR

common/bandwidth-controller/src/ticketbooks.rs
jstuczyn: so if you've done it, resolve the conversation yourself : )

common/bandwidth-controller/src/ticketbooks.rs
jstuczyn: so how about doing that then?

common/bandwidth-controller/src/controller.rs
jstuczyn: I feel it might bite us later when we integrate BC into the gateway client and suddenly it won't work in wasm anymore; I don't think we need JoinSet, we can use other primitives, no?

common/bandwidth-controller/src/controller.rs
jstuczyn: > CredentialFetcher is about "how you get credentials".

precisely, so shouldn't it control how it's getting them?

common/bandwidth-controller/src/controller.rs
jstuczyn: > The fetcher mustn't panic.

neither must any of our code : ) but personally I'm fan of defensive coding for these reasons : )

common/bandwidth-controller/src/ticketbooks.rs
simonwicky: I'd say given the usual vector size and the cost of hashing, it is not worth it

common/bandwidth-controller/src/controller.rs
simonwicky: New fetching loop on the top PR

common/bandwidth-controller/src/controller.rs
simonwicky: New fetching loop on the top PR

common/bandwidth-controller/src/controller.rs
simonwicky: New fetching loop on the top PR

PR #6939, opened 2026-07-06
[BandwidthController] Nyxd Credential Fetcher

simonwicky:
## New `nym-bandwidth-fetcher` crate: nyxd-backed fetcher implementations

Second PR of the stack. Adds the concrete fetchers behind the traits from PR 1,
in a new `common/bandwidth-fetcher` crate, and adjusts credential-storage to
support best-effort ticketbook storage.

### New crate: `nym-bandwidth-fetcher`

- **`NyxdGlobalDataFetcher`** (`public_data.rs`) — `CredentialPublicDataFetcher`
  that retrieves global ecash signing materials (master verification key,
  coin-index & expiration-date signatures) from the nym-apis via a nyxd query
  client. Works on wasm.
- **`NyxdCredentialFetcher`** (`credentials.rs`, native only) — `CredentialFetcher`
  that obtains ticketbooks by depositing on-chain and aggregating wallet
  signatures from the ecash APIs, and recovers ticketbooks from deposits whose
  issuance never completed. Deposits are serialised via a `deposit_lock` so
  concurrent fetches can't race the account sequence number. Backed by a
  sqlite pending-requests store (`storage/`).
- **`NyxdRecoveryFetcher`** (behind the `recovery` feature) — recover-only
  fetcher usable with a plain query client.
- **`EcashApiClientsCache`** — lock-free (`ArcSwapOption`) cache of ecash API
  clients, lazily instantiated and refreshed through `&self`, 30-min validity.

### credential-storage changes

- **Best-effort ticketbook storage.** A ticketbook may now be stored *before*
  its coin-index / expiration-date signatures exist (those are fetched lazily at
  spend time). Migration `20251104120000_drop_ticketbook_signature_fks.sql`
  recreates `ecash_ticketbook` without the signature foreign keys (SQLite can't
  drop constraints in place; all columns/data/uniqueness preserved).
- Removes the now-unused pending-ticketbook types (`RetrievedPendingTicketbook`,
  `StoredPendingTicketbook`) and the corresponding storage-backend plumbing —
  pending-request tracking now lives in the fetcher's own store.
- New migration `20260701120000_add_ecash_tables.sql` and regenerated `.sqlx`
  query cache.

### Notes
- Native-only bits are gated on `cfg(not(target_arch = "wasm32"))`; wasm uses
  only the storage-free `NyxdGlobalDataFetcher`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6939)


common/bandwidth-fetcher/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+[package]
+name = "nym-bandwidth-fetcher"
+description = "Crate for fetching zknym credentials"
+version.workspace = true
+authors.workspace = true
+edition = "2021"
```
jstuczyn, line 6: use workspace edition?

common/bandwidth-fetcher/src/error.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2023 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn, line 1: nit: you copied it over with wrong date

common/bandwidth-fetcher/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+use nym_ecash_time::OffsetDateTime;
+use nym_validator_client::coconut::{all_ecash_api_clients, EcashApiError};
+use nym_validator_client::nym_api::EpochId;
+use nym_validator_client::nyxd::contract_traits::DkgQueryClient;
+use nym_validator_client::EcashApiClient;
+
+#[cfg(all(not(target_arch = "wasm32"), feature = "recovery"))]
+pub use credentials::recovery::NyxdRecoveryFetcher;
+#[cfg(not(target_arch = "wasm32"))]
+pub use credentials::NyxdCredentialFetcher;
```
jstuczyn, line 17: same comment as with previous PR, keep target locked exports/imports (and modules) towards the bottom. I think it makes them more readable

common/bandwidth-fetcher/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+pub use public_data::NyxdGlobalDataFetcher;
+
+use crate::error::NyxdFetcherError;
+
+// credential issuance/recovery is backed by a sqlite pending-requests store, so it's native-only.
+// wasm only uses the storage-free `NyxdGlobalDataFetcher`.
```
jstuczyn, line 23: technically we could have also implemented it for wasm, but I get your feelings about that. it's not worth the effort. at least at this point

common/bandwidth-fetcher/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        // the fetch is idempotent.
+        let cache =
+            Arc::new(EcashApiClientsCacheInner::from_dkg_client(query_client, epoch_id).await?);
+        let clients = cache.clients.clone();
+        self.inner.store(Some(cache));
+        Ok(clients)
+    }
+}
+
+struct EcashApiClientsCacheInner {
```
jstuczyn, line 69: shouldn't this really be a `HashMap<EpochId, AllTheInfoStuff>` in case we (ever...) do a DKG rerun and have to retrieve data across multiple epochs? I know it's an edge case, but shouldn't be too difficult to implement

common/bandwidth-fetcher/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        C: DkgQueryClient,
+    {
+        // fast path: atomic load, return if cached for this epoch
+        if let Some(cache) = self.inner.load_full() {
+            if cache.epoch_id == epoch_id && !cache.is_stale() {
+                return Ok(cache.clients.clone());
+            }
+        }
+
+        // empty or stale - refresh and atomically swap in the new cache, then return it.
```
jstuczyn, line 58: we should probably have some `AtomicBool` flag here to make sure we don't have multiple parallel requests attempting to update the cache at the same time. if there's one in progress, don't start the other one. we use similar pattern in different locations already

common/bandwidth-fetcher/src/public_data.rs
```diff
@@ -0,0 +1,147 @@
...
+{
+    // try apis in pseudorandom way to remove any bias towards the first registered dealer
+    apis.shuffle(&mut thread_rng());
+
+    for api in apis {
+        let disp = api.to_string();
+        match f(api).await {
+            Ok(res) => return Ok(res),
+            Err(err) => {
+                warn!("failed to obtain valid response from API {disp}: {err}")
```
jstuczyn, line 142: given some of our APIs are down, this might create a lot of noise. perhaps use debug rather than warn? what do you think?

common/bandwidth-fetcher/src/storage/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    pub(crate) async fn reset(self) -> Result<(), StorageError> {
+        // First we close the storage to ensure that all files are closed
+        tracing::debug!("Closing pending credential requests storage");
+        self.storage_manager.close().await;
+
+        // Calling close on the storage should be enough to ensure that all files
+        // are closed but just to be sure we wait a bit
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
jstuczyn, line 78: why do we have to do it? isn't call to close() enough?

common/bandwidth-fetcher/src/storage/models.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2024 - Nym Technologies SA <contact@nymtech.net>
```
jstuczyn, line 1: as with other places, wrong date (look through new storage files)

common/bandwidth-fetcher/migrations/20260701120000_add_ecash_tables.sql
```diff
@@ -0,0 +1,19 @@
...
+
+-- data for ticketbooks that have an associated deposit, but failed to get issued
+CREATE TABLE pending_issuance 
+(
+    deposit_id INTEGER NOT NULL PRIMARY KEY,
+    
+    -- introduce a way for us to introduce breaking changes in serialization of data
+    serialization_revision INTEGER NOT NULL,
+    
+    pending_ticketbook_data BLOB NOT NULL UNIQUE,
```
jstuczyn, line 15: it feels weird to have an UNIQUE constraint on the data. wouldn't it be guaranteed through the deposit primary key?

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+};
+
+/// Obtains ticketbooks by depositing on-chain and aggregating wallet signatures from the ecash
+/// APIs, and recovers ticketbooks from deposits whose issuance never completed. Deposits are
+/// serialised via `deposit_lock` so concurrent fetches don't race the account sequence number.
+///
+// Batching them is TODO, it requires extensive changes to the BC
+pub struct NyxdCredentialFetcher<C> {
+    client: Arc<C>,
+    client_id: Zeroizing<Vec<u8>>,
```
jstuczyn, line 44: question: do we have to zeroize client id? isn't it just some public value?

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /// Like [`Self::new`], but places the pending-request DB at the default filename inside `data_dir`.
+    pub async fn new_from_dir(
+        client: Arc<C>,
+        data_dir: impl AsRef<Path>,
+        client_id: Zeroizing<Vec<u8>>,
+    ) -> Result<Self, NyxdFetcherError> {
+        let mut db_path = data_dir.as_ref().to_path_buf();
+        db_path.push("credential_requests.db");
```
jstuczyn, line 86: why not provide the entire path to db in the constructor? we're doing this everywhere. providing just data dir is a bit confusing.
furthermore, the .db file should be set in some constant maybe

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+            }
+
+            match self.obtain_ticketbook(&issuance.pending_ticketbook).await {
+                Err(err) => error!("could not recover deposit {deposit} due to: {err}"),
+                Ok(issued) => {
+                    recovered_books.push(NymCredential::Ticketbook(Box::new(issued)));
+                    info!("managed to recover deposit {deposit}! the ticketbook has been added to the storage");
+                    self.pending_storage
+                        .remove_pending_ticketbook(issuance.pending_id)
+                        .await?;
```
jstuczyn, line 158: should we propagate the error here? we have managed to recover some ticketbooks, but are throwing them away because we run into a db issue

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+        let epoch_id = self.client.get_current_epoch().await?.epoch_id;
+        let threshold = self
+            .client
+            .get_current_epoch_threshold()
+            .await?
+            .ok_or(NyxdFetcherError::NoThreshold)?;
+
+        let apis = self.ecash_api_clients.get(&*self.client, epoch_id).await?;
+
+        log::info!("Querying wallet signatures");
```
jstuczyn, line 184: don't mix `log` and `tracing`

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+        // `make_ticketbook_deposit` broadcasts-and-waits-for-commit, so by the time the tx returns
+        // the sequence has advanced on-chain - holding the lock across it is enough.
+        let _deposit_guard = self.deposit_lock.lock().await;
+
+        let mut rng = OsRng;
+        let signing_key = ed25519::PrivateKey::new(&mut rng);
+
+        let deposit_amount = self.client.get_default_deposit_amount().await?;
+        info!("we'll need to deposit {deposit_amount} to obtain the ticketbook");
+        let result = self
```
jstuczyn, line 211: I'd move the lock to here, i.e. after we had already make chain query for deposit amount, so that we wouldn't need to make for another async call before we could actually start making the tx

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+        // (plus a bit more for tx fees)
+        let raw_deposit_cost = self.client.get_default_deposit_amount().await?;
+        let per_deposit_cost_amount = raw_deposit_cost.amount.u128() + Self::DEPOSIT_TX_FEE_AMOUNT;
+        let total_deposits_cost = Coin::new(
+            per_deposit_cost_amount * nb_deposits,
+            raw_deposit_cost.denom,
+        );
+
+        info!(
+        "this will require {nb_deposits} deposits that will cost approximately {total_deposits_cost}"
```
jstuczyn, line 254: seems like some formatting failed here

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+                self.pending_storage
+                    .insert_pending_ticketbook(&issuance_data)
+                    .await
+                    .inspect_err(|err| {
+                        let deposit = issuance_data.deposit_id();
+                        error!("could not save the recovery data for deposit {deposit}: {err}. the data will unfortunately get lost")
+                    })
+                    .map_err(NyxdFetcherError::from)?;
+
+                Ok(vec![])
```
jstuczyn, line 359: everything failed, yet we still return an `Ok`?

common/bandwidth-fetcher/src/credentials.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+}
+
+#[cfg(feature = "recovery")]
+pub(crate) mod recovery {
+    use super::*;
+    use std::ops::Deref;
+
+    /// Recover-only view over [`NyxdCredentialFetcher`]: reuses all of its query/recovery logic via
+    /// `Deref`, but its `CredentialFetcher` impl only recovers pending deposits (never makes new
```
jstuczyn, line 383: don't abuse `Deref`. just access things via the internal field. the trait should really only be used for smart pointers

common/credential-storage/src/ephemeral_storage.rs
```diff
@@ -121,18 +111,18 @@ impl Storage for EphemeralStorage {
         Ok(self.storage_manager.get_ticketbooks_info().await)
     }
 
-    async fn get_pending_ticketbooks(
-        &self,
-    ) -> Result<Vec<RetrievedPendingTicketbook>, Self::StorageError> {
-        Ok(self.storage_manager.get_pending_ticketbooks().await)
-    }
+    // async fn get_pending_ticketbooks(
```
jstuczyn, line 114: why not just remove them?

common/bandwidth-fetcher/src/storage/mod.rs
simonwicky: That is a question for @trojanfoe and @pronebird

common/bandwidth-fetcher/src/credentials.rs
simonwicky: Yes, we have to, it's not public, it's the ecash secret

common/bandwidth-fetcher/src/credentials.rs
simonwicky: Hack for the vpn-client. The `.db` path is in a const, but not availabe to the vpn-client, it only has a dir

common/bandwidth-fetcher/src/lib.rs
simonwicky: Future work lol. We're closer than we used to be at least

common/bandwidth-fetcher/migrations/20260701120000_add_ecash_tables.sql
simonwicky: You tell me? I reused the bits that were used to create the initial tables in the CredentialStorage. Copied your stuff :P

common/bandwidth-fetcher/src/credentials.rs
simonwicky: Indeed, copied pasted to quickly

common/bandwidth-fetcher/src/credentials.rs
simonwicky: Nice catch

common/credential-storage/src/ephemeral_storage.rs
simonwicky: Forgot to remove them

common/bandwidth-fetcher/src/storage/mod.rs
pronebird: Technically it should be enough.

pronebird (commented):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/6939)* status: 0 of 28 files reviewed, 19 unresolved discussions (waiting on jstuczyn, pronebird, simonwicky, and trojanfoe).

common/bandwidth-fetcher/src/credentials.rs
simonwicky: Too hacky indeed. fixed

PR #6940, opened 2026-07-06
[BandwidthController] wiring and cleanup

simonwicky:
## Wire the new fetcher into consumers; remove dead credential code

Final PR of the stack. Migrates every consumer onto the
`BandwidthController` + `nym-bandwidth-fetcher` API and deletes the code the new
design replaces.

### Removals (net −900 LOC)
- Deletes the whole **`common/credential-utils`** crate (`issue_credential` &
  friends) — dropped from the workspace and `Cargo.toml`.
- Deletes the controller's old inline modules: `acquire/`, `fetcher/`
  (`nyxd_public_data.rs`), and `utils.rs`.

### Consumer wiring
- **`nym-sdk`** — `BandwidthAcquireClient` now holds a `BandwidthController`
  built with a `NyxdCredentialFetcher` instead of calling `issue_credential`
  directly; `new()` is now async, and the nyxd client is shared via `Arc`.
- **`nym-api`** (network monitor), **`nym-gateway-probe`**, and
  **`common/commands`** (`issue_ticket_book`, `recover_ticket_book`) migrated to
  the new fetcher/controller API.
- **`client-core`** — depends on the new crate; disk-persistence config bumped
  (new field, `old_config` variants updated) and `base_client` wiring adjusted.
- **`nym-node`** — persistence config field added across old-config v4/v5/v6.

### Misc
- Workspace: adds `common/bandwidth-fetcher`, drops `common/credential-utils`,
  bumps `si-scale` 0.2 → 0.3, `Cargo.lock` updated.

### Review order
Stacked on top of `nyxd_fetcher` (→ `fetching_logic` → `bandwidth_controller_step_2`).
Best reviewed after the two parent PRs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6940)


common/commands/src/ecash/recover_ticket_book.rs
```diff
@@ -32,10 +36,17 @@ pub async fn execute(args: Args, client: QueryClient) -> anyhow::Result<()> {
...
+    let requests_store = loaded.try_get_credential_requests_store()?;
     let persistent_storage = initialise_persistent_storage(credentials_store).await;
 
-    let recovered = utils::recover_deposits(&client, &persistent_storage).await?;
+    let fetcher = NyxdRecoveryFetcher::new(Arc::new(client), requests_store).await?;
+    let controller = BandwidthController::new(persistent_storage).with_credential_fetcher(fetcher);
 
-    println!("recovered {recovered} ticketbooks");
+    // the recovery fetcher recovers pending deposits per ticket type, so sweep them all
+    for ticketbook_type in TicketType::iter() {
```
jstuczyn, line 46: I think this will make you fetch mixnet EXIT tickets which you really don't want

common/commands/src/utils.rs
```diff
@@ -151,6 +151,21 @@ impl CommonConfigsWrapper {
...
+        match self {
+            CommonConfigsWrapper::NymClients(cfg) => {
+                Ok(cfg.storage_paths.inner.credential_requests_database.clone())
+            }
+            CommonConfigsWrapper::NymApi(cfg) => Ok(cfg
+                .network_monitor
+                .storage_paths
+                .credentials_request_path
+                .clone()),
+            CommonConfigsWrapper::Unknown(_cfg) => todo!(),
```
jstuczyn: missing impl?

nym-node/src/config/persistence.rs
```diff
@@ -332,6 +332,7 @@ impl NetworkRequesterPaths {
 
             // not needed for embedded providers
             credentials_database: Default::default(),
+            credential_requests_database: Default::default(),
```
jstuczyn, line 335: (as well as other old configs) - doesn't this default to an empty path?

jstuczyn (approved):

common/commands/src/ecash/recover_ticket_book.rs
simonwicky: In that precise case, we're trying to recover everything. 
Sure there is no reason to have exit ticket books, but if we do, we want to recover them. 
We are using the `NyxdRecoveryFetcher` so it will never attempt to get some, just recover them

common/commands/src/utils.rs
simonwicky: :homer-bush:

nym-node/src/config/persistence.rs
simonwicky: Yes, but embedded clients will never need those (just like they never need credentials_database

PR #6945, opened 2026-07-07
openspecs for the NS API, NS API agent and gateway probe

mmsinclair:
Adds openspecs for the:

- NS API
- NS API agents
- gateway probe

This is ahead of work to refactor them and change how the tests are done

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6945)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive specifications for the node-status API, monitoring, persistence, test-run orchestration, ticket management, agent, and gateway probe.
  * Documented gateway diagnostics across mixnet, WireGuard, dVPN, SOCKS5, and port checks.
  * Documented CLI workflows, authentication, assignments, result formats, API endpoints, scheduling, retries, timeouts, and operational behavior.
  * Added architecture and end-to-end flow documentation covering measurement, signed result submission, and data storage.


nym-gateway-probe/docs/gateway-probe.md
```diff
@@ -0,0 +1,233 @@
...
+    API-->>Agent: 201 Created
+```
+
+### Key interaction rules
+
+- **Assignment is gated.** The NS API only assigns a test run if the request is authenticated and *fresh* (timestamp within the freshness window, default 120s), the in-progress count is below `agent_max_count`, and there are enough ecash ticketbooks of every buffered type. It claims the **oldest queued** run for a bonded gateway (performance > 0) atomically with `FOR UPDATE ... SKIP LOCKED`, so two agents never get the same run.
+- **Ticket materials pay for the tests.** The NS API's ticketbook manager keeps a buffer of threshold-signed ecash ticketbooks; on assignment it attaches materials the probe imports to pay for mixnet/WireGuard bandwidth during the run.
+- **Log capture is bracketed.** The agent captures only the tracing output emitted between `start()` and `stop_and_drain()` — i.e. during the probe run — and submits it alongside the result.
+- **Submission fans out with version asymmetry.** The primary server gets the v1 endpoint (log truncated to 1024 bytes); secondary servers get the v2 endpoint (full log + gateway identity key). Per-server failures are logged but don't abort the others.
```
jstuczyn, line 226: I guess that's something we want to fix, i.e. they should be using the same endpoint, no?

nym-gateway-probe/docs/gateway-probe.md
```diff
@@ -0,0 +1,233 @@
...
+- **Submission is validated.** The NS API only completes a run if it is in-progress and the submitted `assigned_at_utc` matches the stored assignment. On the v2 endpoint an unknown test-run id causes the API to create an "external" gateway + run before recording the result.
```
jstuczyn, line 227: yes, that bit is quite problematic and I think it could use a bit more of A1 explanation. essentially if some test takes too long, it will internally time out and is going to get re-assigned to a different agent. However, when the original one comes back around with the result, it will get rejected

openspec/specs/architecture/spec.md
```diff
@@ -0,0 +1,58 @@
...
+Each background worker MUST run its own `tokio` timed loop rather than a shared scheduler. On a failed cycle a worker MUST sleep a short retry delay (e.g. 60s for the monitor and metrics scraper) and retry; on success it MUST sleep its configured refresh interval. All loops MUST honour graceful shutdown via the shutdown manager.
+
+#### Scenario: Monitor retries after a failed cycle
+- **GIVEN** the network monitor loop
+- **WHEN** a monitor cycle returns an error
+- **THEN** the loop sleeps the failure-retry delay (60s) and retries, rather than aborting the process
+
+### Requirement: The agent SHALL run as a one-shot CLI, looped externally
+
+The agent process MUST parse its arguments once, execute a single subcommand, and exit. Continuous operation MUST be achieved by an external loop (the container `entrypoint.sh` runs it in a `while true` loop with a `SLEEP_TIME` interval). The agent MUST NOT implement its own daemon loop or internal retry across invocations.
```
jstuczyn: honestly, if we'd be moving gateway proxy things around, I'd loosen that requirement or maybe completely remove it.

openspec/specs/gateway-probe/spec.md
```diff
@@ -0,0 +1,170 @@
...
+> Generated by openlore on 2026-07-07
+> Source files: `nym-gateway-probe/src/main.rs`, `run.rs`, `lib.rs`, `config/{mod,credentials,netstack,socks5,test_mode}.rs`, `common/{mod,types,nodes,probe_tests,netstack,wireguard,icmp,helpers,bandwidth_helpers}.rs`, `common/socks5_test/{mod,json_rpc_client}.rs`, `build.rs`, `Cargo.toml`
+
+## Purpose
+
+`nym-gateway-probe` is a Unix-only CLI diagnostic tool and reusable library that tests the health and connectivity capabilities of Nym mixnet gateway nodes. It connects to a gateway (resolved via the `nym-api` directory, or queried directly by IP for unannounced nodes) and exercises multiple transport paths — mixnet entry/exit routing, WireGuard-over-mixnet, Lewes Protocol registration, SOCKS5 network-requester connectivity, and WireGuard exit-policy TCP port reachability — then emits a structured JSON result.
+
+## Requirements
+
+### Requirement: The probe SHALL expose four run modes selected by subcommand
```
jstuczyn, line 12: just thinking aloud here, I don't think we need the complexity of having to support all the modes. @simonwicky can correct me if I'm wrong, but from what I recall, in particular, `RunLocal` was quite annoying to update to "work" with the current structure and I'm not sure it even supports all the features, like LP

openspec/specs/gateway-probe/spec.md
```diff
@@ -0,0 +1,170 @@
...
+A `TestMode` (default `Core`) MUST gate execution: mixnet tests run in `Core`/`All`; WireGuard tests run in `Core`/`WgMix`/`WgLp`/`All`; LP tests run in `WgLp`/`LpOnly`/`All`; SOCKS5 tests run in `Socks5Only`/`All`. The `mixnet` spelling MUST be accepted as an alias for `core`, and parsing MUST be case-insensitive across accepted spellings.
+
+#### Scenario: LP-only mode skips mixnet and SOCKS5
+- **GIVEN** `--test-mode lp-only`
+- **WHEN** the probe runs
+- **THEN** only LP tests are attempted and mixnet-routing and SOCKS5 phases are skipped
+
+### Requirement: Test phases SHALL run in a fixed order with the mixnet client disconnected before LP/SOCKS5
+
+`do_probe_test` MUST sequence phases as: mixnet ping → WireGuard (holding the mixnet client) → **disconnect the mixnet client** → LP registration → SOCKS5. Each phase MUST be gated by its `TestMode` predicate, and a failure in one phase MUST NOT prevent later phases (SOCKS5 errors in particular are logged and non-fatal); WG/LP results fall back to defaults on failure.
```
jstuczyn, line 48: is this actually what we want? I think it exposes an issue after socks5 has been introduced. since socks5 does require a mixnet client, it should perform that test before disconnecting and doing LP...

openspec/specs/gateway-probe/spec.md
```diff
@@ -0,0 +1,170 @@
...
+- **GIVEN** the entry gateway is under test
+- **WHEN** the self-ping through it fails
+- **THEN** `as_entry` is recorded as `fail_to_connect`
+
+#### Scenario: Exit routing recorded per address family
+- **GIVEN** a reachable exit router
+- **WHEN** ICMP echoes are sent to tun-device and external v4/v6 targets
+- **THEN** the `Exit` result records `can_route_ip_v4`, `can_route_ip_external_v4`, `can_route_ip_v6`, and `can_route_ip_external_v6` from the observed replies
+
+### Requirement: WireGuard tests SHALL register with the authenticator and run tunnel diagnostics via the Go netstack
```
jstuczyn, line 104: when migrating things, shall we make our lives simpler by removing the authenticator? at this point all nodes should support LP registration. if they don't, I'm not sure we want to be using them (I don't think VPN app even allows authenticator registrations anymore)

openspec/specs/node-status-agent/spec.md
```diff
@@ -0,0 +1,104 @@
...
+- **THEN** it logs the condition and exits successfully without running the probe or submitting anything
+
+#### Scenario: Logs captured only during the probe run
+- **GIVEN** an assigned test run
+- **WHEN** the agent brackets the probe run with `log_capture.start()` / `stop_and_drain()`
+- **THEN** only tracing output emitted during the run is captured and submitted with the result
+
+### Requirement: The run-probe submission SHALL use v1 for the primary and v2-with-context for secondaries
+
+When submitting probe results, the agent MUST send the primary server (index 0) a v1 `submit_results` (probe log truncated to 1024 bytes, `POST /internal/testruns/{id}`), and each secondary server (index > 0) a v2 `submit_results_with_context` (`POST /internal/testruns/{id}/v2`, including the gateway identity key, log not truncated). Per-server submission failures MUST be logged as warnings and MUST NOT abort submission to the other servers.
```
jstuczyn, line 56: as mentioned before, we should probably revisit it... note: I have introduced log truncation in May when our probes started failing all of the sudden (and it seemed to ever so slightly make it better). I don't think we're using those logs anywhere, anyway. at least not on harbourmaster, so maybe if we redesign it, we could remove them?

openspec/specs/node-status-api-http/spec.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Purpose
+
+Defines the public read HTTP API of the node status service: an axum server that serves gateway, mixnode, nym-node, dVPN directory, services, summary, metrics, and status data. Read endpoints are served from an in-memory cache backed by PostgreSQL, with OpenAPI/Swagger documentation. (The authenticated `/internal/testruns` endpoints are specified in [[node-status-api-testruns]].)
+
+## Requirements
+
+### Requirement: The API SHALL serve read endpoints from a TTL cache backed by the database
+
+Read-heavy list endpoints (gateways, dVPN gateways, gateway IPs, nym-nodes, mixnode stats, summary history, session stats) MUST be served from per-list `moka` caches with capacity 1 and TTL = the configured `nym_http_cache_ttl` (default 30s). On a cache miss the value MUST be rebuilt from the database and cached. Pagination MUST default to size 10 (max 200) and page 0, returning a `PagedResult { page, size, total, items }`.
```
jstuczyn: "Pagination MUST default to size 10"
that's very aggressive pagination that could be quite detrimental, because naive clients will just use the default meaning they will send a lot of requests to retrieve all the data and pay the TLS handshake cost every single time. 
and one of these naive clients is, you guessed it, our vpn api: https://github.com/nymtech/websites/blob/main/www/vpn-api/src/app/api/public/v1/directory/gateways/getGatewaysFromDirectory.ts#L31

so to actually retrieve all the gateways it makes like 70 queries 😬

openspec/specs/node-status-api-http/spec.md
```diff
@@ -0,0 +1,90 @@
...
+
+The `/v2/services` endpoint MUST derive each service's IP, hostname, and service-provider client id from the gateway `self_described` JSON using JSONPath expressions (host IP, hostname, network-requester address). The default filter MUST keep only gateways that advertise a network-requester service provider; `entry=true` MUST keep all; `wss` and `hostname` filters MUST further restrict the set.
+
+#### Scenario: Default services filter
+- **WHEN** `GET /v2/services` is called with no filters
+- **THEN** only gateways advertising a network-requester service provider are returned
+
+### Requirement: Probe results SHALL deserialize with schema fallback and be scored for display
+
+Stored probe results MUST deserialize by first attempting the latest `nym_gateway_probe` format, then falling back to the crate's older schema, back-filling WireGuard v4 fields into generic fields for compatibility. The API MUST compute a visual gateway score weighting mixnet 40% / download-speed 30% / ping 30% into `Offline|Low|Medium|High`, and bucket gateways into SOCKS5 percentile scores by HTTPS latency (latency 0 → Offline; lowest 50% → High, next 25% → Medium, worst 25% → Low).
```
jstuczyn: what is this scoring? should it even be part of NS?

openspec/specs/node-status-api-monitoring/spec.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Requirements
+
+### Requirement: The network monitor SHALL periodically snapshot the directory, geodata, and delegations
+
+The monitor MUST run on a loop (default refresh 300s; 60s retry on failure). Each cycle MUST: check the ipinfo bandwidth allowance, fetch described nodes v2 from `nym-api`, classify gateways by declared entry/exit-IPR role, fetch bonded/basic/mixing nodes and families, write nym-nodes and families to the database, refresh geodata for every node (geolocating each IP via ipinfo, caching the first success with TTL `geodata_ttl`, else an empty location), build and write gateway records (including an `explorer_pretty_bond` with location and performance), refresh per-node delegations from `nyxd`, and write the network summary (8 summary keys plus `NetworkSummary`). A `run_once` early-exit mode MUST write only nodes and families.
```
jstuczyn: do we really have to do a ipinfo/geolocation lookup every 300s? seems like a huge overkill. how about only do it if
A) it's a new gateway we know nothing about
B) its IP address has updated
C) once a day/week otherwise

also, we should probably move it to a separate, dedicated service (that would **eventually** submit it to contracts)

openspec/specs/node-status-api-monitoring/spec.md
```diff
@@ -0,0 +1,67 @@
...
+- **THEN** the stored description has the HTML sanitized away
+
+#### Scenario: Empty moniker gets a deterministic name
+- **GIVEN** a node whose self-reported moniker is empty
+- **WHEN** the description is stored
+- **THEN** a deterministic three-word name derived from the node id is used
+
+### Requirement: The packet-stats scraper SHALL fetch per-node stats and accumulate daily deltas
+
+The node (packet-stats) scraper MUST run every 1 hour with concurrency from `packet_stats_max_concurrent_tasks` (default 10). Per node it MUST fetch `/stats` (parsing both legacy-mixnode and nym-node formats) and, for entry/exit gateways, fetch bridge client-params. Results MUST be batch-stored, and daily deltas MUST be computed against the previous raw stats row; on a counter decrease (node restart) it MUST use the current value rather than a negative delta.
```
jstuczyn: this should really do some post-processing and remove any outliers, because remember those are the Sve... self-declared endpoints that could easily be polluted

openspec/specs/node-status-api-persistence/spec.md
```diff
@@ -0,0 +1,80 @@
...
+- **THEN** it returns an internal error rather than a partial `NetworkSummary`
+
+### Requirement: Node classification SHALL distinguish gateways from mixing nodes for scraping and writes
+
+`get_nodes_for_scraping` MUST classify a node as an entry/exit nym-node when its ed25519 key matches a bonded gateway, else as a mixing nym-node; scraped descriptions for entry/exit nodes MUST be written to both the nym-node and gateway description tables.
+
+#### Scenario: Gateway node written to both tables
+- **GIVEN** a scraped node whose key is a bonded gateway
+- **WHEN** its description is stored
+- **THEN** the description is written to both the nym-node and gateway description tables
```
jstuczyn: I think this is unnecessary complexity and probably source of problems

openspec/specs/node-status-api-testruns/spec.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Purpose
+
+Defines gateway test-run orchestration: a background queuer that enqueues probe and ports-check test runs for bonded gateways, and the authenticated `/internal/testruns` API that assigns test runs to external agents (attaching ecash ticket materials) and ingests their signed results.
+
+## Requirements
+
+### Requirement: The background queuer SHALL enqueue idempotent probe test runs and re-queue stale ones
+
+The test-run queuer MUST loop: re-queue stale in-progress test runs (those assigned before the stale window, default 7200s), then iterate all gateways ordered by last-testrun time, filter to bonded gateways, and queue a probe run per gateway, then sleep the refresh interval (default 450s). Enqueue MUST be idempotent — if a non-complete Probe run already exists for a gateway it MUST be reused rather than duplicated; otherwise a new `Queued`/`Probe` row is inserted. Unknown gateways MUST be rejected.
```
jstuczyn, line 15: " if a non-complete Probe run already exists for a gateway it MUST be reused rather than duplicated;"

I wonder if it would be better if it was duplicated after all. worst case, we'd just have an additional test run information

openspec/specs/overview/spec.md
```diff
@@ -0,0 +1,36 @@
+# overview Specification
+
+> Generated by openlore on 2026-07-07
+> Source files: `nym-gateway-probe/`, `nym-node-status-api/nym-node-status-api/`, `nym-node-status-agent/`, `nym-node-status-client/`
+
+## Purpose
+
+This overview describes a three-part subsystem that measures and reports the health of the Nym mixnet. The **node status API** persists a snapshot of the network and serves it over HTTP, while also orchestrating gateway "test runs". External **agents** poll the API for assigned test runs, execute network diagnostics against a target gateway using the **gateway probe** library, and submit signed results back. Together they produce the connectivity/performance data that powers dVPN gateway directories and network dashboards.
```
jstuczyn: > that measures and reports the health of the Nym mixnet

well, not just the mixnet, but also, and mostly, dvpn

jstuczyn (approved):

PR #6953, opened 2026-07-11
SDK: add smol-dvpn to provide per app user space tunnels

mmsinclair:
## Summary

Adds `smol-dvpn`: a pure-Rust, userspace WireGuard dVPN datapath that lets a Rust
app tunnel **its own** traffic over the Nym network — 1-hop or 2-hop — with **no
OS `tun` device and no root**. Traffic goes in through ordinary tokio surfaces
(`TcpStream`, `UdpSocket`, `AsyncRead + AsyncWrite`), so `tonic`/`hyper`/`reqwest`
work inside the tunnel unchanged. Because the tunnel is scoped to the app rather
than the whole OS, closing it cuts the app off — a natural per-app kill-switch.

## What's added

Three new crates:

- **`common/smol-core`** — transport-agnostic userspace TCP/IP stack (smoltcp):
  turns a stream of IP packets into `TcpStream` / `UdpSocket` / DNS. Extracted so
  both `smolmix` (5-hop mixnet) and `smol-dvpn` (dVPN) share it.
- **`sdk/rust/nym-sdk-session`** — shared provisioning facade: mnemonic-funded
  zk-nym ticketbook issuance, persistent credential storage, and gateway
  selection/registration. Used by both mixnet and dVPN modes.
- **`sdk/rust/smol-dvpn`** — the WireGuard datapath (`boringtun`) and transport
  strategy.

## Capabilities

- **Three data-plane modes:** one-hop, two-hop (nested `Tunn`s), and
  QUIC-tunnelling two-hop — the entry leg rides a QUIC bridge (via the
  `nym-bridges` client) to evade DPI / UDP blocking.
- **Decentralised, per-hop identity:** each gateway is registered independently
  and issues its own WireGuard identity — no centralised shared public key.
- **Unlinkable payment:** zk-nyms break the link between paying for and using the
  network. Builders can gift credentials to end-users via `nym-credential-proxy`
  so users never hold NYM.
- Gateway selection by identity key, two-letter country code, or random; DNS
  in-tunnel by default; configurable, runtime-adjustable per-hop MTU;
  `CancellationToken` lifecycle.

## Examples (`sdk/rust/smol-dvpn/examples/`)

`smol-dvpn-config` (export a plain WG config), `smol-dvpn-topup`, `smol-dvpn-grpc`
(gRPC through the tunnel), `two-hop-ip` (prove IP relocation), `two-hop-quic`
(QUIC-bridged), and `zcash-sync` (throughput benchmark vs. direct).

## Docs & specs

- Architecture docs at `docs/design/sdk/smol-dvpn/`.
- Five OpenSpec capability specs synced into `openspec/specs/` (`dvpn-tunnel`,
  `dvpn-quic-bridge`, `dvpn-tools`, `dvpn-session`, `smol-core-stack`); the
  `nym-sdk-dvpn` change is archived.

## Testing

- `cargo test -p smol-core` — stack integration tests (TCP/UDP/DNS).
- `cargo test -p nym-smol-dvpn` — QUIC bridge conformance (framing + ed25519-SPKI
  pinning). End-to-end tunnel bring-up is validated against a live Nym gateway via
  the examples (needs a funded mnemonic + network).

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6953)

## Summary by CodeRabbit

* **New Features**
  * Added a pure-Rust dVPN SDK supporting one-hop, two-hop, and optional QUIC-bridged tunnels without an OS TUN device or root privileges.
  * Added tunnel-compatible TCP, UDP, DNS, HTTP, and gRPC connectivity.
  * Added session provisioning, gateway selection, credential storage, cancellation, runtime MTU changes, and bandwidth top-ups.
  * Added WireGuard configuration export, connectivity checks, top-up tools, and performance examples.

* **Documentation**
  * Added SDK architecture, configuration, usage, and design documentation.

* **Compatibility**
  * Preserved the existing smolmix public API while adopting the new networking stack.


docs/design/smoldvpn/design.md
```diff
@@ -0,0 +1,504 @@
...
+  `Vec<u8>` into `tcp_connect() -> TcpStream`, `udp_socket() -> UdpSocket`, and a
+  DNS resolver. `smolmix` (5-hop) and `nym-smol-dvpn` (2-hop) both sit on it.
+- **`nym-sdk-session`** wraps `nym-registration-client` +
+  `nym-bandwidth-controller` + the credential store. It answers "acquire paid
+  access to these gateways" for **both** mixnet and dvpn modes, so it lives beside
+  the mixnet client rather than inside `nym-smol-dvpn`.
+- **`nym-smol-dvpn`** owns only the WireGuard datapath (boringtun) and the
+  transport strategy.
+- **Dependency isolation (required):** every *new* dependency `nym-smol-dvpn` needs
+  — `boringtun`, `quinn`, `quinn-proto`, `nym_bridges` (git, pinned), and anything
```
mfahampshire, line 69: todo remove once bridges published

mfahampshire (commented):
smol-* changes lgtm

common/smol-core/src/stack.rs
```diff
@@ -0,0 +1,168 @@
...
+        Self {
+            ipv4,
+            ipv6: None,
+            mtu: DEFAULT_MTU,
+            tcp_buffer: DEFAULT_TCP_BUFFER,
+        }
+    }
+
+    /// Set the assigned IPv6 address.
+    pub fn with_ipv6(mut self, ipv6: Ipv6Addr) -> Self {
```
jstuczyn, line 67: given lack of dual-stack support (for now) we should mark this method with a warning, because it will not do anything

common/smol-core/src/stack.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.ipv6 = Some(ipv6);
+        self
+    }
+
+    /// Set the interface MTU.
+    pub fn with_mtu(mut self, mtu: usize) -> Self {
```
jstuczyn: nit: all the `with_XXX` methods should be marked with `#[must_use]`

common/smol-core/src/stack.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        Self {
+            net: Net::new(device, net_config),
+            dns: DnsConfig::default(),
+            config,
+        }
+    }
+
+    /// Override the tunnel DNS configuration (upstream server, timeout).
+    pub fn with_dns_config(mut self, dns: DnsConfig) -> Self {
```
jstuczyn, line 120: same nit about `#[must_use]`

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Resolve `host` to a list of IP addresses over the given stack, querying both
+/// A and AAAA records. Every DNS packet is sent through a stack UDP socket.
+pub(crate) async fn resolve(net: &Net, cfg: &DnsConfig, host: &str) -> Result<Vec<IpAddr>> {
+    let name = Name::from_utf8(host).map_err(|e| SmolCoreError::DnsProto(e.to_string()))?;
+
+    let mut addrs = Vec::new();
+    // A and AAAA are independent queries; collect whatever resolves.
+    for rtype in [RecordType::A, RecordType::AAAA] {
+        match query_one(net, cfg, &name, rtype).await {
```
jstuczyn, line 67: possible improvement: we could make the A and AAAA queries concurrently given most of the time is going to be spent waiting on IO

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+/// Default upstream DNS server used inside the tunnel (Cloudflare).
+pub const DEFAULT_DNS_SERVER: SocketAddr =
+    SocketAddr::new(IpAddr::V4(std::net::Ipv4Addr::new(1, 1, 1, 1)), 53);
+
+/// Default per-query timeout.
+pub const DEFAULT_QUERY_TIMEOUT: Duration = Duration::from_secs(5);
+
+// Monotonic-ish DNS transaction id source (single in-flight query per socket,
+// but distinct ids avoid confusing a resolver that reuses a socket).
+static TXN_ID: AtomicU16 = AtomicU16::new(1);
```
jstuczyn: **`static TXN_ID: AtomicU16` — the id is set but never validated**

This global counter is thread-safe and not a live bug, but it's a smell worth addressing because the transaction id it produces is effectively write-only.

`query_one` stamps `id` on the outgoing query (`dns.rs:81-83`) but:
- never compares `response.id()` against it, and
- discards the response source (`let (len, _src) = ...`, `dns.rs:96`), so it doesn't check the datagram came from `cfg.server`.

What actually correlates request↔response is the **socket**: each query binds a fresh ephemeral UDP socket with a single in-flight query (`dns.rs:90-92`), and smoltcp only delivers datagrams addressed to that port. So the id does no work today, and the comment at `dns.rs:28-29` ("single in-flight
query per socket") confirms it doesn't need to be unique or monotonic.

Two concerns:
1. A predictable, monotonic, process-global id is exactly what you'd *not* want if someone later adds id validation assuming it gives spoof resistance — classic DNS randomizes the id precisely for off-path entropy. This is a false-security trap waiting to be sprung.
2. It's shared mutable global state buying zero functional benefit.

The practical spoofing risk is low today (the socket only receives what the tunnel delivers from the configured upstream), so this isn't urgent — but it's the kind of thing that rots into a real gap.

**Suggested fix — pick one coherent design instead of the current half-measure:**
- *Make the id meaningful (preferred):* validate `response.id() == id` **and** `src == cfg.server`, and generate the id randomly per query rather than from a global counter. Closes the in-tunnel spoofing gap and removes the global.
- *Or lean fully on socket demux:* if we trust the one-socket-one-query invariant, the id is arbitrary — use a constant or locally-random value and delete the `static` entirely.

Either way I'd drop the process-global `AtomicU16`.

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+        .answers
+        .iter()
+        .filter_map(|record| match &record.data {
+            RData::A(a) => Some(IpAddr::V4(a.0)),
+            RData::AAAA(aaaa) => Some(IpAddr::V6(aaaa.0)),
+            _ => None,
+        })
+        .collect();
+
+    if addrs.is_empty() {
```
jstuczyn: Server RCODE is never checked, so `SERVFAIL`, `REFUSED`, and `NXDOMAIN` all collapse into the same `DnsNoRecords` as a genuinely empty record set. Two consequences: a transient upstream `SERVFAIL` is indistinguishable from "no such record," and we lose the ability to retry intelligently (SERVFAIL/timeout are retryable; NXDOMAIN is not). Suggest inspecting `response.response_code()` and surfacing a distinct error for the failure codes.

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+        .to_vec()
+        .map_err(|e| SmolCoreError::DnsProto(e.to_string()))?;
+
+    let socket = net
+        .udp_bind("0.0.0.0:0".parse().expect("valid bind addr"))
+        .await?;
+    socket.send_to(&query_bytes, cfg.server).await?;
+
+    let mut buf = vec![0u8; 1500];
+    let (len, _src) = tokio::time::timeout(cfg.timeout, socket.recv_from(&mut buf))
```
jstuczyn: Single-shot: we read exactly one datagram and treat it as the answer, with no retry. Combined with the response id/source not being validated, any stray datagram delivered to this ephemeral port - or a spurious packet arriving before the real reply - defeats the lookup, and there's no second read to recover. The `timeout` only wraps this one `recv_from`, so it doesn't help. Consider a recv loop that validates (id + source) and keeps reading until a match or the timeout.

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+            name: name.to_utf8(),
+        })??;
+
+    let response =
+        Message::from_vec(&buf[..len]).map_err(|e| SmolCoreError::DnsProto(e.to_string()))?;
+
+    let addrs: Vec<IpAddr> = response
+        .answers
+        .iter()
+        .filter_map(|record| match &record.data {
```
jstuczyn: `query_one` returns AAAA results, but the stack is a single-IPv4 `/32` (dual-stack deferred per `StackConfig`). If A fails and AAAA succeeds, callers like `tcp_connect_host` (`addrs[0]`) will hand an IPv6 address to `tcp_connect` on a v4-only datapath and fail. Until dual-stack is wired, suggest filtering AAAA out (or preferring A) so we don't return unroutable addresses.

sdk/rust/nym-sdk-session/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::net::SocketAddr;
+
+/// Errors produced while provisioning a dVPN session.
+#[derive(thiserror::Error, Debug)]
+pub enum SessionError {
+    /// Failed to construct or talk to the nyxd chain client.
+    #[error("chain client error: {0}")]
+    Chain(String),
```
jstuczyn: in all instances of constructing this error we just convert `NyxdError` to string. why not instead just preserve it via `Chain(#[from] NyxdError)`?

sdk/rust/nym-sdk-session/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+    InvalidUrl {
+        which: &'static str,
+        url: String,
+        #[source]
+        source: url::ParseError,
+    },
+
+    /// Credential storage could not be opened.
+    #[error("credential storage error: {0}")]
+    Storage(String),
```
jstuczyn, line 34: this (and some other) variants are never constructed

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+    async fn select_inner(
+        &self,
+        spec: &GatewaySpec,
+        role: WgRole,
+        require_quic: bool,
+        exclude: Option<&ed25519::PublicKey>,
+    ) -> Result<SelectedGateway, SessionError> {
+        let nodes = self
+            .api
+            .get_all_described_nodes_v2()
```
jstuczyn: this is unnecessarily doubling the response size by pulling **ALL** information about **BOTH** gateways **AND** mixnodes - a lot of it is redundant

sdk/rust/nym-sdk-session/src/gateway.rs
```diff
@@ -0,0 +1,428 @@
...
+/// A node is usable for dVPN only if it advertises WireGuard, an authenticator,
+/// and LP data, and fulfils the required role.
+fn wg_capable(desc: &NymNodeDescriptionV2, role: WgRole) -> bool {
+    let d = &desc.description;
+    if d.wireguard.is_none() || d.lewes_protocol.is_none() || d.authenticator.is_none() {
+        return false;
+    }
+    match role {
+        WgRole::Entry => d.declared_role.entry,
+        WgRole::Exit => d.declared_role.can_operate_exit_gateway(),
```
jstuczyn: this restriction is not required as we don't distinguish between entry and exit dvpn nodes; that declared role only makes sense for the mixnet mode

sdk/rust/nym-sdk-session/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+nym-crypto = { workspace = true, features = ["asymmetric"] }
+nym-network-defaults = { workspace = true }
+nym-api-requests = { workspace = true }
+nym-authenticator-requests = { workspace = true }
+nym-kkt-ciphersuite = { workspace = true }
+nym-lp = { workspace = true }
+nym-lp-data = { workspace = true }
+nym-sphinx-addressing = { workspace = true }
+
+bip39 = { workspace = true }
```
jstuczyn, line 41: make sure to enable `zeroize` feature so that the mnemonic gets zeroize on drop

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+
+/// Number of tickets to reserve when checking for / spending a stored ticketbook.
+const TICKETS_TO_SPEND: u32 = 1;
+/// Timeout for nym-api requests.
+const API_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Configuration for creating a [`Session`].
+pub struct SessionConfig {
+    /// Funded chain mnemonic used to deposit NYM and issue ticketbooks.
+    pub mnemonic: bip39::Mnemonic,
```
jstuczyn: ensure the value gets zeroized

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+            which: "nym-api",
+            url: api_url_str.clone(),
+            source,
+        })?;
+
+        // Direct-signing chain client from the mnemonic.
+        let nyxd = DirectSigningHttpRpcNyxdClient::connect_with_mnemonic_and_network_details(
+            nyxd_url.as_str(),
+            network.clone(),
+            mnemonic.clone(),
```
jstuczyn: generate the client_id before creating nyxd client so that we wouldn't need to clone the mnemonic. furthermore, I'm not entirely sure whether using mnemonic bytes doesn't leak some data about it when requesting the ticketbooks

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+            directory,
+        })
+    }
+
+    /// Ensure the required WireGuard ticketbooks are issued and stored, issuing
+    /// (and depositing) only when no usable ticketbook is already stored.
+    pub async fn ensure_ticketbooks(&self, two_hop: bool) -> Result<(), SessionError> {
+        tokio::select! {
+            biased;
+            _ = self.cancel.cancelled() => Err(SessionError::Cancelled),
```
jstuczyn: I'd argue this should not be cancellable at this layer. if, for example, we had already made the deposit, we should be allowed to continue with the rest of issuance and only exit once it's completed, otherwise we lose funds

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+        self.ensure_one(TicketType::V1WireguardEntry).await?;
+        if two_hop {
+            self.ensure_one(TicketType::V1WireguardExit).await?;
+        }
+        Ok(())
+    }
+
+    async fn ensure_one(&self, ticket_type: TicketType) -> Result<(), SessionError> {
+        // Skip issuance if a usable ticketbook of this type is already stored.
+        if let Ok(Some(_)) = self
```
jstuczyn: we should propagate storage error and exit with `SessionError` immediately. because in the current setup on any storage failure (which as we've seen in other places can happen easily), we'll assume we don't have a ticketbook and then unconditionally try to get a new one. which will fail to get stored because our storage has issues thus losing funds.

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+        entry_quic: bool,
+    ) -> Result<Registration, SessionError> {
+        self.ensure_inner(true).await?;
+        let entry_gw = self
+            .select_inner(entry, WgRole::Entry, entry_quic, None)
+            .await?;
+        // Exclude the entry gateway so a two-hop tunnel never uses one gateway twice.
+        let exit_gw = self
+            .select_inner(exit, WgRole::Exit, false, Some(&entry_gw.identity))
+            .await?;
```
jstuczyn: we call `select_inner` twice here, which both internally fetch the ENTIRE topology `self.api.get_all_described_nodes_v2`. why not instead  get it once and pass the values there?

sdk/rust/smol-dvpn/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use thiserror::Error;
+
+/// Errors from the `smol-dvpn` datapath.
+#[derive(Error, Debug)]
+pub enum DvpnError {
+    /// A boringtun WireGuard operation failed.
+    #[error("WireGuard error: {0}")]
+    WireGuard(String),
```
jstuczyn: this variant is never constructed

smoldvpn/src/config.rs
```diff
@@ -0,0 +1,109 @@
...
+//! `smol-dvpn` is decoupled from the provisioning facade: a caller obtains a
+//! [`PeerConfig`] per hop (e.g. by mapping a `nym-sdk-session` registration) and
+//! hands it to the tunnel. The datapath itself needs only raw key material,
+//! endpoints and assigned addresses.
+
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
+
+/// Fixed exit client source port — the reference `DEFAULT_EXIT_WG_CLIENT_PORT`
+/// (two_hop_config.rs:17). Used as a fallback when a dynamic port is not bound.
+pub const DEFAULT_EXIT_WG_CLIENT_PORT: u16 = 54001;
```
jstuczyn, line 18: where did it get this port from? nodes don't really open that port for anything...

sdk/rust/smol-dvpn/src/config.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Fixed exit client source port — the reference `DEFAULT_EXIT_WG_CLIENT_PORT`
+/// (two_hop_config.rs:17). Used as a fallback when a dynamic port is not bound.
+pub const DEFAULT_EXIT_WG_CLIENT_PORT: u16 = 54001;
+
+/// One WireGuard peer/hop. All key material is raw 32-byte x25519, keeping the
+/// datapath independent of any particular crypto wrapper type.
+#[derive(Clone)]
+pub struct PeerConfig {
+    /// The gateway's WireGuard public key (x25519).
+    pub gateway_public_key: [u8; 32],
```
jstuczyn: why not use strongly typed `x25519::PublicKey` instead?

sdk/rust/smol-dvpn/src/config.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl std::fmt::Debug for PeerConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Never print key material.
+        f.debug_struct("PeerConfig")
+            .field("endpoint", &self.endpoint)
+            .field("assigned_ipv4", &self.assigned_ipv4)
+            .field("assigned_ipv6", &self.assigned_ipv6)
+            .field("preshared_key", &self.preshared_key.map(|_| "<redacted>"))
```
jstuczyn: if you're redacting fields anyway, we might as well add other fields here

sdk/rust/smol-dvpn/src/config.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Mobile defaults.
+    pub const MOBILE: MtuConfig = MtuConfig {
+        entry: 1360,
+        exit: 1280,
+    };
+}
+
+impl Default for MtuConfig {
+    fn default() -> Self {
+        Self::DESKTOP
```
jstuczyn: I think it'd be better to target-lock it, i.e. if this code is run on ios/android it'd default to `Self::MOBILE` instead

sdk/rust/smol-dvpn/src/transport.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+}
+
+impl WgReceiver {
+    /// Receive exactly one WireGuard packet.
+    pub(crate) async fn recv(&mut self) -> Result<Vec<u8>> {
+        match self {
+            WgReceiver::Direct(sock) => {
+                let mut buf = vec![0u8; MAX_WG_PACKET];
```
jstuczyn: that's very wasteful. we allocate 64kb buffer which is then zero-initialised. then, we read a single wireguard packet (~1.4kb (bound by MTU)) and truncate the rest of the buffer. let's allocate a smaller value instead?

sdk/rust/smol-dvpn/src/bridge.rs
```diff
@@ -0,0 +1,241 @@
...
+/// Per-WireGuard-packet length prefix width, big-endian.
+const LENGTH_DELIMITER_BYTELEN: usize = 2;
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+static INSTALL_PROVIDER: Once = Once::new();
+
+/// Bridge connection parameters, sourced from the gateway directory / VPN API.
+#[derive(Clone, Debug)]
+pub struct BridgeParams {
+    /// Candidate bridge socket addresses (dialed in order; the directory lists
```
jstuczyn: so which is it? the docs on top of the file say 
```
//! Note: [`nym_bridges`] dials `addresses[0]` (the directory lists the bridge's
//! IPv6 address first) and binds dual-stack `[::]:0`; it does not set QUIC
//! keep-alive/BBR. WireGuard's own persistent-keepalive keeps the long-lived
//! session and its NAT mapping alive.
```

sdk/rust/smol-dvpn/src/bridge.rs
```diff
@@ -0,0 +1,241 @@
...
+//! here, so this crate can never drift from the bridge server. This module only
+//! adds the datapath framing on top: one reliable `open_bi()` stream carrying
+//! WireGuard packets, each prefixed by a 2-byte big-endian length.
+//!
+//! Only ever fronts the two-hop entry leg (the bridge is bound 1:1 to a gateway
+//! and forwards to its WireGuard port); there is no QUIC one-hop mode and no
+//! gateway-selection handshake.
+//!
+//! Note: [`nym_bridges`] dials `addresses[0]` (the directory lists the bridge's
+//! IPv6 address first) and binds dual-stack `[::]:0`; it does not set QUIC
```
jstuczyn: given clients are ipv4-only for now, defaulting to ipv6 will simply fail...

smoldvpn/src/engine.rs
```diff
@@ -0,0 +1,277 @@
...
+    Tunn::new(secret, public, peer.preshared_key, None, index, None)
+}
+
+/// Encapsulate one plaintext packet; returns the single transport datagram if
+/// boringtun produced one (data or a handshake message).
+fn encap(tunn: &mut Tunn, plaintext: &[u8]) -> Option<Vec<u8>> {
+    let mut out = vec![0u8; MAX];
+    match tunn.encapsulate(plaintext, &mut out) {
+        TunnResult::WriteToNetwork(p) => Some(p.to_vec()),
+        _ => None,
```
jstuczyn, line 63: should we at least log errors rather than swallowing them? it seems to be the case throughout this file

sdk/rust/smol-dvpn/src/tunnel.rs
```diff
@@ -0,0 +1,459 @@
...
+                            if stack_in_tx.unbounded_send(pkt).is_err() {
+                                debug!("stack channel closed");
+                                return;
+                            }
+                        }
+                        send_all(&mut sender, out.to_network).await;
+                    }
+                    Err(e) => {
+                        // Transient on Direct UDP; fatal on a closed bridge stream.
+                        debug!("transport recv error: {e}");
```
jstuczyn: if the transport failed, we'll just keep logging this error forever without leaving and burning the ~~oceans~~ CPU

smoldvpn/src/tunnel.rs
```diff
@@ -0,0 +1,459 @@
...
+    pub fn mtu(&self) -> MtuConfig {
+        *self.mtu.read().expect("mtu lock poisoned")
+    }
+
+    /// Change the tunnel MTU at runtime. The WireGuard session is preserved (no
+    /// re-handshake); the smol-core interface is rebuilt with the new MTU, so
+    /// any sockets open at the moment of the change are reset. (A fully seamless
+    /// in-place interface resize is not supported by `tokio-smoltcp`, which fixes
+    /// the interface MTU at construction.)
+    pub fn set_mtu(&self, mtu: MtuConfig) -> Result<()> {
```
jstuczyn, line 406: `set_mtu` sends the new channels (step 1) then publishes the new stack (step 2), and step 2 drops the old `Arc<Stack>`. In the common case nothing else holds it (the `connector` clones the outer `Arc<RwLock<..>>`, not the inner stack), so the old `Stack`-  and its `stack_out_tx` - is dropped synchronously here. If the datapath hasn't processed the swap yet, its old `stack_out_rx` is now closed, so both the `stack_out_rx.next()` branch (→ `None => break`) and the `swap_rx.next()` branch are ready at once.

The `select!` isn't `biased`, so ~half the time it picks the closed rx, hits `None => break`, and the whole datapath task exits - the tunnel silently dies mid-resize.

Fix: mark the datapath `select!` `biased;` with the swap branch first, or don't `break` on `stack_out_rx` closure while a swap is outstanding, or drop the old stack only after the datapath acks the swap.

jstuczyn, line 406: `set_mtu` takes `&self` and does the swap send and the stack publish as two separate steps with no lock spanning both. Two concurrent calls can interleave so the datapath ends on call B's channels while `self.stack` ends on call A's stack - a permanent desync where callers write into a stack whose channels the datapath never reads. Either take `&mut self`, hold a dedicated swap mutex across both steps, or document it as single-caller.

sdk/rust/smol-dvpn/src/connectors.rs
```diff
@@ -0,0 +1,72 @@
...
+        Box::pin(async move {
+            let host = uri
+                .host()
+                .ok_or_else(|| DvpnError::Config("URI missing host".into()))?
+                .to_string();
+            let port = uri.port_u16().unwrap_or(match uri.scheme_str() {
+                Some("http") => 80,
+                _ => 443,
+            });
+            let stream = stack.tcp_connect_host(&host, port).await?;
```
jstuczyn: IP-literal hosts are sent through DNS and fail. `tcp_connect_host` unconditionally resolves its host argument (smol-core's `resolve` does `Name::from_utf8(host)` → DNS query, with no `host.parse::<IpAddr>()` short-circuit), so a URI whose host is already  an IP becomes a bogus lookup:

 `http://10.0.0.1:50051` → `resolve("10.0.0.1")` → A-query for the name `10.0.0.1.` → NXDOMAIN → `DnsNoRecords` → connection fails.

This is the exact example in this module's own doc comment (`from_static("http://10.0.0.1:50051")`), and dVPN targets are commonly raw IPs - so the documented snippet can't actually connect. Root cause is in smol-core's `resolve`/`tcp_connect_host`, but it surfaces here.

Fix: parse `host` as an `IpAddr` first and call `stack.tcp_connect((ip, port).into())` directly, only falling through to `tcp_connect_host` for real hostnames. (While there, strip brackets from IPv6 literals — `Uri::host()` returns `"[::1]"` - or reject v6 with a clear error instead of a confusing DNS failure.)

sdk/rust/smol-dvpn/src/topup.rs
```diff
@@ -0,0 +1,181 @@
...
+        } => Ok(available_bandwidth),
+        _ => Err(DvpnError::Transport(
+            "unexpected response to top-up request".into(),
+        )),
+    }
+}
+
+/// Build a metadata HTTP client for the given endpoint URL.
+pub(crate) fn metadata_client(url: &str) -> Result<MetadataHttpClient> {
+    MetadataHttpClient::new_url(url, None)
```
jstuczyn: **Top-up traffic bypasses the tunnel — likely broken, and a privacy leak if not.**

`metadata_client` builds a plain `MetadataHttpClient::new_url(url, None)` with the default connector, and `run_topup` is given no `Stack`/connector — so every `available_bandwidth` / `topup_once` request goes over the **host network**, not through the userspace tunnel. This contradicts the design (§10), which says the metadata endpoint is "reached through the tunnel ... just another destination reachable via smol-core's `tcp_connect` on the running tunnel."

Either way it's a problem:
- **If the gateway serves metadata in-tunnel** (private tunnel IP inside the entry peer's `allowed_ips`, as §10 describes), a host-network client can't reach it → top-up always fails and the feature is dead.
- **If it's publicly reachable**, top-up works but leaks the client's real IP to the gateway metadata endpoint, correlating the real IP with the tunnel session — a deanonymization hole in a privacy tool.

Fix: route the metadata client through the tunnel — thread the tunnel's `TunnelConnector` into `run_topup` and construct `MetadataHttpClient` over it (or have the caller pass a tunnel-backed HTTP client). As-is I'd treat this as a blocker: the feature probably can't work against an in-tunnel metadata endpoint, and where it does work it undermines the VPN.

jstuczyn (changes requested):
One big issue that I let the A1 friends format in a more readable language, but tl;dr; we don't have a mechanism for automatic top-ups which I think will cause a lot of friction with library consumers:

## Bandwidth top-up isn't automatic - and can't self-sustain even when wired up

The docs (design §10) read as though a live tunnel keeps its own bandwidth topped up. In the code it doesn't, and I think that's a gap against what library users will reasonably expect: bring up a tunnel, have it stay alive without writing plumbing.

Three issues:

1. **Off by default, opt-in only.** `TunnelBuilder.topup` is `None` unless the caller explicitly calls `bandwidth_topup(config, source)` (`tunnel.rs:147`). Without it, the tunnel just stops passing traffic once the registered bandwidth is spent - no warning, no refill.

2. **Caller has to wire two crates together themselves.** The `BandwidthCredentialSource` trait lives in `nym-smol-dvpn`, and `Session` doesn't implement it (dependency direction forbids it). So integrators must hand-roll an adapter over `Session::obtain_wireguard_credential` - and we don't ship
one. The `smol-dvpn-topup` example is one-shot; nothing demonstrates the background task.

3. **Even when enabled, it can't reach back to the chain.** `obtain_wireguard_credential` only spends an *already-stored* ticket and errors with `"no stored ticket available for top-up"` (`session.rs:222`) if the store is dry - it never calls `fetch_ticketbook`. So a long-lived tunnel silently dies
once the pre-stocked tickets run out; the mnemonic is never used to re-issue mid-session.

**Suggestions:**
- Make top-up the default for a long-lived tunnel.
- Provide a built-in `Session`-backed `BandwidthCredentialSource` (or an adapter constructor) so callers don't implement anything.
- Have live top-up fall back to `fetch_ticketbook` (mnemonic re-issuance) when the store runs low.

At minimum, the design doc should be corrected to state that top-up is manual and store-limited.



furthermore:
**Datapath does several needless heap allocations per packet**

The per-packet hot path allocates (and in places zero-fills) far more than needed, across four files. For a single outbound two-hop packet this adds up to well over 100 KiB allocated + memset, and inbound is worse:

- **`engine.rs`** — `encap`/`decap`/`timer`/`handshake_init` each do `let mut out = vec![0u8; 65535]` per call (plus a fresh 64 KiB buffer per iteration of the decap drain loop), then `.to_vec()` the result. WG packets are ~1.4 KB, so these buffers are ~46x oversized and the zero-init is pure waste. A two-hop send is `encap(exit)` + `encap(entry)` = two 64 KiB alloc+zero.
- **`transport.rs`** — `WgReceiver::recv` allocates and zeroes a 65535-byte buffer per received packet, fills ~1.4 KB, then truncates.
- **`bridge.rs`** — `send` does `Bytes::copy_from_slice(packet)` and `recv` does `frame.to_vec()`, a copy each per packet.
- **`framing.rs`** — `build_ipv4_udp` allocates a `Vec` per call and `parse_ipv4_udp` does `.to_vec()` on the payload.

None of these are correctness bugs, but the datapath is exactly where allocator churn shows up under load. Since the engine runs single-task (no locking), it can own reusable scratch buffers and hand `&mut` slices to boringtun; the transport/bridge halves can reuse a buffer sized to a realistic WG max (MTU + overhead) instead of the UDP max. Worth one focused "reduce datapath allocations" pass rather than fixing piecemeal.

docs/design/smoldvpn/design.md
mmsinclair: Done — `nym-bridges` is published to crates.io now (commit 7f3c80ce2), so the git-pin / publish-blocked notes are removed here and in the `Cargo.toml` comment.

common/smol-core/src/stack.rs
mmsinclair: Added a warning doc to `with_ipv6`: dual-stack isn't wired end-to-end yet (tokio-smoltcp 0.5 binds a single interface address), so it currently has no routing effect.

common/smol-core/src/stack.rs
mmsinclair: Done — `#[must_use]` on all `with_*` builders.

common/smol-core/src/stack.rs
mmsinclair: Done — `#[must_use]` added here too.

common/smol-core/src/dns.rs
mmsinclair: Superseded: AAAA is now skipped entirely while the stack is IPv4-only, so there's only a single query. Left a note that `join!` applies once dual-stack lands.

common/smol-core/src/dns.rs
mmsinclair: Removed the process-global `AtomicU16`. Each query now uses a randomly generated id that is validated against `response.id()` (and the datagram source against `cfg.server`), so the id does real work instead of being write-only.

common/smol-core/src/dns.rs
mmsinclair: Now inspects `response_code()`: SERVFAIL/REFUSED surface as a distinct `DnsServerFailure` error, separate from NXDOMAIN / a genuinely empty result (so the retryable ones are distinguishable).

common/smol-core/src/dns.rs
mmsinclair: Replaced the single-shot read with a recv loop that validates id + source and keeps reading until a match or the timeout, so a stray/spurious datagram on the ephemeral port no longer defeats the lookup.

common/smol-core/src/dns.rs
mmsinclair: Fixed: AAAA is not queried while the stack is a single-IPv4 `/32`, so callers never get an unroutable v6 address. (And a failing AAAA no longer discards already-resolved A records.)

sdk/rust/nym-sdk-session/src/error.rs
mmsinclair: Split `Chain(String)` into `Nyxd(#[from] NyxdError)` (preserves the source) and a separate `Api(String)` for nym-api errors.

sdk/rust/nym-sdk-session/src/error.rs
mmsinclair: Cleaned up — `Storage` is now actually constructed (credential-store dir creation), and the remaining never-constructed variants are removed.

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Reduced the immediate cost: `register_two_hop_inner` now fetches the described-node topology once and evaluates both entry and exit selection against it (was two fetches). Open question for the API side: is a role-filtered described-gateways endpoint feasible? `get_all_described_nodes_v2` is the only one exposing LP data today, so full topology is currently required.

sdk/rust/nym-sdk-session/src/gateway.rs
mmsinclair: Done — dropped the declared entry/exit role check for dVPN eligibility; any WireGuard-capable node can serve either hop now.

sdk/rust/nym-sdk-session/Cargo.toml
mmsinclair: The workspace already enables bip39's `zeroize` feature, but that only adds the `Zeroize` impl (not drop-zeroization), so owned mnemonic-derived material is now wrapped in `Zeroizing`.

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Done — wrapped in `Zeroizing` so it's zeroized on drop.

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Done — the client id is now a domain-separated SHA-256 hash of the mnemonic entropy (not the raw entropy), computed before the nyxd client is built, which also removes the `mnemonic.clone()`.

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Agreed — resolved structurally. The session now runs the `BandwidthController` (single writer) and provisions through it; a deposit interrupted by cancellation is recorded in the fetcher's pending-request store and recovered on a later fetch, so cancellation can't lose funds. Cancellation is still raced so callers get a prompt `Cancelled`.

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Fixed — the peek-then-fetch that swallowed storage errors is gone. The controller is the single writer to storage; provisioning goes through `restock_ticketbooks` + `wait_for_ticketbooks`, and storage failures propagate instead of being treated as "no ticketbook".

sdk/rust/nym-sdk-session/src/session.rs
mmsinclair: Fixed by the single-fetch change above — the topology is fetched once and passed to both selections.

sdk/rust/smol-dvpn/src/error.rs
mmsinclair: Removed the never-constructed `WireGuard` variant.

smoldvpn/src/config.rs
mmsinclair: It isn't a node listening port — it's the client-side UDP *source* port used inside the two-hop inner IPv4/UDP frame (the exit tunnel's packets are framed with this source before being encapsulated to the entry gateway), matching the reference `DEFAULT_EXIT_WG_CLIENT_PORT`. Reworded the doc comment to say so.

sdk/rust/smol-dvpn/src/config.rs
mmsinclair: Done — `PeerConfig` now uses `boringtun::x25519::{PublicKey, StaticSecret}` (the datapath's native, `Clone`-able crypto types) instead of raw `[u8; 32]`.

sdk/rust/smol-dvpn/src/config.rs
mmsinclair: Done — `Debug` now lists all fields: the gateway public key is shown (not secret) and the client secret key + PSK are redacted.

sdk/rust/smol-dvpn/src/config.rs
mmsinclair: Done — `MtuConfig::default()` returns `MOBILE` on `target_os = "ios"`/`"android"`, `DESKTOP` elsewhere.

sdk/rust/smol-dvpn/src/transport.rs
mmsinclair: Fixed — the Direct receiver reuses one buffer across recvs instead of allocating (and zeroing) a fresh 64 KiB buffer per packet. (Kept sized to the WG max so nothing is truncated; the win is eliminating the per-packet allocation.)

sdk/rust/smol-dvpn/src/bridge.rs
mmsinclair: Good catch — the docs contradicted each other. Corrected to describe the actual behaviour (the bridge client dials the first address; we reorder candidates to prefer IPv4 for v4-only clients).

sdk/rust/smol-dvpn/src/bridge.rs
mmsinclair: Fixed — `connect` now reorders the candidate addresses to prefer IPv4 before handing them to the bridge client, since clients are IPv4-only for now.

smoldvpn/src/engine.rs
mmsinclair: Fixed — swallowed `TunnResult::Err` results in the engine are now logged.

sdk/rust/smol-dvpn/src/tunnel.rs
mmsinclair: Fixed — transport recv errors are classified: a closed QUIC bridge stream is fatal and exits the datapath; a transient Direct-UDP error logs and backs off, so it no longer spins the CPU.

smoldvpn/src/tunnel.rs
mmsinclair: Fixed — the datapath `select!` is now `biased` with the swap branch first and breaks on swap-channel closure, so it adopts the rebuilt stack's channels before it can observe the old `stack_out_rx` closing. Verified live: a runtime MTU change keeps traffic flowing.

mmsinclair: Fixed — `set_mtu` now serialises the whole swap behind a mutex, so two concurrent calls can't leave the published stack and the datapath's channels referring to different stacks.

sdk/rust/smol-dvpn/src/connectors.rs
mmsinclair: Fixed in both places — smol-core's `resolve`/`tcp_connect_host` short-circuit an IP-literal host (no DNS query), and `TunnelConnector` strips IPv6 brackets from `Uri::host()` and connects directly. The module's `http://10.0.0.1:50051` doc example works now.

sdk/rust/smol-dvpn/src/topup.rs
mmsinclair: Fixed — the metadata client now dials **through the tunnel** (a hyper HTTP/1 client over the tunnel's own `TunnelConnector` / in-tunnel `TcpStream`), never the host network, so top-up can't leak the client's real IP. On the broader point: gateway-side top-up (spending already-stored tickets) is now on by default for session-built tunnels via a shipped `ProviderCredentialSource`; buying new ticketbooks (chain-side restock) is opt-in via `with_automatic_topups` because a ticketbook is ~37.5 GB; and a `Tunnel::bandwidth_events()` stream lets integrators react (e.g. prompt the user) even when auto-top-up is off. Design §10 corrected.

mmsinclair:
Responding to the two review-body items:

**Bandwidth top-up.** Gateway-side top-up (spending already-stored tickets at the gateway `metadata` endpoint) is now on by default for session-built tunnels — a ready-made `ProviderCredentialSource` over `nym-sdk-session`'s controller sender is shipped, so integrators write no plumbing. Buying **new** ticketbooks (chain-side restock) is opt-in via `SessionConfig::with_automatic_topups`, since a ticketbook is ~37.5 GB and over-issuing costs the integrator NYM. A `Tunnel::bandwidth_events()` stream (`Low`/`ToppedUp`/`TopupFailed`/`Exhausted`) lets integrators react even when auto-top-up is off. Design doc §10 is corrected accordingly.

**Datapath allocations.** The engine now owns one reusable scratch buffer that boringtun encrypts/decrypts into (no per-call 64 KiB alloc+zero), and the Direct receiver reuses its recv buffer. The remaining single copies in `framing.rs` / `bridge.rs` are inherent to the smoltcp emit / length-delimited-codec APIs.

common/smol-core/README.md
```diff
@@ -0,0 +1,52 @@
...
+let (outbound_tx, outbound_rx) = mpsc::unbounded::<Vec<u8>>(); // stack -> transport
+let (inbound_tx, inbound_rx) = mpsc::unbounded::<Vec<u8>>();   // transport -> stack
+
+let device = ChannelDevice::new(inbound_rx, outbound_tx, DEFAULT_MTU);
+let stack = Stack::new(device, StackConfig::new("10.0.0.2".parse()?));
+
+let tcp = stack.tcp_connect("1.1.1.1:443".parse()?).await?; // AsyncRead + AsyncWrite
+let udp = stack.udp_socket().await?;
+let addrs = stack.resolve("example.com").await?;            // DNS over the tunnel
+```
```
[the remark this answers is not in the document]
mmsinclair: This is an illustrative README snippet rather than a compiled doctest, kept deliberately concise; leaving it as-is.

common/smol-core/src/dns.rs
```diff
@@ -0,0 +1,121 @@
...
+            RData::AAAA(aaaa) => Some(IpAddr::V6(aaaa.0)),
+            _ => None,
+        })
+        .collect();
+
+    if addrs.is_empty() {
+        return Err(SmolCoreError::DnsNoRecords {
+            name: name.to_utf8(),
+        });
+    }
```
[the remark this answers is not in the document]
mmsinclair: Fixed — `query_one` now rejects a response with the `TC` (truncation) bit set, returning a new `SmolCoreError::DnsTruncated` instead of the partial answers it carries. A datagram-only resolver can't do the RFC 1035 TCP retry, so erroring is the safe choice. Added a test asserting a truncated response doesn't resolve.

common/smol-core/src/stack.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub fn with_mtu(mut self, mtu: usize) -> Self {
+        self.mtu = mtu;
+        self
+    }
+
+    /// Set the per-socket TCP buffer size (the TCP window; see [`DEFAULT_TCP_BUFFER`]).
+    pub fn with_tcp_buffer(mut self, tcp_buffer: usize) -> Self {
+        self.tcp_buffer = tcp_buffer;
+        self
+    }
```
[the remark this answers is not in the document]
mmsinclair: Fixed — removed the `mtu` field and `with_mtu` from `StackConfig`; `Stack::new` never read them. The caller-supplied `AsyncDevice` (e.g. `ChannelDevice`) is the single source of truth for the interface MTU.

common/smol-core/src/stack.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// the host resolver.
+    pub async fn resolve(&self, host: &str) -> Result<Vec<IpAddr>> {
+        dns::resolve(&self.net, &self.dns, host).await
+    }
+
+    /// Resolve `host` and open a TCP connection to the first address on `port`.
+    pub async fn tcp_connect_host(&self, host: &str, port: u16) -> Result<TcpStream> {
+        let addrs = self.resolve(host).await?;
+        let addr = SocketAddr::new(addrs[0], port);
+        self.tcp_connect(addr).await
```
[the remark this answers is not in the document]
mmsinclair: `resolve()` never returns `Ok` with an empty address set — it returns an error (`DnsNoRecords`/`DnsServerFailure`/`DnsTruncated`) when nothing resolves — so `addrs[0]` here is unreachable on an empty answer and can't panic. Happy to switch to `.first()` for defense-in-depth if preferred.

common/smol-core/tests/stack.rs
```diff
@@ -0,0 +1,176 @@
...
+        a.tcp_connect(format!("{B_IP}:9999").parse().unwrap()),
+    )
+    .await;
+
+    // Either the handshake was refused (Err) or it never completed (timeout).
+    // The one thing that must NOT happen is a successful connection.
+    if let Ok(Ok(_stream)) = result {
+        panic!("connect unexpectedly succeeded to a closed port")
+    }
+}
```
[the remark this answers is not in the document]
mmsinclair: Intentional: the closed-port test accepts either a connection-refused error or a connect timeout so it's stable across platforms (some return RST, others just time out). The invariant it guards is "must not connect successfully," which it still enforces.

deny.toml
```diff
@@ -105,6 +105,7 @@ allow = [
     "OpenSSL",
     "Zlib",
     "CDLA-Permissive-2.0",
+    "AGPL-3.0",
```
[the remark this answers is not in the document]
mmsinclair: This global `AGPL-3.0` allowance predates this PR (it's on `develop`), and a crate-scoped exception for `nym-bridges` already exists — which makes the global redundant. Narrowing it is a separate, team-owned `deny.toml` policy change, so flagging it for that owner rather than changing it in this PR.

sdk/rust/nym-sdk-session/src/dvpn.rs
```diff
@@ -0,0 +1,158 @@
...
+                    .filter_map(|a| a.parse::<SocketAddr>().ok())
+                    .collect::<Vec<_>>();
+                if addresses.is_empty() {
+                    return None;
+                }
+                Some(QuicBridge {
+                    addresses,
+                    sni_host: args.host.filter(|h| !h.is_empty()),
+                    id_pubkey_base64: args.id_pubkey,
+                })
```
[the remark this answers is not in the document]
mmsinclair: Fixed — a blank `id_pubkey` is rejected so an unpinned bridge is never advertised as QUIC-capable. While here, `into_quic` now iterates every `quic_plain` transport (a malformed earlier one no longer shadows a valid later one), and the pin/SNI are stored trimmed so a whitespace-padded directory value still base64-decodes at connect time. Added unit tests.

sdk/rust/nym-sdk-session/src/session.rs
```diff
@@ -0,0 +1,503 @@
...
+    pub async fn register_single_hop(
+        &self,
+        gateway: &GatewaySpec,
+    ) -> Result<Registration, SessionError> {
+        tokio::select! {
+            biased;
+            _ = self.cancel.cancelled() => Err(SessionError::Cancelled),
+            res = self.register_single_inner(gateway) => res,
+        }
+    }
```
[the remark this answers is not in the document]
mmsinclair: Fixed — the ticket-spending `register_dvpn`/`handshake_and_register_dvpn` calls no longer race the cancel token, so a cancel firing after the gateway has processed the spend can't drop the future and lose the ticket. Gateway selection and the (ticket-free) LP `perform_handshake` stay promptly cancellable.

sdk/rust/smol-dvpn/src/bridge.rs
```diff
@@ -0,0 +1,241 @@
...
+        r = tokio::time::timeout(CONNECT_TIMEOUT, transport_conn(&options)) => {
+            r.map_err(|_| DvpnError::Bridge("connect timed out".into()))?
+             .map_err(|e| DvpnError::Bridge(format!("connect: {e}")))?
+        }
+    };
+
+    let (send, recv) = conn
+        .open_bi()
+        .await
+        .map_err(|e| DvpnError::Bridge(format!("open_bi: {e}")))?;
```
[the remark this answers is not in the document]
mmsinclair: Fixed — `conn.open_bi()` is now wrapped in the same `CONNECT_TIMEOUT` + cancellation `select!` as the QUIC handshake, so a bridge that withholds bidirectional-stream credit can't leave `connect()` hanging.

openspec/specs/dvpn-tunnel/spec.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Requirement: Bandwidth top-up for a long-lived tunnel
+
+While connected, the tunnel SHALL top up bandwidth by spending stored tickets via the
+gateway `metadata` endpoint before the registered bandwidth is exhausted.
+
+#### Scenario: Top-up before exhaustion
+- **WHEN** remaining bandwidth on a running tunnel falls low and stored tickets exist
+- **THEN** a ticket is spent against the gateway metadata endpoint and available
+  bandwidth increases
```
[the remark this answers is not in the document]
mmsinclair: The full top-up requirements (default-on gateway top-up, opt-in chain restock, monitor-only/disabled, and bandwidth-event semantics) are captured in the `fix-dvpn-review-6953` OpenSpec change delta and propagate into the live `dvpn-tunnel` spec when that change is applied/archived, rather than being hand-edited into the spec here.

sdk/rust/smol-dvpn/README.md
```diff
@@ -0,0 +1,276 @@
...
+tunnel via ordinary tokio socket surfaces (`TcpStream`, `UdpSocket`, and
+`tonic`/`hyper` connectors).
+
+## What can I use `nym-smol-dvpn` for?
+
+Tunnel some or all of your app's internet traffic over the Nym network — in
+**1-hop or 2-hop dVPN mode** — straight from Rust. You don't stand up an
+OS-wide VPN; the tunnel is **scoped to your app's own traffic**, so nothing
+leaks around it and closing the tunnel simply cuts your app off — a **natural
+kill-switch** you get for free.
```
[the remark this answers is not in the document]
mmsinclair: Fixed — reworded to a scoped guarantee: only sockets opened through the tunnel are protected; ordinary sockets and the host's DNS still bypass it, and routing *all* of an app's traffic (preventing leaks around it) is the integrator's responsibility. Now consistent with the design's explicit no-OS-routing / kill-switch non-goal.

sdk/rust/nym-sdk-session/src/error.rs
jstuczyn: but we could have similarly wrapped `NymAPIError` within Api variant... let me just push it myself.

jstuczyn (approved):

PR #6961, opened 2026-07-21
service provider ephemeral credential storage

simonwicky:
### Problem

The gateway-embedded service providers (network requester, IP packet router, authenticator) are built as mixnet clients, so they're forced to carry a `credentials_database` — even though they never spend ecash for their own traffic. They worked around this with placeholder paths, which break when they are actually used.

### Fix

Add `EmbeddedProviderStorage` in `nym-service-providers-common`: a `MixnetClientStorage` impl that persists keys, reply-SURBs and gateway registrations on disk while using an **ephemeral, in-memory credential store**. All three providers now pass it to `MixnetClientBuilder::new_with_storage(...)` instead of `new_with_default_storage(...)`, so "no credentials" is enforced by the type rather than a magic path.

### Changes

- **New:** `service-providers/common/src/storage.rs`.
- Switched NR / IPR / authenticator mixnet-client setup to `new_with_storage`.
- `nym-node`: reverted the `":memory:"` credential paths to `Default::default()` (now inert).
- **Deps (cycle-free):** sp-common → `nym-client-core` + `nym-credential-storage`; gateway → `nym-service-providers-common`.

### Notes

- Runtime behaviour unchanged — these providers never used the credential store.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6961)


service-providers/common/Cargo.toml
```diff
@@ -24,3 +24,8 @@ thiserror = { workspace = true }
 serde = { workspace = true, features = ["derive"]}
 serde_json = { workspace = true }
 
+# storage backend for gateway-embedded service providers (native only)
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
```
jstuczyn: I like how you're trying to add wasm support, but no sane person would ever compile service providers for wasm, so you don't need to target lock it

jstuczyn (commented):
you forgot about the authenticator service provider

jstuczyn (approved):

PR #6962, opened 2026-07-21
Fix/ipr selection and mtu

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6962)

## Summary by CodeRabbit

* **New Features**
  * Added support for IP packet protocol version 10, including v10 connect/data/response handling.
  * IPR tunnels can now negotiate and apply the tunnel MTU, and v10 clients report the negotiated MTU.
  * Mobile-friendly MTU handling and an `NYM_MTU_SIZE` override for IPR egress MTU.
* **Compatibility**
  * Automatic fallback to v9 continues when v10 isn’t available.
* **Bug Fixes**
  * IPR gateway discovery now excludes gateways that don’t explicitly advertise exit IPR support.
  * Exit node selection improves by accounting for directory/gateway version compatibility.


common/ip-packet-requests/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod codec;
 pub mod icmp_utils;
 pub mod response_helpers;
 pub mod sign;
+pub mod v10;
```
jstuczyn, line 9: nit: can you keep the version modules in order?

common/ip-packet-requests/src/response_helpers.rs
```diff
@@ -25,6 +25,9 @@ pub enum IprResponseError {
 
     #[error("connect denied: {0:?}")]
     ConnectDenied(crate::v8::response::ConnectFailureReason),
+
+    #[error("expected v10 connect control response")]
+    UnexpectedV10Response,
```
jstuczyn: does V10 require a dedicated error variant? couldn't we use `VersionMismatch`?

common/ip-packet-requests/src/response_helpers.rs
```diff
@@ -68,6 +71,27 @@ pub fn parse_connect_response(response: IpPacketResponse) -> Result<IpPair, IprR
     }
 }
 
+/// Parse a v10 connect response, returning the allocated IPs and the IPR's MTU.
+/// Separate from [`parse_connect_response`] as it decodes the v10 response tree.
+pub fn parse_connect_response_v10(
```
jstuczyn, line 81: similarly, why do we need a dedicated v10 handler. can't we just use the previous logic of what we were doing with v6,v7,v8, etc?

common/ip-packet-requests/src/v10/mod.rs
```diff
@@ -0,0 +1,30 @@
+pub const VERSION: u8 = 10;
+
+/// Minimum nym-node release that supports v10. Older nodes fall back to v9.
+pub const MIN_RELEASE_VERSION: semver::Version = semver::Version::new(1, 37, 0);
+
+// Same request wire format as v8; only the connect response differs (it carries
+// the IPR's MTU). Use the wrapper constructors below to set the version byte.
+pub use super::v8::request;
```
jstuczyn, line 16: I think there were some issues with v8/v9 due to re-using the types as opposed to redefining them. @simonwicky do you happen to remember what it was and whether it's something that could bite us here?

common/tun/src/lib.rs
```diff
@@ -5,3 +5,18 @@ pub mod tun_task_channel;
...
 pub use linux::tun_device;
+
+/// Default IPR egress TUN MTU (bytes), also reported to v10 clients as the
+/// largest IP packet the IPR will accept.
+pub const DEFAULT_IPR_TUN_MTU: u16 = 1500;
+
+/// The IPR TUN MTU: `NYM_MTU_SIZE` if set and valid, else [`DEFAULT_IPR_TUN_MTU`].
+/// Single source so the TUN and the value reported to clients agree.
+pub fn configured_ipr_tun_mtu() -> u16 {
+    std::env::var("NYM_MTU_SIZE")
```
jstuczyn: can we define this env var name in `nym-network-defaults`?  similarly, we should perhaps do the same for the default MTU

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
```diff
@@ -64,6 +64,13 @@ pub async fn retrieve_exit_nodes_with_performance(
     let mut described = Vec::new();
 
     for exit in exit_gateways {
+        // Only select nodes that declare the exit-IPR role. A mixnode can carry a
+        // stale `ip_packet_router` self-description; its IPR is not served, so the
+        // v9 connect just times out.
+        if !exit.supported_roles.exit_ipr {
```
jstuczyn, line 79: btw, looking at it, the issue wasn't that mixnodes could have been running IPR. it was that we were (and still are) retrieving exit gateways via `let exit_gateways = client.get_all_basic_nodes_with_metadata().await?.nodes;` and `get_all_basic_nodes_with_metadata` retrieves EVERYTHING. there's another endpoint for just the gateways

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -106,9 +124,65 @@ impl IpMixStream {
         }
     }
 
-    async fn connect_tunnel(stream: &mut MixnetStream) -> Result<IpPair, Error> {
+    /// Connect, trying v10 (which reports the IPR's MTU) then falling back to v9.
+    /// Returns the allocated IPs and, on the v10 path, the reported MTU.
+    async fn connect_tunnel(stream: &mut MixnetStream) -> Result<(IpPair, Option<u16>), Error> {
+        match Self::connect_v10(stream).await {
```
jstuczyn: I fear this will rot if we attempt to connect with explicit versions (v10 -> v9), I think we had some logic elsewhere that attempts to choose the correct version based on gateway version and then it tried to downgrade accordingly

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -106,9 +124,65 @@ impl IpMixStream {
...
+
+        loop {
+            tokio::select! {
+                _ = &mut timeout => {
+                    return Err(Error::IPRConnectResponseTimeout);
+                }
+                result = stream.recv() => {
+                    let data = result.ok_or(Error::IPRClientStreamClosed)?;
+
+                    // Ignore non-v10 frames; the probe timeout handles a silent IPR.
```
jstuczyn: in this instance, if it's a non-v10, shouldn't we stop the stream immediately?

service-providers/ip-packet-router/src/messages/request.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+mod v10;
```
jstuczyn, line 4: nit: keep them in order

service-providers/ip-packet-router/src/messages/response.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2025 - Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+mod v10;
```
jstuczyn, line 4: same nit

service-providers/ip-packet-router/src/messages/response/v10.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl From<DynamicConnectResponse> for ConnectResponseReplyV10 {
+    fn from(reply: DynamicConnectResponse) -> Self {
+        match reply {
+            DynamicConnectResponse::Success(DynamicConnectSuccess { ips }) => {
+                // Report the egress TUN MTU so the client can size its MTU to fit.
+                ConnectResponseReplyV10::Success(ConnectSuccessV10 {
+                    ips,
+                    mtu: nym_tun::configured_ipr_tun_mtu(),
```
jstuczyn, line 66: is env the only way of configuring the MTU? could the gateway have set it with cli flag or something?

smolmix/core/src/device.rs
```diff
@@ -29,10 +29,23 @@ impl NymAsyncDevice {
...
         rx: mpsc::UnboundedReceiver<Vec<u8>>,
         tx: mpsc::UnboundedSender<Vec<u8>>,
+        negotiated_mtu: Option<u16>,
     ) -> Self {
         let mut capabilities = DeviceCapabilities::default();
         capabilities.medium = Medium::Ip;
-        capabilities.max_transmission_unit = 1500;
+        // Client MTU precedence: SMOLMIX_MTU override > Android's mobile MTU > the
+        // IPR-negotiated MTU (v10) > 1420 fallback for pre-v10 IPRs.
+        #[cfg(target_os = "android")]
```
jstuczyn: can we extract all of this to some helper?

wasm/smolmix/src/device.rs
```diff
@@ -28,14 +28,40 @@ pub struct WasmDevice {
     tx_stats: TxStats,
 }
 
+/// Client MTU: Android keeps its mobile MTU (1280); elsewhere use the IPR-reported
+/// MTU (v10) or 1420 for pre-v10 IPRs. Native tests can't sniff, so take the else.
+fn client_mtu(negotiated: Option<u16>) -> usize {
+    #[cfg(target_arch = "wasm32")]
+    {
+        if is_android() {
+            return 1280;
```
jstuczyn: could we extract these to some constants (maybe `nym-network-defaults` or somewhere within `smolmix`?) so that we could more easily adjust it in the future shall we need it?

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
mfahampshire: Good point, then we just keep a list of the minimum gw version that could support that v - will hunt for that.

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
mfahampshire: It would preclude the use of v9 supporting IPRs, which I didn't think we wanted to do necessarily

wasm/smolmix/src/device.rs
mfahampshire: good point, will do

service-providers/ip-packet-router/src/messages/response/v10.rs
mfahampshire: I was just following the existing convention, we totally could though.

common/ip-packet-requests/src/v10/mod.rs
simonwicky: Can't remember exactly what from the top of my head, but that is correct. Let's not do that

simonwicky: If the wire format is the same, why do we need v10?

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: Which one do you mean? Since we're going to add proper IPR testing in the future I'll have to rewrite the query logic anyway, so I could leave it for the moment?

service-providers/ip-packet-router/src/messages/request.rs
mfahampshire: iirc this is `rustfmt`'s fault but will double check and fix if not

common/ip-packet-requests/src/v10/mod.rs
mfahampshire: We need a v10 because I'm adding the IPR's MTU to the response type, and I thought it was better to just make a new one rather than mod the existing one

mfahampshire: But I can look @ redefining the types instead of re-use to avoid the previous issues we had yes

service-providers/ip-packet-router/src/messages/response/v10.rs
mfahampshire: Should I make a ticket for this to do on its own, due to the time constraints?

common/ip-packet-requests/src/lib.rs
mfahampshire: `rustfmt` is moving it around

common/ip-packet-requests/src/response_helpers.rs
mfahampshire: Because I'm adding the mtu field - do you mean could I just mod the existing one to check if there *is* an MTU field present?

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
jstuczyn: I mean, it's a minor thing. I'm just saying we're getting nodes from nym-api that are **guaranteed** to not have an exit_ipr so we just needless transfer that data

common/ip-packet-requests/src/lib.rs
jstuczyn: that pesky rustfmt : D

common/ip-packet-requests/src/lib.rs
```diff
@@ -20,6 +21,28 @@ pub const MAX_NON_STREAM_VERSION: u8 = v8::VERSION;
...
+
+/// Client MTU fallback for IPRs that predate MTU negotiation (the v10 connect
+/// response); at or below the historic 1420-byte IPR TUN.
+pub const CLIENT_MTU_FALLBACK: u16 = 1420;
+
+/// Highest IPR protocol version a node's release supports, from each version's
+/// `MIN_RELEASE_VERSION`. Lets a client pick the protocol up front from the
+/// node's directory version instead of probing. `None` means the node is too old
+/// for even v9.
+pub fn best_supported_version(node_version: &semver::Version) -> Option<u8> {
```
jstuczyn, line 37: so are we completely dropping support for v8 (and older?) - I actually don't mind, especially if they're that old. but in that case, let's maybe completely remove the v6,v7,v8 code too

service-providers/ip-packet-router/src/messages/response/v10.rs
jstuczyn: I think that makes sense. so that we'd feel good about wanting to do something about it in 2031

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -106,9 +157,101 @@ impl IpMixStream {
...
+    async fn connect_tunnel(
+        stream: &mut MixnetStream,
+        node_version: Option<&semver::Version>,
+    ) -> Result<(IpPair, Option<u16>), Error> {
+        // SMOLMIX_IPR_VERSION overrides the node version used for protocol gating
+        // (semver, e.g. "1.37.0"). Test/debug knob for exercising a protocol
+        // version against a node the directory reports as older, such as a local
+        // IPR that isn't a directory node. Read here in the native SDK only; the
+        // wasm/smolmix build cannot read process env. Mirrors SMOLMIX_MTU.
+        let forced = match std::env::var("SMOLMIX_IPR_VERSION") {
```
jstuczyn: what's with this magic env variable? I think it should be defined as a constant and retrieval of its value moved to a helper

sdk/rust/nym-sdk/src/ipr_wrapper/ip_mix_stream.rs
```diff
@@ -106,9 +157,101 @@ impl IpMixStream {
...
+                }
+            },
+            Err(_) => None,
+        };
+        let node_version = forced.as_ref().or(node_version);
+        if node_version.is_none() {
+            debug!("no directory version for IPR node; connecting v9 (MTU unreported)");
+        }
+
+        let use_v10 = node_version.and_then(best_supported_version) == Some(v10::VERSION);
```
jstuczyn: I think this will rot...

can we at least do a match on `best_supported_version` and then connect_v10 on v10, connect v9 on v9 and log warning on anything else?

wasm/smolmix/src/ipr.rs
```diff
@@ -61,36 +62,157 @@ impl Default for SurbsConfig {
...
 
-    // 2. Send v9 ConnectRequest as LP Data frame (seq=0).
-    // Data frames have their own seq space; Open's seq field is independent.
+    // 2. Connect (Data seq=0) at the version the node's release supports.
+    if node_version.is_none() {
+        crate::util::debug_log!(
+            "[ipr] no directory version for node; connecting v9 (MTU unreported)"
+        );
+    }
+    let use_v10 = node_version.and_then(nym_ip_packet_requests::best_supported_version)
```
jstuczyn: similarly to the previous comment, could we do a match on the version instead so that we wouldn't be tied to just v10/v9?

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: Yeah, I didn't see an immediate way of doing it without using the a deprecated endpoint, but I know what you mean.

common/ip-packet-requests/src/lib.rs
mfahampshire: We still need v8 (`nym-vpn-client` uses it iirc), could drop v7 and below tho

sdk/rust/nym-sdk/src/ip_packet_client/discovery.rs
mfahampshire: Ok sorted this

jstuczyn (approved):

PR #6984, opened 2026-07-25
SDK: nym-swizzle

mmsinclair:
Adds `nym-swizzle`, an SDK utility crate for **application-layer traffic-shape obfuscation**, aimed at wallets and any client that fetches sequential, index-addressed data or broadcasts at meaningful moments — plus `nym-swizzle-zcash`, a Zcash-specific policy layer that packages those primitives into a library Zcash wallet developers can depend on directly.

## Why

A mixnet hides *who* is talking. It does not hide *what an application's query pattern says about it*, and two leaks survive perfect transport anonymity:

- **Timing.** The destination sees wall-clock arrival. A wallet that broadcasts the moment it reaches chain tip is trivially correlatable with its own sync activity.
- **Index / start height.** A light client asking for exactly blocks `4_120_000..4_120_010` states its resume point, and because today's start is yesterday's end, the start height acts as a **linking key that chains otherwise-unlinkable sessions together**.

Every nym-powered wallet currently has to reinvent these mitigations. They are pure application-layer transforms with no dependency on nym internals, so they belong in one small reusable crate.

## What's in it

Two composable primitives plus a shared, seedable randomness configuration:

**`delay`** — schedules an async action after a sampled delay (uniform, Poisson-process, or normal).

```rust
let mut s = Delay::uniform(Duration::ZERO, Duration::from_secs(10));
let result = s.run(async move { broadcast_tx(tx).await }).await;
```

**`range`** — decomposes an index range into randomly sized, deliberately overlapping, randomly ordered chunks that cover it exactly, with start-edge obfuscation.

```rust
Range::new(resume_height, tip)
    .snap_start(Snap::Spacing(1000))   // anonymity by collision
    .start_jitter(2500)                // anonymity by noise
    .plan()
    .for_each_concurrent(4, |start, end| get_blocks(start, end))
    .await;
```

## Design decisions worth reviewing

These are the choices a reviewer should push back on if they disagree — each is recorded with its rationale in `openspec/changes/add-nym-swizzle/design.md`.

1. **Rejection-resampling, never clamping.** Clamping an out-of-bounds sample to `max` piles probability mass into a spike at exactly `max` — itself a fingerprint. Out-of-bounds samples are redrawn instead, with a bounded-retry fallback so a pathological config still terminates.
2. **"Poisson" means exponential inter-arrival times**, matching `sample_poisson_duration` in `common/nymsphinx`. Deliberate: delays drawn from the same family as mixnet cover traffic blend into traffic the adversary already observes.
3. **Snapping is deterministic and consumes no randomness.** Its entire value is that independent clients collide on an identical start; any RNG involvement destroys that. There is a test asserting the RNG sample stream is byte-identical with and without snapping enabled.
4. **Jitter composes with snapping in whole checkpoint intervals**, so the emitted start stays on-grid. Naive composition (jitter in raw indexes) would smear starts off-grid and quietly destroy the collision property.
5. **The end of a range is never extended.** The crate cannot know which indexes exist (chain tip, array bounds), so callers widen ranges themselves. The *downward* start edge is the one sanctioned exception, because earlier indexes always exist.
6. **No VRF machinery.** A VRF is keyed and produces a proof; nothing here needs a third party to verify the schedule. Seeded ChaCha20 gives the reproducibility that tests and replay actually want, and a caller who needs verifiable randomness feeds a VRF output in as opaque seed material.

## Wasm is a hard constraint, not a goal

The crate is designed to be wrapped, unmodified, by a `wasm-pack` wrapper with JS conveniences — wallets are the target and wallets are often wasm. Every non-dev dependency compiles to `wasm32-unknown-unknown` (`tokio` time gated to non-wasm, `wasmtimer` to wasm32, following `common/http-api-client`). This is enforced in CI rather than trusted: `nym-swizzle` is added to `WASM_CRATES` so `sdk-wasm-lint` clippies it for wasm32 with `-Dwarnings`.

It also has **zero nym dependencies**, so it stays independently publishable and honest about what it is.

## nym-swizzle-zcash: the Zcash policy layer

`sdk/rust/nym-swizzle-zcash` started life on this branch as a benchmark example; it is now a **library crate for Zcash wallet developers**, implementing the baseline-hygiene mechanisms from the external privacy review of the crate (Diaz, 2026-07-27) as a thin policy layer over `nym-swizzle`. Artifacts for this part live in `openspec/changes/add-nym-swizzle-zcash-crate/`.

The adversary here is the lightwalletd itself, and two content/timing leaks survive perfect transport anonymity: **resume-point chaining** (a client resuming at `previous end + 1` is trackable by height alone) and **sync-then-send** (a broadcast seconds after a sync is attributed to it, and via chaining to the wallet's whole history). The crate ships one module per countermeasure plus the transport slots:

- **`grid` — quantized sync ranges, pure arithmetic.** Every queued range is widened to a network-wide grid: spacing from the ladder `S_j = 144·2^j` (nesting with ZIP 318's anchor grid) with floor 1152 (~one day), start rounded down, end rounded up, capped at the public tip. Every wallet resuming in the same grid cell emits **identical boundaries** — anonymity by collision. The reorg-verify window (10 blocks below the resume point) rides *inside* the widened range, with a one-cell widening rule when the resume point sits too close to a boundary; a separate tiny request would have named the resume point exactly. `classify` tags every delivered block (`CoverBelow` / `VerifyWindow` / `Requested` / `CoverAbove`) so cover is discarded cheaply.
- **`sync` — the driver and the `BlockSource` slot.** The wallet author implements one trait method (`block_range`) against their existing transport; the driver puts the quantized range on the wire **deterministically** — ascending, disjoint, 1152-block grid-aligned cells, no random sizes, no overlap, no shuffle (reviewer follow-up: randomization among wallets that already emit an identical union buys nothing against this adversary, costs bandwidth, and per-wallet variation is itself a distinguishing dimension) — and withholds `Committed` until every request landed **and** the verify window matched stored hashes; a mismatch resolves `ReorgDetected` so the wallet rewinds exactly as its SDK does today.
- **`broadcast` — decoupled, restart-safe send scheduling.** Delays are exponential with mean 144 blocks (~3 h) rejection-resampled above 576 (~12 h) — ZIP 318's transfer-scheduling parameters, so wallet sends pool with migration traffic (opt-in `fast()` profile: mean 24 blocks, smaller anonymity set, documented as such). Because those delays outlive a phone process, scheduling is **two-phase**: `schedule()` samples exactly once into a plain-old-data `BroadcastPlan` the wallet persists (optional `serde` feature); `resume()` sleeps only the remainder — resampling on restart would bias frequently-restarted wallets toward short delays. The transaction is **built at fire time**, so its on-chain-visible expiry derives from a fresh tip instead of leaking the last sync height; `needs_refresh_sync` implements the anchor-age decision (anchors live ~2 days, so most sends need no preceding request). `TxBroadcaster` is a **separate trait** from `BlockSource`, so no API ever couples one session object to both roles.

Design decisions worth reviewing, recorded in `openspec/changes/add-nym-swizzle-zcash-crate/design.md`:

1. **No librustzcash dependency, not even feature-gated.** The crate aims to be upstreamed into librustzcash, so any dependency would become circular. It speaks bare `u64` heights and two small enums; mandatory deps are exactly `nym-swizzle` + `futures`, and the wasm CI check now covers this crate too (gRPC/TLS lives in dev-dependencies only).
2. **Constants are named and fixed, not knobs.** `SHARD = 144`, `S_FLOOR = 1152`, `VERIFY_LOOKAHEAD = 10`, profile parameters — all documented with their public derivations (ZIP 318, ZIP 203, librustzcash), none configurable on the standard path: a custom floor or delay distribution is a fingerprint.
3. **The README and example address Zcash wallet developers only** — standalone, no internal-process references, and every privacy claim is paired with a way to verify it locally, because that audience is (rightly) skeptical of claims they can't check.
4. **The network-wide rule is pinned, not just implemented.** Ranges are half-open and the ladder is selected by half-open length, boundary-tested at rung multiples (a wallet exactly 1152 blocks behind takes `S = 2304`; an inclusive reading would give 1152 and silently partition the collision sets); `S_FLOOR` is asserted at compile time to sit on the `144·2^j` ladder, since shard divisibility alone would admit a floor that stops nesting.

## Evidence

**Profiling harness** (`--example profiling`, dev-deps only) streams 10M samples per distribution through fixed-size accumulators and renders each against its theoretical density as SVG. The plots are evidence, not the gate — every one is paired with a moment check that fails the run on drift:

| check | measured | expected |
|---|---|---|
| uniform mean / std dev | 3.0005 / 1.1547 | 3.0 / 1.1547 |
| truncated-exponential mean | 1.9320 | 1.9322 |
| normal mean / std dev | 5.0000 / 0.9997 | 5.0 / 1.0 |
| chunk size / overlap mean | 199.97 / 27.504 | 200 / 27.5 |
| start-jitter mean | 500.28 | 500 |

It also asserts the *absence* of a spike at the max bound — the artefact a clamping implementation would produce.

**Live example** (`cargo run --release -p nym-swizzle-zcash --example wallet_sync`) runs the whole story against the public `zec.rocks` lightwalletd with real chain data and no wallet seed: it prints the naive request a resuming wallet would send next to the grid-aligned range actually emitted, issues the deterministic ascending grid-cell requests through a `BlockSource` implementation, asserts the wire union equals the quantized range, then schedules a broadcast — printing what the plan commits to (profile, sampled delay in blocks and hours, projected fire height, the literal bytes persisted to disk) — and plays out five wallet "restarts" a couple of seconds apart: each wake-up restores the plan from disk and goes back to sleep with **no network calls** (the hours of delay and the chain drift they imply are simulated so the demo stays interactive), and the final round builds a dummy transaction at fire time (fresh **real** tip fetched inside the builder) into a mock `TxBroadcaster`.

**Opt-in live suite** (`cargo test -p nym-swizzle-zcash -- --ignored`) verifies on real chain data: the reorg rule (stored hashes matching the chain ⇒ `Committed`; one corrupted stored hash ⇒ `ReorgDetected`, with the resume point deliberately placed within 10 blocks of a grid boundary so the widening rule runs live), that every wire request is exactly the deterministic grid-aligned plan, and per-regime overhead. Requests are disjoint, so grid cover is the *only* overhead — the suite asserts streamed blocks equal the emitted length exactly. Measured on a representative run:

| regime | cover overhead (total) |
|---|---|
| daily incremental (1152-block gap) | 1.96x |
| long catch-up (34,560-block gap) | 1.33x |

The exact ratios move with where the resume point falls in its grid cell; the suite asserts the structural bound rather than a point value, and prints the numbers so a reviewer can eyeball both regimes.

## Testing

`nym-swizzle`: 39 tests (unit + integration) and 4 doctests, covering the coverage/overlap/permutation invariants across many seeds, the delay laziness guarantee (poll-counting future under paused time), rejection-resampling bounds, floor and on-grid behaviour, determinism, and the concurrency limit.

`nym-swizzle-zcash`: 34 unit tests (quantization worked examples, half-open-convention boundary pins, and property sweeps; sync driver against an in-memory fake chain — exact coverage, request determinism/ascending order/grid alignment, commit-withholding, reorg early-stop, censored-verify-window rejection; broadcast single-sampling, remainder arithmetic, fire-time building under paused time, profile bounds/mean, serde round-trip) plus the 3 ignored live tests above, and a compile-time assertion pinning `S_FLOOR` to the ladder family.

Clippy clean with `-Dwarnings` on native and wasm32 for both crates; `tokio` and the gRPC/TLS stack are asserted absent from both resolved wasm dependency graphs.

## Not in scope

Transport and destination splitting (never broadcast over the sync session; sync from one server and broadcast through another) and deduplication of overlapping results stay with the application — both documented in the crate docs, and the two-trait split in `nym-swizzle-zcash` makes the session separation structural rather than advisory.

**Tuning numbers for the chunking layer are deliberately unvalidated.** Wider overlaps and checkpoint spacing buy a larger anonymity set at the cost of re-downloaded data, and there are no settled values for that trade-off. Defaults are conservative percentage-of-range derivations, exposed as knobs and documented as such, not presented as validated anonymity parameters. (The `nym-swizzle-zcash` grid and broadcast constants, by contrast, follow ZIP 318's published design and are fixed on purpose.) Quantifying anonymity-set size against overlap distribution is left as open research in `design.md`.

Real transaction broadcasting is mocked everywhere — exercising it needs a funded wallet seed, which neither the example nor CI should hold.

This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6984)

## Summary by CodeRabbit

* **New Features**
  * Added the `nym-swizzle` SDK utility for randomized delays and obfuscated, overlapping range chunking with shuffled execution and configurable concurrency.
  * Added WASM-compatible support and examples for sampling, chunk fetching, profiling, and deterministic plans.
  * Added `nym-swizzle-zcash` for quantized sync ranges, reorg verification, deterministic fetching, and persistable broadcast scheduling, including a lightwalletd example.
* **Documentation**
  * Added SDK usage guides, specifications, design notes, and integration examples.
* **Tests**
  * Added extensive coverage for sampling, determinism, range invariants, synchronization, and broadcast scheduling.
* **Chores**
  * Expanded CI and build targets for the new SDK components.


sdk/rust/nym-swizzle/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+repository = "https://github.com/nymtech/nym/sdk/rust/nym-swizzle/"
+homepage.workspace = true
+documentation.workspace = true
+rust-version.workspace = true
+readme = "README.md"
+publish = true
+
+[dependencies]
+futures = { workspace = true }
+log = { workspace = true }
```
jstuczyn: `tracing` would be preferable

sdk/rust/nym-swizzle/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+rand = { workspace = true }
```
jstuczyn: current monorepo uses a rather outdated `rand 0.8` (we are blocked by some upstream deps). I reckon, for this library consumers, it'd be more useful if it was using the most recent `rand 0.10` (we use it in the monorepo as `rand010`)

sdk/rust/nym-swizzle/src/rng.rs
```diff
@@ -0,0 +1,282 @@
...
+
+use rand::rngs::OsRng;
+use rand::{CryptoRng, Rng, RngCore, SeedableRng};
+use rand_chacha::ChaCha20Rng;
+use rand_distr::{Distribution as _, Exp, Normal};
+
+/// Object-safe combination of [`RngCore`] and [`CryptoRng`], so caller-supplied
+/// generators can be stored without generic plumbing. Blanket-implemented for
+/// every suitable RNG.
+pub trait CryptoRngCore: RngCore + CryptoRng + Send {}
```
jstuczyn: `CryptoRngCore` already exists in rand_core` that does this very thing

sdk/rust/nym-swizzle/src/range.rs
```diff
@@ -0,0 +1,485 @@
...
+        F: Fn(u64, u64) -> Fut,
+        Fut: Future<Output = ()>,
+    {
+        let mut delay = self.delay.take();
+        let jobs: Vec<((u64, u64), Option<Duration>)> = (&mut self)
+            .map(|chunk| (chunk, delay.as_mut().map(|d| d.sample())))
+            .collect();
+
+        futures::stream::iter(jobs)
+            .for_each_concurrent(limit.max(1), |((start, end), delay)| {
```
jstuczyn, line 402: wouldn't `limit.max(1)` essentially cause everything to run sequentially?

sdk/rust/nym-swizzle/src/range.rs
```diff
@@ -0,0 +1,485 @@
...
+            }
+        }
+    }
+
+    /// Execute `f` for up to `limit` chunks concurrently, in the plan's
+    /// random order, until every chunk has been executed. When an inter-chunk
+    /// delay is configured, each chunk's delay is pre-sampled and elapses
+    /// inside its own task, so the schedule stays randomized under
+    /// concurrency.
+    pub async fn for_each_concurrent<F, Fut>(mut self, limit: usize, f: F)
```
jstuczyn, line 389: I think it would be useful if there was also some adapter resembling `buffer_unordered` to return the results of the futures

sdk/rust/nym-swizzle-zcash/README.md
```diff
@@ -0,0 +1,194 @@
...
+```rust
+use nym_swizzle_zcash::{Scheduler, BroadcastPlan, TxBroadcaster, expiry_height};
+
+// phase 1: at "user pressed send" — sample ONCE, persist, forget
+let plan = Scheduler::standard().schedule();
+my_db.save(("plan", plan.delay_secs, plan.profile), ("scheduled_at", now));
+
+// phase 2: at every wallet startup (and right after phase 1)
+let plan = my_db.load_plan();
+plan.resume(now - scheduled_at, &mut my_broadcaster, || async {
```
jstuczyn: [SHOULD FIX] the code snippet does not show implementation of the `TxBroadcaster` trait for `my_broadcaster` nor even declaration of `my_broadcaster`

sdk/rust/nym-swizzle-zcash/README.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+And the send path, in two phases so it survives restarts:
+
+```rust
+use nym_swizzle_zcash::{Scheduler, BroadcastPlan, TxBroadcaster, expiry_height};
+
+// phase 1: at "user pressed send" — sample ONCE, persist, forget
+let plan = Scheduler::standard().schedule();
+my_db.save(("plan", plan.delay_secs, plan.profile), ("scheduled_at", now));
```
jstuczyn: how is one meant to create `my_db`? it's not obvious from this example

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    /// Fetch every compact block in the half-open range `[start, end)`.
+    ///
+    /// The half-open range is converted to lightwalletd's inclusive
+    /// `BlockRange` on the wire.
+    pub async fn block_range(&mut self, start: u64, end: u64) -> Result<Vec<CompactBlock>, Status> {
+        debug_assert!(start < end, "empty range {start}..{end}");
+        let request = BlockRange {
+            start: Some(BlockId::at(start)),
+            end: Some(BlockId::at(end - 1)),
```
jstuczyn: this looks like a landmine. you request end at `end`, yet end up retrieving only up to `end-1`

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
```diff
@@ -0,0 +1,142 @@
...
+            .await?;
+        Ok(response.into_inner().height)
+    }
+
+    /// Fetch every compact block in the half-open range `[start, end)`.
+    ///
+    /// The half-open range is converted to lightwalletd's inclusive
+    /// `BlockRange` on the wire.
+    pub async fn block_range(&mut self, start: u64, end: u64) -> Result<Vec<CompactBlock>, Status> {
+        debug_assert!(start < end, "empty range {start}..{end}");
```
jstuczyn: [MUST] this should return an error instead. in a release build the assert won't be hit (and if it did, it would have cause a panic anyway) and it is possible to make nonsense requests

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
```diff
@@ -0,0 +1,142 @@
...
+                Request::new(request),
+                PathAndQuery::from_static(GET_BLOCK_RANGE),
+                ProstCodec::<BlockRange, CompactBlock>::default(),
+            )
+            .await?
+            .into_inner();
+
+        let mut blocks = Vec::new();
+        while let Some(block) = stream.message().await? {
+            blocks.push(block);
```
jstuczyn, line 143: [SHOULD - maybe?] - would it be useful to either change that or add additional method that instead returns the stream so that library consumer was able to process the blocks as they are being delivered rather than having to wait for the entire range to get resolved, especially if the range is quite big?

sdk/rust/nym-swizzle-zcash/examples/wallet_sync.rs
```diff
@@ -0,0 +1,361 @@
...
+        );
+        Ok(())
+    }
+}
+
+/// A `BroadcastPlan` is plain data. This example persists it as two lines of
+/// text to show there's nothing up its sleeve — with the crate's `serde`
+/// feature you'd `serde_json::to_string(&plan)` instead. A real wallet also
+/// stores *when* it scheduled, to compute elapsed time after a restart.
+fn save_plan(path: &std::path::Path, plan: &BroadcastPlan) -> Result<(), BoxError> {
```
jstuczyn: [SHOULD] looking at this example, I wonder if perhaps plan saving/loading should also be a trait to be more closely integrated with the `TxBroadcaster` (or its construction)? and perhaps the plan itself should be a concrete serialisable struct (proto, postcard or even stupid json) rather than what we have in this example

sdk/rust/nym-swizzle-zcash/src/broadcast.rs
```diff
@@ -0,0 +1,397 @@
...
+    ///
+    /// `elapsed` is how much time has passed since [`Scheduler::schedule`] —
+    /// you persisted the scheduling moment, so you know. Freshly scheduled
+    /// and never restarted? Pass `Duration::ZERO`.
+    ///
+    /// `build` runs only after the delay has fully elapsed — fetch the
+    /// current tip *inside it* and set expiry with [`expiry_height`]. Its
+    /// output goes straight to `broadcaster`.
+    pub async fn resume<Bx, F, Fut, E>(
+        self,
```
jstuczyn, line 231: why are we consuming the plan itself on resumption? shouldn't it be mutated instead?

sdk/rust/nym-swizzle-zcash/src/sync.rs
```diff
@@ -0,0 +1,543 @@
...
+    V: FnMut(u64, &S::Block) -> bool,
+{
+    let quantized = quantize(range, tip);
+    let mut window = WindowTracker::new(&quantized);
+
+    let mut results = futures::stream::iter(quantized.requests().map(|(start, end)| {
+        let mut source = source.clone();
+        async move { source.block_range(start, end).await }
+    }))
+    .buffer_unordered(limit.max(1));
```
jstuczyn, line 219: [MUST] what's the point of using `buffer_unordered` if we're setting it to `.max(1)` anyway?

sdk/rust/nym-swizzle-zcash/tests/live.rs
```diff
@@ -0,0 +1,263 @@
...
+    assert_eq!(*heights.last().unwrap(), ee - 1);
+}
+
+/// Measured overhead for the two regimes that occur in practice. The daily
+/// incremental sync is fetched live (grid cover, in actual streamed
+/// blocks); the long catch-up is arithmetic only, to keep the test polite
+/// to the public server. Requests are disjoint, so cover is the only
+/// overhead — streamed equals the emitted length exactly.
+#[tokio::test]
+#[ignore = "talks to a public lightwalletd; run with -- --ignored"]
```
jstuczyn: [MUST]: unfortunately we abuse the `ignore` in our codebase to mark EXPENSIVE tests, not tests that are not meant to run at all; there is a separate step in CI to run all `ignore'd` tests so this one will run.

gate it behind, perhaps, an env value instead like within smol-dvpn

sdk/rust/nym-swizzle/src/delay.rs
```diff
@@ -0,0 +1,236 @@
...
+        self.max = Some(max);
+        self
+    }
+
+    /// Set both bounds at once.
+    ///
+    /// Panics if `min > max`.
+    pub fn bounds(self, min: Duration, max: Duration) -> Self {
+        self.max(max).min(min)
+    }
```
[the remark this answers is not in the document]
mmsinclair: yes, will fix

sdk/rust/nym-swizzle/src/delay.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// Sample one delay from the configured distribution within the
+    /// configured bounds.
+    pub fn sample(&mut self) -> Duration {
+        let min = self.min.as_nanos() as f64;
+        let max = self
+            .max
+            .map(|m| m.as_nanos() as f64)
+            .unwrap_or(f64::INFINITY);
+        let nanos = sample_bounded(&mut self.rng, &self.sampling, min, max);
+        Duration::from_nanos(nanos.round() as u64)
```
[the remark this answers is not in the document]
mmsinclair: yes, will fix

sdk/rust/nym-swizzle/Cargo.toml
mmsinclair: yes, will use `tracing` instead

sdk/rust/nym-swizzle/src/rng.rs
mmsinclair: yes, will use that instead

sdk/rust/nym-swizzle/src/range.rs
mmsinclair: not sure, will investigate

sdk/rust/nym-swizzle/src/range.rs
mmsinclair: yes, good idea, will change

sdk/rust/nym-swizzle-zcash/README.md
mmsinclair: yes, will fix

sdk/rust/nym-swizzle-zcash/README.md
mmsinclair: yes, similar to https://github.com/nymtech/nym/pull/6984#discussion_r3734898707 will fix

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
mmsinclair: yes, that is a landmine, will fix

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
mmsinclair: yes, will fix

sdk/rust/nym-swizzle-zcash/examples/wallet_sync.rs
mmsinclair: yes, that is a good idea, will fix

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
mmsinclair: hmmmm, that might mean a fairly big change, no? will need to see what the impact is

sdk/rust/nym-swizzle-zcash/src/broadcast.rs
mmsinclair: isn't it because the plan is stored outside the library, so it needs to be passed back in? I think if the containing code has lost its state, then it needs to be passed back in again?

sdk/rust/nym-swizzle-zcash/src/sync.rs
mmsinclair: yes good point, will fix

sdk/rust/nym-swizzle-zcash/tests/live.rs
mmsinclair: yes, good idea, will fix

sdk/rust/nym-swizzle-zcash/src/sync.rs
mmsinclair: Turns out this is a misreading — `limit.max(1)` is `std::cmp::max(limit, 1)`, a *lower* clamp: `limit = 4` stays 4 and concurrency works (the concurrent-coverage test exercises it). It exists because `buffer_unordered(0)` panics in `futures`. So behavior is correct; I've added a clarifying comment at the site so the next reader doesn't trip on the same thing.

sdk/rust/nym-swizzle/src/range.rs
mmsinclair: Investigated: it doesn't — `limit.max(1)` is `std::cmp::max(limit, 1)`, a lower clamp, so `limit = 4` stays 4. It only raises a zero limit to 1, because `futures`' concurrency adapters panic on 0. Added a clarifying comment at both sites (this one and the equivalent in nym-swizzle-zcash).

sdk/rust/nym-swizzle-zcash/src/broadcast.rs
mmsinclair: Consuming is deliberate, and free: `BroadcastPlan` is `Copy`, so `resume(self)` costs nothing — and it signals that firing is terminal. A plan has no further meaning once its broadcast has gone out; what outlives the process is the *persisted* copy, and the right move there is to clear it after firing. That lifecycle is now explicit: there's a `PlanStore` trait (per your other comment) and a `resume_pending` helper that does load → wait remainder → build at fire time → send → clear in one call. Also added a doc note on `resume` stating the consume-on-fire semantics.

sdk/rust/nym-swizzle-zcash/examples/support/lightwalletd.rs
mmsinclair: Looked at the impact: with the deterministic split, every wire request is capped at S_FLOOR = 1152 compact blocks (~1–2 MB), and the driver already chops long catch-ups into these cells — so the `Vec` buffering is bounded regardless of range size. Changing `BlockSource` to a stream would be an API-wide change for little gain at that cap, so I'd defer it; worth revisiting if this integrates with librustzcash and its scanning sink wants streams. (The support client itself could trivially expose the tonic stream, but the trait is the contract that matters.) Documented the bound on the trait.

sdk/rust/nym-swizzle-zcash/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+## What it does
+
+**Quantized sync ranges** (`grid`, `sync`). Instead of requesting exactly
+`[resume, tip]`, the wallet requests a range widened to a network-wide grid:
+start rounded down, end rounded up, grid spacing scaled to the range and
+never below one day of blocks. Every wallet resuming anywhere in the same
+grid cell emits **identical boundaries** — anonymity by collision, not by
+noise. The range goes on the wire deterministically: ascending, disjoint,
+day-sized grid cells, no random sizes, no overlap, no shuffling. That
```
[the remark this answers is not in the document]
mmsinclair: Two different grids are in play, deliberately: the *quantization spacing* (ladder-selected, scales with the range, one-day minimum) sets the emitted boundaries, while the *wire-split unit* that chops the emitted range into individual requests is always the fixed 1152-block cell. So "every request is a 1152-block cell" is accurate for every range size, and boundary expectations against the spacing remain correct for the emitted union. Added a sentence to the README distinguishing the two grids so readers don't conflate them either.

jstuczyn (approved):

PR #6988, opened 2026-07-27
 [BandwidthController] Background global-data fetching + global-data-aware readiness

simonwicky:
## Background global-data fetching + global-data-aware readiness

Builds on the `BandwidthController` background-fetch work: the controller now provisions the global ecash signing data (master key, coin-index & expiration-date signatures) **off the event loop**, and `wait_for_ticketbooks` only reports a type ready once the data needed to actually *spend* it is local too.

### Why
- Global-data fetches can be slow nym-api round-trips. Doing them inline blocked the controller's `select!` loop (timer sweep, post-fetch store), stalling every other request.
- Readiness only checked ticket stock, so a waiter could be told "ready" while the signatures needed to spend were still missing — a problem for callers that set a fetcher, wait, then remove it and spend offline.

### What
- **Background prefetch.** Global-data fetches go through the same in-flight tracker as ticketbooks (spawn off-loop, persist on completion). Triggered on ticketbook store, the top-up timer, and fetcher install. The spend path keeps a blocking `ensure_*` fallback (a non-running controller has no loop to drain).
- **Storage query.** New `Storage::get_available_global_data()` returns the epochs / `(epoch, date)` present without loading the blobs — replaces per-piece `is_some()` probing. (sqlx cache regenerated.)
- **Readiness covers global data.** The readiness snapshot gains three maps — master key & coin-index by epoch, expiration by `(epoch, date)` — each built uniformly (present ⇒ Ready, in-flight ⇒ InFlight, absent ⇒ Unavailable). A type is `Ready` only when it's stocked **and** all three pieces for its next ticketbook are present; `InFlight` while any is being fetched. Snapshot assembly lives on the snapshot side; the controller just gathers inputs.
- **`in_flight` module** split into `in_flight/{mod,global_data}.rs`; a single `InFlightFetches` map keyed by a `FetchKey` enum handles both ticketbook and global-data fetches.

### Notes
- Only `PersistentStorage` / `EphemeralStorage` implement the storage trait, both updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/6988)

## Summary by CodeRabbit

* **Improvements**
  * Global signing data is now prefetched proactively in the background to reduce credential delays.
  * Readiness checks now cover all required global credential components in addition to ticketbooks.
  * Ticketbook availability is improved by coordinating ticketbook fetch progress with global-data prefetching.
  * Stored global-data availability metadata is supported consistently across memory and persistent storage.
  * Ticketbook metadata now includes an associated epoch identifier.
* **Bug Fixes**
  * More reliable wait-and-serve behavior when required data is missing or global fetches fail.
* **Tests**
  * Removed an end-to-end signer-failure test scenario.
* **Documentation**
  * Updated a crate documentation example for channel device construction.


common/bandwidth-controller/src/in_flight/global_data.rs
```diff
@@ -0,0 +1,74 @@
...
+    },
+}
+
+impl GlobalDataRequest {
+    /// The pieces of global data needed to spend a ticketbook of the given epoch and expiration.
+    pub(crate) fn for_ticketbook(
+        epoch_id: EpochId,
+        expiration_date: Date,
+    ) -> [GlobalDataRequest; 3] {
+        [
```
jstuczyn, line 31: this enum makes it things weird because technically the type system allows for `[GlobalDataRequest::MasterVerificationKey(epoch_id), GlobalDataRequest::MasterVerificationKey(epoch_id), GlobalDataRequest::MasterVerificationKey(epoch_id)]

what if instead it was a struct with `Option` field for each request type? though then, I guess, the type system would allow for multiple `None`... 🤔

common/bandwidth-controller/src/readiness.rs
```diff
@@ -39,44 +43,145 @@ pub(crate) struct FetchFailure {
...
+/// Each map holds only what's known to be `Ready` or `InFlight`; anything absent is `Unavailable`
+/// at resolution time. Global data is tracked per piece, keyed as it's stored: master key and
+/// coin-index signatures by epoch, expiration-date signatures by `(epoch, date)`.
 #[derive(Debug)]
 pub(crate) struct ReadinessSnapshot {
-    pub(crate) upgrade_mode: bool,
-    pub(crate) tickets_readiness: HashMap<TicketType, ReadinessStatus>,
+    upgrade_mode: bool,
+    tickets_readiness: HashMap<TicketType, ReadinessStatus>,
+    master_key_readiness: HashMap<EpochId, ReadinessStatus>,
```
jstuczyn, line 54: is there really a point of separating it per type? why not have a single HashMap<EpochId, GlobalData> instead?

common/bandwidth-controller/src/controller.rs
```diff
@@ -292,11 +304,28 @@ impl<St: Storage> BandwidthController<St> {
         self.get_available_ticketbooks().await
     }
 
-    async fn handle_wait_for_ticketbooks(&mut self, request: ReadinessRequest) {
+    async fn handle_wait_for_ticketbooks(
```
jstuczyn, line 306: I agree with the following finding which was written in a more readable language (albeit more robotic 🤖 )

A failed global-data prefetch strands wait_for_ticketbooks for up to 3 hours

handle_wait_for_ticketbooks (controller.rs:307) evaluates global-data readiness but never triggers a fetch. Neither does check_and_restock, which is the only other thing the SDK calls before waiting (sdk/rust/nym-sdk-session/src/session.rs:305-314:
restock_ticketbooks then wait_for_ticketbooks). The only prefetch triggers are fetcher install, ticketbook-store completion, and the topup tick, whose default is 3 hours (config.rs:34).

So: ticketbooks already stocked, the startup prefetch of a global-data piece fails transiently → absent and not in flight → the snapshot resolves Unavailable → the waiter gets TicketbooksUnavailable immediately. A caller retrying ensure_ticket_types gets
the same answer every time, because nothing in that path re-spawns the fetch. Before this PR the waiter would have returned Ready and the spend path's inline ensure_* would have retried and most likely succeeded - so this is a resilience regression on
exactly the flow the PR targets.

Fix is one line: self.prefetch_global_data().await; at the top of handle_wait_for_ticketbooks (turns the state into InFlight → parked → resolved on completion). Adding it to check_and_restock too would make the SDK's restock-then-wait pair self-healing.

jstuczyn (approved):

common/bandwidth-controller/src/controller.rs
simonwicky: Adding a global data check to the check_and_restock path. 
It has nothing to do with wait_for_ticketbooks and the sdk session thing is using it wrong

PR #7008, opened 2026-07-31
bugfix: allow NMv3 to correctly test nodes with multiple ip addresses

jstuczyn:
## Problem

An NM agent announced exactly one mixnet address, and the orchestrator's node refresher picked one of a node's self-described IPs **at random** every refresh cycle (`node_refresher.rs`: `ip_address.choose(&mut rand::thread_rng())`). So whenever the dice landed on a dual-stack node's ipv6 address, the agent connected from its ipv6 egress, which had never been authorised on-chain, and the node rejected the probe on all three of its gates (routing filter, Noise responder, and the sphinx replay/bloomfilter bypass) because they key on source IP. The run reported 0 received out of the full expected count, indistinguishable from a dead node, with nothing recording which address had been measured.

## What this changes

**Agents announce a pair of addresses, both authorised in one transaction.** `AgentMixAddresses { v4, v6 }` lives in the requests crate and is embedded in both the announce and request-testrun bodies. The orchestrator authorises both addresses in a single`execute_multiple` transaction, so an agent is never left with only one family authorised, and the write lock on the signing client is taken once instead of twice.

**The orchestrator's agent cache is one record per agent.** `KnownAgents` is now `HashMap<SocketAddr, KnownAgent>` keyed by the agent's ipv4 socket address (the port matters: several agents share a host IP) with the ipv6 address held in the record and a single `announced` flag. The contract can't be redeployed, so it still holds one entry per socket address with nothing linking the two; on restart the orchestrator recovers the pairs by grouping entries on their x25519 noise key, which is unique per agent, and drops anything that isn't exactly one ipv4/ipv6 pair rather than guessing. That's safe because the cache only exists to skip redundant transactions, and an agent always announces before requesting work.

**Test targets are deliberate and attributable.** The refresher now stores every announced address (canonicalised, deduplicated and sorted, because the rotation is positional and a node may report them in any order), and assignment rotates onto the node's next address inside the existing `BEGIN IMMEDIATE` transaction, advancing the pointer at hand-out so an abandoned run still moves on. The agent reports which address it tested, which is persisted on the testrun row and exposed as `tested_address` on the results API, so a broken ipv6 path is attributable instead of being averaged into the node's single result series. Note the staleness gate stays per node while coverage is per address, so a node announcing N addresses has each one tested every N x `test_interval`.

**The agent no longer assumes the node replies from the address it dialled.** The assignment carries `node_ips`, and the agent treats every one of them as the node itself. Two gates needed this, not one: the listener's source check (`listener/mod.rs`), and the responder-side noise map, because `upgrade_noise_responder` looks the initiator up by source IP and silently downgrades to raw TCP on a miss, which the listener then rejects as "does the node support the protocol?". Both comparisons are canonicalised, which also fixes a live bug: with the default `[::]:9000` bind, an inbound ipv4 peer presents as `::ffff:a.b.c.d` and never matched the stored plain-ipv4 address. The sphinx return hop now follows the tested address's family, so a run exercises one family in both directions rather than measuring a node's ipv6 ingress against its ipv4 egress.

**Validation.** Announcements must be one plain ipv4 address and one non-ipv4-mapped ipv6 address, rejected with a 400 rather than normalised: an `::ffff:` address in the v6 slot collapses onto the ipv4 entry once a node canonicalises it, leaving one authorised ingress while the contract and orchestrator both believe there are two. The agent applies the same check at config construction so a misconfigured deployment fails at startup rather than at its first announce.

## Breaking changes

Agent and orchestrator must be redeployed together. The contract is unchanged.

## Deployment notes

- Every agent deployment now needs an announced ipv6 egress address; there is no single-address fallback.
- The agent's mixnet listener must stay dual-stack (`[::]:port`, the default) or ipv6 return traffic has nowhere to land.
- Migration `02_node_announced_ips.sql` is additive with nullable columns and needs no backfill: nodes fall back to their single stored address until their next refresh cycle.
- Each agent now occupies two contract entries, so the agent map holds roughly two per agent. Worth keeping in mind against the known cascade-delete gas limitation in `RevokeNetworkMonitorOrchestrator`.

### Practical notes:
1. Rebuild images for both the orchestrator and the agent
2. Shut everything down
3. On each agent update env value of `NYM_NETWORK_MONITOR_AGENT_HOST_IP` to `NYM_NETWORK_MONITOR_AGENT_HOST_IP_V4`
4. On each agent set `NYM_NETWORK_MONITOR_AGENT_HOST_IP_V6` value to its ipv6 static address
5. Start orchestrator with the new version
6. Start the agents with the new version

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7008)

## Summary by CodeRabbit

- **New Features**
  - Added dual-stack IPv4 and IPv6 support for agent announcements, assignments, and network tests.
  - Tests now select compatible address families and record the specific address tested.
  - Node testing supports multiple announced addresses with canonicalization, deduplication, and rotation.
  - Added metrics for changed agent details and malformed address announcements.
- **Bug Fixes**
  - Improved validation of malformed, duplicate, or mismatched address pairs.
  - Enhanced connection verification across known node addresses.
- **Documentation**
  - Updated specifications for dual-stack configuration, authorization, persistence, and address-aware testing.


nym-network-monitor-v3/nym-network-monitor-agent/src/agent/tested_node.rs
```diff
@@ -6,16 +6,21 @@ use nym_crypto::asymmetric::x25519;
...
 pub(crate) struct TestedNodeDetails {
     pub(crate) node_id: Option<u32>,
 
     /// TCP socket address of the node's mixnet listener, used for the egress connection.
     pub(crate) address: SocketAddr,
 
+    /// Every ip address the node is known by, canonicalised. The node may return the test packets
+    /// from an address other than the one it was reached on (it may be multi-homed, or be replying
+    /// over a different family), so all of them have to be treated as the node itself.
+    pub(crate) known_ips: Vec<IpAddr>,
```
simonwicky, line 22: Given its use, could this be a `HashSet` instead?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
```diff
@@ -102,25 +118,51 @@ impl KnownAgents {
...
+/// changed one of its addresses. Dropping them is safe because this cache exists purely to skip
+/// redundant contract transactions - agents always announce before requesting work, which
+/// re-creates the entry at the cost of one extra transaction.
 impl TryFrom<Vec<AuthorisedNetworkMonitor>> for KnownAgents {
     type Error = anyhow::Error;
 
     fn try_from(agents: Vec<AuthorisedNetworkMonitor>) -> Result<Self, Self::Error> {
-        let mut agents_map = HashMap::new();
-
+        let mut by_noise_key: HashMap<String, Vec<SocketAddr>> = HashMap::new();
```
simonwicky, line 133: That is...shaky no?

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
```diff
@@ -8,91 +8,107 @@ use crate::storage::models::NewTestRun;
...
 use std::sync::Arc;
 use std::time::Duration;
 use time::OffsetDateTime;
 use tokio::sync::{Mutex, RwLock};
-use tracing::error;
+use tracing::{error, warn};
 
-/// Thread-safe cache of all agents known to this orchestrator, keyed by host IP.
-/// Used to short-circuit the contract tx for already-announced agents.
+/// Thread-safe cache of all agents known to this orchestrator, keyed by the agent's IPv4 mixnet
```
simonwicky, line 26: What's the argument against keying by both?

simonwicky (approved):

nym-network-monitor-v3/nym-network-monitor-agent/src/agent/tested_node.rs
jstuczyn: for small number of entries Vec is actually more efficient

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
jstuczyn: what do we do if only ipv6 address changes? ironically ipv4 is guaranteed to be more stable

nym-network-monitor-v3/nym-network-monitor-orchestrator/src/http/state.rs
jstuczyn: unfortunately best we can do for now : /

PR #7005, opened 2026-07-29
Max/docs threat model overhaul

mfahampshire:
CLOSE THIS WHEN #7027 IS MERGED 

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive threat-model documentation section covering actors, vectors, unlinkability, configurations, comparisons, and worked examples.
  * Added interactive visualizations for network configurations, packet anatomy, mixnet mixing, dVPN cover traffic, and latency.
  * Added wallet, messaging, and web-browsing privacy examples, plus a hidden-services deep dive.
* **Documentation**
  * Updated navigation, redirects, developer guidance, and transport-mode explanations.
  * Added Simplified Technical English guidance and a dismissible threat-model banner.
* **Tests**
  * Added automated latency and navigation validation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7005)

PR #7027, opened 2026-08-04
Max/docs ai assistant mcp + threat model overhaul

mfahampshire:
~~A note on this - the Voyage key has not been uploaded to Vercel, so the 'Ask AI' button doesn't work on the Vercel preview. I've run it locally as an env var, works well.~~

TODO
- [ ] Replace testing `VOYAGE` key with prod key in Vercel 

  Combines two docs branches: the threat-model overhaul (#7005) and the MCP retrieval server. #7005 is contained in
  this branch and should auto-close on merge.

  Threat model

  - /network/threat-model rebuilt around a two-layer model (transport vs hygiene). Actors, vectors and properties
  render from a typed spine in lib/privacy-model/, so every card, table and scenario comes from one source.
  - /developers restructured into a runtime x E2E/proxy front door. integrations.mdx removed, ~11 redirects added in
  next.config.js.
  - New pages: limitations.mdx (workloads Nym cannot serve), swizzle.mdx + swizzle/zcash.mdx, exit security,
  use-with-ai.mdx.

  MCP server

  - lib/mcp/ serves the docs over Streamable HTTP JSON-RPC. Tools: search_docs, search_code, get_page, network_summary.
  - lib/retrieval/ builds the index. Voyage embeddings, content-hash cache, MDX chunker that expands partials,
  substitutes version constants and projects React components into text.
  - ChatWidget removed in favour of McpPanel, which hands out connection config. MCP answers were consistently better
  than the chat.
  - scripts/check-mcp-server.sh runs 37 assertions against a deployed instance. Coverage checks match on symbols rather
  than prose, because prose queries decay as the corpus grows.
  - lib/retrieval/coverage.test.ts fails the build if a page or partial contributes no indexable text.

  Notes for review

  - The 24k deletions are almost entirely documentation/docs/public: generated per-page markdown and the retrieval
  index, now built rather than committed. Nothing authored was removed.
  - documentation/agent-scenarios.md records the scenario suite used to validate retrieval, including known open
  findings.
  - documentation/MCP-SERVER.md holds the architecture writeup that was previously inline in the README.
  - nym-swizzle is documented but not yet released; the page says so.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7027)


mmsinclair (approved):
🚀

PR #7036, opened 2026-08-06
Clean up usage of nym-bridges

jmwample:
Three crates independently redefined the same "QUIC bridge connect params" shape (addresses/host/id_pubkey) and smoldvpn hand-rolled its own dial+open_bi+framing on top of `nym-bridges` 0.1. This collapses all of it onto the canonical `nym-bridges-types::ClientConfig` and the newer `nym-bridges::BridgeConn`:

- **nym-node-status-api**: `node_scraper::models::BridgeInformation` is now a
  re-export of `nym_bridges_types::PersistedClientConfig` instead of a hand-copied
  struct.
- **nym-sdk-session**: removed the `QuicBridge` type entirely; `SelectedGateway`/
  `HopConfig` now carry `nym_bridges_types::ClientConfig` directly, and directory
  parsing picks the first transport `ClientConfig::is_usable`/`usable_transports`
  accepts instead of hard-coding `quic_plain`. Renamed the QUIC-specific surface
  to generic bridge terms: `has_quic`→`has_bridge`, `NoQuicGateway`→
  `NoBridgeGateway`, `register_two_hop_quic`→`register_two_hop_bridge`, etc.
- **smoldvpn**: `bridge.rs`'s `BridgeParams` is now `ClientConfig` (any transport,
  not just QUIC); `connect()` is a thin wrapper over `BridgeConn::try_connect` +
  `.into_parts()`, dropping the manual `open_bi`/timeout-select plumbing and the
  now-redundant IPv4-reorder hack (nym-bridges 0.2 does IPv4 selection itself).
  `QuicBridgeSender`/`Receiver` → `BridgeSender`/`Receiver`.
- Depends on `nym-bridges`/`nym-bridges-types` `v0.2.1`

Going forward, adding a new bridge transport (e.g. TLS, and others upcoming) only need lib version changes — nym-sdk-session and smoldvpn already operate on the transport-generic
`ClientConfig`/`BridgeConn`.

## Testing
- [x] `cargo test -p nym-sdk-session` — dvpn/gateway/session unit tests (29) pass
- [x] `cargo test -p nym-smoldvpn bridge::` — real cert-pinning connect/reject
      round-trip against a live mock bridge server, via `BridgeConn`, passes
- [x] `cargo check -p nym-node-status-api --all-targets`
- [x] `cargo clippy -p nym-smoldvpn -p nym-sdk-session --all-targets` — clean

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7036)

## New Features
- Added support for multiple bridge transport types beyond QUIC.
- Bridge discovery selects the first usable supported transport.
- Added generic bridge registration and gateway selection.
- Bridge connections use shared transport configuration.

## Improvements
- Invalid or unsupported bridge data is handled safely.
- Updated examples and SDK documentation for generic bridge usage.
- Improved bridge transport reporting and error messages.
- Replaced QUIC-specific bridge APIs and errors with generic bridge terminology.


nym-node-status-api/nym-node-status-api/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "nym-node-status-api"
-version = "4.6.2"
+version = "4.6.3-rc1"
```
mfahampshire: Why an `-rc` version bump instead of a proper one? 

If we want to test it with these changes before rolling it out wholesale we can do that by coordinating with @benedettadavico during the QA process.

mfahampshire (approved):
Overall this looks good to me. 

I would say that I would prefer if we could move the bridges crate into the monorepo sooner rather than later; as soon as we're making changes elsewhere that require (in this case) major version bumps in the SDK crates (since we keep all the monorepo crates in lockstep wrt versioning) I'd rather that instead of being caught in a PR (or after release) we had more visibility on changes per release. 

One other thing - I'm not sure how this affects the nymvpn repo / how they're importing the bridge crate, so I think we should get an opinion from one of the vpn team devs as well to check. Since it is affecting the nym-smoldvpn it might be affecting them too, good to check.

nym-node-status-api/nym-node-status-api/Cargo.toml
jmwample: This is an `rc` because CI was failing over a version collision and I figured changes in progress were `rc` until we do release changes -- It doesn't actually make a difference to me.

The changes should make no difference to the client as the types moved, but through re-exports they should be the same. It wouldn't hurt to test that quic still works though.

jmwample:
> I'm not sure how this affects the nymvpn repo / how they're importing the bridge crate, so I think we should get an opinion from one of the vpn team devs as well to check.

Just so it is noted here  -- I don't believe that this PR effects the vpn-client usage. @pronebird if you want to take a look to confirm. 

From the VPN perspective they are using the `nym-bridges` crate(s) starting at: https://github.com/nymtech/nym-vpn-client/pull/5928

And I made a PR to update to apply the `0.2.1` version here https://github.com/nymtech/nym-vpn-client/pull/6211

PR #7061, opened 2026-08-14
bugfix: reject packets to unknown clients

jstuczyn:
do not store mixnet packets for clients that never registered with this gateway. eventually there won't be any offline storage at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7061)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented messages for unknown recipients from being stored.
  * Continued forwarding acknowledgments even when message delivery is skipped.
  * Preserved handling and reporting of genuine storage failures.

* **Monitoring**
  * Added metrics for final-hop packets dropped because the recipient is unknown.
  * Added logging and counters distinguishing stored, skipped, and failed messages.

* **Tests**
  * Added coverage for valid recipients, unknown recipients, and repeated delivery attempts.

### DIFF
--- a/common/gateway-storage/src/traits.rs
+++ b/common/gateway-storage/src/traits.rs
@@ -46,11 +46,15 @@ pub trait SharedKeyGatewayStorage {
 
 #[async_trait]
 pub trait InboxGatewayStorage {
+    /// Store a message for the given recipient, returning whether it was stored.
+    ///
+    /// A recipient that has never registered with this gateway could never retrieve the message,
+    /// so nothing is stored for one and `false` is returned.
     async fn store_message(
         &self,
         client_address: DestinationAddressBytes,
         message: Vec<u8>,
-    ) -> Result<(), GatewayStorageError>;
+    ) -> Result<bool, GatewayStorageError>;
     async fn retrieve_messages(
         &self,
         client_address: DestinationAddressBytes,
--- a/common/gateway-storage/src/traits.rs
+++ b/common/gateway-storage/src/traits.rs
@@ -47,9 +47,6 @@ pub trait SharedKeyGatewayStorage {
 #[async_trait]
 pub trait InboxGatewayStorage {
     /// Store a message for the given recipient, returning whether it was stored.
-    ///
-    /// A recipient that has never registered with this gateway could never retrieve the message,
-    /// so nothing is stored for one and `false` is returned.
     async fn store_message(
         &self,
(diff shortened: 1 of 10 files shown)


nym-node/src/node/mixnet/shared/final_hop.rs
```diff
@@ -8,6 +8,16 @@ use nym_sphinx_types::DestinationAddressBytes;
 use tokio::time::Instant;
 use tracing::{debug, warn};
 
+/// Outcome of trying to persist a payload for a recipient with no live session.
+pub(crate) enum InboxOutcome {
```
simonwicky: Imo, this is a bit overkill, it could just be `true`/ `false` and spare quite some loc in a couple of places

common/gateway-storage/src/traits.rs
```diff
@@ -46,11 +46,15 @@ pub trait SharedKeyGatewayStorage {
 
 #[async_trait]
 pub trait InboxGatewayStorage {
+    /// Store a message for the given recipient, returning whether it was stored.
+    ///
+    /// A recipient that has never registered with this gateway could never retrieve the message,
```
simonwicky: I'm gonna nitpick here and say that the second part of the comment shouldn't be on the trait. 
Implementation should return wether it's stored or not, why it was/wasn't is up to the implementation

nym-node/src/node/mixnet/handler.rs
```diff
@@ -302,6 +303,24 @@ impl ConnectionHandler {
...
+                            .metrics
+                            .mixnet
+                            .egress
+                            .add_unknown_recipient_dropped_packet();
+                        self.shared
+                            .dropped_final_hop_packet(self.remote_address.ip());
+
+                        // no ack: unlike the disk fallback, this payload is gone for good, so the
+                        // sender must not be told it landed
+                        return;
```
simonwicky: Actually, this can leak information about whether a client is registered or not. 
Say I know a client's key, I can send a message to all gateways, and the one where I will get ACKs from are the exact ones where it is registered.

simonwicky (changes requested):

nym-node/src/node/mixnet/shared/final_hop.rs
jstuczyn: good idea. done

simonwicky (approved):

PR #7085, opened 2026-08-21
Add prune command to BandwidthController

neacsu:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7085)

## Summary by CodeRabbit

- **New Features**
  - Expired ticketbooks are automatically removed during periodic bandwidth top-up checks.
  - Credential fetching now cleans up expired pending ticketbooks before requesting new ones.
  - Cleanup failures are logged as warnings without interrupting normal processing.

- **Tests**
  - Added coverage for pruning empty, expired, and active ticketbook storage.

### DIFF
--- a/common/bandwidth-controller/src/controller.rs
+++ b/common/bandwidth-controller/src/controller.rs
@@ -504,12 +504,6 @@ impl<St: Storage> BandwidthController<St> {
         } else {
             tracing::debug!("Finished pruning of expired ticketbooks");
         }
-
-        if let Some(fetcher) = &self.credential_fetcher {
-            if let Err(err) = fetcher.prune().await {
-                tracing::warn!("Could not prune fetcher expired ticketbooks: {err}");
-            }
-        }
     }
 
     /// Spawns a background fetch for `ticket_type` unless one is already in flight for it.
--- a/common/bandwidth-controller/src/controller.rs
+++ b/common/bandwidth-controller/src/controller.rs
@@ -155,6 +155,7 @@ impl<St: Storage> BandwidthController<St> {
                     break;
                 }
                 _ = topup_interval.tick() => {
+                    self.prune_expired().await;
                     let _ = self.print_info().await;
                     self.check_and_restock(self.config.managed_ticket_types.clone()).await;
                 }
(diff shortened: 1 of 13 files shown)


simonwicky:
Why can't BC prune outdated credentials on its own? Why do we need a command for it? 
Same thing for the fetcher, let them be independent, BC doesn't care about what they are doing with their storage

common/bandwidth-controller/src/controller.rs
```diff
@@ -296,6 +299,37 @@ impl<St: Storage> BandwidthController<St> {
...
 
+    // Removes expired ticketbooks from storage and expired ticketbooks that are still in pending.
+    async fn handle_prune(&mut self) -> Result<(), BandwidthControllerError> {
+        let expired_stored_ret = self
+            .storage
+            .cleanup_expired()
+            .await
+            .map_err(BandwidthControllerError::credential_storage_error);
+
+        let expired_pending_ret = if let Some(fetcher) = &self.credential_fetcher {
```
simonwicky: Let the fetcher handle pruning itself, this is very much breaking the abstraction

simonwicky (changes requested):

common/bandwidth-controller/src/traits.rs
```diff
@@ -98,6 +98,9 @@ pub trait CredentialFetcher: CredentialPublicDataFetcher + Send + Sync {
         ticketbook_type: TicketType,
     ) -> Result<Vec<NymCredential>, CredentialFetcherError>;
 
+    /// Stops in-flight retrieval of blinded ticketbook shares whose validity has passed.
+    async fn prune(&self) -> Result<(), CredentialFetcherError>;
```
simonwicky: I feel like this is a bit too much. 
IMO the BC should not stop in-flight retrieval, if it has a fetcher, it's not supposed to do that. 
Fetcher do need a prune, but each should have their own behavior no?

simonwicky (commented):
Looks nice :)
I still wonder about that `prune` method on the fetcher, it should be handled internally as well. Fetcher deciding what to do

simonwicky (approved):

PR #7099, opened 2026-08-26
Feat/allow future lp compat

jstuczyn:
Adjust LP version passing to allow for introduction of a v2 in the future

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7099)

## Summary by CodeRabbit

* **New Features**
  * Added protocol-version negotiation based on node versions.
  * Added compatibility handling for supported older node versions.
  * Nodes with unsupported or unrecognized protocol versions are rejected.

* **Bug Fixes**
  * Sessions now enforce the negotiated packet version when processing packets.
  * Improved packet-version errors to distinguish negotiation mismatches from unsupported layouts.
  * Prevented replay-window updates for packets with mismatched versions.

* **Tests**
  * Added coverage for negotiation, unsupported versions, compatibility, and malformed packets.


common/nym-lp-data/src/packet/mod.rs
```diff
@@ -15,8 +15,78 @@ pub mod frame;
...
+            return None;
+        }
+        // currently there are no other branches known to the client
+        // once a new version is introduced, follow the pattern implemented in
+        // `common/authenticator-requests/src/version.rs`
+        Some(V1)
+    }
+
+    #[cfg(test)]
+    mod tests {
```
simonwicky: I know these are tests for the version module, but for readability, could we either have those at the end of the file, or in a new file? 
I'm not fond of unit tests before actual code in a file

common/nym-lp/src/codec.rs
```diff
@@ -55,17 +55,20 @@ pub(crate) fn encrypt_lp_packet(
...
+    expected_version: u8,
 ) -> Result<LpPacket, LpError> {
     if packet.ciphertext().len() < InnerHeader::SIZE + SANE_DEC_OVERHEAD {
         return Err(LpError::InsufficientBufferSize);
     }
 
     let plaintext = decrypt_data(packet.ciphertext(), transport)?;
 
-    let inner_header = InnerHeader::parse(&plaintext)?;
+    let inner_header = InnerHeader::parse(&plaintext, expected_version)?;
```
simonwicky: Failing here seems a bit out of place. 
Since we have to decrypt anyway, how about do that actual check where the counter checks are done? It feels more natural. 
This is like a low level decryption tool, it shouldn't fail because of the version

common/nym-lp/src/session.rs
simonwicky, line 353: IMO, do the version check here

nym-gateway-probe/src/common/nodes.rs
```diff
@@ -142,15 +142,20 @@ impl DirectoryNode {
 
                 let version: semver::Version =
                     description.build_information.build_version.parse()?;
-                let Some(ciphersuite) = Ciphersuite::from_node_version(version) else {
+                let Some(ciphersuite) = Ciphersuite::from_node_version(version.clone()) else {
                     bail!("failed to identify valid ciphersuite for {identity}");
                 };
+                // nodes don't advertise their LP version, so derive it from the build version
+                // the same way the ciphersuite is
+                let Some(lp_version) = version::from_node_version(version) else {
```
simonwicky, line 150: looking at this `version::` call, maybe it could benefit from being an enum?

jstuczyn: currently it's just one value. when we do a V2, we should change it then.

common/nym-lp/src/codec.rs
jstuczyn: there are actually two checks here. The "no layout implemented for this version" one has to stay in the parser because `parse` needs to know which byte layout it's decoding (ehm. possible TSL encoding changes in V2), so there's no way around failing there.

the other, i.e.  got-vs-negotiated check could move up, sure, but the version byte is inside the ciphertext anyway, so the earliest it can run is right after decryption - same place it runs now. Moving it to the session buys nothing behaviour-wise, it just makes it possible to call `decrypt_lp_packet` and forget the check. With the parameter you can't so I'd rather leave it.

common/nym-lp/src/session.rs
jstuczyn: same thing as the codec comment - by the time we're marking counters the packet is already decrypted, so the check would run at the same point anyway

simonwicky (approved):

PR #7105, opened 2026-08-26
HTTP front / retry policy

jmwample:
Improves domain-fronting resilience in `http-api-client`: rate-limit errors now trigger host rotation without forcing fronting on, fronting policy is more configurable, and shared-resource test flakiness is fixed.

* More configurable fronting policy:
  * Even while fronting is enabled, the client still rotates through non-fronting hosts rather than sticking exclusively to fronted ones.
  * Added recovery logic: if non-fronting domains start succeeding again, the client can transition back from fronting-enabled to fronting-disabled instead of being stuck once triggered.
* Test stability: isolated tests that share a common resource so they no longer interfere with each other's state when run concurrently.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7105)

## Summary by CodeRabbit

* **New Features**
  * Added configurable domain-fronting retries with failure thresholds, domain rotation, and recovery options.
  * Non-fronted hosts can participate in rotation when enabled.
  * Fronting can recover automatically after successful direct requests.

* **Bug Fixes**
  * Improved detection of rate limits from HTTP 429 responses, throttled 503 responses, and Vercel challenges.
  * Improved fronting state reset after policy changes and successful requests.

* **Tests**
  * Added coverage for retry thresholds, rotation, recovery, and rate-limit detection.

### DIFF
--- a/common/http-api-client/src/fronted.rs
+++ b/common/http-api-client/src/fronted.rs
@@ -666,18 +666,36 @@ mod tests {
         // println!("{response:?}");
         assert_eq!(response.status(), 200);
     }
+}
 
-    // sends real requests and relies on host rotation not being suppressed, which depends on
-    // the process-wide SHARED_NETWORK_RECONFIGURATION marker - must not run concurrently with
-    // tests that mutate it.
+#[cfg(test)]
+mod mocked_tests {
+    use super::*;
+    use crate::{ApiClientCore, HickoryDnsResolver, NO_PARAMS, Url};
+    use hickory_resolver::{
+        net::{DnsError, NetError, NoRecords},
+        proto::{
+            op::{Query, ResponseCode},
+            rr::{Name as HickoryName, RecordType},
+        },
+    };
+    use reqwest::dns::{Addrs, Name, Resolve, Resolving};
+    use std::{
+        net::{IpAddr, SocketAddr},
+        str::FromStr,
+    };
+
     #[tokio::test]
-    #[serial]
     async fn fallback_on_failure() {
+        // `fake-front-1`/`fake-front-2` are pinned to deterministic DNS failures via
+        // `MockResolver` below (one NXDOMAIN, one SERVFAIL - exercising both branches of
+        // `might_be_network_interference`'s DNS classification) rather than relying on any live
+        // DNS infrastructure to keep answering consistently for hostnames that don't exist.
         let url1 = Url::new(
-            "https://fake-domain.nymtech.net",
+            "https://fake-domain.invalid",
             Some(vec![
-                "https://fake-front-1.nymtech.net",
-                "https://fake-front-2.nymtech.net",
+                "https://fake-front-1.invalid",
+                "https://fake-front-2.invalid",
             ]),
         )
         .unwrap();
@@ -687,20 +705,25 @@ mod tests {
         )
         .unwrap(); // fastly
 
+        let mock_resolver = MockResolver::new()
+            .with_nxdomain("fake-front-1.invalid")
+            .with_servfail("fake-front-2.invalid");
+
         let client = ClientBuilder::new_with_urls(vec![url1, url2])
             .expect("bad url")
             .with_fronting(Some(FrontPolicy::Always))
+            .dns_resolver(std::sync::Arc::new(mock_resolver))
             .build()
             .expect("failed to build client");
 
         // Check that the initial configuration has the broken domain and front.
         assert_eq!(
             client.current_url().as_str(),
-            "https://fake-domain.nymtech.net/",
+            "https://fake-domain.invalid/",
         );
         assert_eq!(
             client.current_url().front_str(),
-            Some("fake-front-1.nymtech.net"),
+            Some("fake-front-1.invalid"),
         );
 
         let result = client
@@ -716,11 +739,11 @@ mod tests {
         // Check that the host configuration updated the front on error.
         assert_eq!(
             client.current_url().as_str(),
-            "https://fake-domain.nymtech.net/",
+            "https://fake-domain.invalid/",
         );
         assert_eq!(
             client.current_url().front_str(),
-            Some("fake-front-2.nymtech.net"),
+            Some("fake-front-2.invalid"),
         );
 
         let result = client
@@ -742,17 +765,99 @@ mod tests {
             client.current_url().front_str(),
             Some("yelp.global.ssl.fastly.net"),
         );
+    }
 
-        let response = client
-            .send_request::<_, (), &str, &str>(
-                reqwest::Method::GET,
-                &["api", "v1", "network", "details"],
-                NO_PARAMS,
-                None,
-            )
-            .await
-            .expect("failed get request");
+    /// Deterministic outcome a [`MockResolver`] returns for one hostname.
+    #[derive(Clone)]
+    enum MockOutcome {
+        /// Resolve to these addresses, as if the lookup succeeded.
+        #[allow(dead_code)]
+        Addrs(Vec<IpAddr>),
+        /// Fail as a clean NXDOMAIN (the domain does not exist).
+        NxDomain,
+        /// Fail as a SERVFAIL (the server could not process the query).
+        ServFail,
+    }
 
-        assert_eq!(response.status(), 200);
+    /// A [`Resolve`] implementation for tests that need a specific, deterministic DNS outcome
+    /// for one or more hostnames without depending on any live DNS infrastructure for them.
+    ///
+    /// Any hostname not explicitly configured falls through to a real (independent, non-shared)
+    /// resolver, so a test can mock only the hosts it cares about (e.g. hosts standing in for a
+    /// network failure) while still reaching the network for the rest (e.g. a host the test
+    /// expects to actually succeed against).
+    struct MockResolver {
+        outcomes: HashMap<String, MockOutcome>,
+        fallback: HickoryDnsResolver,
+    }
+
+    impl MockResolver {
+        fn new() -> Self {
+            Self {
+                outcomes: HashMap::new(),
+                fallback: HickoryDnsResolver::thread_resolver(),
+            }
+        }
+
+        fn with_nxdomain(mut self, host: &str) -> Self {
+            self.outcomes
+                .insert(host.to_string(), MockOutcome::NxDomain);
+            self
+        }
+
+        fn with_servfail(mut self, host: &str) -> Self {
+            self.outcomes
+                .insert(host.to_string(), MockOutcome::ServFail);
+            self
+        }
+
+        #[allow(dead_code)]
+        fn with_addrs(mut self, host: &str, addrs: Vec<IpAddr>) -> Self {
+            self.outcomes
+                .insert(host.to_string(), MockOutcome::Addrs(addrs));
+            self
+        }
+    }
+
+    /// Wraps a [`NetError`] the same way the real resolver does (see `dns::ResolveError`), so it
+    /// reaches [`crate::might_be_network_interference`] in exactly the shape a genuine DNS
+    /// failure takes: a `crate::ResolveError::ResolveError(NetError)`, one `.source()` hop below
+    /// the `reqwest::Error` the client actually sees.
+    fn mock_dns_error(net_err: NetError) -> Box<dyn std::error::Error + Send + Sync> {
+        Box::new(crate::ResolveError::ResolveError(net_err))
+    }
+
+    impl Resolve for MockResolver {
+        fn resolve(&self, name: Name) -> Resolving {
+            let host = name.as_str().to_string();
+            match self.outcomes.get(&host) {
+                Some(MockOutcome::Addrs(addrs)) => {
+                    let addrs = addrs.clone();
+                    Box::pin(async move {
+                        let addrs: Addrs =
+                            Box::new(addrs.into_iter().map(|ip| SocketAddr::new(ip, 0)));
+                        Ok(addrs)
+                    })
+                }
+                Some(MockOutcome::NxDomain) => {
+                    // the exact query contents don't matter to `NetError::is_nx_domain()`, only
+                    // the response code does, but a real `Query` is required to build one.
+                    let query = HickoryName::from_str(&host)
+                        .map(|name| Query::query(name, RecordType::A))
+                        .unwrap_or_default();
+                    let err = mock_dns_error(NetError::Dns(DnsError::NoRecordsFound(
+                        NoRecords::new(Box::new(query), ResponseCode::NXDomain),
+                    )));
+                    Box::pin(async move { Err(err) })
+                }
+                Some(MockOutcome::ServFail) => {
+                    let err = mock_dns_error(NetError::Dns(DnsError::ResponseCode(
+                        ResponseCode::ServFail,
+                    )));
+                    Box::pin(async move { Err(err) })
+                }
+                None => self.fallback.resolve(name),
+            }
+        }
     }
 }
--- a/common/http-api-client/src/fronted.rs
+++ b/common/http-api-client/src/fronted.rs
@@ -3,8 +3,9 @@
 
 //! Utilities for and implementation of request tunneling
 
+use std::collections::HashMap;
 use std::sync::{
-    Arc, LazyLock, RwLock,
+    Arc, LazyLock, Mutex, RwLock,
     atomic::{AtomicBool, Ordering},
 };
 use tracing::warn;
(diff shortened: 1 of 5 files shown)


common/http-api-client/src/fronted.rs
```diff
@@ -47,46 +53,185 @@ impl Front {
...
             enabled: AtomicBool::new(false),
             policy,
+            domain_failures: Mutex::new(HashMap::new()),
         }
     }
 
     pub(crate) fn set_policy(&self, policy: FrontPolicy) {
         *self.policy.write().unwrap() = policy;
         self.enabled.store(false, Ordering::Relaxed);
+        self.domain_failures.lock().unwrap().clear();
```
jstuczyn, line 62: given our http-api-client can be used in a lot of places at once that realistically could fail, I wonder if we should always unwrap on that poison error. perhaps we could attempt to get data out of `PoisonError` with `into_inner`? though then our mutex is still stuck forever... 🤔

common/http-api-client/src/url.rs
```diff
@@ -59,6 +59,13 @@ pub struct Url {
     url: url::Url,
     fronts: Option<Vec<url::Url>>,
     current_front: Arc<AtomicUsize>,
+
+    // Used only by the `include_non_fronted_in_rotation` rotation policy: each host gets a turn
+    // shown directly (slot 0) before cycling through its configured fronts (slot `i + 1` for
+    // `fronts[i]`) one at a time. `rotation_seen` tracks whether this host's direct turn has
+    // already happened during the current lap.
+    rotation_slot: Arc<AtomicUsize>,
```
jstuczyn, line 67: these field don't feel like they belong here. they're not really part of the `Url` itself. more like part of the client which deals with rotating the urls

common/http-api-client/src/fronted.rs
```diff
@@ -47,46 +53,185 @@ impl Front {
...
+/// - `failures_per_domain: 0` - a domain counts as "failed" on its very first failure.
+/// - `num_domains_failed: 0` - fronting is enabled as soon as `retry_enable` is called with a
+///   domain, even before any domain has reached `failures_per_domain`. This makes
+///   `failures_per_domain` irrelevant, so setting `num_domains_failed` to `0` alongside a
+///   non-zero `failures_per_domain` is almost certainly a configuration mistake - use
+///   [`FrontPolicy::OnRetry`] instead if "enable on the first failure" is actually what's wanted.
+/// - `FrontingConfig::default()` (both `0`) therefore enables fronting on the very first failure
+///   seen, same as `num_domains_failed: 0` above.
+#[derive(Debug, Default, PartialEq, Clone)]
+pub struct FrontingConfig {
```
jstuczyn, line 174: this is never exported so you can't build it outside the crate (the `frontend` mod is declared as private)

common/http-api-client/src/url.rs
```diff
@@ -228,8 +245,16 @@ impl Url {
 
     /// Return the serialization of this URL.
     ///
-    /// This is fast since that serialization is already stored in the inner url::Url struct.
+    /// This is fast since that serialization is already stored in the inner url::Url struct,
+    /// unless [`Self::take_rotation_turn`] has advanced this host onto one of its fronts, in
+    /// which case the active front's serialization is returned instead.
     pub fn as_str(&self) -> &str {
+        let slot = self.rotation_slot.load(Ordering::Relaxed);
```
jstuczyn, line 252: this can result in reporting the front domain after fronting has been turned off because `rotation_slot` is only reset within `update_host` **IF** `self.front.is_enabled()` 


```rust
 #[test]
  #[cfg(feature = "tunneling")]
  fn as_str_reports_front_after_fronting_disabled() {
      use crate::fronted::{FrontPolicy, FrontingConfig};

      let url = Url::new("https://a.test", Some(vec!["https://f0.test"])).unwrap();
      let mut client = ClientBuilder::new(url)
          .unwrap()
          .with_fronting(Some(FrontPolicy::ConfiguredRetry(
              FrontingConfig::new(1, 1).with_include_non_fronted_in_rotation(true),
          )))
          .build()
          .unwrap();

      client.front.retry_enable(Some("a.test"));
      assert!(client.front.is_enabled());

      // rotate until the host advances off its direct turn and onto a front (rotation_slot != 0)
      client.update_host(None);
      client.update_host(None);
      assert_eq!(client.current_url().as_str(), "https://f0.test/");

      // now turn fronting off - a plain public API call, no race required
      client.set_front_policy(FrontPolicy::Off);
      assert!(!client.front.is_enabled());

      // the request correctly goes out direct, unfronted...
      let mut req = reqwest::Request::new(reqwest::Method::GET, client.current_url().clone().into());
      let (domain, front_used) = client.apply_hosts_to_req(&mut req);
      assert_eq!(req.url().host_str(), Some("a.test"));
      assert_eq!(domain, Some("a.test"));
      assert_eq!(front_used, None);

      // ...but as_str() still claims we are talking to the CDN.
      assert_eq!(
          client.current_url().as_str(),
          "https://a.test/",
          "as_str() reports the front while requests go out direct"
      );
  }
  ```

File: common/http-api-client/src/fronted.rs [hunk 5]
```diff
@@ -47,46 +53,185 @@ impl Front {
...
+                }
+            }
+            FrontPolicy::Off | FrontPolicy::Always => {}
+        }
+    }
+
+    /// Whether domains without a front configured should still be considered when rotating
+    /// hosts while fronting is enabled. Only meaningful for [`FrontPolicy::ConfiguredRetry`] -
+    /// other policies keep the existing behavior of only rotating between fronted domains.
+    pub(crate) fn include_non_fronted_in_rotation(&self) -> bool {
```
jstuczyn, line 111: this changes behaviour of `front_str`: it introduces a second front cursor (`rotation_slot`) but never advances the original `current_front`, since `url.update()` is only called in the other branch of `update_host`. So `front_str()`, `Display`, `PartialEq` and `Hash` stay pinned to `fronts[0]` while the request goes out via whichever front `active_rotation_front_str()` picked. That makes the public `front_str()` accessor wrong, and the `debug!("http client rotating host {} -> {}")` log at `lib.rs:1069` prints the wrong front on every rotation.

```rust

  #[cfg(feature = "tunneling")]
  fn rotating_client(fronts: Vec<&str>) -> Client {
      use crate::fronted::{FrontPolicy, FrontingConfig};
      let url = Url::new("https://a.test", Some(fronts)).unwrap();
      let client = ClientBuilder::new(url)
          .unwrap()
          .with_fronting(Some(FrontPolicy::ConfiguredRetry(
              FrontingConfig::new(1, 1).with_include_non_fronted_in_rotation(true),
          )))
          .build()
          .unwrap();
      client.front.retry_enable(Some("a.test"));
      assert!(client.front.is_enabled());
      client
  }

  #[test]
  #[cfg(feature = "tunneling")]
  fn front_str_tracks_the_front_actually_used() {
      let client = rotating_client(vec!["https://f0.test", "https://f1.test"]);

      // advance onto the SECOND front
      for _ in 0..3 {
          client.update_host(None);
      }

      let url = client.current_url();
      let mut req = reqwest::Request::new(reqwest::Method::GET, url.clone().into());
      let (_domain, front_used) = client.apply_hosts_to_req(&mut req);

      assert_eq!(front_used, Some("f1.test"), "sanity: request goes via f1");
      assert_eq!(
          url.front_str(),
          front_used,
          "front_str() disagrees with the front on the wire; Display says {url}"
      );
  }
  ```

File: common/http-api-client/src/url.rs [hunk 6]
```diff
@@ -246,6 +271,49 @@ impl Url {
...
 
+    /// Used by [`crate::Client::update_host`] when a policy allows non-fronted domains into the
+    /// rotation alongside fronted ones (`include_non_fronted_in_rotation`). Each host gets one
+    /// turn shown directly before cycling through its configured fronts, one at a time.
+    ///
+    /// Returns `true` if this host just advanced onto one of its fronts (the caller should stay
+    /// on this host rather than rotating away), or `false` if this was the host's direct turn,
+    /// or if the direct turn and every front have each already had a turn - in the latter case
+    /// the cursor is reset so the next lap starts fresh with the direct turn again.
+    pub(crate) fn take_rotation_turn(&self) -> bool {
```
jstuczyn, line 282: The `rotation_seen` reset at the end of a lap (`url.rs:301-303`) returns `false`, so the caller rotates away, comes back to `slot 0` with `seen = false`, and the next `take_rotation_turn` burns a second consecutive turn on the direct host. The result is that the unfronted host gets twice the airtime of any front - backwards, given fronting was enabled precisely because direct access is failing.


  ```rust
  #[cfg(feature = "tunneling")]
  fn rotating_client(fronts: Vec<&str>) -> Client {
      use crate::fronted::{FrontPolicy, FrontingConfig};
      let url = Url::new("https://a.test", Some(fronts)).unwrap();
      let client = ClientBuilder::new(url)
          .unwrap()
          .with_fronting(Some(FrontPolicy::ConfiguredRetry(
              FrontingConfig::new(1, 1).with_include_non_fronted_in_rotation(true),
          )))
          .build()
          .unwrap();
      client.front.retry_enable(Some("a.test"));
      assert!(client.front.is_enabled());
      client
  }

  #[test]
  #[cfg(feature = "tunneling")]
  fn rotation_visits_each_slot_once_per_lap() {
      let client = rotating_client(vec!["https://f0.test", "https://f1.test"]);

      let mut seq = Vec::new();
      for _ in 0..6 {
          seq.push(client.current_url().as_str().to_string());
          client.update_host(None);
      }

      assert_eq!(
          seq,
          vec![
              "https://a.test/",
              "https://f0.test/",
              "https://f1.test/",
              "https://a.test/",
              "https://f0.test/",
              "https://f1.test/",
          ],
      );
  }
  ```

  ```
    left: ["https://a.test/", "https://a.test/", "https://f0.test/", "https://f1.test/", "https://a.test/", "https://a.test/"]
   right: ["https://a.test/", "https://f0.test/", "https://f1.test/", "https://a.test/", "https://f0.test/", "https://f1.test/"]
  ```

common/http-api-client/src/fronted.rs
```diff
@@ -47,46 +53,185 @@ impl Front {
...
+#[derive(Debug, Default, PartialEq, Clone)]
+pub struct FrontingConfig {
+    /// Allow N failures per domain before fronting enables. `0` means a single failure is
+    /// enough for a domain to count as "failed" - see the type-level docs for details.
+    failures_per_domain: usize,
+
+    /// Allow N domains to fail before fronting is enabled. `0` means fronting enables as soon as
+    /// any failure is recorded, bypassing `failures_per_domain` entirely - see the type-level
+    /// docs for details.
+    num_domains_failed: usize,
```
jstuczyn, line 182: If `num_domains_failed` exceeds the number of distinct base urls, the threshold is unreachable and fronting silently never engages. `with_fronting` already warns when none of the supplied urls have fronts (`fronted.rs:246`); this misconfiguration deserves the same warning, or clamping to the url count.

```rust
  #[test]
  #[cfg(feature = "tunneling")]
  fn unsatisfiable_num_domains_failed_is_reported() {
      use crate::fronted::{FrontPolicy, FrontingConfig};

      // one base url, but the policy demands two distinct domains fail
      let url = Url::new("https://a.test", Some(vec!["https://f0.test"])).unwrap();
      let client = ClientBuilder::new(url)
          .unwrap()
          .with_fronting(Some(FrontPolicy::ConfiguredRetry(FrontingConfig::new(1, 2))))
          .build()
          .unwrap();

      for _ in 0..100 {
          client.front.retry_enable(Some("a.test"));
      }

      assert!(
          client.front.is_enabled(),
          "fronting can never engage: only 1 base url but num_domains_failed = 2"
      );
  }
  ```

  ```
  fronting can never engage: only 1 base url but num_domains_failed = 2
  ```

File: common/http-api-client/src/lib.rs [hunk 8]
```diff
@@ -1095,7 +1139,14 @@ impl Client {
                 )
             }
         }
-        (url.as_str(), None)
+
+        // Ensure no stale fronting headers survive from a prior fronted attempt on this
+        // (possibly cloned/retried) request -- e.g. after a host rotation or a recovery to
+        // fronting-disabled.
+        r.headers_mut().remove(reqwest::header::HOST);
```
jstuczyn, line 1129: while this might not happen too frequently, this would remove **any** custom HOST header we might want to set regardless of fronting, i.e. 

  ```rust
  #[test]
  fn caller_supplied_host_header_is_preserved() {
      let url = Url::new("https://a.test", None).unwrap();
      let client = ClientBuilder::new(url).unwrap().build().unwrap();

      let mut req = client
          .create_request(reqwest::Method::GET, &["x"], NO_PARAMS, None::<&()>)
          .unwrap()
          .header(reqwest::header::HOST, "caller-supplied.test")
          .build()
          .unwrap();

      client.apply_hosts_to_req(&mut req);

      assert_eq!(
          req.headers().get(reqwest::header::HOST).map(|v| v.to_str().unwrap()),
          Some("caller-supplied.test"),
          "caller's Host header was silently dropped",
      );
  }
  ```

  ```
  assertion `left == right` failed: caller's Host header was silently dropped
    left: None
   right: Some("caller-supplied.test")
  ```

common/http-api-client/src/lib.rs
```diff
@@ -1273,15 +1349,31 @@ impl ApiClientCore for Client {
...
+
+fn is_rate_limit_response(status: StatusCode, headers: &HeaderMap) -> bool {
+    status == StatusCode::TOO_MANY_REQUESTS
+        || is_throttled_service_unavailable(status, headers)
+        || is_vercel_rate_limit_challenge(status, headers)
+}
+
+/// A 503 accompanied by `Retry-After` indicates the server is asking us to back off, as opposed
+/// to a plain 503 which may just mean the service is down.
+fn is_throttled_service_unavailable(status: StatusCode, headers: &HeaderMap) -> bool {
```
jstuczyn, line 1349: `Retry-After` is only presence-checked; the value is parsed nowhere in the crate and nothing ever sleeps. Combined with the new blanket "any 429 rotates hosts", a client whose base urls all front the same rate-limited origin will cycle through every host at full speed and then hand the 429 back anyway. Either honour the delay, or drop the header check and treat the status alone as the signal so the code does not imply a backoff it does not implement.

common/http-api-client/src/fronted.rs
pronebird: Poisoned lock means there was a panic somewhere, so I guess it can't get any worse than that?

pronebird (commented):
> @pronebird made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/7105)* status: 0 of 5 files reviewed, 9 unresolved discussions (waiting on jmwample, jstuczyn, and mmsinclair).

common/http-api-client/src/url.rs
pronebird: Exactly. Everything including already existing `current_front` should be moved out of `Url`. The type itself is better be renamed to something like `FrontedUrl` to avoid confusion with `url::Url`. Rotation should be moved to separate manager type shared across HTTP clients. `Url` should be plain descriptors.

pronebird (commented):
> @pronebird reviewed 2 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/7105)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on jmwample, jstuczyn, and mmsinclair).

pronebird (approved):
> @pronebird reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nymtech/nym/7105)* status: all files reviewed, 9 unresolved discussions (waiting on jmwample, jstuczyn, and mmsinclair).

common/http-api-client/src/lib.rs
jmwample: Added a change to allow the host header provided by a calling client to take precedence. However, I would like to be clear that this will break things if fronting is enabled. The HOST header is how the CDN identifies the service responsible for handling the request - and if it is overwritten by the caller it likely won't be handled properly without careful coordination. I added a warning log when this happens.

common/http-api-client/src/url.rs
jmwample: fixed. The host string returned now matches when fornting gets disabled.

common/http-api-client/src/url.rs
jmwample: This has been moved out to a rotation manager that the client itself tracks now.

common/http-api-client/src/fronted.rs
jmwample: For now the solution that I have added is just to clarify documentation and add a warning when  `num_domains_failed` exceeds the number of distinct base urls.

common/http-api-client/src/fronted.rs
jmwample: While I think it is technically possible for the lock to become poisoned I think something pretty awful would have to be happening. Locks can only become poisoned while they are held, so there would have to be a panic in `set_policy`, `recover`, or `retry_enable`. The first two jsut call `HashMap::clear()` which should never panic. `retry_enable` does valid safe `HashMap` (and atomic) operations which should never be able to panic before the lock is safely dropped. 

So if the that lock gets to a poisoned state, then we have significantly bigger problems elsewhere - and trying to recover the state of the mutex is probable not a primary concern.

PR #7123, opened 2026-09-02
Max/nym sdk feature gating

mfahampshire:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7123)

## Summary by CodeRabbit

- **New Features**
  - Added configurable SDK feature flags for filesystem storage, SOCKS5, IPR, bandwidth, streams, TCP proxying, client pooling, and CLI functionality.
  - SDK subsystems remain enabled by default, while selected features can support lighter builds.
  - Added clearer documentation for available features and configuration options.

- **Bug Fixes**
  - API clients now preserve fronted URL settings when connecting to Nym API endpoints.

- **Compatibility**
  - Storage and credential functionality requires its corresponding feature to be enabled.


common/bandwidth-fetcher/src/error.rs
```diff
@@ -6,7 +6,7 @@ use nym_credentials::error::Error as CredentialsError;
 use nym_validator_client::coconut::EcashApiError;
 use thiserror::Error;
 
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(all(not(target_arch = "wasm32"), feature = "credentials"))]
 use crate::storage::error::StorageError;
```
jstuczyn, line 10: rather than locking away credentials, it looks like you're locking away persistent storage?

common/bandwidth-fetcher/Cargo.toml
```diff
@@ -45,8 +55,14 @@ sqlx = { workspace = true, features = [
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 
 [features]
-# exposes `NyxdRecoveryFetcher`, a recover-only fetcher usable with a plain query client
-recovery = []
+# native credential issuance/import, backed by a sqlite pending-requests store.
+# exposes `NyxdCredentialFetcher` and pulls in sqlx. The storage-free
+# `NyxdGlobalDataFetcher` is always available and needs none of this.
+credentials = ["dep:sqlx", "dep:nym-sqlx-pool-guard"]
```
jstuczyn, line 61: is bandwidth-fetcher even usable without credentials feature? maybe we should conditionally import fetcher itself within the SDK instead?

sdk/rust/nym-sdk/src/api_client.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[allow(clippy::result_large_err)]
+pub fn create_nym_api_client(
+    nym_api_urls: Vec<ApiUrl>,
+) -> Result<nym_http_api_client::Client, Error> {
+    if nym_api_urls.is_empty() {
+        return Err(Error::NoNymAPIUrl);
+    }
+
+    let user_agent = format!("nym-sdk/{}", env!("CARGO_PKG_VERSION"));
```
jstuczyn: you can instead use `UserAgent::from(bin_info!())` for more consistent formatting

sdk/rust/nym-sdk/src/error.rs
```diff
@@ -60,12 +61,15 @@ pub enum Error {
     #[error("socks5 channel could not be started")]
     Socks5NotStarted,
 
+    #[cfg(feature = "bandwidth")]
```
jstuczyn: just thinking aloud here, why would you ever want to use nym-sdk WITHOUT bandwidth feature? it would be unusable, no?

sdk/rust/nym-sdk/src/lib.rs
```diff
@@ -80,22 +80,48 @@
...
+//! Every feature below is on by default, so `nym-sdk = "..."` reproduces the full
+//! SDK. Set `default-features = false` and enable only what you need to drop the
+//! dependencies you do not use.
+//!
+//! | Feature | Enables |
+//! |---------|---------|
+//! | `fs-storage` | On-disk persistent storage (keys, credentials, SURBs, gateways); pulls in the sqlx stack |
+//! | `socks5` | SOCKS5 client mode and network-requester country discovery |
+//! | `ipr` | IP packet router client and IP-over-mixnet stream (implies `stream`) |
+//! | `bandwidth` | Paid coconut bandwidth credentials |
```
jstuczyn: nit: ecash credentials. not coconut.

sdk/rust/nym-sdk/src/mixnet/client.rs
```diff
@@ -26,8 +33,10 @@ use nym_client_core::error::ClientCoreError;
 use nym_client_core::init::helpers::gateways_for_init;
 use nym_client_core::init::types::{GatewaySelectionSpecification, GatewaySetup};
 use nym_client_core::init::{refresh_gateway_published_data, setup_gateway};
+#[cfg(feature = "bandwidth")]
```
jstuczyn: can you put all feature locked imports below the rest of them to improve readability?  what I have in mind is something like this: https://github.com/nymtech/nym/blob/a9df556ac9fe7e7e9eac1df561a6b00458fdae68/common/client-libs/validator-client/src/nyxd/mod.rs#L80-L88

mfahampshire: That's nicer, yes I will do that

sdk/rust/nym-sdk/src/error.rs
mfahampshire: I'm pretty sure I ran it, but at this point I can't quite remember. Will double check :smiling_face_with_tear:

mfahampshire: My naming was fuzzy, renamed it to `credentials` which is a bit clearer (?) - it only gates the paid ticketbook acquire/import stuff, which is separate from the bandwidth controller. Without this you still connect and send fine, the Entry GW just grants free bandwidth to the client

common/bandwidth-fetcher/Cargo.toml
mfahampshire: There is already a conditional import in the sdk, but `client-core` uses it for `NyxdGlobalDataFetcher` (`common/client-core/src/client/base_client/mod.rs`), and that's unconditional

common/bandwidth-fetcher/src/error.rs
mfahampshire: I'm finding it tricky with the naming, since its the storage *for* the credentials it made sense to gate it with them?

mmsinclair (approved):

PR #7134, opened 2026-09-04
Lock-free topology

simonwicky:
Migrate the topology reading from a RwLock to an Arcswap.
Main changes in `accessor.rs` the rest is updating to the new signatures

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7134)

## Summary by CodeRabbit

* **Performance**
  * Improved topology access with faster, synchronous route-provider snapshots.
  * Reduced coordination overhead during topology refreshes and network traffic processing.

* **API Updates**
  * Topology inspection and manual topology replacement are now synchronous in Rust, SOCKS5, and WASM clients.
  * Current route providers are returned as independent snapshots rather than lock-bound values.

* **Reliability**
  * Improved consistency when topology updates occur during message processing and route selection.
  * Improved handling when no usable network topology is available.


common/client-core/src/client/real_messages_control/message_handler.rs
```diff
@@ -509,9 +508,8 @@ where
         // TODO: I really dislike existence of this assertion, it implies code has to be re-organised
         debug_assert!(!matches!(message, NymMessage::Reply(_)));
 
-        // TODO2: it's really annoying we have to get topology permit again here due to borrow-checker
-        let topology_permit = self.topology_access.get_read_permit().await;
-        let topology = self.get_topology(&topology_permit)?;
+        // TODO2: it's really annoying we have to grab the topology again here due to borrow-checker
+        let topology = self.get_topology()?;
```
jstuczyn: is it still the case? iirc that was originally due to the lock guard

common/client-core/src/client/topology_control/accessor.rs
```diff
@@ -1,67 +1,76 @@
...
-            topology: RwLock::new(initial),
+            topology: ArcSwap::from_pointee(initial),
         }
     }
 
-    async fn update(&self, new: Option<NymTopology>) {
-        let mut guard = self.topology.write().await;
+    fn update(&self, new: Option<NymTopology>) {
+        // Only the topology is updated, the flag stays
+        let ignore_egress_epoch_roles = self.topology.load().ignore_egress_epoch_roles;
```
jstuczyn: that feels flaky. maybe the thing behind ArcSwap should be the inner topology instead that would get accessed with the `update()` method like we do here? because attempting to reapply all the configs, like `ignore_egress_epoch_roles` sounds like a recipe for disaster

common/client-core/src/client/topology_control/accessor.rs
```diff
@@ -80,79 +89,39 @@ impl<'a> TopologyReadPermit<'a> {
...
-        self.current_route_provider()
-            .await
-            .as_ref()
-            .map(|p| p.topology.clone())
-    }
-
-    pub async fn current_route_provider(&self) -> Option<RwLockReadGuard<'_, NymRouteProvider>> {
-        let provider = self.inner.topology.read().await;
+    pub fn current_route_provider(&self) -> Option<Arc<NymRouteProvider>> {
+        let provider = self.inner.topology.load_full();
```
jstuczyn: I think in most of these cases a normal cheaper `load()` would be preferred over the `load_full`. like if you just want to get metadata out of it

common/client-core/src/client/topology_control/accessor.rs
```diff
@@ -14,31 +14,33 @@ pub struct TopologyAccessorInner {
...
 
     fn update(&self, new: Option<NymTopology>) {
-        // Only the topology is updated, the flag stays
-        let ignore_egress_epoch_roles = self.topology.load().ignore_egress_epoch_roles;
-
-        let updated = match new {
-            Some(topology) => NymRouteProvider::new(topology, ignore_egress_epoch_roles),
-            None => NymRouteProvider::new_empty(ignore_egress_epoch_roles),
-        };
+        self.topology.store(Arc::new(new.unwrap_or_default()))
```
jstuczyn, line 38: doesn't this change the semantics? there's a difference between empty topology (Some(NymTopology::default()) and no topology (None)

simonwicky: Previous code was already doing that. 
Internally, it isn't an option, so updating it with `None` was effectively storing an empty one, same as `update(Some(NymTopology::default()))`

The accessor is returning an `Option`, which is `None` when the topology was empty. 

Happy to change the paradigm if needed, but that update doesn't change the code behavior

jstuczyn (approved):

PR #7145, opened 2026-09-09
bug fixes

benedettadavico:
- Cred proxy quorum gate was epoch-blind, it now remembers what epoch checks were made in.
- Transient failures came back as 500s, they are now 503.
- Cred proxy could not start right after a ceremony, now the warm-up is best-effort with a warn

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7145)

### DIFF
--- a/clients/native/src/client/config/template.rs
+++ b/clients/native/src/client/config/template.rs
@@ -57,6 +57,10 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
+# Path to the database of pending credential requests, i.e. deposits made for
+# ticketbooks whose issuance hasn't completed yet.
+credential_requests_database = '{{ storage_paths.credential_requests_database }}'
+
 # Path to the persistent store for received reply surbs, unused encryption keys and used sender tags.
 reply_surb_database = '{{ storage_paths.reply_surb_database }}'
 
--- a/clients/native/src/client/config/template.rs
+++ b/clients/native/src/client/config/template.rs
@@ -57,10 +57,6 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
-# Path to the database of pending credential requests, i.e. deposits made for
-# ticketbooks whose issuance hasn't completed yet.
-credential_requests_database = '{{ storage_paths.credential_requests_database }}'
-
 # Path to the persistent store for received reply surbs, unused encryption keys and used sender tags.
 reply_surb_database = '{{ storage_paths.reply_surb_database }}'
 
--- a/clients/socks5/src/config/template.rs
+++ b/clients/socks5/src/config/template.rs
@@ -57,6 +57,10 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
+# Path to the database of pending credential requests, i.e. deposits made for
+# ticketbooks whose issuance hasn't completed yet.
+credential_requests_database = '{{ storage_paths.credential_requests_database }}'
+
 # Path to the persistent store for received reply surbs, unused encryption keys and used sender tags.
 reply_surb_database = '{{ storage_paths.reply_surb_database }}'
 
--- a/clients/socks5/src/config/template.rs
+++ b/clients/socks5/src/config/template.rs
@@ -57,10 +57,6 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
-# Path to the database of pending credential requests, i.e. deposits made for
-# ticketbooks whose issuance hasn't completed yet.
-credential_requests_database = '{{ storage_paths.credential_requests_database }}'
-
 # Path to the persistent store for received reply surbs, unused encryption keys and used sender tags.
 reply_surb_database = '{{ storage_paths.reply_surb_database }}'
 
--- a/common/credential-proxy/src/error.rs
+++ b/common/credential-proxy/src/error.rs
@@ -223,12 +223,8 @@ impl CredentialProxyError {
     /// Conditions that resolve on their own, and which a caller should therefore retry, are
     /// `503`; everything else is reported as a server fault. The distinction matters to clients:
     /// a mid-ceremony refusal is over within minutes, and a client that reads it as a `500` gives
-    /// up on something that was about to succeed.
-    ///
-    /// These two are the cases the ticketbook routes already document as `503`. Others are
-    /// arguably transient too - an unavailable signing quorum, most obviously - but they are not
-    /// what those routes advertise, so they are deliberately left alone rather than reclassified
-    /// on the way past.
+    /// up on something that was about to succeed. The same applies to a temporarily unavailable
+    /// signing quorum and shortfalls collecting shares or responses.
     pub fn status_code(&self) -> StatusCode {
         match self {
             CredentialProxyError::CredentialsNotYetIssuable { .. }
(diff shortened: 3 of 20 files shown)


clients/native/src/client/config/template.rs
```diff
@@ -57,6 +57,10 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
+# Path to the database of pending credential requests, i.e. deposits made for
```
simonwicky: This is not wrong, but it is very much a lost bullet

clients/socks5/src/config/template.rs
```diff
@@ -57,6 +57,10 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
 # Path to the database containing bandwidth credentials
 credentials_database = '{{ storage_paths.credentials_database }}'
 
+# Path to the database of pending credential requests, i.e. deposits made for
```
simonwicky: Same lost bullet comment

clients/native/src/client/config/template.rs
```diff
@@ -57,6 +57,10 @@ keys.ack_key_file = '{{ storage_paths.keys.ack_key_file }}'
...
+# ticketbooks whose issuance hasn't completed yet.
+credential_requests_database = '{{ storage_paths.credential_requests_database }}'
```
simonwicky: It looks like we do have a config issue indeed, but I reckon it's out of scope of this PR. 
(Same thing for the three other config files changes)

common/credential-proxy/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2025 Nym Technologies SA <contact@nymtech.net>
 // SPDX-License-Identifier: GPL-3.0-only
 
+use std::time::Duration;
+
+pub const SIGNER_REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
```
simonwicky: That is definitely misplaced

common/credential-proxy/src/ticketbook_manager/mod.rs
```diff
@@ -87,10 +88,12 @@ impl TicketbookManager {
...
         };
 
-        // since this is startup,
-        // might as well do all the needed network queries to establish needed global signatures
-        // if we don't already have them
-        this.build_initial_cache().await?;
+        // a warm cache is an optimisation, not a precondition - every entry fills lazily on first
+        // use. signers keep serving the previous epoch for a few minutes after a ceremony
+        // concludes, and a proxy restarting in that window still has to come up.
+        if let Err(err) = this.build_initial_cache().await {
```
simonwicky, line 94: Given the description, it looks like one of the bug fixes of the PR. I cannot comment on that behavior change but it does look reasonable

common/credential-proxy/src/error.rs
```diff
@@ -229,7 +232,12 @@ impl CredentialProxyError {
     pub fn status_code(&self) -> StatusCode {
         match self {
             CredentialProxyError::CredentialsNotYetIssuable { .. }
-            | CredentialProxyError::CallerCannotIssueMidCeremony => StatusCode::SERVICE_UNAVAILABLE,
+            | CredentialProxyError::CallerCannotIssueMidCeremony
```
simonwicky, line 231: I'm not familiar with how the cred proxy works. @jstuczyn can you double check that new classification is correct

common/credential-proxy/src/quorum_checker.rs
```diff
@@ -4,30 +4,43 @@
...
 impl QuorumState {
-    pub fn available(&self) -> bool {
-        self.available.load(Ordering::Acquire)
+    pub fn rules_out(&self, epoch_id: EpochId) -> bool {
+        !self.available.load(Ordering::Acquire)
+            && self.checked_epoch.load(Ordering::Acquire) == epoch_id
+    }
+
+    fn record(&self, epoch_id: EpochId, available: bool) {
+        self.available.store(available, Ordering::Release);
```
simonwicky: I wouldn't mix Ordering with Atomics. 
Also, since you introduced a second one, they aren't atomics across them anymore. Overlapping `rules_out` and `record` calls  can read an inconsistent state, where `available` was updated, but not `checked_epoch`

common/credential-proxy/src/error.rs
jstuczyn: don't change this behaviour. currently VPN-API upon seeing a 503 will assign upstream unavailable due to DKG transition error message which will cause the client APP to change its polling behaviour. we do not really want it for other error variants (at least not like this)

common/credential-proxy/src/shared_state/ecash_state.rs
```diff
@@ -82,7 +82,7 @@ fn construct_ecash_api_client(share: ContractVKShare) -> Result<EcashApiClient,
 
     let api_client = nym_http_api_client::Client::builder(url_address)
         .map_err(|e| EcashApiError::ClientError(e.to_string()))?
-        .with_timeout(Duration::from_secs(5))
+        .with_timeout(Duration::from_secs(15))
```
jstuczyn: note: this timeout doesn't actually do anything if it's not wasm 🫠 
but did we find timeout to be insufficient? what made you make this change?

common/credential-proxy/src/ticketbook_manager/mod.rs
jstuczyn: sure a warm cache is an optimisation, but failing to build it means credential proxy can't do its job anyway (because it failed to get data for the latest issuable epoch), so what do we want to achieve with this change now?

common/credential-proxy/src/quorum_checker.rs
```diff
@@ -4,30 +4,60 @@
...
+#[derive(Copy, Clone)]
+struct QuorumSnapshot {
+    epoch_id: EpochId,
+    available: bool,
+}
+
 #[derive(Clone)]
 pub struct QuorumState {
-    available: Arc<AtomicBool>,
+    inner: Arc<RwLock<QuorumSnapshot>>,
```
jstuczyn, line 22: I reckon we should probably use tokio's async RwLock for consistency and to prevent possibly holding it across await points (i.e. risking deadlocks)

jstuczyn: perhaps it could even be a map between `EpochId` and `QuorumSnapshot` in case clients ask for different epochs during/after transition? though it might be an overkill?

common/credential-proxy/src/quorum_checker.rs
```diff
@@ -4,30 +4,60 @@
...
 }
 
 impl QuorumState {
-    pub fn available(&self) -> bool {
-        self.available.load(Ordering::Acquire)
+    pub fn rules_out(&self, epoch_id: EpochId) -> bool {
```
jstuczyn, line 26: since we only hold information about a single epoch (going to my previous comment about having a map), if client asks for epoch information for an older epoch just after transition, this will fail, no?

nym-api/src/support/nyxd/mod.rs
```diff
@@ -70,48 +70,32 @@ use serde::Deserialize;
...
+        };
+        let guard = lock.write().await;
+        guard.$($op)*
     }};
 }
 
-#[derive(Clone)]
 pub(crate) struct Client {
-    inner: Arc<RwLock<ClientInner>>,
+    query: QueryHttpRpcNyxdClient,
```
jstuczyn, line 88: that's a nice optimisation, but were the queries really blocked for so long waiting for single transactions?

common/credential-proxy/src/shared_state/ecash_state.rs
benedettadavico: a typo :)

common/credential-proxy/src/quorum_checker.rs
benedettadavico: no, that's what used to happen though.. it was the bug. It no longer happens now

jstuczyn (approved):

common/credential-proxy/src/quorum_checker.rs
benedettadavico: we only save one epoch so a map would be overkill imo

PR #7146, opened 2026-09-09
nym-api fixes

benedettadavico:
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7146)

### DIFF
--- a/nym-api/src/support/nyxd/mod.rs
+++ b/nym-api/src/support/nyxd/mod.rs
@@ -74,56 +74,28 @@ use tokio::sync::RwLock;
 use tracing::warn;
 use url::Url;
 
-#[macro_export]
-macro_rules! query_guard {
-  ($guard:expr, $($op:tt)*) => {{
-        match &*$guard {
-            $crate::support::nyxd::ClientInner::Signing(client) => client.$($op)*,
-            $crate::support::nyxd::ClientInner::Query => {
-                panic!("query_guard on a query-only client; use Client::query_client() instead")
-            }
-        }
-    }};
-}
-
-// Queries never take the signing lock. Holding that lock across broadcast + inclusion polling
-// is what made blind-sign time out for minutes during DKG dealing exchange (and during any
-// other concentrated chain writes). Signing still serialises through the RwLock so account
-// sequence numbers stay ordered.
-macro_rules! nyxd_query {
-    ($self:expr, $($op:tt)*) => {{
-        $self.query.$($op)*
-    }};
-}
-
 macro_rules! nyxd_signing {
     ($self:expr, $($op:tt)*) => {{
-        let guard = $self.inner.write().await;
-        match &*guard {
-            $crate::support::nyxd::ClientInner::Signing(client) => client.$($op)*,
-            $crate::support::nyxd::ClientInner::Query => panic!("attempted to use a signing method on a query client"),
-        }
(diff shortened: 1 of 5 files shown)


nym-api/src/support/nyxd/mod.rs
```diff
@@ -79,18 +79,20 @@ macro_rules! query_guard {
   ($guard:expr, $($op:tt)*) => {{
         match &*$guard {
             $crate::support::nyxd::ClientInner::Signing(client) => client.$($op)*,
-            $crate::support::nyxd::ClientInner::Query(client) => client.$($op)*,
+            $crate::support::nyxd::ClientInner::Query => {
```
simonwicky: That macro is now unused and can panic, that's a landmine, you can just delete it

nym-api/src/support/nyxd/mod.rs
```diff
@@ -79,18 +79,20 @@ macro_rules! query_guard {
...
+// other concentrated chain writes). Signing still serialises through the RwLock so account
+// sequence numbers stay ordered.
 macro_rules! nyxd_query {
     ($self:expr, $($op:tt)*) => {{
-        let guard = $self.inner.read().await;
-        match &*guard {
-            $crate::support::nyxd::ClientInner::Signing(client) => client.$($op)*,
-            $crate::support::nyxd::ClientInner::Query(client) => client.$($op)*,
-        }
+        $self.query.$($op)*
```
simonwicky: That now looks like a very straightforward that could just be inlined everywhere no?

nym-api/src/support/nyxd/mod.rs
```diff
@@ -99,19 +101,29 @@ macro_rules! nyxd_signing {
...
             $crate::support::nyxd::ClientInner::Signing(client) => client.$($op)*,
-            $crate::support::nyxd::ClientInner::Query(_) => panic!("attempted to use a signing method on a query client"),
+            $crate::support::nyxd::ClientInner::Query => panic!("attempted to use a signing method on a query client"),
         }
     }};
 }
 
-#[derive(Clone)]
 pub(crate) struct Client {
+    query: QueryHttpRpcNyxdClient,
```
simonwicky, line 88: I get the intention, the resulting form is a bit weird. 
We either just have a query client, or a query client + a signing one? 
That makes sense to separate reads and writes, but then we should use an option no? 

Something like 
```
pub(crate) struct Client {
    query: QueryHttpRpcNyxdClient,

    signing: Arc<Option<RwLock<DirectSigningHttpRpcNyxdClient>>>,
}
```

nym-api/src/support/nyxd/mod.rs
```diff
@@ -125,33 +137,29 @@ impl Client {
...
     }
 
-    pub(crate) async fn query_client(&self) -> QueryHttpRpcNyxdClient {
-        nyxd_query!(self, clone_query_client())
-    }
-
-    pub(crate) async fn read(&self) -> RwLockReadGuard<'_, ClientInner> {
-        self.inner.read().await
+    pub(crate) fn query_client(&self) -> QueryHttpRpcNyxdClient {
+        self.query.clone_query_client()
```
simonwicky, line 131: Yes, like this, since we always have one, maybe all other `nyxd_query` call can be inlined like that

PR #7150, opened 2026-09-10
Apply clock skew to LP registration timestamps

tommyv1987:
Bandwidth claims already subtract cached VPN API skew. LP request stamps still used raw local now, so a device more than 30s fast was rejected as invalid timestamp. Same subtract-ahead convention - nodes and the wire format are unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7150)

## Summary by CodeRabbit

- **Bug Fixes**
  - VPN registration now accounts for clock differences between the device and the VPN API when validating registration timestamps.
  - Devices with clocks more than 30 seconds ahead are no longer incorrectly rejected as having invalid timestamps.
  - Clock-skew adjustments are applied consistently during both initial registration and registration finalization.

### DIFF
--- a/common/registration/src/lp_messages.rs
+++ b/common/registration/src/lp_messages.rs
@@ -113,26 +113,18 @@ impl RegistrationStatus {
     }
 }
 
-fn current_timestamp() -> u64 {
+fn current_timestamp() -> std::time::Duration {
     std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
         .inspect_err(|_| error!("the current timestamp predates unix epoch!"))
         .unwrap_or_default()
-        .as_secs()
 }
 
 impl LpRegistrationRequest {
-    /// `spend_time_skew_secs` is local-ahead-of-api-call seconds; gateways allow 30s.
-    fn new(
-        registration_data: LpRegistrationRequestData,
-        spend_time_skew_secs: i64,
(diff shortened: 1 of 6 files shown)


common/registration/src/lp_messages.rs
```diff
@@ -122,36 +122,49 @@ fn current_timestamp() -> u64 {
...
 
 impl LpRegistrationRequest {
-    /// Helper wrapping timestamp extraction
-    fn new(registration_data: LpRegistrationRequestData) -> LpRegistrationRequest {
+    /// `spend_time_skew_secs` is local-ahead-of-api-call seconds; gateways allow 30s.
+    fn new(
+        registration_data: LpRegistrationRequestData,
+        spend_time_skew_secs: i64,
+    ) -> LpRegistrationRequest {
+        let timestamp = (current_timestamp() as i64)
```
simonwicky: That is some AI type gymnastics, there's no need to cast a timestamp into a i64. 
Worst case scenario, make `current_timestamp` return a duration and handle durations until the end. 
Also maybe leave room for a timestamp free version of the below fn? 
It's gonna be annoying to set that 0 is other places that can use this

tommyv1987:
@coderabbitai review

simonwicky (approved):

PR #7158, opened 2026-09-16
feat(coconut-dkg): make a ceremony's length controllable

jstuczyn:
## Why

Mainnet's stored `time_configuration` is 5 days for registration and 6 hours for each of the four following phases, and two of those phases cannot end early: registration, because the contract had no way to know how many dealers to expect, and validation, because the voting happens in the multisig. Every other phase already short-circuits once everyone is in. So a reset ceremony on mainnet takes at least 5 days 6 hours and realistically 6 days, during which every client that cannot state an epoch is refused issuance, a ceremony that ends sub-threshold costs another 5 days to retry, and the outgoing signers have to stay reachable for the ceremony plus the 7-day ticketbook validity.

None of that was tunable: `TimeConfiguration` is instantiate-only and `next_ceremony` carries it into every later ceremony. The defaults, meanwhile, dated from a ceremony that was expected to be waited out, and at 60 seconds the validation default cannot even hold 17 dealers' votes.

## What changes

- **Registration ends once every voting member of the cw4 group has registered** (`check_state_completion`, `epoch_state/utils.rs`). Only a voting member can register, so the group is the eligibility set and "everyone eligible is in" is exactly the right condition. The mistake is only allowed in the safe direction: a member that never turns up leaves the phase to its deadline, a group the contract cannot read counts as "not complete" rather than failing the advance, and `>=` rather than `==` means a member removed after registering cannot wedge the phase. Members are paged the way cw4-group serves them. `CanAdvanceState` gives the same answer nym-api's tick acts on.
- **`ExecuteMsg::ForceAdvanceEpochState {}`**, admin-only, for the two phases the contract cannot see the end of. It skips only the completion-or-deadline gate and then runs the ordinary transition (the threshold is still fixed on entering `DealingExchange`, a sub-threshold finish still resets). It is refused where the contract can already tell the ceremony would end sub-threshold: nobody registered out of registration (`NoDealersToAdvance`), fewer than threshold-many shares out of submission or verified out of finalization (`ForcedAdvanceBelowThreshold`), and in `WaitingInitialisation` and `InProgress`. The transaction carries a `forced_advance` attribute naming the phase cut short. Forcing out of validation with proposals still open loses the votes of any api that has not cast them yet, so the operator checks the multisig first; the contract cannot.
- **Defaults sized against the work**: 3600 / 3600 / 600 / 1800 / 600 s, roughly one block per transaction per api with 5-10x headroom. The long ones are fallbacks for a participant going quiet. Nothing deployed changes; defaults are only read at instantiation. The cw-multi-test harness now instantiates with explicit short timings because it walks phases with a 601 s jump.
- **`MigrateMsg { time_configuration: Option<TimeConfiguration> }`**. Absent or `None` writes nothing; a payload replaces the stored timings and records the previous and new values as attributes, in the comma-separated form `DKG_TIME_CONFIGURATION` reads. Deadlines are computed per transition, so a retiming mid-ceremony leaves the running phase's deadline alone and applies from the next transition. A retune is a second migrate to the same code id. Both instantiate and migrate now validate timings, where instantiate used to accept anything: no zero-length phase, no phase over 30 days (past which `Timestamp::plus_seconds` overflows at the next transition and takes every advance and the forced reset with it), and submission + validation + finalization under the 24 h verification-proposal lifetime, or a share committed at the start of submission could not be executed at the end of finalization.
- **The deprecated `in_progress_time_secs` is `#[serde(default)]`**, so a payload carries only the five real durations.

The `seed_keys_in_service` migration step is removed: the previous build's migration will already have recorded the epoch in service by the time this one runs. If that deploy order changes, the fourth commit is where the assumption lives.

## Deployment and operations

- **Mainnet payload values are a decision for the deploy**, not fixed here. The defaults are a reasonable starting point; anything with validation at least ~10x the vote count is fine. A payload needs only the five durations:

```json
{ "time_configuration": { "public_key_submission_time_secs": 3600, "dealing_exchange_time_secs": 3600, "verification_key_submission_time_secs": 600, "verification_key_validation_time_secs": 1800, "verification_key_finalization_time_secs": 600 } }
```

- **Prune the cw4 group to the participant set before triggering**, otherwise the registration short-circuit never fires and the configured timer stands. Pruning also stops absent members' weight counting against the validation proposals.
- Confirm the multisig's `max_voting_period` covers submission + validation + finalization at the chosen values.
- Rehearse in sandbox at the exact chosen values, with the group pruned, and once through the forced path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7158)


contracts/coconut-dkg/src/epoch_state/transactions/advance_epoch_state.rs
```diff
@@ -55,6 +56,85 @@ pub fn try_advance_epoch_state(deps: DepsMut<'_>, env: Env) -> Result<Response,
...
+pub(crate) fn try_force_advance_epoch_state(
+    deps: DepsMut<'_>,
+    env: Env,
+    info: MessageInfo,
+) -> Result<Response, ContractError> {
+    DKG_ADMIN.assert_admin(deps.as_ref(), &info.sender)?;
+    let current_epoch = load_current_epoch(deps.storage)?;
+
+    // forcing moves a ceremony along; it neither starts one nor rotates keys
+    if current_epoch.state == EpochState::WaitingInitialisation {
```
simonwicky, line 78: nit : why are those `if`s not grouped with the match on the same value below?

simonwicky (approved):

PR #7165, opened 2026-09-17
HTTP Rate-limit Rotation

jmwample:
Rotate on rate limiting, don't force fronting

HTTP errors indicating rate limiting are now treated as a trigger to rotate to a different host, but no longer automatically enable domain fronting — rotation and fronting are decoupled so a rate-limited (but otherwise healthy) endpoint doesn't unnecessarily flip on fronting.

Also fixes an inconsistent test that relied on network resources -- now determinitic using a mocked item.

This separates a concern out of #7105 in order to get independent changes merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7165)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTTP rate-limit detection for 429 responses, qualifying 503 responses with `Retry-After`, and recognized challenge responses.
  - Plain 503 responses are no longer incorrectly classified as rate limits.
  - Improved host rotation and fallback when DNS resolution encounters failed or unavailable hosts.
  - Updated rate-limit warnings and errors with clearer, provider-neutral wording.

- **Tests**
  - Added coverage for rate-limit detection and deterministic fallback behavior across failed hosts.


jmwample:
in #7105 @jstuczyn said

> Retry-After is only presence-checked; the value is parsed nowhere in the crate and nothing ever sleeps. Combined with the new blanket "any 429 rotates hosts", a client whose base urls all front the same rate-limited origin will cycle through every host at full speed and then hand the 429 back anyway. Either honour the delay, or drop the header check and treat the status alone as the signal so the code does not imply a backoff it does not implement.

I have resolved this by removing the check for Retry-After as handling delay / backoff would add unrelated complexity.

common/http-api-client/src/lib.rs
```diff
@@ -1270,18 +1270,33 @@ impl ApiClientCore for Client {
...
+/// rate limit challenge response from the vercel API.
+pub(crate) fn is_http_rate_limit_err(resp: &Response) -> bool {
+    is_rate_limit_response(resp.status(), resp.headers())
+}
+
+fn is_rate_limit_response(status: StatusCode, headers: &HeaderMap) -> bool {
+    // A 503 is sometimes sent as a way to indicate that the service is rate limiting requests from
+    // the client. While this usually comes with a `RETRY_AFTER` header we are not explicitly
+    // handling the retry delay - so the status is just presence checked here.
+    let service_unavailable = status == StatusCode::SERVICE_UNAVAILABLE;
```
jstuczyn: but should a 503 result in rotating host?

jstuczyn (approved):

PR #7172, opened 2026-09-18
chore: update deps

jstuczyn:
blocked waiting for `nym-bls12_381-fork` to get published

### RustCrypto cipher/AEAD generation
| crate | from | to |
|---|---|---|
| `aead` | 0.5.2 | 0.6.1 |
| `aes` | 0.8.1 | 0.9.3 |
| `aes-gcm` | 0.10.1 | 0.11.1 |
| `aes-gcm-siv` | 0.11.1 | 0.12.1 |
| `argon2` | 0.5.0 | 0.6.0 |
| `chacha20` | 0.9.0 | 0.10.2 |
| `chacha20poly1305` | 0.10.1 | 0.11.0 |
| `cipher` | 0.4.3 | 0.5.2 |
| `ctr` | 0.9.1 | 0.10.1 |
| `generic-array` | 0.14.7 | **removed**, replaced by `hybrid-array` 0.4.15 |

### Curve / signature stack
| crate | from | to |
|---|---|---|
| `curve25519-dalek` | 4.1.3 | 5.0.0 |
| `ed25519-dalek` | 2.1 | 3.0.0 |
| `x25519-dalek` | 2.0.0 | 3.0.0 |
| `ecdsa` | 0.16 | 0.17.0 |
| `ff` | 0.13.1 | 0.14.0 |
| `group` | 0.13.0 | 0.14.0 |
| `subtle` | 2.5.0 | 2.6.1 |
| `sphinx-packet` | =0.6.0 | =0.7.0 |
| `nym-bls12_381-fork` | 0.8.0-forked | 0.9.0-forked |

### rand family
| crate | from | to |
|---|---|---|
| `rand` | 0.8.6 | 0.10.2 |
| `rand_chacha` | 0.3 | =0.10.0 |
| `rand_distr` | 0.4 | 0.6 |
| `rand_core` | 0.6.3 | **removed** (unused) |
| `getrandom` | 0.2.10 | 0.4 |

### WebSocket / networking
| crate | from | to |
|---|---|---|
| `tungstenite` | 0.20.1 | 0.30.0 |
| `tokio-tungstenite` | 0.20.1 | 0.30.0 |
| `async-tungstenite` | 0.24 | 0.35.0 |
| `smoltcp` | 0.12 | 0.13.1 |
| `tokio-smoltcp` | 0.5 | 0.6.0 |
| `tokio-tun` | 0.11.5 | 0.15.2 |
| `simple-dns` | 0.7 | 0.12.0 |

`smoltcp` is pinned to 0.13.1 rather than 0.14 because it must match the version `tokio-smoltcp` 0.6 is built against, otherwise `smol-core` sees two incompatible copies of the wire types.

### Other
| crate | from | to |
|---|---|---|
| `bip39` | 2.0.0 | 3.0.0 |

### Deliberately not bumped
`digest`, `sha2`, `sha3`, `hmac`, `hkdf` and `blake3` stay on the digest 0.10 generation, because `nym-bls12_381-fork` is built against digest 0.10 and its `ExpandMsgXmd` has to accept our `sha2::Sha256`. digest 0.11 is still present in the graph via `curve25519-dalek` 5, `ed25519-dalek` 3 and `sphinx-packet` 0.7, so both generations coexist by construction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/nymtech/nym/7172)


common/wasm/client-core/src/helpers.rs
```diff
@@ -23,7 +23,7 @@ use nym_topology::{EpochRewardedSet, NymTopology, RoutingNode};
 use nym_validator_client::client::IdentityKey;
 use nym_validator_client::{nym_api::NymApiClientExt, UserAgent};
 use nym_wasm_utils::error::PromisableResult;
-use rand::thread_rng;
+use rand::rng as thread_rng;
```
simonwicky: A bit inconsistent here, I don't think that alias is really helpful

common/gateway-requests/src/registration/handshake/state.rs
```diff
@@ -338,7 +338,7 @@ impl<'a, S, R> State<'a, S, R> {
     {
         let handshake_message = types::RegistrationHandshake::new_error(message);
         self.ws_stream
-            .send(WsMessage::Text(handshake_message.into()))
+            .send(WsMessage::Text(String::from(handshake_message).into()))
```
simonwicky: Isn't that a double `into` ? (Same below)

common/crypto/src/hkdf.rs
```diff
@@ -1,18 +1,12 @@
...
-        digest::{Digest, crypto_common::BlockSizeUser},
-    },
-};
+use hkdf::hmac::digest::{Digest, crypto_common::BlockSizeUser};
 use sha2::{Sha256, Sha512};
-
-pub use hkdf::InvalidLength;
 use zeroize::ZeroizeOnDrop;
 
+pub use hkdf::{InvalidLength, *};
```
simonwicky, line 8: Either `InvalidLenght` isn't needed, or the `*` isn't needed. 
Do we need to re-export the whole crate?

common/crypto/src/rng.rs
```diff
@@ -0,0 +1,15 @@
...
+pub use rand::rand_core::UnwrapErr;
+pub use rand::rngs::SysRng;
+
+/// OS entropy as an infallible `CryptoRng`, panicking on entropy failure just as rand 0.8's
+/// `OsRng` did. Unlike `rand::rng()` it is `Send + Sync + Copy`, so this is the one to reach
+/// for when the rng lives in a struct or across an `.await`; prefer `rand::rng()` otherwise.
+pub type OsRng = UnwrapErr<SysRng>;
+
+/// Constructs an [`OsRng`].
+pub fn os_rng() -> OsRng {
```
simonwicky, line 13: Isn't that a bit ugly? 
I get the idea, I don't like the style of it though

common/test-utils/src/helpers.rs
```diff
@@ -69,22 +63,10 @@ pub fn deterministic_rng_09() -> DeterministicRng010 {
...
 
-pub fn deterministic_rng() -> DeterministicRng {
-    seeded_rng([42u8; 32])
-}
-
-pub fn seeded_rng(seed: [u8; 32]) -> DeterministicRng {
-    DeterministicRng::from_seed(seed)
-}
-
 pub fn seeded_rng_09(seed: [u8; 32]) -> DeterministicRng010 {
```
simonwicky: This is gonna get confusing real quick. I reckon we should drop the 010 prefix, and move it to the older version

common/test-utils/src/helpers.rs
```diff
@@ -9,21 +9,15 @@ use nym_bin_common::logging::tracing_subscriber::EnvFilter;
...
 use tokio::time::error::Elapsed;
 
-// 'current' rand crate
-pub use rand_chacha::ChaCha20Rng as DeterministicRng;
-pub use rand_chacha::rand_core::{CryptoRng, RngCore};
-
-// rand010 compat
-pub use rand_chacha010::ChaChaRng as DeterministicRng010;
-pub use rand_chacha010::rand_core::{CryptoRng as CryptoRng010, Rng as Rng010};
+pub use rand_chacha::ChaChaRng as DeterministicRng010;
```
simonwicky: Same comment regarding the version prefix no?

common/inclusion-probability/src/lib.rs
```diff
@@ -164,8 +164,9 @@ fn sample_candidate<R>(list_cumul: &[u128], rng: &mut R) -> Result<usize, Error>
 where
     R: Rng + ?Sized,
 {
-    use rand::distributions::{Distribution, Uniform};
-    let uniform = Uniform::from(0..*list_cumul.last().ok_or(Error::EmptyListCumulStake)?);
+    use rand::distr::{Distribution, Uniform};
+    let uniform = Uniform::try_from(0..*list_cumul.last().ok_or(Error::EmptyListCumulStake)?)
+        .expect("cumulative stake range is empty");
```
simonwicky: I reckon we could take the chance and get rid of that panic since we are in a fallible function already

common/crypto/src/hkdf.rs
jstuczyn: the reason for re-export of the whole crate is so that hkdf could be used without explicit dependency in consumers (Andrei has asked for this)

common/crypto/src/rng.rs
jstuczyn: removes UnwrapErr all over the place though

common/test-utils/src/helpers.rs
jstuczyn: I think that was an omission. let me fix it.

common/wasm/client-core/src/helpers.rs
jstuczyn: true. let me revisit it.

simonwicky (approved):

